PR #1, opened 2020-02-14
Inline SVG

nucleogenesis:
# WORK IN PROGRESS

Here is an implementation that I came up with that will pull in SVG files and inline them without using webpack.

How?

- `require('@material-icons/svg/svg/MATERIAL_NAME/ICON_TYPE.svg')` gets the SVG file URL back. We strip the `data:svg...` nonsense off of it, leaving us with the Base64 representation.
- We decode the B64 to a string of the HTML `<svg ...` element. 
- We compile that as a template using the `vue-template-compiler`
- We use a `render()` function in KIcon to render the icon without any wrapping elements (avoiding using `v-html` on a div or span).
- Eliminated the need for the `svg-icon-inline-loader` altogether (semi in-progress) - KIcon _will_ allow targeting specific material icons outside of our own internal design definitions.

Additional notes:

- We inline the a11y attrs which are inserted in the `svg-icon-inline-loader` package here: https://github.com/learningequality/svg-icon-inline-loader/blob/9f137b273af978eda901c6f92b648be30322a15d/index.js#L121

- I removed the CSS for the `svg` element for now - but that's because the styling was different between an SVG in Kolibri being inserted using `mat-svg` rather than `KIcon` - I'll probably return it but this is something we'll need to consider if we go through and replace every `<mat-svg` in Kolibri with `<KIcon>` - basically the CSS there imposes an `18px` width/height but the SVGs are default given in `24px` size - so when you render using `mat-svg` you get `24px` but with `KIcon` you would get `18px` instead.

- The class for isRtl didn't seem to be used? I made sure it works with this implementation.

- It is worth noting that we can override styles and classes when using `KIcon` a la `<KIcon icon="correct" color="green" style="width: 100px; height: 200px;" class="my-fancy-class" />` and they are directly applied to the resulting `<svg>` element.

TODOs

- I think that I'll create an dictionary (object) where keys are "Kolibri Design Names" which refer to the material names. This would be easy to update and refer to

- I'll add some props to allow for targeting of specific types (outline, baseline, etc) and to allow for us to target specific material icons if we want.

- Figure out whether we reimplement the `svg {}` css styles or not and how to address the differences between items rendered with `mat-svg` in Kolibri.

### DIFF
--- a/lib/KIcon.vue
+++ b/lib/KIcon.vue
@@ -1,66 +1,10 @@
-<template>
-
-  <!-- eslint-disable max-len -->
-
-  <!-- tracking -->
-  <mat-svg v-if="icon === 'correct'" name="check_circle" category="action" :style="style" />
-  <mat-svg v-else-if="icon === 'helpNeeded'" name="error" category="alert" :style="style" />
-  <mat-svg v-else-if="icon === 'hint'" name="lightbulb_outline" category="action" :style="style" />
-  <mat-svg v-else-if="icon === 'incorrect'" name="close" category="navigation" :style="style" />
-  <mat-svg v-else-if="icon === 'inProgress'" name="access_time" category="device" :style="style" />
-  <mat-svg v-else-if="icon === 'mastered'" name="stars" category="action" :style="style" />
-  <mat-svg v-else-if="icon === 'notStarted'" name="brightness_1" category="image" :style="style" />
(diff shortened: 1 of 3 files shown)


nucleogenesis:
I screwed up my fork somehow - don't know what I did but my commits weren't being pushed right...

Will reference this in relevant PR

PR #4, opened 2020-02-26
updates to icons page and slight refactor of KIcon

indirectlylit:
incorporating aspects of 

https://docs.google.com/document/d/1xf1JWYR9U44Ikbfjb4xnu7ve6CpOEwhML9PILsTP1hg/edit

view preview here:

https://deploy-preview-4--kolibri-design-system.netlify.com/icons/


docs/pages/icons.vue
```diff
@@ -5,79 +5,114 @@
...
       <p>
-        Note that many of these icons also are used with  <InternalLink href="colors" text="conventional colors" />
-        in the design system.
+        <KLabeledIcon icon="lesson" label="Paragraph" />
       </p>
+    </PageSection>
 
+    <PageSection title="Color" anchor="#color">
+      <p>
+        When inline with text, icons should be the same color as the text.
```
nucleogenesis: This isn't always true. I'm thinking about the icons in Coach Reports for help needed, completed, etc. That's the only spot I can think of it differing in Kolibri.

docs/pages/icons.vue
```diff
@@ -5,79 +5,114 @@
...
+      </p>
+      <p>
+        Note that many of these icons also are used with <InternalLink href="colors" text="conventional colors" /> in the design system. For example, the <InternalLink href="#tokens-coach" text="coach" code /> icon is often shown using the <InternalLink href="/colors#tokens-coachContent" text="coachContent" code /> color, e.g. <KIcon icon="coach" :color="$themeTokens.coachContent" />.
```
nucleogenesis: Should we cover icons that are specific colors all the time? Help needed, completed, incorrect answers, etc all have specific colors. Also when they're used inline, they're not the same color as the text.

indirectlylit: good points thanks

indirectlylit: > icons that are specific colors all the time

Perhaps this is something we should bake into `KIcon`? specifying a particular `icon` prop could set a default color, which could be overwritten?

nucleogenesis: Yeah I think that's a good call.

indirectlylit:
@nucleogenesis I've pushed some updates - ready for re-review.

This includes a slight refactor which makes it so that:

* only some icons should be RTL-flipped
* add default colors
* removes the props on the generated SVG icons which didn't seem necessary

indirectlylit:
@jtamiace @khangmach see new icon guidance here: https://deploy-preview-4--kolibri-design-system.netlify.com/icons/

Please let me know if you have any feedback. This page will very likely need to be updated by you two

docs/pages/icons.vue
indirectlylit: updated

PR #5, opened 2020-02-27
kuitextbox autofocus directive expects object

nucleogenesis:
KeenUITextbox has the `autofocus` directive. The value for `autofocus` in vanilla KeenUIText box imports from a file that exports an object like the one that is defined in the changes in this PR.

The issue was that when trying to login with a username that requires a password, I was not being autofocused to the password textbox and the password textbox was disabled.


nucleogenesis:
Addressed here: https://github.com/learningequality/kolibri-design-system/pull/1

PR #6, opened 2020-02-28
Component Review Prework

nucleogenesis:
Continuation of https://github.com/learningequality/kolibri-design-system/pull/1 which I closed to combine everything into one place.

I messed up my commits - and also wanted every one of these changes in one branch & PR for easier import and testing in Kolibri.

*KIcon, Material SVGs*

`lib/utils/precompileSVGs.js` compiles all of the SVGs in the `material-icons` package into a `lib/material-svg` folder.

KIcon directly refers to those items.

If we add an icon, KIcon should be the place where we add it.

Note that in Kolibri - we have to add an exclusion to the buble loader to avoid trying to parse the JS files that are prerendered ([see this](https://github.com/learningequality/kolibri/pull/6606/files#r385926207))

*Bugfix*
This also includes a quick bugfix for  autofocuse in KeenUITextbox

*Brand colors*

Kolibri purple brand colors are hardcoded here now.


nucleogenesis:
@indirectlylit 

- Your Vue template generation idea is awesome and feels far less hacky. It was a bugger getting to work for minor but not immediately obvious reasons but in the end I'm happy for the experience.

I think I cleaned everything up well and the deploy preview looks good.

PR #13, opened 2020-03-20
Update vendored keen

nucleogenesis:
Keen components updated to latest.

Related: https://github.com/learningequality/kolibri/pull/6691


README.md
```diff
@@ -129,3 +129,30 @@ Icons are drawn from https://github.com/material-icons/material-icons and then c
 ```bash
 yarn run precompile-svgs
 ```
+
```
nucleogenesis: Oops - I wrote this up but meant to include it in a separate PR. I just `git add .`'ed so that I could rebase on `develop` - should've `stashed`... I'll remove this change here and make a new with it later as it is my first pass.

PR #16, opened 2020-03-26
Buttons! KButtonGroup, KIconButton, etc

nucleogenesis:
# Overview

~**WIP**~

** Ready for Review**

HOW TO:

- Checkout this branch locally,
- Checkout https://github.com/learningequality/kolibri/pull/6696 this branch locally
- `cd /path/to/design-system` and: `yarn link`
- `cd /path/to/kolibri` and remove the entry for `kolibri-design-system` from the `kolibri/core/package.json`
- `yarn link kolibri-design-system` then `yarn install`

Now you can run the devserver and will be using this branch :+1: 

## KButtonGroup

Related Issue: https://github.com/learningequality/kolibri/issues/5900
Related Kolibri PR: https://github.com/learningequality/kolibri/pull/6696

### What was done

- Remove 8px base margin on buttons.
- Create KButtonGroup - puts 8px margin on all `<a>` and `<button>` direct children of the `<span>` wrapper (with a CSS `.class > a, .class > button` selector).
... a lot of other stuff...
- Updates to KLabeledButton, KButton, KRouterLink, KExternalLink

_In Kolibri_

- Wrapped adjacent `KButton`s with the new `KButtonGroup`.
- *Strong note* that buttons which previously were 8px offset from - say - the edge of their wrapping container are now aligned with them. A good example of this is any button at the top-right of a table in Coach or Facility. There are likely other places where this is the case.

## KIconButton

Circular buttons replacing all UIIconButtons in Kolibri

## KRouterLink, ExternalLink, Button Icon slots/props

### DIFF
--- a/lib/KBreadcrumbs.vue
+++ b/lib/KBreadcrumbs.vue
@@ -9,7 +9,11 @@
           appearance="raised-button"
           @click="showDropdown = !showDropdown"
         />
-        <div v-if="showDropdown" class="breadcrumbs-dropdown">
+        <div
+          v-if="showDropdown"
+          class="breadcrumbs-dropdown"
+          :style="{ background: $themeTokens.surface }"
+        >
           <ol class="breadcrumbs-dropdown-items">
             <li
               v-for="(crumb, index) in collapsedCrumbs"
@@ -270,13 +274,12 @@
   }
 
   .breadcrumbs-dropdown {
-    @extend %dropshadow-16dp;
+    @extend %dropshadow-8dp;
 
     position: absolute;
-    z-index: 100;
+    z-index: 8;
     padding: 16px;
     font-weight: bold;
-    background: white;
   }
 
   .breadcrumbs-dropdown-items {
--- a/lib/KBreadcrumbs.vue
+++ b/lib/KBreadcrumbs.vue
@@ -3,25 +3,28 @@
   <div v-show="showSingleItem || crumbs.length > 1">
     <nav class="breadcrumbs">
       <div v-show="collapsedCrumbs.length" class="breadcrumbs-dropdown-wrapper">
-        <UiIconButton :hasDropdown="true" size="small">
-          <KIcon icon="arrow_down" />
-          <div slot="dropdown" class="breadcrumbs-dropdown">
-            <ol class="breadcrumbs-dropdown-items">
-              <li
-                v-for="(crumb, index) in collapsedCrumbs"
-                :key="index"
-                class="breadcrumbs-dropdown-item"
-              >
-                <KRouterLink
-                  :text="crumb.text"
-                  :to="crumb.link"
-                  :style="{ maxWidth: `${collapsedCrumbMaxWidth}px` }"
-                  dir="auto"
-                />
-              </li>
-            </ol>
-          </div>
-        </UiIconButton>
+        <KIconButton
+          size="small"
+          :icon="showDropdown ? 'arrow_up' : 'arrow_down'"
+          appearance="raised-button"
+          @click="showDropdown = !showDropdown"
+        />
+        <div v-if="showDropdown" class="breadcrumbs-dropdown">
+          <ol class="breadcrumbs-dropdown-items">
+            <li
+              v-for="(crumb, index) in collapsedCrumbs"
+              :key="index"
+              class="breadcrumbs-dropdown-item"
+            >
+              <KRouterLink
+                :text="crumb.text"
+                :to="crumb.link"
+                :style="{ maxWidth: `${collapsedCrumbMaxWidth}px` }"
+                dir="auto"
+              />
+            </li>
+          </ol>
+        </div>
       </div>
 
       <ol class="breadcrumbs-visible-items">
@@ -93,7 +96,6 @@
   import throttle from 'lodash/throttle';
   import every from 'lodash/every';
   import keys from 'lodash/keys';
-  import UiIconButton from './keen/UiIconButton';
   import KResponsiveElementMixin from './KResponsiveElementMixin';
 
   const DROPDOWN_BTN_WIDTH = 55;
@@ -110,7 +112,6 @@
    */
   export default {
     name: 'KBreadcrumbs',
-    components: { UiIconButton },
     mixins: [KResponsiveElementMixin],
     props: {
       /**
@@ -145,6 +146,7 @@
       // Each object contains:
       // text, router-link 'to' object, vue ref, a resize sensor, and its collapsed state
       crumbs: [],
+      showDropdown: false,
     }),
     computed: {
       collapsedCrumbs() {
@@ -247,6 +249,8 @@
 
 <style lang="scss" scoped>
 
+  @import './styles/definitions';
+
   .breadcrumbs {
     margin-top: 8px;
     margin-bottom: 8px;
@@ -266,6 +270,10 @@
   }
 
   .breadcrumbs-dropdown {
+    position: absolute;
+    @extend %dropshadow-16dp;
+    background: white;
+    z-index: 100;
     padding: 16px;
     font-weight: bold;
   }
--- a/lib/KBreadcrumbs.vue
+++ b/lib/KBreadcrumbs.vue
@@ -270,12 +270,13 @@
   }
 
   .breadcrumbs-dropdown {
-    position: absolute;
     @extend %dropshadow-16dp;
-    background: white;
+
+    position: absolute;
     z-index: 100;
     padding: 16px;
     font-weight: bold;
+    background: white;
   }
 
   .breadcrumbs-dropdown-items {
--- a/lib/KCheckbox.vue
+++ b/lib/KCheckbox.vue
@@ -25,16 +25,19 @@
         <KIcon
           v-if="isCurrentlyIndeterminate"
           :style="notBlank"
+          class="indeterminate"
           icon="indeterminate_check_box"
         />
         <KIcon
           v-else-if="!isCurrentlyIndeterminate && isCurrentlyChecked"
           :style="[ notBlank, activeOutline ]"
+          class="checked"
           icon="check_box"
         />
         <KIcon
           v-else
           :style="[ blank, activeOutline ]"
+          class="unchecked"
           icon="check_box_outline_blank"
         />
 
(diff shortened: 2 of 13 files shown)


indirectlylit:
> Strong note that buttons which previously were 8px offset from - say - the edge of their wrapping container are now aligned with them. A good example of this is any button at the top-right of a table in Coach or Facility. There are likely other places where this is the case.

Perfect, this is desired behavior. In general, buttons should be aligned with text above and below it, not offset.

There are many places where we added CSS to remove this margin which can now be removed. Here are the ones I could find:

* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/plugins/coach/assets/src/views/common/QuizStatus.vue#L13-L19
* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/plugins/coach/assets/src/views/common/QuizStatus.vue#L30-L36
* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/plugins/device/assets/src/views/ManageContentPage/ManageChannelContentsPage/index.vue#L23-L27
* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/plugins/learn/assets/src/views/ExamPage/index.vue#L97-L103
* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/core/assets/src/views/AppError/TechnicalTextBlock.vue#L20-L26
* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/core/assets/src/views/language-switcher/LanguageSwitcherList.vue#L29-L35
* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/plugins/coach/assets/src/views/plan/GroupsPage/index.vue#L14-L19
* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/plugins/coach/assets/src/views/plan/LessonContentPreviewPage/SelectOptions.vue#L11-L25
* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/plugins/device/assets/src/views/AvailableChannelsPage/index.vue#L30-L37
* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue#L77-L83
* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/plugins/device/assets/src/views/ManageContentPage/index.vue#L27-L32
* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/plugins/device/assets/src/views/ManageContentPage/ChannelPanel/WithSizeAndOptions.vue#L42-L47
* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/plugins/device/assets/src/views/UserPermissionsPage/index.vue#L75-L82
* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/plugins/facility/assets/src/views/DataPage/index.vue#L19-L24
* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/plugins/facility/assets/src/views/DataPage/index.vue#L51-L56
* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/index.vue#L69-L75
* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/plugins/learn/assets/src/views/AssessmentWrapper/index.vue#L77-L84
* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/OnboardingForm.vue#L24-L29
* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/plugins/user/assets/src/views/ProfileEditPage.vue#L38-L44
* https://github.com/learningequality/kolibri/blob/27d11d9c838655f43a77aa297c0d5429e8e1cf65/kolibri/plugins/user/assets/src/views/SignUpPage.vue#L85-L91


Let's delete some CSS!

indirectlylit:
switched label: this is "library" code (in the `/lib` dir), not "supporting" code (e.g. nuxt or docs parsers)

lib/KBreadcrumbs.vue
```diff
@@ -145,6 +146,7 @@
       // Each object contains:
       // text, router-link 'to' object, vue ref, a resize sensor, and its collapsed state
       crumbs: [],
+      showDropdown: false,
```
indirectlylit, line 149: nice - better to bring this functionality here

lib/KBreadcrumbs.vue
```diff
@@ -266,8 +270,13 @@
   }
 
   .breadcrumbs-dropdown {
+    @extend %dropshadow-16dp;
+
+    position: absolute;
+    z-index: 100;
```
indirectlylit, line 276: Generally we try to match dropshadow and z-index together based on default material elevations:

https://material.io/design/environment/elevation.html#default-elevations

In theory then this would be `dropshadow-8dp` and `z-index: 8`

lib/KBreadcrumbs.vue
```diff
@@ -266,8 +270,13 @@
...
     padding: 16px;
     font-weight: bold;
+    background: white;
```
indirectlylit, line 279: technically, this should be set using the `surface` token in the theme:

https://kolibri-design-system.netlify.com/colors/#tokens-surface

lib/KCheckbox.vue
```diff
@@ -228,4 +231,13 @@
...
   }
 
+  // KIcon overrides
+  .indeterminate,
+  .unchecked,
+  .checked {
+    top: 0;
+    width: $checkbox-height;
+    height: $checkbox-height;
+  }
```
indirectlylit, line 241: I'm a bit confused why there are three classes defined rather than just a single `checkbox` class?

lib/KLabeledIcon.vue
```diff
@@ -29,11 +36,21 @@
       KIcon,
     },
     props: {
-      // If provided, will render a KIcon with the same 'icon' prop
+      // If provided, will render a KIcon using the value given as the `icon` prop. The icon will be prepended to the label.
```
indirectlylit, line 39: nice, thanks for adding the Markdown

```suggestion
      // An icon that will be prepended to the label. Uses the same icon names as the `KIcon` component
```

lib/KLabeledIcon.vue
```diff
@@ -29,11 +36,21 @@
...
       icon: {
         type: String,
         required: false,
       },
+      // If provided, will render a KICon using the value given as the `icon` prop. The iconAfter will be appended to the label.
```
indirectlylit, line 44:
```suggestion
      // An icon that will be appended to the label. Uses the same icon names as the `KIcon` component
```

lib/KLabeledIcon.vue
```diff
@@ -29,11 +36,21 @@
...
+      iconAfter: {
+        type: String,
+        required: false,
+      },
+      // If provided, determines the color of the label and any icons that are provided.
```
indirectlylit, line 49: Following writing convention: Skip periods and unnecessary punctuation

https://kolibri-design-system.netlify.app/writing/#punctuation

```suggestion
      // If provided, determines the color of the label and any icons that are provided
```

lib/KLabeledIcon.vue
```diff
@@ -57,6 +74,19 @@
           (defaultSlot.children && defaultSlot.children.length)
         );
       },
+      labelStyle() {
+        let styles = {};
+        // Margin for icons - use em to match parent font size
+        if (this.iconAfter || this.$slots['iconAfter']) {
+          styles['marginRight'] = '1.975em';
```
indirectlylit, line 81:
```suggestion
          styles['marginRight'] = '1.975em'; // scale with parent font size
```

lib/KLabeledIcon.vue
```diff
@@ -57,6 +74,19 @@
...
+        }
+
+        if (this.icon || this.$slots['icon']) {
+          styles['marginLeft'] = '1.975em';
```
indirectlylit, line 85:
```suggestion
          styles['marginLeft'] = '1.975em'; // scale with parent font size
```

lib/buttons-and-links/KButtonGroup.vue
```diff
@@ -0,0 +1,38 @@
...
+
+  .button-group {
+    overflow: hidden;
+  }
+
+  /* KButton can produce <a> or <button> tags */
+  .button-group > a,
+  .button-group > button {
+    margin: 8px;
+  }
```
indirectlylit, line 36: I think this should only be adding internal margins, not external.

Something like

```css
  .button-group > a:not(:first-child) ,
  .button-group > button:not(:first-child)  {
    margin-left: 8px;
  }
```

(untested)

lib/buttons-and-links/KExternalLink.vue
```diff
@@ -6,8 +6,26 @@
     :href="href"
     :download="download"
     dir="auto"
+    @mouseenter="hovering = true"
+    @mouseleave="hovering = false"
```
indirectlylit, line 10: too bad that this is necessary, but I think it's fine.

We could add a css transition inside `KIcon` so this uses the same fade transition as defined in buttons.scss

lib/KCheckbox.vue
nucleogenesis: @indirectlylit honestly because I saw the same convention used in KRadioButton's template. Changed them now to `checkbox-icon`

PR #18, opened 2020-04-01
adding functionality for displaying component APIs

indirectlylit:
This is a first pass at supporting component API documentation in the design system. It uses `vue-docgen-api` from https://github.com/vue-styleguidist/vue-styleguidist to parse our components.

For testing, see the `KCheckbox` and `KEmptyPlaceholder` pages:

* https://deploy-preview-18--kolibri-design-system.netlify.com/kcheckbox/
* https://deploy-preview-18--kolibri-design-system.netlify.com/kemptyplaceholder/


docs/pages/KCheckbox.vue
```diff
@@ -0,0 +1,47 @@
...
+          label="Another label"
+          :checked="false"
+        />
+      </DocsShow>
+      <p>
+        The checkbox is generally used to select one of two possible values in a form. It should not be used to make real-time changes; for this situation, use a <code>KSwitch</code> component instead.
+      </p>
+      <p>Layout:</p>
+      <ul>
+        <li>Always left-aligned</li>
```
nucleogenesis, line 22: This is a bit confusing. In RTL this would be right-aligned. If you say "always" I'd think that this is true even when in RTL.

Is there a proper term for "text-direction aligned"?

docs/common/DocsPageTemplate/SideNav/index.vue
```diff
@@ -25,17 +25,11 @@
...
-          <li v-for="(page, i) in visiblePatternRoutes" :key="i">
-            <NavLink :page="page" />
-          </li>
-        </NavSectionList>
-        <NavSectionList v-show="showComponents" :title="componentsText">
-          <li v-for="(page, i) in visibleComponentRoutes" :key="i">
-            <NavLink :page="page" />
-          </li>
-        </NavSectionList>
+        <NavSectionList
```
nucleogenesis, line 28: Pretty refactor here overall :+1:

indirectlylit:
>  everything except the comment-derived description

Are you referring to the top-level descriptions, e.g. 

```html
<script>
  /**
   * Used for toggling boolean user input
   */
  export default {
``` 

We actually have access to those and could display them if we wanted to. However, I was leaning towards actually deleting these, and instead leaning on all high-level documentation being in the `/pages/K*.vue` files in order to keep things DRY.

Thoughts?

nucleogenesis:
@indirectlylit 

By this:

> everything except the comment-derived description

I was referring to the extracted API (props, slots, etc) comments that are put into the 'description' column for each.

As for the high-level descriptive comments giving an overview of the component altogether - agreed - would be better to keep them in the `/pages` path.

indirectlylit:
>  It's great that it generates from code everything except the comment-derived description. 

Ah I understand your point: adding things in comment-derived descriptions is a convention that we'll need to document and train people on, whereas things in code will always be there.

This is a good point: we'll need to make sure people are aware how to use these. I don't even necessarily know the full extend of the DSL that you can use in these. I believe it's a [Vue extension](https://vue-styleguidist.github.io/docs/Documenting.html#using-jsdoc-tags) of [JSDoc](https://en.wikipedia.org/wiki/JSDoc)

docs/pages/KCheckbox.vue
indirectlylit: updated

indirectlylit:
* updated as per comments
* also added documentation on using the new functionality

PR #27, opened 2020-06-02
kbreadcrumbs_single_tag_remove

rationality6:
single stylelint warnning fix from Kolibri isssue #6614

this is one single fix for #6614 for test


nucleogenesis:
Thanks @rationality6 !

PR #29, opened 2020-06-07
UiTextbox & KTextbox autocapitalize

Lewiscowles1986:
Adds `autocapitalize` support to Kolibri design system, RE: https://github.com/learningequality/kolibri/pull/7002 & https://github.com/learningequality/kolibri/issues/6831

### DIFF
--- a/utils/pregenerate.js
+++ b/utils/pregenerate.js
@@ -18,14 +18,23 @@ export default `;
 
 async function writeApi() {
   const componentFolder = path.resolve('./lib/');
-  const files = await globby('**/K*.vue', { cwd: componentFolder });
+  const simpleFiles = await globby('**/K*.vue', { cwd: componentFolder });
   const output = {};
 
-  for (const filename of files) {
+  for (const filename of simpleFiles) {
     const filepath = path.resolve(componentFolder, filename);
     const val = await docGenAPI.parse(filepath);
     output[val.displayName] = val;
   }
+
+  const complexFiles = await globby('**/K*/index.vue', { cwd: componentFolder });
+
+  for (const filename of complexFiles) {
+    const filepath = path.resolve(componentFolder, filename);
+    const val = await docGenAPI.parse(filepath);
+    output[val.displayName] = val;
+  }
+
   const outputPath = path.resolve('./docs/jsdocs.js');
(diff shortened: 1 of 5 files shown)


utils/pregenerate.js
```diff
@@ -18,14 +18,23 @@ export default `;
 
 async function writeApi() {
   const componentFolder = path.resolve('./lib/');
-  const files = await globby('**/K*.vue', { cwd: componentFolder });
+  const simpleFiles = await globby('**/K*.vue', { cwd: componentFolder });
```
Lewiscowles1986, line 21: these changes were to allow the auto-building / extraction of both named (simple) files and files with a hierarchical folder structure. I've no idea why you have that or why this is so different to the keen build system, but it was necessary to stop errors when building.

Lewiscowles1986:
**Important:** I've raised a bug with Google Chrome https://bugs.chromium.org/p/chromium/issues/detail?id=1092224 which may apply to other browsers on a mobile. It seems to be a bug if non-stock keyboard is used (Specifically Microsoft on-screen keyboard is known not to work for me)

It might be a good idea to add something to the docs / manual for Kolibri if this merges as-is to call out that third party keyboards are known to interact poorly with on-screen keyboard devices.

Lewiscowles1986:
Thanks 🥇

indirectlylit:
lol @ https://bugs.chromium.org/p/chromium/issues/detail?id=1092224

> What steps will reproduce the problem?
> (1) select one of the autocapitalize inputs
> (2) notice it starts with a capitalized first letter
> (3) cry

😢

indirectlylit:
> It might be a good idea to add something to the docs / manual for Kolibri if this merges as-is to call out that third party keyboards are known to interact poorly with on-screen keyboard devices.

I agree in principle, but we typically wait to document things until we get at least a few questions in the forums or bug reports. Otherwise our docs will get very large and hard to maintain!

Lewiscowles1986:
@indirectlylit there is a workaround for the autocapitalize of Microsoft swiftkey if it comes up (turn off autocapitalize in it's settings, although that is global and not per-control). Without combing through chromium source code I am unsure if this is poor coding from Microsoft, or if Google does not provide a straightforward way to communicate to such options to third-party keyboards.

Issue #32, opened 2020-06-10
KButtonGroups hide shadows of buttons


jonboiser:
Example in KDP Sync modal:

![image](https://user-images.githubusercontent.com/10248067/84307861-6e821000-ab12-11ea-8ce1-8d87725a29eb.png)

indirectlylit:
I'm still seeing this in the 0.14 beta 3:

![image](https://user-images.githubusercontent.com/2367265/85078873-4df13000-b17a-11ea-8d00-e3e961c66d53.png)

nucleogenesis:
I'm no longer seeing this in the `release-v0.14.x` branch - with confirmation from another we can close this again.

PR #37, opened 2020-06-18
Fix for 0.14 to KButtonGroup

nucleogenesis:
The overflow is unnecessary and hides the shadow. If it is not unnecessary we'll hear about it for sure but I haven't seen any issues.


nucleogenesis:
Also - going to merge this so it's ready for bug bash.

nucleogenesis:
Related issues: 

https://github.com/learningequality/kolibri-design-system/issues/32
https://github.com/learningequality/kolibri/issues/7016

PR #38, opened 2020-06-18
remove undefined prop reference

indirectlylit:
seeing lots of errors like this in current 0.14 Kolibri:

```
KCircularLoader.vue?7265:9 [Vue warn]: Property or method "disableTransition" is not defined on the instance but referenced during render. Make sure that this property is reactive, either in the data option, or for class-based components, by initializing the property. See: https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties.

found in

---> <KCircularLoader> at node_modules/kolibri-design-system/lib/loaders/KCircularLoader.vue
       <TaskPanel> at kolibri/plugins/device/assets/src/views/ManageTasksPage/TaskPanel.vue
         <TransitionGroup>
           <ManageTasksPage> at kolibri/plugins/device/assets/src/views/ManageTasksPage/index.vue
             <WithAuthMessage>
               <KPageContainer> at node_modules/kolibri-design-system/lib/KPageContainer.vue
                 <CoreBase> at kolibri/core/assets/src/views/CoreBase/index.vue
                   <Root>
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ yarn run precompile-svgs
 
 ### Development in parallel with other applications
 
-If you are working on this repository and want to see live updates to the component library here in other applications that are using this library, then you will need to do a few things.
+If you are working on the design system library code in this repo and want to see local updates reflected in other applications that are using this library, then you will need to do a few things.
 
 1. While in the root of your local `kolibri-design-system` repository, run `yarn link`.
 2. In the application where you intend to use `kolibri-design-system` remove the reference to `kolibri-design-system` from the `package.json` of that project.
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ If you are working on the design system library code in this repo and want to se
 
 1. While in the root of your local `kolibri-design-system` repository, run `yarn link`.
 2. In the application where you intend to use `kolibri-design-system` remove the reference to `kolibri-design-system` from the `package.json` of that project.
-3. In the root of the application where you intend to use `kolibri-design-system` run `yarn link kolibri-design-system && yarn install`.
+3. In the root of the application where you intend to use `kolibri-design-system` run `yarn link kolibri-design-system` and then `yarn install`.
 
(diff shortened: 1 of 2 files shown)


nucleogenesis:
I thought I fixed that one but in the other direction here: https://github.com/learningequality/kolibri-design-system/pull/30

This is actually probably better a quick search of Kolibri doesn't show "disableTransition" anywhere so it's probably safe to remove it from the props definition as well

indirectlylit:
> it's probably safe to remove it from the props definition as well

oh strange, now I'm super confused by the error `Property or method "disableTransition" is not defined on the instance but referenced during render.` because it very clearly _is_ defined as a prop!

What am I missing? 🤔  

```js
      disableTransition: {
        type: Boolean,
        default: false,
      },
```

MisRob:
> oh strange, now I'm super confused by the error Property or method "disableTransition" is not defined on the instance but referenced during render. because it very clearly is defined as a prop!

@indirectlylit I ran into the same issue yesterday :)

Although `disableTransition` is defined in the code, it's not present in the result bundle. I think that the problem might be [this line's syntax](https://github.com/learningequality/kolibri-design-system/blob/10457e9147e5a373535d7bf7ff470b4d78d1d71d/lib/loaders/KCircularLoader.vue#L106). My assumption is that `stroke: Number` needs to be replaced by `stroke: { type: Number }`.

MisRob:
So in Kolibri I needed to also remove _yarn.lock_ before reinstalling modules to be able to get up-to-date `develop` version that contains the prop. This is my lockfiles dif:

Selection_001

I don't have much experience with referencing another repository directly though I suppose that we might need to raise versions after each update here so there's no need to delete the lockfile?

I guess this should not cause any problems because it's related to another _package.json_ nested in _kolibri/packages_ though what's the reason for referencing [one of our fork's repository](https://github.com/learningequality/kolibri/blob/develop/packages/kolibri-core-for-export/package.json#L20) (I guess that we will need to be careful about its updates too)?

indirectlylit:
self-merging to test updating betas

PR #40, opened 2020-06-23
New file for icon definitions & some code to sort and present icons

nucleogenesis:
I needed the icon definitions cleanly available both in the Docs and our `<KIcon>` component and place to put the sorting logic. I honestly like the idea of having a file where the icons are defined. 

Maybe the logic I added could be extracted to the Docs component since that's the only place it will be used so the iconDefinitions file lives up to its name?

---

This PR adds iconDefinitions.js which exports:

`KolibriIcons` - Our object of icon definitions removed from `KIcon/index.vue`

`sortedColoredIcons` & `sortedBlackIcons` - These are each an array of arrays. Each subarray is a list of aliases which refer to the same icon sorted alphabetically. Then, each of those subarrays is sorted alphabetically by their first element. See the comment above in `iconDefinitions.js` line 207.

The `*Colored*` version is our aliased icons which include a specific color. The `*Black*` are those which use the default text color.

You can see how I made a `KIconDebug` component here: https://gist.github.com/nucleogenesis/62a1a32337338193c36df567b3e16470 that uses the `sorted*Icons` to render all of them separately.

I didn't include that component in the PR because there's no use for it once we have the icons documentation in place.


lib/KIcon/iconDefinitions.js
```diff
@@ -0,0 +1,268 @@
...
+
+    if (hasColorToken) {
+      coloredPaths = thePathsWeCareAbout;
+    } else {
+      blackPaths = thePathsWeCareAbout;
+    }
+  });
+
+  return { blackPaths, coloredPaths };
+})();
```
indirectlylit, line 263: might be able to use ES2015 classes for this?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes

lib/KIcon/iconDefinitions.js
```diff
@@ -0,0 +1,268 @@
...
+
+const sortedColoredIcons = sortedIcons(allIcons.coloredPaths);
+const sortedBlackIcons = sortedIcons(allIcons.blackPaths);
```
indirectlylit, line 266: I'm a little confused why we're exporting the file paths of the components, e.g. `KolibriIcons[alias].icon.__file`. Where would this be used?

nucleogenesis: `allIcons.coloredPaths` and `allIcons.blackPaths` use that file path as a unique identifier. All aliases that point to the same `KolibriIcons[alias].icon.__file` are aliases for the same icon. Then there's some accounting for whether it has a `defaultColor` defined on it.

```
{
    '/path/to/a/specific/icon_type/Baseline.vue': [ 'alias1', 'alias2' ],
    '/path/to/a/different/icon_type/Outline.vue': ['differentAlias'],
    ...
}
```

We're exporting the return value of passing those objects to `sortedIcons()` which turns it into a sorted array of sorted arrays:

```
[
    [ 'channel'],
    [ 'coach', 'coachContents'],
    ...
    ['zebra']
]
```

Ultimately, this is all just so that you can `v-for` that outermost array where the iterator is an array of aliases which all refer to the same KIcon icon.

```
<div v-for="aliasArray in sortedBlackIcons">
    <KIcon :icon="aliasArray[0]" />
    <span> {{ aliasArray.join(", ") }} </span>
</div>
```

nucleogenesis: FWIW here is what KIconDebug looks like:

image

indirectlylit:
retargeted to new release branch, v0.2.x

nucleogenesis:
@indirectlylit there are a few other changes in this PR right now that could use a quick look before merge.

nucleogenesis:
@indirectlylit merged this to make https://github.com/learningequality/kolibri/pull/7190 pass and I will fix any issues in KDS that arise from there (hopefully none).

PR #41, opened 2020-06-25
Fix breadcrumbs overflow

bjester:
See regression issue here https://github.com/learningequality/kolibri/issues/7140

- Added `text` slot to `<KRouterLink>`
- Added `maxWidth` prop to `<KRouterLink>` and `<KLabeledIcon>`
- Added `.breadcrumbs-crumb-text` styling to handle text overflow
- Reduced specificity of global selectors styling the link text

Screenshot from 2020-07-01 10-16-47
Screenshot from 2020-07-01 10-16-20
Screenshot from 2020-07-01 10-16-04

### DIFF
--- a/lib/KBreadcrumbs.vue
+++ b/lib/KBreadcrumbs.vue
@@ -55,6 +55,7 @@
             <span
               :style="{ maxWidth: `${lastCrumbMaxWidth}px` }"
               dir="auto"
+              class="link-text"
             >
               {{ crumb.text }}
             </span>
@@ -313,7 +314,8 @@
   .breadcrumbs-visible-item {
     display: inline-block;
     vertical-align: middle;
-    &:first-child {
+
+    /deep/ .link-text {
       display: inline-block;
       max-width: 300px;
       overflow: hidden;
--- a/lib/KBreadcrumbs.vue
+++ b/lib/KBreadcrumbs.vue
@@ -1,12 +1,19 @@
 <template>
 
-  <div v-show="showSingleItem || crumbs.length > 1">
+  <div
+    v-show="showSingleItem || crumbs.length > 1"
+    :class="{'breadcrumbs-collapsed': collapsedCrumbs.length}"
+  >
     <nav class="breadcrumbs">
-      <div v-show="collapsedCrumbs.length" class="breadcrumbs-dropdown-wrapper">
+      <div
+        v-show="collapsedCrumbs.length"
+        class="breadcrumbs-dropdown-wrapper"
+      >
         <KIconButton
           size="small"
           :icon="showDropdown ? 'arrow_up' : 'arrow_down'"
           appearance="raised-button"
+          :style="{ verticalAlign: 'middle' }"
           @click="showDropdown = !showDropdown"
         />
         <div
@@ -24,9 +31,11 @@
                 class="krouter-item"
                 :text="crumb.text"
                 :to="crumb.link"
-                :style="{ maxWidth: `${collapsedCrumbMaxWidth}px` }"
-                dir="auto"
-              />
+              >
+                <template #text="{ text }">
+                  <span class="breadcrumbs-crumb-text" dir="auto">{{ text }}</span>
+                </template>
+              </KRouterLink>
             </li>
           </ol>
         </div>
@@ -44,7 +53,11 @@
               :text="crumb.text"
               :to="crumb.link"
               dir="auto"
-            />
+            >
+              <template #text="{ text }">
+                <span class="breadcrumbs-crumb-text">{{ text }}</span>
+              </template>
+            </KRouterLink>
           </li>
 
           <li
@@ -53,9 +66,8 @@
             class="breadcrumb-visible-item-last breadcrumbs-visible-item"
           >
             <span
-              :style="{ maxWidth: `${lastCrumbMaxWidth}px` }"
+              class="breadcrumbs-crumb-text"
               dir="auto"
-              class="link-text"
             >
               {{ crumb.text }}
             </span>
@@ -75,7 +87,11 @@
             :key="index"
             class="breadcrumbs-visible-item breadcrumbs-visible-item-notlast"
           >
-            <KRouterLink :text="crumb.text" :to="crumb.link" tabindex="-1" />
+            <KRouterLink :text="crumb.text" :to="crumb.link" tabindex="-1">
+              <template #text="{ text }">
+                <span class="breadcrumbs-crumb-text">{{ text }}</span>
+              </template>
+            </KRouterLink>
           </li>
 
           <li
@@ -84,7 +100,9 @@
             :key="index"
             class="breadcrumb-visible-item-last breadcrumbs-visible-item"
           >
-            <span :style="{ maxWidth: `${lastCrumbMaxWidth}px` }">{{ crumb.text }}</span>
+            <span class="breadcrumbs-crumb-text">
+              {{ crumb.text }}
+            </span>
           </li>
         </template>
       </ol>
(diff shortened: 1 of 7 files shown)


lib/KBreadcrumbs.vue
```diff
@@ -313,7 +314,8 @@
   .breadcrumbs-visible-item {
     display: inline-block;
     vertical-align: middle;
-    &:first-child {
+
+    /deep/ .link-text {
```
indirectlylit, line 318: This change is a little bit confusing/brittle because it's using a `/deep/` link to refer to both a class defined in the current component, as well as a class defined in a completely separate component.

If at all possible it's best to avoid `/deep/` because it exposes the internal implementation details of `KRouterLink` as an implicit API, breaking the otherwise `scoped` styles. If someone ever refactors that class, the thumbnails will break.

A way to address this using more explicit component API mechanisms might be to expose a new prop like `linkTextStyles` which allows you to pass in the desired styles as an object. Then the button `KRouterLink` can apply it to the right DOM node. We can make the `linkTextStyles` prop private for now so it's only used within design system internals.

bjester: @indirectlylit I see what you're saying but I also think that it might be putting developer convenience and preference above the end user who will end up seeing this rendered. I think that approach makes it more complex, and therefore more expensive in a way for users. I think perhaps defining some global CSS pattern for components and descendant elements would make more sense and is not at the cost of our users. And also, this will be a versioned package right? So I would assume some "breaking" changes would occur at some point and that it won't be backward compatible forever.

nucleogenesis: @bjester Could you clarify how @indirectlylit 's suggestion would affect the end user?

I'll toss my two cents in. I dislike `/deep/` in general - especially when we own the source of both components that we're targeting with styles because I see using `/deep/` as the same as reaching into a class and changing a private field or method. If we're trying to make a change in a 3rd party component's styles, then there's no other way beside vendoring the component's source.

If we have a need for adjusting the text, then I think maybe just adding a slot might work because you could just slot it in like:

```
<KRouterLink>
    <TextTruncator slot="text" :text="text" class='link-text' />
</KRouterLink>
```

and in KRouterLink

```
<slot></slot>
<span v-if="!$slots.text">{{ text }}</span>
```

This opens up the API for KRouterLink quite a bit because you could pass anything in there that you want to have wrapped in a router link (which would almost always be exception to the rules we're defining in KDS) - but it gives some subtle flexibility if and when we need it to.

This way you don't have to mess with object-based styles when using something other than plaintext in your router link because you can just define styles on it directly that override any inherited styles easily.

indirectlylit: We've been working hard to ban global CSS classes and patterns from the Kolibri code base, and are almost there. It _greatly_ improves the composability and modularity of components when there is no CSS leakage across component boundaries.

indirectlylit: I like the `slot`-based proposal above - it's more flexible and more intuitive than the object-based styles I proposed

bjester: @nucleogenesis I totally agree it's great design, I just question applying it to CSS. I think it adds complexity, and complexity means more code, and more code means larger file sizes and perhaps delayed loading, rendering, or additional processing. I like your slot based approach and I think it solves this situation nicely.

On the subject of more styling in JS generally speaking, I also feel that it can be somewhat anti-vue. I see the single file component structure as a coupling of individual technologies to produce a single solution (component). Mixing too much of either HTML or styling with the JS I think breaks that down, and make me question using Vue.js over something like React where that would be more cleanly implemented in my opinion. I think this [article section](https://medium.com/@s.molinari/templating-separation-of-concerns-or-separation-of-technology-or-something-else-123a3d41f0b4#2bfb) has some good thoughts. I like the point at the end:

> If I am going to go to a component for a reason to change it, i.e. the structure, the design or the display (or even data) logic, what is going to be simpler for me? Looking at a wall of JSX (maybe), or the clear separation of HTML, CSS and JS? I’ll let you all make the final decision on that for yourselves

Something that I think would be helpful and would feel less awkward to me would be leveraging the CSS technologies of LESS or SASS to battle some of these CSS challenges.

nucleogenesis: RE: SFCs

> I see the single file component structure as a coupling of individual technologies to produce a single solution (component).

I frame SFC's differently because when you build a Vue app the SFC is converted to a JS object. You're instantiating an object of a class by passing it 3 things that just happen to live in the same file:

- A valid Vue object  (the <script>)
- A string that is a valid Vue template (the <template>)
- A string that is your styles - whatever kind of `lang=<type>` you defined. (the <style>)

So a Vue SFC is just a prettied up version of two strings and a JS object. With consideration of this quote from that article:

> The Single Responsibility Principle (SRP) states that each software module should have one and only one reason to change

The SFC is not the 'module' - it's just a nice way to look at what conceptually should be seen as 3 modules at once IMO.

*RE: Styling in JS generally*

I just came to terms with the fact that I would be writing styles closer to JS the first time I wrote JS that affected styles.

I'm personally not opposed to having a set of common SCSS files for each of our Kolibri plugins and in KDS to accommodate shared component styles overall but less so because of a SOC problem, but more so that our template logic can better just use class name conditionals rather than having to do it all in JS.

A complication is our theming system is all in JS - so anything that uses a theme token has to use it in JS somewhere (template or script).

Someday I will find a way to convert our theme definitions into a SCSS file and our lives will be better for it.

bjester: @nucleogenesis 
> The SFC is not the 'module' - it's just a nice way to look at what conceptually should be seen as 3 modules at once IMO.

I'm pretty sure that's what he's saying in the article, and is inline with what I'm thinking. He says:

> Now, anyone can argue that we are also already doing “modules” with SFCs. But...

His "but" goes on to ask some rhetorical questions and that leads to each of those softwares, HTML, JS, and CSS, being the modules we're directing change towards. So if I need to change the structure, I would go to the HTML module, design -> CSS module. When you starting mixing, you might as well be using JSX/React, which furthermore seems awkward when we're using Vue.js

lib/buttons-and-links/buttons.scss
```diff
@@ -41,7 +41,7 @@ a.button {
 }
 
 .link,
-.link .label > span > span {
```
nucleogenesis, line 44: We may need to add the `link-text` class to other components which use the `buttonClasses` mixin property like

https://github.com/learningequality/kolibri-design-system/blob/64f92dbeb39d92035ed4f4112928f35b95863bab/lib/buttons-and-links/KExternalLink.vue#L20

and 

https://github.com/learningequality/kolibri-design-system/blob/64f92dbeb39d92035ed4f4112928f35b95863bab/lib/buttons-and-links/KButton.vue#L25

lib/KBreadcrumbs.vue
indirectlylit: Yes, we've incurred a lot of extra complexity from the need to be able to dynamically re-style Kolibri _after_ the webpack bundles have been generated. If we could keep CSS out of JS that would be a big win.

This should actually be possible in Studio and KDP at least, where custom themes are not a requirement!

CSS is JS is a hacky way of emulating `scoped` styles, and one of the reasons it's so popular in JSX is because there aren't any other truly effective ways of limiting CSS leakage across components. I've always felt that `scoped` styles alone is the killer feature that makes me prefer Vue over React. Was a little surprised not to see it mentioned in the article.

At the end of the article, the author also mentions the new Vue 3 composition API. One thing I find interesting about the composition API is that it allows us to deprecate Vue mixins. Mixins add behavior to components from the outside, but they do it implicitly and it becomes difficult to track where these side effects are coming from. In this way, Vue mixins are very similar to global styles in the sense of having side effects implicitly imposed from outside the component.

PR #51, opened 2020-07-01
clear trailing slashes

indirectlylit:
netlify seems to add trailing slashes which don't appear locally


indirectlylit:
tested on netlify and works

PR #53, opened 2020-07-02
updating docs for icons

indirectlylit:
* Updates icon docs to use new icon definitions file
* Migrates logic related to clustering and sorting aliases to docs, outside of library code
* Makes the filter component re-usable and applies it to icons
* Fixes the 'external link' component to use new icons
* Updates mechanism for disambiguating icons by generating a unique hashed ID as a name

Demo online at https://deploy-preview-53--kolibri-design-system.netlify.app/icons/#icons

Added highlighting of current anchor:

I'm noticing some discrepancies around duplicate icons which I'm looking at now


jtamiace:
Looks great and the text filter is a nice touch!

For documentation, can we add something along the lines of this for custom icons:

## Custom icons

If the default Material icon library and/or community-contributed libraries are insufficient, use [Material's guidelines](https://material.io/design/iconography/system-icons.html) to create a new one.

indirectlylit:
> can we add something custom icons

thanks, done

Issue #22, opened 2020-04-09
Updates to flat button behaviors


jtamiace:
Based on a [Slack thread](https://learningequality.slack.com/archives/C0LKG14NL/p1586458645105100)

Issue summary: We have inconsistent ways of handling the alignment of flat button components in our projects. Should the text or component be aligned to a page container?

A couple of takeaways (copy pasted)

1. Add the rules “For flat buttons: align button text, not the button component to container content boundaries; in button rows, keep 8px margin between components” to general Button Patterns documentation
2. Add a new Boolean prop on flat buttons offset which, when set to true, adds -8px left and right margins, which will align the button based on text rather than outer boundary

Implications for 2 current PRs
https://github.com/learningequality/kolibri-design-system/pull/16
https://github.com/learningequality/kolibri-design-system/pull/21

PR #68, opened 2020-07-08
Button examples

nucleogenesis:
Implements examples for the Buttons based on @jtamiace 's initial documentation docs for buttons and the feedback in the comments.

Includes code changes surrounding tooltips on KIconButtons and `icon` prop alignment in KButton.

Fixes https://github.com/learningequality/kolibri-design-system/issues/69

### DIFF
--- a/docs/pages/buttons.vue
+++ b/docs/pages/buttons.vue
@@ -70,11 +70,13 @@
         <KButtonGroup style="margin-bottom: 8px;">
           <KButton text="Primary" :primary="true" appearance="raised-button" />
           <KButton text="Secondary" :primary="false" appearance="raised-button" />
+          <KIconButton tooltip="KIconButton" icon="plus" :primary="true" appearance="raised-button" />
         </KButtonGroup>
         <br>
         <KButtonGroup>
           <KButton text="Primary" :primary="true" appearance="flat-button" />
           <KButton text="Secondary" :primary="false" appearance="flat-button" />
+          <KIconButton tooltip="KIconButton" icon="plus" :primary="false" appearance="flat-button" />
         </KButtonGroup>
       </DocsShow>
 
--- a/docs/pages/buttons.vue
+++ b/docs/pages/buttons.vue
@@ -51,11 +51,13 @@
           hyperlink-like appearance for deemphasized actions, or actions inline within text
         </li>
       </ul>
-      <!-- TODO: add a flat secondary icon button to these examples below. See Google doc for example -->
       <DocsShow>
-        <KButton text="Raised button" :primary="true" appearance="raised-button" />
-        <KButton text="Flat button" :primary="true" appearance="flat-button" />
-        <KButton text="Basic link" :primary="true" appearance="basic-link" />
+        <KButtonGroup>
+          <KButton text="Raised button" :primary="true" appearance="raised-button" />
+          <KButton text="Flat button" :primary="true" appearance="flat-button" />
+          <KIconButton tooltip="A tooltip" text="Flat button" :primary="false" icon="plus" appearance="flat-button" />
+          <KButton text="Basic link" :primary="true" appearance="basic-link" />
+        </KButtonGroup>
       </DocsShow>
 
       <p>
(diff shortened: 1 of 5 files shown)


indirectlylit:
This looks good.

My only request is that we update the text here, because the premise that "There are 3 main appearances" is no longer accurate with the 4 styles shown below:

> image

jtamiace:
Looks great Jacob!

I can write some draft text to address Devon's comment - guessing that new style would just be "icon button"? I had originally left it out because I had considered it to just be "flat" since there's the icon+text flat button. But I guess icon buttons are visually distinct, since the border radius is completely round in contrast with the others.


And a couple things:

I realize this was misleading in the google doc, so my bad here. The gray button was meant to be the hover state and not a raised button - I think we're excluding raised icon buttons from the design system since we don't currently use them anywhere. So I think the change can be to just remove the raised button?
Screen Shot 2020-07-08 at 9 56 16 AM


And then here, can you move the icon about 1-2px down so that it's in line with the text? 
Screen Shot 2020-07-08 at 9 58 52 AM

jtamiace:
Some proposed text
Screen Shot 2020-07-08 at 10 51 28 AM

nucleogenesis:
@jtamiace @indirectlylit @khangmach 

I was going through the Issues and saw this one that I forgot about:

https://github.com/learningequality/kolibri-design-system/issues/22

Which says that we made decisions for the documentation and the buttons themselves.

1. Rule: "Flat-buttons should have their text aligned with things vertically"
2. Code: Add a flag to KButton so that the left (or right in RTL) margin is able to be toggled.

---

My questions:

- Are we still good with these decisions?
- Could someone come up with the right wording and placement of that button rule?
- Can the new code flag be pushed to our next release? It doesn't seem like it would have any meaningful effect for Kolibri development right now as we push to finish 0.14.

jtamiace:
I think that documentation bit could be placed under the `Placement` section here? Added some draft text as a suggestion

https://docs.google.com/document/d/1ullAljzj7w813xpn2t6DUYzRKBLQsg7W4k6ahASLv6k/edit#

lib/KDropdownMenu.vue
```diff
@@ -6,7 +6,9 @@
       ref="buttonContainer"
       class="button-container dib"
     >
+      <slot name="button"></slot>
       <KButton
+        v-if="!$slots.button"
```
indirectlylit, line 11: cool

docs/pages/buttons.vue
```diff
@@ -65,11 +67,15 @@
...
+          <KButton text="Primary" :primary="true" appearance="raised-button" />
+          <KButton text="Secondary" :primary="false" appearance="raised-button" />
+        </KButtonGroup>
         <br>
-        <KButton text="Primary" :primary="true" appearance="flat-button" />
-        <KButton text="Secondary" :primary="false" appearance="flat-button" />
+        <KButtonGroup>
+          <KButton text="Primary" :primary="true" appearance="flat-button" />
+          <KButton text="Secondary" :primary="false" appearance="flat-button" />
+        </KButtonGroup>
```
indirectlylit, line 78: could we add the icon button, e.g.:

image

docs/pages/buttons.vue
```diff
@@ -51,11 +51,13 @@
...
       </ul>
-      <!-- TODO: add a flat secondary icon button to these examples below. See Google doc for example -->
       <DocsShow>
-        <KButton text="Raised button" :primary="true" appearance="raised-button" />
-        <KButton text="Flat button" :primary="true" appearance="flat-button" />
-        <KButton text="Basic link" :primary="true" appearance="basic-link" />
+        <KButtonGroup>
+          <KButton text="Raised button" :primary="true" appearance="raised-button" />
+          <KButton text="Flat button" :primary="true" appearance="flat-button" />
+          <KIconButton tooltip="A tooltip" text="Flat button" :primary="false" icon="plus" appearance="flat-button" />
```
indirectlylit, line 58: @jtamiace @nucleogenesis realizing that 'primary' might not be implemented for `KIconButton`?

image

indirectlylit: it might be that in fact we don't want to create a primary state for icon buttons. I think that would be totally reasonable.

jtamiace: Yeah at the moment we don't currently use raised or primary icon buttons anywhere across all the apps

jtamiace: At least, not ones that we'd want to standardize

PR #80, opened 2020-07-22
update git logic

indirectlylit:
* specific links weren't working because some pages were defined in directories and others in vue files. now just link to top-level branch or PR
* better handling of local dev environments
* better handling of netlify environment


nucleogenesis:
Curious about what it will look like when merged into the release branch and deployed there since it says "pull/80" for the link text on the deploy link for this PR.

Links work as expected so as long as that's what you're expecting to see for the label in this context, then that's great.

indirectlylit:
> Curious about what it will look like when merged into the release branch and deployed there since it says "pull/80" for the link text on the deploy link for this PR.

Did some testing with netlify env vars, and as far as I can tell it will produce things like `v0.2.x` or `develop`. If that turns out to be incorrect, will revisit :)

indirectlylit:
looks good

image

looks bad

image


will follow up

Issue #12, opened 2020-03-19
cherry pick latest changes from kolibri repo


indirectlylit:
The following changes to shared Kolibri Components have been made recently in Kolibri `develop`. They need to be added here _before_ https://github.com/learningequality/kolibri/pull/6595 is merged in

* https://github.com/learningequality/kolibri/commit/a024934233efc5538a951b6aba79d429c850b5ee#diff-7d9c37dc9a600932d07e64be8ff7b11e
* https://github.com/learningequality/kolibri/commit/c037b6e73f2f93e1a8ba741242935f1b9fa049eb#diff-053a10dd1d40a8a0dabc76d4f84cb32a
* https://github.com/learningequality/kolibri/commit/af2c8fc252ecc05fe7af2e29b1321e6d1773d8bc#diff-053a10dd1d40a8a0dabc76d4f84cb32a
* https://github.com/learningequality/kolibri/pull/6701/commits/e0b154ec901a8945eaa52dbfb5aa2ed5bddbd799
* https://github.com/learningequality/kolibri/pull/6660/files#diff-2425a6167bea4c6f2460022a0cb7a512
* https://github.com/learningequality/kolibri/pull/6830

indirectlylit:
One thing to watch out for from https://github.com/learningequality/kolibri/commit/e0b154ec901a8945eaa52dbfb5aa2ed5bddbd799

The theming currently doesn't work for icons in links - these should be the same color:

![image](https://user-images.githubusercontent.com/2367265/78411982-a1fe6680-75c6-11ea-8014-b824be2dfad5.png)

(cc @nucleogenesis)

nucleogenesis:
@indirectlylit do you recall if any of the merges referencing this issue close it?

indirectlylit:
no, I don't believe so. I can take a look at this tomorrow

PR #87, opened 2020-08-04
cherry picks changes from Kolibri

indirectlylit:
fixes https://github.com/learningequality/kolibri-design-system/issues/12

### DIFF
--- a/lib/KModal.vue
+++ b/lib/KModal.vue
@@ -291,23 +291,23 @@
         this.$refs.modal.focus();
       },
       focusElementTest(event) {
-        // switching apps - not relevant
-        if (event.target === window) {
+        const { target } = event;
+        const noopOnFocus =
+          target === window || // switching apps
+          !this.$refs.modal || // if $refs.modal isn't available
+          target === this.$refs.modal || // addresses #3824
+          this.$refs.modal.contains(target.activeElement);
+        if (noopOnFocus) {
           return;
         }
-        // not sure when this would be true
-        if (!this.$refs.modal) {
-          return;
-        }
-        // addresses https://github.com/learningequality/kolibri/issues/3824
-        if (
-          event.target === this.$refs.modal ||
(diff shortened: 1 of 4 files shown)


lib/KIcon/iconDefinitions.js
```diff
@@ -129,6 +129,9 @@ const KolibriIcons = {
   quiz: {
     icon: require('./precompiled-icons/mdi/file-document-edit-outline.vue').default,
   },
+  exam: {
```
nucleogenesis, line 132: I think that I removed this altogether based on suggestion from @jtamiace - are there any missing icons in Kolibri that were found?

Whenever I removed a key from this file, I made corresponding changes in Kolibri (or tried to - if there are things trying to use the `exam` key still)

lib/KModal.vue
```diff
@@ -329,7 +329,10 @@
     position: fixed;
     top: 0;
     left: 0;
-    z-index: 24;
```
nucleogenesis, line 332: Ah - I fixed this differently here: https://github.com/learningequality/kolibri/pull/7360/files in `CoreSnackbar.index.vue` by making the z-index of that to 24 (which seemed right for Material guidelines since 24 is the highest it expects).

Kind of still hacky - but, by design, we load the snackbar component very low in the DOM - which means it shows over sibling elements and anything inside of them that has the same z-index.

Note that this change won't break anything, but if we want it one less than the snackbar, it should be 23.

indirectlylit: okay great - that's the way it should be. 

ref https://github.com/learningequality/kolibri/pull/6618/files#r387355934

indirectlylit: will revert my change

lib/KIcon/iconDefinitions.js
indirectlylit: okay awesome, will remove

indirectlylit:
updated

Issue #82, opened 2020-07-22
KButton, KExternalLink, KRouterLink should look identical with identical inputs


nucleogenesis:
I think that, perhaps, the only difference between them ought to really just be how we handle events.

- KRouterLink takes a `:to` route which must be a valid router route.
- KExternalLink takes a `download` prop (among others) to manage external URL links
- KButton goes "click" if you click it and you can do what you want with that information.

All of them should have:

- Primary or no?
- Appearance `basic-link` `raised-button` `flat-button`
- `icon` and `iconAfter` props & slots w/ a consistent sizing/styling
- `disabled` prop

And I'm suggesting that If any of the components receive the same values for these props, they should look 100% identical. The only difference being how they act.

indirectlylit:
yes please

sairina:
Closing this issue, as it is continued in #166

PR #91, opened 2020-08-06
Fix display for `KLabeledIcon` with `iconAfter` & overflowing text

micahscopes:
## Summary

Using a `KLabeledIcon` component with the `iconAfter` property and label or slot text that overflows the maximum width of the `KLabeledIcon` causes the icon to overlap the text.

## Screenshots
Before:
image

After:
image

### DIFF
--- a/lib/KLabeledIcon.vue
+++ b/lib/KLabeledIcon.vue
@@ -125,6 +125,9 @@
 
   .label {
     display: inline-block;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    vertical-align: top;
   }
 
   .debug-warning > svg {


nucleogenesis (changes requested):
Hey Micah! Sorry for such a lengthy time since PR creation to review - but I went in and tested this and it gave some undesirable results. 

Part of this is due to how we set `max-width` but not `width` on the element. We cannot say `width: 100%` because then any `iconAfter` will be pushed too far right relative to the text.

I'm not sure of the best approach to resolving this but I suspect it will require some significant refactoring of the component - which may also do well to be done in line with addressing  https://github.com/learningequality/kolibri-design-system/issues/82

aggressive-overflow-ellipsis

sairina:
Hi all, I'd like to close this PR because it does seem like making sure #82 is squared away will help here. Is there any opposition to closing this one (and if fixing #82 doesn't actually fix this issue, to reopen this)?

indirectlylit:
makes sense thanks

PR #95, opened 2020-08-26
Add design principles

jtamiace:


docs/pages/principles.vue
```diff
@@ -0,0 +1,51 @@
+<template>
+
+  <DocsPageTemplate>
+
+    <DocsPageSection title="" anchor="">
```
indirectlylit, line 5: Could this have simply been 

```html
<DocsPageSection>
```

Alternatively, was the section necessary at all?

jtamiace: I was getting linting errors without adding that part in

indirectlylit: okay thanks - opened a follow-up issue: https://github.com/learningequality/kolibri-design-system/issues/96

indirectlylit: FYI, this is addressed in https://github.com/learningequality/kolibri-design-system/pull/114

You can now have a `DocsPageSection` with no title and anchor

Issue #3, opened 2020-02-25
Firefox link issues


indirectlylit:
Side nav isn't highlighted:

![image](https://user-images.githubusercontent.com/2367265/75276744-99db2c00-57bb-11ea-906e-790ba9a378c1.png)

The header anchor links don't work in firefox. For some reason the first click goes to the top of the page, and the second click goes to the actual anchor:

![2020-02-25 10 40 26](https://user-images.githubusercontent.com/2367265/75276763-a2cbfd80-57bb-11ea-9cc5-c461869386ce.gif)

PR #112, opened 2020-09-23
remove router override behavior

indirectlylit:
fixes https://github.com/learningequality/kolibri-design-system/issues/3

removes override behavior which was disrupting post-load JS routing behavior


indirectlylit:
self-merge

PR #115, opened 2020-09-23
update link usage

indirectlylit:
This PR changes nearly all instances of internal and external links to use props instead of slots, from this form:

```html
Follow the <DocsExternalLink href="https://material.io/design">
  Material design guidelines
</DocsExternalLink>.
```

to this form

```html
Follow the <DocsExternalLink text="Material design guidelines" href="https://material.io/design" />.
```

This has two advantages:

A) the previous form causes undesirable underline behaviors:

| before | after |
|--|--|
| image  | image |

B) The code is more concise.

Note while possibly desirable for consistency with `<a href="...">...</a>`, it is not possible to write this form:

```html
Follow the <DocsExternalLink href="https://material.io/design">Material design guidelines</DocsExternalLink>.
```

It appears that Vue-ESLint [always inserts additional whitespace](https://stackoverflow.com/questions/60286109/vue-cli-eslint-injects-additional-whitespace-into-multi-line-spans), and I can't determine any way of disabling this behavior.

### DIFF
--- a/docs/common/DocsColorBlock.vue
+++ b/docs/common/DocsColorBlock.vue
@@ -6,9 +6,12 @@
       <code>{{ name }}</code><DocsAnchorTarget v-if="definition" :anchor="anchor" />
     </div>
     <div class="code value">
-      <DocsInternalLink v-if="isToken && showTokenCrossLink" :href="tokenAnchor">
-        <code>{{ tokenSource }}</code>
-      </DocsInternalLink>
+      <DocsInternalLink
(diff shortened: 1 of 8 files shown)


indirectlylit:
also relevant - here is how we eslint from messing up whitespace internally within the component:

https://github.com/learningequality/kolibri-design-system/blob/3e56c3f52561715f2801a46333cbc75e880f21bb/docs/common/DocsExternalLink.vue#L3-L6

PR #116, opened 2020-09-24
versions

indirectlylit:
thoughts on this as a structure for version tracking?

image


nucleogenesis (commented):
I love the style/approach here. 

A few thoughts/questions:

- When a user selects a previous version - I think we should explicitly say what version of the docs they're looking at. Perhaps we can put the version somewhere in the sidebar or another prominent place?
- I like the changelog style page for a historical reference, but for being the interface by which a user finds a specific version of the docs I think it's lacking. It's easy to miss a link called "Version History" in a list of similar links.  I'd love if there was a dropdown or some other persistently visible "Version you're on:" indicator that would then provide either a link to a page where you select a different version or maybe a dropdown or other way to select.

indirectlylit:
For both comments, my hope was that the information in the header should serve that purpose.  The text which says "Local environment" below would show the current release in tagged versions:

image

See the current behavior at:

https://kolibri-design-system.netlify.app/ 

image

Then the addition in this PR is the "version history" link to the right.

Thoughts? Happy to try and make these header links more prominent

indirectlylit:
Ready for re-review. Now includes github links:

indirectlylit:
polite bump on review - @jtamiace @nucleogenesis

nucleogenesis (approved):
This is excellent! @indirectlylit

jtamiace (approved):

PR #119, opened 2020-09-28
Changelog

nucleogenesis:
Adds a changelog and some hopefully hidden comments on how/when to use it.

Would definitely appreciate additional input on the changelog comments about what to detail and when to make use of it.

Or - if you have a suggestion for how better to format it, happy to adjust.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,3 @@
+# 28 September 2020
+
+  - Added changelog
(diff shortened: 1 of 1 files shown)


indirectlylit:
What do you think of tracking the changelog as an actual page, e.g. https://github.com/learningequality/kolibri-design-system/pull/116

nucleogenesis:
Closing in favor of #116

PR #122, opened 2020-10-15
Fix invalid property type check

MisRob:
Fixes

```
Invalid prop: type check failed for prop "zIndex". Expected Number with value 24, got String with value "24".
```


MisRob:
@nucleogenesis 

We'll need to merge this to get rid of Studio errors and update version there again:

Selection_001

PR #134, opened 2020-11-17
adding glossary to design system

indirectlylit:
Adds glossary based on Crowdin tbx file to design system

image


radinamatic:
Thank you for pushing on this! 

I'll have to go back and check my notes from a few years ago, before we made a first glossary, I remember for some reason not being happiest with the TBX.

indirectlylit:
I believe TBX is the only format Crowdin offers downloads in

https://support.crowdin.com/api/download-glossary/

radinamatic:
No, it offers download and upload as CSV & XLSX too, and we've been using the latter as it allows for easier import and share as a GDoc for translation.

Selection_100

nucleogenesis:
@indirectlylit 

I think that the table's two left-most columns look squished. This may be a more of a critique of the design system layout itself though - perhaps a max-width of 900 or so might look better (the grid layout of the columns could then be adjusted so at least the left-most column has more breathing room).

I'd be curious to see how that might affect other parts of the app's appearance though. It's not a big deal honestly just looked off to me a bit.

---

A couple thoughts:

- I wonder if it would be easy to add a filtering search bar like we have for the icons
- How often will we need to update the glossary? Will we just - when needed - download the latest glossary? I wonder if we could securely load the XML dynamically (server side?) and just always use the latest TBX file that way.

indirectlylit:
> No, it offers download and upload as CSV & XLSX

Okay - hopefully the API supports it also.  I had these functions written for import/export, but it could also be done as a manual UI process

https://github.com/learningequality/kolibri/blob/68f33c3f429f55c60af23d6752ba389a6ac20a0d/build_tools/i18n/crowdin.py#L451-L477

> I wonder if it would be easy to add a filtering search bar like we have for the icons

quite easy

> How often will we need to update the glossary? 

When we first created it, we discussed updating it once per translation push on Kolibri (once per major release)

indirectlylit:
@nucleogenesis added filter and widened overall table and left column:

indirectlylit:
@radinamatic @nucleogenesis we should decide what the process for managing the glossary "source of truth" should be.

See the discussion in this PR to refresh your memory about work and stop-gap decisions made previously:

https://github.com/learningequality/kolibri/pull/6019

nucleogenesis (approved):
LGTM!

nucleogenesis:
>  we should decide what the process for managing the glossary "source of truth" should be.

@indirectlylit @radinamatic this PR looks good to me - but I think we should merge this and discuss that process in a quick meeting or on Slack.

Issue #123, opened 2020-10-19
KExternalLink should have an open_in_new icon


jayoshih:
Links that open a page in a new tab need an icon (also note that icon needs to flip for rtl)

What it looks like now
![image](https://user-images.githubusercontent.com/7447496/96492914-2b8b2400-11f9-11eb-873d-98260a8f13e9.png)

What it should look like
![image](https://user-images.githubusercontent.com/7447496/96493037-4e1d3d00-11f9-11eb-80cc-62faf0641f1a.png)

PR #137, opened 2020-12-01
Add boolean prop, 'openInNewTab' to KExternalLink

### LINKED ISSUE #94: KExternalLink ought to open in new tab safely
We removed the `target="_blank" rel="noopener noreferrer"` due to https://github.com/learningequality/kolibri/issues/7480

However, `KExternalLink` ought to have a boolean prop `openInNewTab` when true will apply the proper attrs.

It also should be tested in Android and other special-build versions of Kolibri designed to run in a web view.

First, this must be finished: https://github.com/learningequality/kolibri/issues/7482 because Kolibri uses `KExternalLink` for internal routing at time of writing.

sairina:
**Summary:**
- Added boolean prop `openInNewTab` to `KExternalLink`
- Added `:target` to anchor tag in `<template>` to open up new tab if `openInNewTab` is given `true`

Fixes #94 

**Testing:**
- Tested with Chrome on MacOSX, but not yet with other versions of Kolibri.
Screen-Recording-2020-11-30-at-7 28 34-PM

### DIFF
--- a/lib/buttons-and-links/KExternalLink.vue
+++ b/lib/buttons-and-links/KExternalLink.vue
@@ -5,6 +5,8 @@
     :class="buttonClasses"
     :href="href"
     :download="download"
+    :openInNewTab="openInNewTab"
+    :target="openInNewTab ? '_blank' : false"
     dir="auto"
     @mouseenter="hovering = true"
     @mouseleave="hovering = false"
@@ -68,6 +70,13 @@
        */
       iconAfter: {
         type: String,
+        required: false,
+      },
+      /**
+       * If provided, opens link in new tab
+       */
+      openInNewTab: {
+        type: Boolean,
         required: false,
       },
     },
--- a/lib/buttons-and-links/KExternalLink.vue
+++ b/lib/buttons-and-links/KExternalLink.vue
@@ -77,7 +77,7 @@
        */
       openInNewTab: {
         type: Boolean,
-        required: false,
+        default: false,
       },
     },
     data() {


lib/buttons-and-links/KExternalLink.vue
```diff
@@ -70,6 +72,13 @@
         type: String,
         required: false,
       },
+      /**
+       * If provided, opens link in new tab
+       */
+      openInNewTab: {
+        type: Boolean,
+        required: false,
+      },
```
indirectlylit, line 81:
```suggestion
      openInNewTab: {
        type: Boolean,
        default: false,
      },
```

Generally in prop definitions you should provide one of either `required: true`, or provide a `default`.

Providing a `default` value ensures the prop value is well-defined when nothing is provided from outside, and makes setting `required: false` superfluous.

docs/rstIconReplacements.txt
```diff
@@ -11,7 +11,7 @@
...
 .. |clear| replace:: :raw-html:`<span class="design-system-icon"><svg viewBox="0 0 24 24" role="presentation" focusable="false" xmlns="http://www.w3.org/2000/svg" width="24" height="24"><path d="M19 6.41L17.59 5 12 10.59 6.41 5 5 6.41 10.59 12 5 17.59 6.41 19 12 13.41 17.59 19 19 17.59 13.41 12z"/></svg></span>`
-.. |clipboard| replace:: :raw-html:`<span class="design-system-icon"><svg viewBox="0 0 24 24" role="presentation" focusable="false" width="24" height="24" xmlns="http://www.w3.org/2000/svg"><path d="M19 2h-4.18C14.4.84 13.3 0 12 0c-1.3 0-2.4.84-2.82 2H5c-1.1 0-2 .9-2 2v16c0 1.1.9 2 2 2h14c1.1 0 2-.9 2-2V4c0-1.1-.9-2-2-2zm-7 0c.55 0 1 .45 1 1s-.45 1-1 1-1-.45-1-1 .45-1 1-1z"/></svg></span>`
+.. |clipboard| replace:: :raw-html:`<span class="design-system-icon"><svg viewBox="0 0 24 24" role="presentation" focusable="false" width="24" height="24" xmlns="http://www.w3.org/2000/svg"><path d="M19 2h-4.18C14.4.84 13.3 0 12 0c-1.3 0-2.4.84-2.82 2H5c-1.1 0-2 .9-2 2v16c0 1.1.9 2 2 2h14c1.1 0 2-.9 2-2V4c0-1.1-.9-2-2-2zm-7 0c.55 0 1 .45 1 1s-.45 1-1 1-1-.45-1-1 .45-1 1-1zm7 18H5V4h2v3h10V4h2v16z"/></svg></span>`
```
indirectlylit, line 14: @sairina this change wasn't technically part of your PR

@nucleogenesis I think we may not have committed this regenerated file after updating the clipboard icon. We should figure out how to ensure that this gets committed - I could imagine a linter rule for example

indirectlylit: opened a new PR to catch this in the future:

https://github.com/learningequality/kolibri-design-system/pull/138

@sairina you can revert the change to this file

indirectlylit:
This looks good, thanks!

It wasn't in the scope of the original issue, but I'd like to suggest that while we're doing this work, we include an iconographic representation of the "new tab" behavior. We do something similar in the design system itself:

image

thoughts?

nucleogenesis:
@indirectlylit There is an issue for that https://github.com/learningequality/kolibri-design-system/issues/123 - but the one thing not noted in that issue is that we want it to be conditional upon a prop - and we probably want to default it to not showing.

@sairina maybe we could add the icon. You'll find the name of the icon here: https://kolibri-design-system.netlify.app/icons/#icons that you'll want to pass to the `:icon` prop of `<KIcon>`

Additionally - we'll want to make sure that it displays with the correct color and on the right side of the link (just like in https://github.com/learningequality/kolibri-design-system/issues/123) - but when we're RTL it should show on the left side.

nucleogenesis (changes requested):
We'll want to also add `rel="noopener noreferrer"` attr to that `<a>` tag for security purposes

https://mathiasbynens.github.io/rel-noopener/ has a great breakdown of why we need to add this to external links.

sairina:
@nucleogenesis Thanks for that explanation! I'll add that in.
@indirectlylit I'll work on adding the icon!

PR #139, opened 2020-12-02
Material background on KTextbox

### LINKED ISSUE #124: KTextbox needs box styling instead of underline
According to the [documentation](https://docs.google.com/document/d/1w6v-CGAM72hg5QncT913koxZQLO-n8e7_PCibBc6d-c/edit#), text fields should use box styling. This also applies to selects and text areas

image

### LINKED ISSUE #167: upgrade text boxes to new Material style
### Observed behavior

We're currently using the old Material style of textboxes. In addition to having [proven UX issues](https://link.medium.com/hDuR66wnq1), it also leads to inconsistent vertical rhythm:

| image |   image |
|--|--|
| image |image |

### Expected behavior

upgrade to the new style:

image

### User-facing consequences

According to the study, this has improvements in three dimensions:

* Identifiability: The number of correct versus incorrect clicks
* Find-ability / scan-ability: The time it took a participant to find and click on a requested element
* Preference: Participants rank-ordered each text field variation

### Context

0.12.x

nucleogenesis:
This is a draft for now - this should not be merged until KSelect is also available in KDS and has the same or similar styles applied.

Fixes #124
Fixes #167 

- Gray #e9e9e9 background on KTextbox
- CSS changes to align things roughly like the sample field https://material.io/components/text-fields
- Changed the value given to the `translate()` CSS function to ensure the label is centered (original value pushed the label to the baseline)
- Add a `&nbsp;` empty div fallback for the part of the DOM flow where the error/help message would show. A counter shows on the same horizontal line but is `position: absolute;` so when there is no help/error text, subsequent elements are closer to the previous one than they ought to be and they look squished. 

I took this first screenshot w/ a delay feature on my screenshot extension so I could properly focus the field to show that the theme colors work still.

screenshot-localhost_8000-2020 12 02-13_55_14

These are just a few more examples with error messages, text, no text.+

screenshot-localhost_8000-2020 12 02-13_50_31
screenshot-localhost_8000-2020 12 02-13_52_40
screenshot-localhost_8000-2020 12 02-13_52_59

### DIFF
--- a/lib/keen/UiTextbox.vue
+++ b/lib/keen/UiTextbox.vue
@@ -103,6 +103,12 @@
           </slot>
         </div>
 
+        <!-- Placeholder to keep the text height spacing in place even when
+          not showing any errors -->
+        <div v-else class="ui-textbox-feedback-text">
+          &nbsp;
+        </div>
+
         <div v-if="maxlength" class="ui-textbox-counter">
           {{ valueLength + '/' + maxlength }}
         </div>
@@ -433,7 +439,8 @@
         &.is-inline {
           color: $ui-input-label-color; // So the hover styles don't override it
           cursor: text;
-          transform: translateY($ui-input-label-top--inline) scale(1.1);
+          // 1em here is custom to keep text centered
+          transform: translateY(1em) scale(1.1);
         }
 
         &.is-floating {
@@ -490,8 +497,10 @@
     display: flex;
     flex-direction: column-reverse;
     width: 100%;
-    padding: 0;
+    padding: 4px 0 0 0;
     margin: 0;
+    background: #e9e9e9;
+    border-radius: 4px 4px 0 0;
   }
 
   .ui-textbox-icon-wrapper {
@@ -509,7 +518,7 @@
   }
 
   .ui-textbox-label-text {
-    margin-bottom: $ui-input-label-margin-bottom;
+    margin: 0 8px 0;
     font-size: $ui-input-label-font-size;
     line-height: $ui-input-label-line-height;
     color: $ui-input-label-color;
@@ -522,7 +531,7 @@
   .ui-textbox-textarea {
     display: block;
     width: 100%;
-    padding: 0;
+    padding: 0 0 0 8px;
     margin: 0;
     font-size: $ui-input-text-font-size;
     font-weight: normal;
--- a/lib/keen/UiTextbox.vue
+++ b/lib/keen/UiTextbox.vue
@@ -36,7 +36,7 @@
           :readonly="readonly"
           :required="required"
           :step="stepValue"
-          :style="isActive ? { borderBottomColor: $themeTokens.primary } : {}"
+          :style="isActive ? { borderBottomColor: $themeTokens.primaryDark } : {}"
           :tabindex="tabindex"
           :type="type"
           :value="value"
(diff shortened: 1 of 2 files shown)


nucleogenesis:
I don't recall why this is a draft "Until we get KSelect" in. I actually am okay now with merging it. The fact that KSelect will still be without the background (like in the image below) isn't super offensive compared to the images included at the top.

screenshot-localhost_8000-2021 03 03-13_56_03

@indirectlylit @jtamiace do you agree?

jtamiace:
I'm interpreting that the main benefit of waiting would be to have complete consistency once this background color change is merged in? 

I guess my question is how prevalent KSelect is. If like half of them would end up with different backgrounds, may be worth waiting. If there are only a few KSelects, probably ok?

nucleogenesis:
@jtamiace I don't think it's very prevalent. The only places I can think of are in table filters and a couple of device and/or facility settings.

jtamiace:
Cool. It's a 👍 for me then!

indirectlylit:
same here

radinamatic:
Can't but notice that if we implement the text field background, **contrast** becomes insufficient for labels and error message placeholders.

Label | Error
--- | ---
 |

indirectlylit:
@radinamatic can you give us the darkest compliant grey value?

radinamatic:
With the red 700 (`#d32f2f`) foreground color, darkest compliant grey value for the background would be `#F5F5F5`, which is already the main background color.



Our main purple was already on the edge, so the only compliant background color would be `#FAFAFA`, just a notch off white 😕 

fafafa

(_Crazy idea_) 
How about we make the text fields background color full white...? 😈

indirectlylit:
>  How about we make the text fields background color full white...? 😈

@radinamatic for the context of this change see https://github.com/learningequality/kolibri-design-system/issues/167

jtamiace:
We could also choose a darker purple from our palette that's compliant with `#F5F5F5`?

nucleogenesis:
@radinamatic @jtamiace @indirectlylit I've updated to use more accessible colors - but I did not change the background to a lighter grey for reasons to be mentioned below:

| WebAIM | Color | After (Before can be seen above or at https://kolibribeta.learningequality.org) |
| --- | --- | --- |
| errortoken | [Primary Dark](https://kolibri-design-system.netlify.app/colors/#tokens-primaryDark) |image |
| screenshot-webaim org-2021 03 18-09_25_30 | [Error](https://kolibri-design-system.netlify.app/colors/#palette-red-v_700) | token-error





So - I darkened the purple to `primaryDark` which is improved - but using `#f5f5f5` can have product implications because that's the color in Kolibri we use for the background of the sign in page - so you don't actually see the background that we're trying to implement.

An alternative thought is that we could change the [themeTokens.error](https://kolibri-design-system.netlify.app/colors/#tokens-error) to use the slightly darker red that [themeTokens.incorrect](https://kolibri-design-system.netlify.app/colors/#tokens-incorrect) uses, which passes the WebAIM AA test (but not the AAA):

screenshot-webaim org-2021 03 18-09_25_44

jtamiace (approved):
Thanks for doing this exploration @nucleogenesis!

indirectlylit:
> we could change the `themeTokens.error` to use the slightly darker red that `themeTokens.incorrect` uses

I chose these tokens somewhat arbitrarily – there was not an intentional design decision to use different red shades for the two meanings.

nucleogenesis:
Thanks @indirectlylit - I've changed the theme to use the darker red now - will await an approval from @radinamatic and if she's good with this then will merge

radinamatic (approved):
`primaryDark` is looking much better!!! 💯 

Thank you, Jacob! 👍🏽

PR #140, opened 2020-12-08
move material icons from normal to dev dependency and update readme

indirectlylit:
This should make KDS 30 MB smaller for users, e.g. Kolibri, Studio, KDP

ref: https://learningequality.slack.com/archives/CTG8XDTCL/p1607447862001200

cc @nucleogenesis @jonboiser


indirectlylit:
Actually, some of the other dependencies also look like dev. For example:

* `svgo`
* `@mdi/svg`?
* `@babel/core` and `@babel/preset-env`?
* also, where do we use `fuzzysearch`?

nucleogenesis:
`fuzzysearch` is used by Keen's UIAutocomplete component.

The others ought to be good added to `devDependencies` though.

indirectlylit:
thanks @nucleogenesis - moved those dependencies also.

Ready for review & merge

nucleogenesis (approved):
Thanks @indirectlylit !

PR #142, opened 2020-12-24
Update <KExternalLink> with icon and noopener noreferrer attr

sairina:
### Summary
Background: `openInNewTab` attr was recently added to `<KExternalLink>` so that a value of `true` would allow for the link to be opened in another tab. Previously, `<KExternalLink>` used `target="_blank"` and did not have an icon to differentiate it as an external link as opposed to a link to another URL internally.

### In this PR
1. The `noopener noreferrer` attr was also added as an extra security measure to ensure `window.opener = null` because `target="_blank"` is used in the component.
2. The ["open in new tab" icon](https://kolibri-design-system.netlify.app/icons/#token-openNewTab) was added.

#### Issues:
Addresses #123 
Addresses #137 
Addresses #94

### Expected Behavior
Use of this component should:
1. Allow for an external link to be opened safely in a new tab
2. The "open in new tab" icon should open correctly depending on the language - see below.
-- to the **Left** of the link when link is written in RTL languages (e.g. Arabic)
-- to the **Right** of the link when link is written in LTR languages (e.g. English)

### RTL/LTR Screenshots
| Icon on Left for Arabic link (Arabic) | Icon on Right for English link (English) |
|--|--|
|Screen Shot 2020-12-23 at 10 22 53 PM| Screen Shot 2020-12-23 at 10 23 12 PM

| Icon on Right for English link (Arabic) | Icon on Right for English link (English) |
|--|--|
Screen Shot 2020-12-23 at 10 23 57 PM| Screen Shot 2020-12-23 at 10 23 34 PM
- the screenshot above is for demonstration purposes of `<KExternalLink>`; from my understanding the above should actually be a use of the `<KRouterLink>` though in Kolibri, it's currently a `<KExternalLink>` and could potentially need to be fixed (see: learningequality/kolibri#7482)

**Follow-up Notes**
- There are still places in Kolibri that still use `target=_blank`, that should be updated.
- https://github.com/learningequality/kolibri/issues/7712 can continue


indirectlylit:
Looking great, and thanks for testing the RTL behaviors!

Two small notes:

* Should probably have a few more pixels of margin between the icon and the text so it feels less cramped
* When inline with an RTL-language link, the icon should be flipped so it points up and to the left. This can be achieved by adding `rtlFlip: true` to [iconDefinitions.js](https://github.com/learningequality/kolibri-design-system/blob/v0.2.x/lib/KIcon/iconDefinitions.js)

lib/buttons-and-links/KExternalLink.vue
```diff
@@ -64,11 +59,11 @@
         required: false,
       },
       /**
-       * If provided, shows a KIcon after the text
+       * If provided, opens link in new tab
```
indirectlylit, line 62:
```suggestion
       * If provided, opens link in new tab and displays a "pop out" icon
```

sairina:
Updated screenshots:
| Margin added (Arabic) | Margin added (English) |
|--|--|
|Screen Shot 2020-12-24 at 8 35 23 PM| Screen Shot 2020-12-24 at 8 35 42 PM|

lib/buttons-and-links/KExternalLink.vue
```diff
@@ -78,18 +66,18 @@
     },
     computed: {
       /**
-       * If icon is provided, add 8px margin between the icon and the text
+       * If link opens in new tab, add 8px margin between the icon and the text
        */
       spanStyle() {
-        if (this.icon) {
```
nucleogenesis, line 84: This bit of the logic needs to be applied to your changes to avoid the margin showing when we don't have an icon. During testing I see this on the Usage & Privacy on Kolibri (you can find the Usage & Privacy modal clicking the link at the bottom of the main sidebar).

image

nucleogenesis (changes requested):
This is looking good except one thing I noted. I tested on Studio as well and it looked great! Thanks @sairina

sairina:
**Updated so that margin no longer shows when there is no icon:** `this.icon` was not defined in my code, so I used `this.openInNewTab` to check for whether or not the icon existed (since we only have the icon when `openInNewTab` is `true`.

|Arabic (No icon) |English (No icon)|
|--|--|
|Screen Shot 2021-01-04 at 2 49 34 PM|Screen Shot 2021-01-04 at 2 49 59 PM|

|Arabic (With icon) |English (With icon)|
|--|--|
|Screen Shot 2021-01-04 at 2 48 44 PM|Screen Shot 2021-01-04 at 2 50 27 PM|

sairina:
@nucleogenesis I also noticed, however that in Arabic, for the Update Notification modal (`UpdateNotification.vue`), the external link icon appears on the left side of the link (as opposed to the right side in, say, the privacy modal. This seems only to be an issue for `KExternalLink` for this particular modal.

Screen Shot 2021-01-04 at 2 51 20 PM

nucleogenesis (approved):
Great work! Thanks @sairina !

sairina:
@nucleogenesis - it's all fixed now. Here is a summary of what was fixed:

### In either LTR or RTL languages, the direction of the icon matches the language of the text of the link
|English text in English|Arabic text in Arabic|
|--|--|
|Screen Shot 2021-01-05 at 9 57 34 AM|Screen Shot 2021-01-05 at 1 35 25 PM|

|Now: English text in Arabic|(Previously: English text in RTL)|
|--|--|
|Screen Shot 2021-01-05 at 1 06 08 PM|Screen Shot 2021-01-05 at 9 56 58 AM|

### If the user does not choose to open in a new tab, then the link has no left margin.
|No margins, no icon - English|No margins, no icon - Arabic|
|--|--|
|Screen Shot 2021-01-05 at 9 53 19 AM|Screen Shot 2021-01-05 at 9 53 48 AM|

### If the user uses `openInNewTab`, icon appears correctly and has no left margin.
|No margins, with icon - English|No margins, with icon - Arabic|
|--|--|
|Screen Shot 2021-01-05 at 9 51 56 AM|Screen Shot 2021-01-05 at 9 52 28 AM|

PR #143, opened 2021-01-05
<KExternalLink> final fixes for icons and margins

sairina:
@nucleogenesis - it's all fixed now! Here is a summary of what was updated:

### In either LTR or RTL languages, the direction of the icon matches the language of the text of the link
|English text in English|Arabic text in Arabic|
|--|--|
|Screen Shot 2021-01-05 at 9 57 34 AM|Screen Shot 2021-01-05 at 1 35 25 PM|

|Now: English text in Arabic|(Previously: English text in RTL)|
|--|--|
|Screen Shot 2021-01-05 at 1 06 08 PM|Screen Shot 2021-01-05 at 9 56 58 AM|

### If the user does not choose to open in a new tab, then the link has no left margin.
|No margins, no icon - English|No margins, no icon - Arabic|
|--|--|
|Screen Shot 2021-01-05 at 9 53 19 AM|Screen Shot 2021-01-05 at 9 53 48 AM|

### If the user uses `openInNewTab`, icon appears correctly and has no left margin.
|No margins, with icon - English|No margins, with icon - Arabic|
|--|--|
|Screen Shot 2021-01-05 at 9 51 56 AM|Screen Shot 2021-01-05 at 9 52 28 AM|

### DIFF
--- a/lib/buttons-and-links/KExternalLink.vue
+++ b/lib/buttons-and-links/KExternalLink.vue
@@ -17,7 +17,7 @@
       <KIcon
         v-if="openInNewTab"
         icon="openNewTab"
-        :style="iconStyle"
+        style="top: 4px;"
         :color="hovering ? $themeTokens.primaryDark : $themeTokens.primary"
       />
     </slot>
@@ -80,16 +80,6 @@
           } else {
             styles['marginRight'] = '8px';
           }
-        }
-        return { ...styles };
-      },
-      /**
-       * Changes icon direction according to language of text of URL
-       */
-      iconStyle() {
-        let styles = { top: '4px' };
-        if (this.text === this.href) {
-          styles['transform'] = 'scaleX(1)';
         }
         return { ...styles };
       },
--- a/lib/buttons-and-links/KExternalLink.vue
+++ b/lib/buttons-and-links/KExternalLink.vue
@@ -12,7 +12,23 @@
     @mouseenter="hovering = true"
     @mouseleave="hovering = false"
   >
+    <slot name="icon">
+      <KIcon
+        v-if="icon"
+        :icon="icon"
+        style="top: 4px;"
+        :color="hovering ? $themeTokens.primaryDark : $themeTokens.primary"
+      />
+    </slot>
     <span class="link-text" :style="spanStyle">{{ text }}</span>
+    <slot name="iconAfter">
+      <KIcon
+        v-if="iconAfter"
+        :icon="iconAfter"
+        style="top: 4px;"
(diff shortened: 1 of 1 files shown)


lib/buttons-and-links/KExternalLink.vue
```diff
@@ -72,19 +72,29 @@
...
           }
         }
         return { ...styles };
       },
+      /**
+       * Changes icon direction according to language of text of URL
+       */
+      iconStyle() {
+        let styles = { top: '4px' };
+        if (this.text === this.href) {
```
nucleogenesis, line 91: I think that we can remove this `iconStyle` computed property although I totally see why you added it.

I switched my Firefox to Arabic to confirm this, but people who are viewing RTL content would probably just expect the icon's arrow always to point left because that's where their new tab would open (note where the + is along the tabs at the top):

image

So even if the link's text is English or another LTR language, if the page is displayed in RTL (aka `this.isRtl === true`) then the icon should always be flipped as if the text associated is RTL even if it looks kinda funny next to some LTR text.

---

As for the margins, that looks good to me!

nucleogenesis (changes requested):

lib/buttons-and-links/KExternalLink.vue
nucleogenesis: Following up on the above - some of this also kind of confirms my thoughts about the icon I think: https://material.io/design/usability/bidirectionality.html#mirroring-layout (at least the bit about leftward arrows meaning forward motion to RTL readers)

sairina: That makes sense! Fixed.

sairina:
###Update
I realized I had **deleted** the `icon` and `iconAfter` props completely (instead of just adding the `openInNewTab` prop). So, I restored that functionality and double-checked that external page icons still appear correctly. See images below. Sorry about that!

|Arabic `icon`|Arabic `iconAfter`|
|--|--|
|Screen Shot 2021-02-03 at 9 18 39 PM|Screen Shot 2021-02-03 at 9 18 26 PM|

|English `icon`|English `iconAfter`|
|--|--|
|Screen Shot 2021-02-03 at 9 21 59 PM|Screen Shot 2021-02-03 at 9 21 33 PM|

indirectlylit:
In both chromium and firefox on mac I see this:

image

for

```vue
        <KExternalLink
          text="external (google)"
          href="https://google.com"
          icon="facility"
          iconAfter="facility"
        />
```

indirectlylit (approved):
merging in so we can see everything together. Will file follow-up for cross-component consistency

PR #145, opened 2021-01-12
Added minWidth to prevent squished/oval effect in KIconButton

marcellamaki:
This PR re-adds a minWidth value, which prevents the "squished" effect that is seen in Studio with some browser window resizes, especially at small sizes (~600px) in the tree views

### DIFF
--- a/lib/buttons-and-links/KIconButton.vue
+++ b/lib/buttons-and-links/KIconButton.vue
@@ -75,9 +75,10 @@
           ...this.sizeStyles,
           // Circle
           borderRadius: '50%',
-          // Remove mins & padding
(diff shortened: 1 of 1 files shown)


nucleogenesis (approved):
Looked good to me. I do note that there may be overlap -and that we should be sure that we include this in some part of the our specs @jtamiace that our KIconButtons are 32px and require enough padding/margin that they don't get cut off:

image

But for the KDS purposes, this works for me. In general, I'd like to avoid hard-coding sizes into our components but we already do that for KIcon to a degree, so it makes sense that a button relying on KIcon would ensure enough space around it.

PR #149, opened 2021-02-05
Update issue and PR templates

sairina:
Created templates for opening an issue and submitting a PR in KDS.


indirectlylit (commented):
awesome thanks - will review this soon!

FYI coincidentally I just filed this related issue: https://github.com/learningequality/studio/issues/2919

I'll probably end up comparing templates from both Studio and Kolibri

indirectlylit (approved):
sorry to hold this up. I haven't had a chance to review but this is clearly a positive step, so no need for me to block it. Might revisit in the future with updates

Issue #85, opened 2020-07-29
KRouterLink - Add `replace` prop


nucleogenesis:
`<router-link>` is like `router.push()` - in that it creates a new history 

We can extend the `replace` prop from `<router-link>` to `KRouterLink` so that we have the option.

https://router.vuejs.org/api/#replace

I noticed this while looking into the first listed issue in this: https://github.com/learningequality/kolibri/issues/7366

Where the back button "doesn't work" but it's really just that you're getting a bunch of history items pushed to the history stack.

sairina:
@indirectlylit So ideally, we'd want to be able to do something like this?

`<KRouterLink
      :text="facility.name"
      :to="coachClassListPageLink(facility)"
      replace
  />`

nucleogenesis:
Yep and `KRouterLink` should pass the value of that new boolean prop (default: false) to the `<router-link>` in its template.

sairina:
With this as our code: `<KRouterLink  :text="facility.name"   :to="facilityLink(facility)"  replace />`

The way I currently have it, the prop is in the `<router-link>` in the template.

Is this how we want it to come through? 

This is what the Vue devtools shows for `KRouterLink`, and then nested within it, `router-link`:
|`KRouterLink`|`router-link`|
|--|--|
|![Screen Shot 2021-02-09 at 9 55 57 PM](https://user-images.githubusercontent.com/13563002/107471200-9900ee80-6b21-11eb-9a64-d227630277fd.png)|![Screen Shot 2021-02-09 at 9 52 42 PM](https://user-images.githubusercontent.com/13563002/107470973-26900e80-6b21-11eb-98f1-c7bd7e30dd33.png)|

PR #152, opened 2021-02-10
Add `replace` prop to KRouterLink

sairina:
## Summary
Added `replace` prop to `KRouterLink` by making `replace` accessible within `<router-link>` component.

Addresses #85

### DIFF
--- a/lib/buttons-and-links/KRouterLink.vue
+++ b/lib/buttons-and-links/KRouterLink.vue
@@ -1,7 +1,7 @@
 <template>
 
   <!-- no extra whitespace inside link -->
-  <router-link :class="buttonClasses" :to="to" dir="auto">
+  <router-link :class="buttonClasses" :to="to" replace dir="auto">
     <KLabeledIcon :maxWidth="maxWidth" @mouseenter="hovering = true" @mouseleave="hovering = false">
       <KIcon
         v-if="icon"
--- a/lib/buttons-and-links/KRouterLink.vue
+++ b/lib/buttons-and-links/KRouterLink.vue
@@ -1,7 +1,12 @@
 <template>
 
   <!-- no extra whitespace inside link -->
-  <router-link :class="buttonClasses" :to="to" replace dir="auto">
+  <router-link
+    :class="buttonClasses"
+    :to="to"
+    :replace="replace"
+    dir="auto"
+  >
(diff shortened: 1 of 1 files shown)


nucleogenesis:
Hey @sairina - we'll want to have the `replace` prop defined on the `KRouterLink` component as well - then the `:replace="replace"` in the template. I wasn't clear in the original issue, but the desire is to give the option to the implementer whether or not they want to replace.

I think that since the default is `false` on `<router-link>` we should make it the default in our `KRouterLink` as well

sairina:
## Showing use of `replace` prop:

### `replace="true"`
#### Expected behavior: the route to `/test` doesn't get added as a new history object; it replaces the current one (`/info`), so clicking the browser's back button takes it back to `/content`, and clicking the forward button takes it to `/test`, as `/info` has been replaced by `/test`.

** Flow: **
1. To - Channels `(/content)`
2. To - Info `(/info)`
3. To - Test `(/test)`
4. Back Arrow - Channels `(/content)`
5. Forward Arrow - Test `(/test)`



---
### `replace="false"`
#### Expected behavior: the route to `/test` does get added as a new history object, so clicking the browser's back button takes it back to `/info`.

** Flow: **
1. To - Channels `(/content)`
2. To - Info `(/info)`
3. To - Test `(/test)`
4. Back Arrow - Info `(/info)`
5. Forward Arrow - Test `(/test)`

nucleogenesis (approved):
LGTM!

lib/buttons-and-links/KRouterLink.vue
```diff
@@ -54,6 +59,13 @@
         type: String,
         required: false,
       },
+      /**
+       * If provided, calls window.history.replaceState() so navigation will not leave history record
+       */
+      replace: {
+        type: Boolean,
+        required: false,
+      },
```
indirectlylit, line 68: To avoid ambiguity and `undefined` input values, component props should either be `required: true` or provide a `default` value.

In this case, the prop should be:

```js
      replace: {
        type: Boolean,
        default: false,
      }
```

sairina: Thank you! I fixed it in a separate branch as a PR #162. Here. Not sure why that wasn't caught?

nucleogenesis: Ah I missed that - good eye @indirectlylit !

Issue #103, opened 2020-09-09
basic-link KButton icons are not colored properly


nucleogenesis:
This does not account for the fact that icons associated with a `basic-link` appearance ought to show as `$themeTokens.primary`.

Additionally, it seems that icons cause the underline of the associated text to extend out a character in either direction. 

![screenshot-localhost_8080-2020 09 09-16_07_35](https://user-images.githubusercontent.com/6356129/92663614-98380800-f2b6-11ea-8a85-1d8c33a031dc.png)

sairina:
When I used a `KButton` with a `basic-link` appearance and added icons before and after, I couldn't seem to replicate the space between the icons. However, I'll work on getting the icons to be: `$themeTokens.primary`

![Screen Shot 2021-02-11 at 2 57 48 PM](https://user-images.githubusercontent.com/13563002/107709224-b8129400-6c79-11eb-9d93-d87f9e4b7819.png)

PR #161, opened 2021-02-12
KButton - basic-link: add spacing between link and icons and change color of icons

sairina:
## Summary

- Updated the icons' colors to be `$themeTokens.primary`
- Increased spacing between text and icon to be 8px.

Addresses #103 

|Before|After|
|--|--|
|Screen Shot 2021-02-11 at 2 57 48 PM|Screen Shot 2021-02-11 at 10 28 32 PM|

### DIFF
--- a/lib/buttons-and-links/KButton.vue
+++ b/lib/buttons-and-links/KButton.vue
@@ -13,9 +13,9 @@
   >
     <!-- icon may come by slot or by prop -->
     <slot name="icon"></slot>
-    <KIcon 
-      v-if="icon" 
-      :icon="icon" 
+    <KIcon
+      v-if="icon"
+      :icon="icon"
       :color="iconColor"
       class="prop-icon"
     />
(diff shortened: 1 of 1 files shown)


lib/buttons-and-links/KButton.vue
```diff
@@ -108,13 +108,7 @@
...
     computed: {
       iconColor() {
-        if (this.primary) {
-          return this.appearance === 'raised-button'
-            ? this.$themeTokens.textInverted
-            : this.$themeTokens.primary;
-        } else {
-          return this.$themeTokens.text;
-        }
+        return this.primary ? this.$themeTokens.primaryDark : this.$themeTokens.primary;
```
nucleogenesis, line 111: Because this can be a link (as in <a> tag) or a button we need to account for whether the `appearance` is `raised-button` - in which case, the text and icon colors should be `$themeTokens.textInverted`.

lib/buttons-and-links/KButton.vue
```diff
@@ -13,23 +13,23 @@
...
+      :icon="icon"
       :color="iconColor"
       class="prop-icon"
     />
 
     <slot v-if="$slots.default"></slot>
 
     <template v-else>
-      <span class="link-text">{{ text }}</span>
+      <span class="link-text" :style="spanStyle">{{ text }}</span>
```
nucleogenesis, line 26: `spanStyle` is a big vague. Maybe something like `textStyle` would be a bit clearer?

nucleogenesis (changes requested):
Will need to address the case for when the text itself is supposed to be `$themeTokens.textInverted` as well.

indirectlylit:
In chromium and firefox on mac I see this:

image


for 

```vue
        <KButton
          text="button (alert)"
          appearance="basic-link"
          icon="facility"
          iconAfter="facility"
          @click="console.log('hello')"
        />
```

indirectlylit (approved):
merging in so we can see everything together. Will file follow-up for cross-component consistency

PR #162, opened 2021-02-15
Fix default=false on boolean prop

sairina:
Updated `required=false` to `default=false` - [the issue that was not caught with the linter](https://github.com/learningequality/kolibri-design-system/pull/152#discussion_r575730153).


indirectlylit (approved):
thank you!

PR #171, opened 2021-02-20
big bang of dependency upgrades

indirectlylit:
Upgrading dependencies.

I was encountering some nuxt hot reload timeout issues, and this appears to gave addressed them.

Open question: does this also upgrade dependencies for our component library that risks introducing regressions to Kolibri or Studio if not thoroughly tested?


indirectlylit:
closing as it did not actually address the issue I was encountering with nuxt.js hot reload timeouts

PR #163, opened 2021-02-16
update default and required props

indirectlylit:
Updates behavior of `default` and `required` props.

Refs:
* https://github.com/learningequality/kolibri/issues/7794
* https://github.com/learningequality/kolibri-design-system/pull/152#discussion_r575730153

PR #173, opened 2021-02-20
Update PR template and fix default props in KExternalLink

sairina:
## Description

1. Updates the location of the PR template. It was originally, erroneously housed in the `ISSUES_TEMPLATES` folder, so it was showing up as an issue template rather than a PR template.
2. Makes an adjustment so users can decide if the issue is a bug report or a feature request when they click on "New issue"
3. `KExternalLink` was missing default props. Threw those in here (since most were fixed in #163 )

### Before/After Screenshots (if applicable)

Before
- Currently, opening an issue only gives us an option to add an "ISSUE_TEMPLATE" or a "PULL_REQUEST" (?!!)

Screen Shot 2021-02-19 at 8 17 40 PM

After:
1. We should still have two options, but they will be *Bug Report* and *Enhancement Request*
2. We should see a template when we open up a PR

## Checklist for Reviewer


indirectlylit (approved):
Seems there's no way to test the templates without merging them?

Let's see what happens!

indirectlylit:
Looks pretty good I think?

Issue #92, opened 2020-08-14
Add spec to modals documentation - do not use gerunds in modal or page titles


jtamiace:
Adding a gerund to a title gives the user the impression that the task is in progress when it's not. Several anecdotal accounts of this happening from team members

Slack thread which contains other links
https://learningequality.slack.com/archives/CTG8XDTCL/p1597436796011600

indirectlylit:
could also be noted in writing guidelines

PR #176, opened 2021-02-24
Adds guidelines to not use gerunds in page titles and modal titles

jtamiace:
---
name: Pull Request
about: Create a pull request
title: "[TITLE]: Brief description of PR"
labels: ''
assignees: ''

---

## Description

Adds guidelines to not use gerunds in page titles and modal titles titles because they give the false impression that a task is in progress. Added to Writing page and Modals page.

#### Issue Addressed

Addresses https://github.com/learningequality/kolibri-design-system/issues/92

### Writing page 

| Before  | After |
| Screen Shot 2021-02-23 at 5 42 25 PM | Screen Shot 2021-02-23 at 5 42 31 PM |

### Modals page

| Before  | After |
| Screen Shot 2021-02-23 at 5 44 26 PM | Screen Shot 2021-02-23 at 5 44 20 PM |

### DIFF
--- a/docs/pages/writing.vue
+++ b/docs/pages/writing.vue
@@ -49,6 +49,9 @@
           Avoid creating new Proper Nouns in casual messages. For example,
           "You are a coach", not "You are a Coach"
         </li>
+        <li>
+          Avoid using gerunds in modal or page titles. This creates the impression that the task is in progress when it may not be. For example, "Publish channel", not "Publishing channel"
+        </li>
       </ul>
     </DocsPageSection>
 
--- a/docs/pages/modals/index.vue
+++ b/docs/pages/modals/index.vue
@@ -27,7 +27,7 @@
         Title
       </h3>
       <p>
-        Keep this short, concise, and easy for the user to understand. Should not be structured as a sentence or question, and should not contain user-generated text.
+        Keep this short, concise, and easy for the user to understand. Titles should not be structured as a sentence or question, and should not contain user-generated text. Titles should also not have verbs that end in "-ing". 
       </p>
       <DocsDoNot>
         <template v-slot:do>


docs/pages/writing.vue
```diff
@@ -49,6 +49,9 @@
           Avoid creating new Proper Nouns in casual messages. For example,
           "You are a coach", not "You are a Coach"
         </li>
+        <li>
+          Avoid using gerunds in modal or page titles. This creates the impression that the task is in progress when it may not be. For example, "Publish channel", not "Publishing channel"
```
indirectlylit, line 53: This sort of reads to me like "don't use gerunds incorrectly" :) We might similarly say something like

> Avoid using gerunds in modal or page titles prior to initiating a background task because a gerund communicates that the task is already in progress

but this seems obvious when phrased like that....

I believe @radinamatic had mentioned wanting to avoid gerunds more generally because they tend to cause challenges for translators, but I think that might be a slightly different sort of guidance we're giving above.

rtibbles: Noting that it's not a gerund that communicates that the task is in progress, but that in English the gerund form `-ing` and the present participle form `-ing` are identical. So using a gerund `Publishing channel` can easily imply to the reader that this is a contracted form of `The channel is publishing`.

I feel like it would be easier to describe what ought to be done:

> Use imperative verbs in modal or page titles.

indirectlylit: @radinamatic posted this http://asd-ste100.org/faq.html

image

But it leaves me wondering (:rimshot:) how we would express things like "Downloading", "Exporting", etc. "The download is in progress" seems so verbose

jtamiace: "Use imperative verbs in modal or page titles" is _much_ simpler than what's there now. I'm going to update the guideline with that.

This PR is meant to only address the first point
> 1. this guidance specifically for modal and page titles (in which case I would expect it go on a more specific page) 

I'm thinking I should remove it from the Writing page and add it to the title guidelines in Page headers and Modals only.

For the second point
> 2. in fact there there is blanket avoidance of -ing forms even for things that are actually in-progress?

If we do want to go in that direction, which I'd rather let Radina speak on, I think it's a separate issue that shouldn't be addressed by this PR, and maybe that's the level of guidance that should go in the Writing page.

indirectlylit: > I'm thinking I should remove it from the Writing page and add it to the title guidelines in Page headers and Modals only.

that would be my preference. otherwise we're sort of bubbling up a one-off writing mistake into a highly-visible new global writing rule

jtamiace:
Closing this because I messed up my local git setup while trying to address feedback.

Opened in a PR here: https://github.com/learningequality/kolibri-design-system/pull/184

Issue #178, opened 2021-02-25
KDropdownMenu not keyboard navigable


marcellamaki:
## Product
KDS

## Expected behavior
When using tab navigation, and opening a dropdown menu, a user should be able to enter the menu using only the keyboard.

## Actual behavior
Although the KDropdownMenu allows the user to open the dropdown, the focus never enters the opened menu, and the user cannot get to any of the items.

## Steps to reproduce the issue
1. On the Channels page in Kolibri, navigate to the Options dropdown using tab. 
2. When the button is in focus, use enter or space to open the menu. Using tab or the arrow keys, you will not be able to navigate into the menu, and the focus will continue to the following elements

## Additional information
![tab-nav](https://user-images.githubusercontent.com/17235236/109225568-6eeb3580-778b-11eb-842b-28b311edef02.gif)

Issue #164, opened 2021-02-16
KDropdownMenu issues


indirectlylit:
The `KDropdownMenu` component implements full buttons with menus like these:

![image](https://user-images.githubusercontent.com/2367265/108136537-16bb8180-706f-11eb-9938-57b058160fc0.png)

A couple issues:

* The name implies that it would just implement a menu, but it actually implements buttons that happen to have a menu
* It doesn't play nicely with `KButtonGroup`, which doesn't add internal spacing as expected

Recommend deprecating this component entirely, and adding an `options` property to `KButton` and removing `hasDropdown`.

MisRob:
Closed by https://github.com/learningequality/kolibri-design-system/pull/346

PR #180, opened 2021-02-26
Eliminate need to double-tab into dropdown

### LINKED ISSUE #136: KDropdown menu is wrapped by a div with tabindex="0", which means it takes 2 tabs to get to the inner button
Example from Device > Channels page

image

Because the outer container has `tabindex="0"`, it takes two tabs to focus on the actual `KButton` inside the dropdown component.

I'm not sure why this is happening. I don't see a `tabindex="0"` attr in the KDropdownMenu code.

marcellamaki:
## Description

This PR prevents the loss of focus/double tab necessary to enter KDropdownModal

#### Issue Addressed
Fixes #136 
tab-nav-after-2

## Steps to Test

1. In Kolibri, on the channels page, use the tab key to navigate. You should be able to proceed from the main horizontal nav and move to the "Options" menu without a break/loss of focus.

## Comments

There is a [separate issue with `<KDropdownMenu />` not being keyboard navigable](https://github.com/learningequality/kolibri-design-system/issues/178) when the menu is open (meaning, there is no way to enter the menu). This PR does not resolve that. I've documented and assigned that issue to myself.

### DIFF
--- a/lib/KDropdownMenu.vue
+++ b/lib/KDropdownMenu.vue
@@ -1,29 +1,17 @@
 <template>
-
-  <div class="dib">
-    <!-- Added because the trigger could not be a vue element smh-->
-    <div
-      ref="buttonContainer"
-      class="button-container dib"
-    >
-      <slot name="button"></slot>
-      <KButton
-        v-if="!$slots.button"
-        ref="button"
-        :text="text"
-        :appearance="appearance"
-        :disabled="disabled"
-        :hasDropdown="true"
-        :primary="$attrs.primary"
-      />
-    </div>
-
-    <UiPopover
+  <KButton
+    ref="button"
+    :appearance="appearance"
+    :disabled="disabled"
+    :hasDropdown="true"
+    :primary="$attrs.primary"
+  >
+  <span>{{text}}</span>
+  <UiPopover
       v-if="!disabled"
       ref="popover"
       :z-index="12"
-      :trigger="$refs.buttonContainer"
-      :containFocus="false"
+      :containFocus="true"
       :dropdownPosition="position"
       @close="handleClose"
       @open="handleOpen"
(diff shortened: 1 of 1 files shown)


lib/KDropdownMenu.vue
```diff
@@ -1,40 +1,28 @@
...
-
-  <div class="dib">
-    <!-- Added because the trigger could not be a vue element smh-->
-    <div
-      ref="buttonContainer"
-      class="button-container dib"
-    >
-      <slot name="button"></slot>
-      <KButton
-        v-if="!$slots.button"
```
nucleogenesis, line 11: The change here fundamentally changes the API of the component by removing the allowance of a named `buttons` slot.

After looking into Kolibri, I see that we don't use that slot anywhere ( or in Studio - where we dont' use this component at all )

I don't think we actually need this part of the API - but it's worth noting in any case (and in our CHANGELOG) that the API for this component has changed.

nucleogenesis: I think that it's actually good to remove this from the API in this case because the component (I believe) is already designed with good consideration for our design system guidance on buttons and offering the slot makes it easier to use the component while incorrectly styling the button.

nucleogenesis (changes requested):
I'm not sure what the linter is mad about but I'm going to guess that it's that everything between the `<style>` tags is dead code now.

Once the linter passes this is good to go!

lib/KDropdownMenu.vue
indirectlylit: Relevant: https://github.com/learningequality/kolibri-design-system/issues/164

Long story short, I think `KDropdownMenu` should not exist in it's current form. However, deprecating it could only happen in a "major" (technically minor in the [semver](https://semver.org/) sense) bump because this would truly be a breaking change.

In this case I agree with @nucleogenesis that we can drop the API because we know with high confidence that we won't break apps for library users. Once KDS gets more widely used (and probably once we call it 1.0.0) we'll want to get a less cavalier about this kind of change

nucleogenesis (approved):
LGTM thanks @marcellamaki

PR #184, opened 2021-03-02
Adds imperative verb guidance to modals and page headers

jtamiace:
## Description

I messed up my local git somehow... reposting [this PR](https://github.com/learningequality/kolibri-design-system/pull/176) here

Adds guidelines to not use verbs ending in "-ing" on page titles and modal titles because they give the false impression that a task is in progress. Added to Page headers and Modals pages

#### Issue Addressed

Addresses https://github.com/learningequality/kolibri-design-system/issues/92

### Page headers

| Before  | After |
| Screen Shot 2021-03-01 at 4 27 04 PM | Screen Shot 2021-03-01 at 4 26 02 PM |

### Modals

| Before  | After |
| Screen Shot 2021-03-01 at 4 27 11 PM | Screen Shot 2021-03-01 at 4 26 34 PM |

### DIFF
--- a/docs/pages/modals/index.vue
+++ b/docs/pages/modals/index.vue
@@ -27,7 +27,7 @@
         Title
       </h3>
       <p>
-        Keep this short, concise, and easy for the user to understand. Should not be structured as a sentence or question, and should not contain user-generated text.
+        Keep this short, concise, and easy for the user to understand. Titles should not be structured as a sentence or question, and should not contain user-generated text. Use verbs in the imperative form. Do not use verbs that end in "-ing"; doing so may give users the impression that a task is in progress when it is not.
       </p>
       <DocsDoNot>
         <template v-slot:do>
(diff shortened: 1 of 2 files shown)


radinamatic:
Coming late to the party, apologies...

In agreement with the proposed change, and glad to see the discussion took in consideration the recommendation to avoid verbal tenses with `-ing` forms 👏🏽 

On the topic of:
> how we would express things like "Importing", "Exporting", etc. "The export is in progress" seems so verbose

When writing user strings for the global audience we must favor `verbosity` over `minimalism` whenever it improves clarity. 

My major _beef_ with English verbal tenses with `-ing` is that they so often **obfuscate** the subject of the clause, which then creates hurdles for localization. Probably no translator will ask themselves (or us for more context) on how to translate **The X is in progress**, while many may scratch heads with single-word-strings like **X-ing** (_Who is X-ing what?_ 😕).

jtamiace:
Thanks for clarifying @radinamatic 

I'm wondering what level of clarity is sufficient? Do we need to be as verbose as "The export is in progress," or would something like "Exporting data" be enough - just being sure to indicate _what_ the action is referring to? 

"Loading" --> "Loading page"
"Saving" --> "Saving changes"
"Uploading" --> "Uploading files"

nucleogenesis (approved):
lgt us

PR #185, opened 2021-03-09
Kdropdown keyboard navigation

### LINKED ISSUE #178: KDropdownMenu not keyboard navigable

marcellamaki:
## Description

This PR adds tab navigability for KDropdownModal. Currently, a user cannot use keyboard navigation to enter into the dropdown.

#### Issue addressed

Fixes #178 

### Before/after screenshots

BEFORE

tab-nav (1)

AFTER

tab-nav-updated

## Steps to test

1. Use the tab key to navigate to a dropdown menu
2. Use the enter key to open the dropdown
3. The focus should be directed to the first item in the menu
4. Using the tab key should direct you down each item in the menu, and then a double-tab at the end should bring you back to the first element
5. Using escape key when you are inside the menu should close the menu, and bring the focus back to the dropdown menu button

## Comments 

Any feedback for general code improvement (and if I'm using some of the existing KDS patterns correctly) or about how this affects other components that I might not be thinking of would be very helpful.

Also, @jtamiace / @khangmach, I am not sure if there is already a design convention for the focus on menu items. I did review the existing docs but know they're a WIP and it's possible you have some work on this that I'm just unaware of. If you have design feedback or recommendations of how I can improve this, please let me know! So far, I just tried to adapt existing patterns and styles, but I'm sure there is room for improvement.

### DIFF
--- a/lib/KDropdownMenu.vue
+++ b/lib/KDropdownMenu.vue
@@ -98,12 +98,12 @@
       },
     },
     beforeDestroy() {
-      window.removeEventListener('keyup', this.handleKeyUp, true);
+      window.removeEventListener('keydown', this.handleArrowKeys, true);
     },
     methods: {
       handleOpen() {
         this.$nextTick(() => this.setFocus());
-        window.addEventListener('keyup', this.handleKeyUp, true);
+        window.addEventListener('keydown', this.handleArrowKeys, true);
       },
       setFocus() {
         this.$refs.menu.$el.querySelector('li').focus();
@@ -121,14 +121,23 @@
         }
         window.removeEventListener('keyup', this.handleKeyUp, true);
       },
-      handleKeyUp(event) {
-        if (event.shiftKey && event.keyCode == 9) {
-          const popover = this.$refs.popover.$el;
-          const popoverIsOpen = popover.clientWidth > 0 && popover.clientHeight > 0;
-          if (popoverIsOpen && !popover.contains(document.activeElement)) {
-            this.closePopover();
-            this.focusOnButton();
-          }
+      handleArrowKeys(event) {
+        // identify the menu state and length 
+        const popover = this.$refs.popover.$el;
+        const menuElements = this.$refs.menu.$el.querySelector('div').children
+        const lastChild = menuElements[menuElements.length - 1]
+        const popoverIsOpen = popover.clientWidth > 0 && popover.clientHeight > 0;
+        // set current element and its siblings
+        let focusedElement = document.activeElement;
+        let sibling = focusedElement.nextElementSibling;
+        let prevSibling = focusedElement.previousElementSibling;
+        // manage rotating through the options
+        if (event.keyCode == 38 && popoverIsOpen) {
+          event.preventDefault()
+          prevSibling ? this.$nextTick(() => prevSibling.focus()) : this.$nextTick(() => lastChild.focus())
+        } else if (event.keyCode == 40 && popoverIsOpen) {
+          event.preventDefault()
+          sibling ? this.$nextTick(() => sibling.focus()) : this.$nextTick(() => this.setFocus());
         }
       },
       handleSelection(selection) {
--- a/lib/KDropdownMenu.vue
+++ b/lib/KDropdownMenu.vue
@@ -122,10 +122,10 @@
         window.removeEventListener('keyup', this.handleKeyUp, true);
       },
       handleArrowKeys(event) {
-        // identify the menu state and length 
+        // identify the menu state and length
         const popover = this.$refs.popover.$el;
-        const menuElements = this.$refs.menu.$el.querySelector('div').children
-        const lastChild = menuElements[menuElements.length - 1]
+        const menuElements = this.$refs.menu.$el.querySelector('div').children;
+        const lastChild = menuElements[menuElements.length - 1];
         const popoverIsOpen = popover.clientWidth > 0 && popover.clientHeight > 0;
         // set current element and its siblings
         let focusedElement = document.activeElement;
@@ -133,10 +133,12 @@
         let prevSibling = focusedElement.previousElementSibling;
         // manage rotating through the options
         if (event.keyCode == 38 && popoverIsOpen) {
-          event.preventDefault()
-          prevSibling ? this.$nextTick(() => prevSibling.focus()) : this.$nextTick(() => lastChild.focus())
+          event.preventDefault();
+          prevSibling
+            ? this.$nextTick(() => prevSibling.focus())
+            : this.$nextTick(() => lastChild.focus());
         } else if (event.keyCode == 40 && popoverIsOpen) {
-          event.preventDefault()
+          event.preventDefault();
           sibling ? this.$nextTick(() => sibling.focus()) : this.$nextTick(() => this.setFocus());
         }
       },
--- a/lib/KDropdownMenu.vue
+++ b/lib/KDropdownMenu.vue
@@ -18,6 +18,7 @@
       @open="handleOpen"
     >
       <UiMenu
+        ref="menu"
         :options="options"
         @select="handleSelection"
       />
@@ -101,7 +102,11 @@
     },
     methods: {
       handleOpen() {
+        this.$nextTick(() => this.setFocus());
         window.addEventListener('keyup', this.handleKeyUp, true);
+      },
+      setFocus() {
+        this.$refs.menu.$el.querySelector('li').focus();
       },
       handleClose() {
         const focusedElement = document.activeElement;
@@ -116,16 +121,16 @@
         }
         window.removeEventListener('keyup', this.handleKeyUp, true);
       },
-      handleKeyUp(event) {
-        if (event.shiftKey && event.keyCode == 9) {
-          const popover = this.$refs.popover.$el;
-          const popoverIsOpen = popover.clientWidth > 0 && popover.clientHeight > 0;
-          if (popoverIsOpen && !popover.contains(document.activeElement)) {
-            this.closePopover();
-            this.focusOnButton();
-          }
-        }
-      },
+      // handleKeyUp(event) {
+      //   if (event.shiftKey && event.keyCode == 9) {
+      //     const popover = this.$refs.popover.$el;
+      //     const popoverIsOpen = popover.clientWidth > 0 && popover.clientHeight > 0;
+      //     if (popoverIsOpen && !popover.contains(document.activeElement)) {
+      //       this.closePopover();
+      //       this.focusOnButton();
+      //     }
+      //   }
+      // },
       handleSelection(selection) {
         /**
          * Emitted when an option is selected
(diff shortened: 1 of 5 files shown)


lib/keen/UiMenu.vue
```diff
@@ -101,6 +101,12 @@
           'has-secondary-text': this.hasSecondaryText,
         };
       },
+      activeOutline() {
+        return this.isActive ? this.$coreOutline : {};
+      },
+      redirectFocus() {
+        console.log(document.activeElement)
+      },
```
nucleogenesis, line 109: Should remove this function

nucleogenesis (changes requested):
The code looks good aside from the console.log

@radinamatic might have some more insights here on the a11y.

When I got focused into the first element of the dropdown, I tried using the arrow keys to navigate the menu items which scrolled the whole page and I realized you can only tab through the children.

I did some digging and found this: https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html  - if you scroll down you see some specs for Keyboard Navigation.

Perhaps the use case there is slightly different than ours? In any case - I'll let @radinamatic weigh in on that aspect.

marcellamaki:
@nucleogenesis - thanks for your notes! You are right about the arrow keys and not the tab. I had read some resources saying either was fine, but I think it's a better idea to go with the w3 standard. I'll make those changes today or tomorrow and hopefully get this wrapped up before we move into all-product sprints 😊

lib/keen/UiMenu.vue
marcellamaki: Good catch!

jtamiace (approved):
Approving from the design side - the behaviors you implemented make a lot of sense, and so does the highlight itself

marcellamaki:
@nucleogenesis - I made some changes to the code since you last reviewed. One making sure the arrow keys worked and two adding a `$focus-outline` variable. Previously I had been using some javascript and active styles, but it was getting unnecessarily complicated to be creating so many computed properties just to add some basic css. That said, I'm not sure if there is a reason why there _wasn't_ this variable already, and why the value was mostly being accessed in different ways. I'd be happy to switch back to using javascript if either you or Devon could give me a few hints on a simpler approach. Thanks for your feedback!

lib/keen/styles/variables.scss
```diff
@@ -25,6 +25,9 @@ $disabled-text-color: rgba(black, 0.38) !default;
 // Divider color
 $divider-color: rgba(black, 0.12) !default;
 
+// Focus outline color
+$focus-outline-color: #8dc5b6 !default;
```
indirectlylit, line 29: Unfortunately, because Kolibri supports theming we cannot hard-code colors as scss variables. Instead, use the `$coreOutline` computed style provided by the `KThemePlugin` mixin:

https://github.com/learningequality/kolibri-design-system/search?q=%24coreOutline

indirectlylit (changes requested):

lib/KDropdownMenu.vue
```diff
@@ -116,14 +121,25 @@
...
-          const popoverIsOpen = popover.clientWidth > 0 && popover.clientHeight > 0;
-          if (popoverIsOpen && !popover.contains(document.activeElement)) {
-            this.closePopover();
-            this.focusOnButton();
-          }
+      handleArrowKeys(event) {
+        // identify the menu state and length
+        const popover = this.$refs.popover.$el;
+        const menuElements = this.$refs.menu.$el.querySelector('div').children;
+        const lastChild = menuElements[menuElements.length - 1];
```
nucleogenesis, line 128: When we're on the lastChild and hit tab (keycode 9) then the focus outline is lost altogether and I'm not sure where it is. The menu stays open as well.

On that W3 link I shared in my previous comment, if you keyboard navigate to the dropdown (you'll have to eyeball it because it doesn't show a focus indicator on that site) - but if you hit space to open it, arrows to navigate within it, and then hit tab - the tab key moves the focus to the next element in the DOM.

So the listed items in the drop-down should not be a part of the tab navigation.

nucleogenesis (changes requested):
With @indirectlylit's note about the CSS and the fiddly a11y tab/keyboard navigation stuff I noted both resolved this should be good to go, thanks @marcellamaki !

marcellamaki:
I've made some changes in this PR that were not changes when I started it, but now are reverting other changes that have been made in the interim around focus and keyboard navigation. Would especially appreciate review from @sairina and @MisRob in regards to these changes, as it essentially undid some recent changes. 

**For context and the thought process:** 
- Most of the elements that are accessed through keyboard navigation are through tab navigation
- However, a handful of elements such as dropdown menus, radio buttons, and scolling, use the arrow keys (as @nucleogenesis mentioned above)[ Further details](https://webaim.org/techniques/keyboard/)
- The original logic applied a focus outline to the default elements but did not check for whether or not there was a keyboard event, which caused the issue with the focus outline appearing around textbox elements on the sign in page, [even when the user was not navigating with keyboard](https://github.com/learningequality/kolibri/issues/8786)
- The other consideration is when the focus is added through javascript, and is not tied to a keyboard event, but we _want_ to set the focus outline (i.e. for the select dropdown menu), then we need to manage this as well
- The original logic also [did not include focus outline for `KSwitch`](https://github.com/learningequality/kolibri-design-system/pull/191) which was addressed

**General approach to changes:** 
- Some elements should always default to having a focus outline, and this should be associated only with keyboard events 
- Focus outline should not appear around "focusable elements" when navigated with the keyboard 
- There may be use cases where we want to add focus outline and keyboard navigation ability, especially around arrow keys, so we should have an option to add this explicitly
- Goal: to reduce the number of places in future development that we need to manage an "activeClass" that requires an outline as a one off within Kolibri, especially for non-tabbable elements.

**Outcomes of implementation:**

| Element and interaction  |  Functionality |
|---|---|
| Navigating checkboxes with keyboard  |  checkboxes-keyboard |
| Download resource button with keyboard | download-keyboard |
| Dropdown menu with keyboard (arrow keys) | dropdown-keyboard |
|  Dropdown with mouse click (no focus) | dropdown-mouseclick |
| Logging in using mouse (no focus outline) | login-mouseclick |
|  Logging in when starting to use the tab key to navigate | login-start-tabbing |


**NOT SOLVED BY THIS**

The user profile dropdown menu in the top right corner does not use this same style of menu. It is currently navigable with tab keys, which might be confusing to users... However it is being removed in 0.16. Worth discussing whether there should be a follow up issue to change this... I think it might be a low lift and the consistency might be beneficial?

MisRob (commented):
Thank you @marcellamaki, this is great work. I've only checked the code so far. To be honest, I don't recall exactly all updates to the tracking input modality script I did and reasons for them but I think that your updates with all background information and comments in this PR make sense overall. Thank you for sharing the WebAIM resource on keyboard testing, that's very helpful. I learned a lot from the discussion and work in this PR.

I left some minor comments, neither of them is blocking.

The only thing that needs to be resolved is kolibri-tools upgrade. I saw that you posted [a message on Slack ](https://learningequality.slack.com/archives/CTG8XDTCL/p1652713780133189)about it so I responded there with some issues I experienced.

lib/KDropdownMenu.vue
```diff
@@ -116,14 +121,31 @@
...
+        const popoverIsOpen = popover.clientWidth > 0 && popover.clientHeight > 0;
+        // set current element and its siblings
+        let focusedElement = document.activeElement;
+        let sibling = focusedElement.nextElementSibling;
+        let prevSibling = focusedElement.previousElementSibling;
+        // manage rotating through the options
+        if (event.keyCode == 38 && popoverIsOpen) {
```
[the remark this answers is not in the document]
marcellamaki: Thank you @MisRob -- this is very helpful. I have updated based on these references

lib/styles/trackInputModality.js
```diff
@@ -115,7 +115,12 @@ function setUpEventHandlers(disableFocusRingByDefault) {
         (hadKeyboardEvent && focusTriggersKeyboardModality(e.target)) ||
         (focusSetExplicitly(e.target) && !hadClickEvent)
       ) {
+        // both the JS state and the body attribute for keyboard modality
         globalThemeState.inputModality = 'keyboard';
+        document.body.setAttribute('modality', 'keyboard');
+      } else {
+        globalThemeState.inputModality = null;
+        document.body.setAttribute('modality', '');
```
indirectlylit, line 123: See https://github.com/learningequality/kolibri/issues/4822#issuecomment-458761889 and https://github.com/learningequality/kolibri/pull/4847 for some historical context on how the code got into its current state.

lib/styles/trackInputModality.js
```diff
@@ -33,25 +88,39 @@ function setUpEventHandlers(disableFocusRingByDefault) {
...
 
       isHandlingKeyboardThrottle = setTimeout(() => {
-        recentKeyboardEvent = null;
+        hadKeyboardEvent = false;
       }, 100);
     },
     true
   );
 
+  document.body.addEventListener('mouseup', () => {
```
rtibbles, line 106: Was this copied from the previous implementation? Otherwise, would be good to include a comment explaining the rationale for using `mouseup` versus `mousedown` etc.

lib/styles/trackInputModality.js
```diff
@@ -33,25 +88,39 @@ function setUpEventHandlers(disableFocusRingByDefault) {
...
+      if (
+        (hadKeyboardEvent && focusTriggersKeyboardModality(e.target)) ||
+        (focusSetExplicitly(e.target) && !hadClickEvent)
+      ) {
+        // both the JS state and the body attribute for keyboard modality
         globalThemeState.inputModality = 'keyboard';
+        document.body.setAttribute('modality', 'keyboard');
+      } else {
+        globalThemeState.inputModality = null;
+        document.body.setAttribute('modality', '');
```
rtibbles, line 123: We need to also remove the `modality` attribute from the body element in the `blur` listener below, otherwise, we will only clear it when there are focus events after a mouse click.

rtibbles (changes requested):
One thing still missing for the body modality attribute.

lib/styles/trackInputModality.js
marcellamaki: This should be `mousedown` and has been changed. I had been reading some guidance on this, but realized from your comment and revisiting it that I had gotten it backwards. fixed!

rtibbles (approved):

Issue #131, opened 2020-11-05
All components, including KTextBox, should have a standard focus outline


indirectlylit:
Currently textboxes don't have a standard focus outline. For accessibility and consistency, when tabbing through the UI, _all_ keyboard-focusable elements should use the same standard outline.

Ref decision made in design system tactical:

https://www.notion.so/learningequality/we-will-try-adding-a-standard-blue-focus-outline-to-text-boxes-4f86ba39d14a4a07a9df5ea20bb717d9

Components with the issue - issue completed once all items are marked off:
- [x] KTextbox
- [x] KSwitch

sairina:
@jtamiace and @khangmach - Is there a place where I can see what the standard blue focus outline looks like for the textboxes?

indirectlylit:
No, the focus outline refers to the `secondary`-color border which appears around most K* components on tab-focus, but not on the text box. This outline should only appear when using tab keyboard navigation, and not when selecting UI elements with the mouse or touchpad.


![2021-02-11 17 17 37](https://user-images.githubusercontent.com/2367265/107719766-cff41300-6c8d-11eb-9dc0-8b439ab644f9.gif)

#124 is about applying a permanent background color to the text box, even in a non-focused state. 

Adding this tab focus outline will be in addition to any existing focus behaviors such as adjustments to the label and underline color/thickness.

sairina:
@indirectlylit, @jtamiace, @khangmach This is what I have so far. Is this what we're looking for? And, also, when there's a character count (like when editing a user's full name), should the count be inside the outline?

![2021-03-03 22 41 44](https://user-images.githubusercontent.com/13563002/109922097-b8e38800-7c71-11eb-8ae0-b1e4f27a04e9.gif)

indirectlylit:
wow, much a11y :)

And just to be clear, this does not show up when you simply click in the box with the mouse, correct?

One thing I'm seeing is that the vertical layout is shifting slightly on focus – though this might be due to the validation and not the new focus outline. Preferably, neither of these would cause a shift in height.

>  should the count be inside the outline?

I don't think it matters that much? Since validation is already inside, I might expect the count inside also.

nucleogenesis:
@sairina is there a PR incoming for this?

Cronus1007:
@nucleogenesis Shall I work upon this issue.

sairina:
[...]

PR #191, opened 2021-03-15
Fix outlines for KTextbox and KSwitch

sairina:
#### Issue addressed

Addresses #131 

### Before/after screenshots

|`KTextbox` Before | `KTextbox` After| `KTextbox` (after background fixed) |
|--|--|--|
| (Basically, before there was no outline) | ||

|`KSwitch` Before | `KSwitch` After|
|--|--|
|||

## Steps to test

`KTextbox`
1. On the sign-in page, tab until you get into the username textbox.
2. Check to see that there is an outline around the textbox.

`KSwitch`
1. Sign in as a coach
2. Click on "Reports" for your class
3. Under "Lessons," toggle the switch on/off for "Visible to learners"
4. Check to see if there is an outline that appears as you are toggling.
5. Please check for an RTL language as well by changing the language in the address bar in your browser from `/en/` to `/ar/` for example.

For both of these components, I added `:style` and a ternary for whether or not there was a focus on the element. I also made updates to `trackInputModality.js` in order to differentiate between keyboard tab and mouse click.

This in itself does not introduce any tech-debt items. However, it should be noted that there is a new CSS pseudo-class called `:focus-visible` that simplifies a lot of the logic that was introduced in order to track the use of keyboard vs. mouse, and it will likely be a refactor that we will want to look into in the design system. I will be opening an issue and linking this issue.

## Reviewer guidance

## Comments

I accidentally forgot and added both `KTextbox` and `KSwitch` together in this PR 😩


sairina:
@indirectlylit From [here](https://github.com/learningequality/kolibri-design-system/issues/131#issuecomment-790770612):

> One thing I'm seeing is that the vertical layout is shifting slightly on focus – though this might be due to the validation and not the new focus outline. Preferably, neither of these would cause a shift in height.

I double-checked with both clicking by mouse or tabbing with the keyboard and it seems to shift with focus, which I think means that it is validation-related. I will open a separate issue for this problem.

indirectlylit:
That sounds right, thanks!

One question though – I thought we'd switched textboxes to the new style in https://github.com/learningequality/kolibri-design-system/issues/167

These look like it's still using the old styling? (cc @nucleogenesis who worked on the other work)

nucleogenesis:
1) Sorry for not seeing this sooner
2) Also sorry for removing the line that added the background somehow in a follow-up commit to my original KTextbox update PR...

Now it will be fixed here: https://github.com/learningequality/kolibri-design-system/pull/240 @indirectlylit

sairina:
@indirectlylit With the new style for `KTextbox`:

indirectlylit:
hooray! thanks guys

MisRob (approved):
Thank you @sairina, code looks good to me, and I haven't noticed any problems when testing locally.

PR #194, opened 2021-03-16
Adds 'Filters' pattern page

jtamiace:
## Description

Part of the [In docs we trust](https://www.notion.so/learningequality/Design-System-e29b22eb65274d51b8354dfb404db1c1?p=ce7be497ba544881b0cea636fcf6e67b) project

Adds specs and behaviors for these components as they relate to filtering:

* Dropdown menus
* Text input dropdowns
* Text filters
* Checkbox lists

## Reviewer guidance

This new page documents specs for our current use of filters. Looking for feedback on anything that's missing, inaccurate, or under-specced. 

## Comments

Later I'll add two follow up issues that I couldn't document for now because of inconsistencies & undesirable behaviors, and the fact that library implementation would need to happen first or in tandem:

* Multi-select dropdown filters: we need a better representation of multiple selections at one time and how long text is handled 
* This new page states that when there are '0 results' for a filter in a dropdown, we should hide that option from the dropdown. In the future, we should also communicate the number of missing options in a filter dropdown menu and why they are hidden via a tooltip.


indirectlylit (approved):
this looks great!

PR #195, opened 2021-03-17
Fixes Safari text-wrapping in KLabeledIcon and KRouterLink

sairina:
## Description

Fixes the text-wrapping that occurs from webkit/Safari in `KLabeledIcon` and `KRouterLink`. PR in Kolibri addresses the specific issue that is found in the user dropdown menu for "Change language" option.

#### Issue addressed

Addresses learningequality/kolibri#7378

### Before/after screenshots
|`KLabeledIcon` fix for "Sign out"|
|--|
|Screen Shot 2021-03-16 at 9 04 50 PM|

|`KRouterLink` Before|`KRouterLink` After|
|--|--|
|Screen Shot 2021-03-16 at 9 05 57 PM|Screen Shot 2021-03-16 at 9 06 14 PM|

|`KLabeledIcon` Before|`KLabeledIcon` After|
|--|--|
|Screen Shot 2021-03-16 at 9 06 49 PM|Screen Shot 2021-03-16 at 9 22 36 PM|

## Steps to test (at least the features that were created)

1. Log in on Safari as admin or superadmin
2. Click on the user at the upper right hand to open the user menu dropdown - check to see if "Change language" is on one line.
3. Click on the hamburger menu on the left to open the sidenav - check to see if "Sign out" is on one line
4. Click on Coach
5. Click on a facility
6. Check to see if "All facilities" is on line line
7. Click on a class (preferably one with a long name!)
8. Check to see that "All classes" is on one line
9. Check to see that the class name is on one line
10. Repeat with other browsers (Chromium, Firefox, Chrome, etc.)

Please feel free to check other places where there are `KRouterLink`s or `KLabeledIcon`s

Added `whitespace: no-wrap` to both `KLabeledIcon` and `KRouterLink`

Please check for regressions in other browsers!

## Reviewer guidance

- [ ] Does this fix cause regressions in other browsers?

### DIFF
--- a/lib/KLabeledIcon.vue
+++ b/lib/KLabeledIcon.vue
@@ -65,9 +65,9 @@
     data() {
       return {
         noWhiteSpace: {
-          'white-space': 'nowrap'
+          'white-space': 'nowrap',
         },
-      }
+      };
     },
     computed: {
       labelEmpty() {
--- a/lib/KLabeledIcon.vue
+++ b/lib/KLabeledIcon.vue
@@ -6,7 +6,7 @@
         <KIcon v-if="icon" :icon="icon" :color="color || $themeTokens.text" style="top: 2px;" />
       </slot>
     </span>
-    <span class="label" :style="labelStyle">
+    <span class="label" :style="[ labelStyle, noWhiteSpace ]">
       <!-- nest slot inside span to get alignment and flow correct for mixed RLT/LTR -->
       <span dir="auto" class="debug-warning">
         <!-- Use zero-width space when empty -->
@@ -62,6 +62,13 @@
         default: '100%',
       },
     },
+    data() {
+      return {
+        noWhiteSpace: {
+          'white-space': 'nowrap'
+        },
+      }
+    },
     computed: {
       labelEmpty() {
         if (this.label) {
@@ -109,8 +116,6 @@
   .labeled-icon-wrapper {
     position: relative;
     display: inline-block;
-    min-width: 100%;
-    // Fixes the text-wrapping in Safari
     max-width: 100%;
   }
 
--- a/lib/KLabeledIcon.vue
+++ b/lib/KLabeledIcon.vue
@@ -6,7 +6,7 @@
         <KIcon v-if="icon" :icon="icon" :color="color || $themeTokens.text" style="top: 2px;" />
       </slot>
     </span>
-    <span class="label" :style="[ labelStyle, noWhiteSpace ]">
+    <span class="label" :style="[ labelStyle, { 'white-space': 'nowrap' } ]">
       <!-- nest slot inside span to get alignment and flow correct for mixed RLT/LTR -->
       <span dir="auto" class="debug-warning">
         <!-- Use zero-width space when empty -->
@@ -61,13 +61,6 @@
         type: String,
         default: '100%',
       },
-    },
-    data() {
-      return {
-        noWhiteSpace: {
-          'white-space': 'nowrap',
-        },
-      };
     },
     computed: {
       labelEmpty() {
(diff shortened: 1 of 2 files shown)


lib/KLabeledIcon.vue
```diff
@@ -62,6 +62,13 @@
         default: '100%',
       },
     },
+    data() {
```
jonboiser, line 65: Could you just add this to the `styles` object in `labelStyle`? It doesn't need to be reactive.

jonboiser:
This looks good, but might cause problems when the label _does_ need to break...

I like that this is a big visual improvement, but I'm not confident it scales to all labeled icons in all locales:

image

Did you do any research explaining why this particular behavior happens in Safari?

sairina:
Oh no - that's great catch. :sob: 

> Did you do any research explaining why this particular behavior happens in Safari? I just don't understand why the display: inline-block rules seem to get ignored.
Excellent question. My understanding (so far) seems to be aligned to what @nucleogenesis wrote in the initial issue - Safari tends to ignore certain attributes, like `min-width: auto`, for example (though I couldn't get the `min-width` fix to work). As for the why, I haven't yet found an answer to that. `display: inline-block` might have some whitespace issues, and there might be a `font-size: 0` fix or for some reason, `vertical-align: top`, which I'll play around with more tomorrow.

sairina:
Updated `KLabeledIcon` to remove the reactivity from `white-space`. That fix in `KLabeledIcon` seems to fix the issue that we were seeing in the menus (sidenav and user drop down).

nucleogenesis (approved):
I can only test in non-Safari browsers but I saw no regressions there - LGTM!

sairina:
@jonboiser Made some updates that should fix the issue. The only question I have is whether or not these change should live in KDS or in Kolibri code, and I have brought that up to the KDS team on Slack. Let me know if you see any other edge cases!

lib/KLabeledIcon.vue
```diff
@@ -108,7 +108,7 @@
 
   .labeled-icon-wrapper {
     position: relative;
-    display: inline-block;
+    display: block;
```
jonboiser, line 111: This _might_ be okay since we tend to use labeled icons as block-level elements, but I don't see any reason why this would always be the case. For example, if we decide to have a horizontal list of links, etc.

lib/buttons-and-links/KRouterLink.vue
```diff
@@ -6,6 +6,7 @@
     :to="to"
     :replace="replace"
     dir="auto"
+    style="display: block;"
```
jonboiser, line 9: Links can't be displayed as block elements, since we have too many places where they need to be displayed as inline elements:

image

sairina: Ah. Yes, thank you! Removed.

jonboiser (changes requested):
Changing the rule to `display: block` causes too many regressions in Kolibri. I think we should just selectively fix the problematic links in the Kolibri code-base for now.

jonboiser:
@nucleogenesis When you made this comment https://github.com/learningequality/kolibri/issues/7378#issuecomment-668800860

exactly where are you supposed to apply the `max-width: 0` rule?

jonboiser:
Here's something that seems to work: change the `max-width: 100%` rule to `width: 100%` on the `labeled-icon-wrapper` class. 

Here's what it looks like (`labeled-icon-wrapper` is in blue, the label span is in red)

**Before**

CleanShot 2021-03-23 at 16 32 16


**After**

CleanShot 2021-03-23 at 16 31 25

sairina:
@jonboiser You are amazing! 🥳  The only thing that didn't get updated with that was the user dropdown menu on the upper right-hand side under "Change language"! 

Screen Shot 2021-03-23 at 9 12 12 PM

jonboiser:
I think the word-wrapping on "Change Language" is fine, because there wouldn't be enough space in that CoreMenu. So the line is breaking in the natural way.

I did see just one possible regression in Kolibri in the Facility Syncing page. I'll make a follow up issue on that. Otherwise, I think this change will fix the majority of Safari issues in Kolibri!

jonboiser (approved):

Issue #70, opened 2020-07-08
examples in docs don't show proper keyboard focus highlighting


indirectlylit:
The docs intentionally have their own styling and focus highlighting behaviors independent of the library code. However the examples don't render properly:

![2020-07-08 08 34 53](https://user-images.githubusercontent.com/2367265/86939392-5e9c2280-c0f6-11ea-8437-de0ce7809dc9.gif)

The examples should use proper focus highlighting:

![2020-07-08 08 39 20](https://user-images.githubusercontent.com/2367265/86939623-a91d9f00-c0f6-11ea-9c4c-e32596556014.gif)

indirectlylit:
related: https://github.com/learningequality/kolibri-design-system/issues/131

PR #200, opened 2021-03-19
Show proper keyboard focus ring in KDS examples in documentation

MisRob:
## Description

Track input modality on documentation pages because `KThemePlugin` used by KDS components in documentation examples needs this information to be able to decide if a keyboard focus ring should be displayed (input modality information is used only to inform KDS components, not the documentation pages themselves).

See commit messages for details.

#### Issue addressed

Addresses #70

### After

kds

Because I made some updates to a code that is used in projects that import KDS, I also checked that there are no changes in Kolibri by linking the updated KDS package locally to Kolibri:

kolibri

## Steps to test

1. Use the keyboard to navigate documentation

I added a new option to the `trackInputModality` that allows us to not disable the default focus ring and then imported `trackInputModality` to documentation pages using this new setting so that there is a correct input modality saved in `KThemePlugin`'s `globalThemeState.inputModality` that is then used in KDS components in documentation examples.

An alternative approach to adding a new `disableDefaultFocusRing` parameter would be to remove style with ID `"disable-focus-ring"` as suggested in the documentation of [alice/modality](https://github.com/alice/modality), a library that's a template for our `trackInputModality` script:

> This is added in a <style> element with the ID "disable-focus-ring", to allow easy removal if different behavior is desired.

I consider configuration through a parameter easier.

~~I still need to solve [a problem with logging](https://github.com/learningequality/kolibri-design-system/pull/200/files#r597740463) in this PR.~~

~~Other than that,~~ nothing that I am aware of related to this particular task. However, we might consider moving to [`:focus-visible` ](https://matthiasott.com/notes/focus-visible-is-here) and [`focus-visible` polyfill](https://github.com/WICG/focus-visible) instead of the solution based on alice/modality that has been deprecated in favor of the focus-visible polyfill. That would require updates of `KThemePlugin` and Kolibri testing.

## Reviewer guidance

Please follow the steps to test

### DIFF
--- a/lib/styles/trackInputModality.js
+++ b/lib/styles/trackInputModality.js
@@ -10,7 +10,7 @@ import globalThemeState from './globalThemeState';
 
 const logging = logger.getLogger(__filename);
 
-function setUpEventHandlers() {
+function setUpEventHandlers(disableDefaultFocusRing) {
   let hadKeyboardEvent = false;
   const keyboardModalityWhitelist = [
     'input:not([type])',
@@ -76,7 +76,9 @@ function setUpEventHandlers() {
     return triggers;
   };
 
-  disableFocusRingByDefault();
+  if (disableDefaultFocusRing) {
+    disableFocusRingByDefault();
+  }
 
   document.body.addEventListener(
     'keydown',
@@ -113,9 +115,18 @@ function setUpEventHandlers() {
   );
 }
 
-export default function trackInputModality() {
+/**
+ * Update `globalThemeState.inputModality` to true/false based on
+ * whether the user is currently navigating with the keyboard or not.
+ *
+ * @param {Boolean} [disableDefaultFocusRing=true] Set focus outline to none
+ *                                                 when not navigating with keyboard
+ */
+export default function trackInputModality({ disableDefaultFocusRing = true } = {}) {
   // skip for SSR
   if (typeof document !== 'undefined') {
-    document.addEventListener('DOMContentLoaded', setUpEventHandlers);
+    document.addEventListener('DOMContentLoaded', function() {
+      setUpEventHandlers(disableDefaultFocusRing);
+    });
   }
 }
--- a/lib/styles/trackInputModality.js
+++ b/lib/styles/trackInputModality.js
@@ -7,7 +7,7 @@
 
 import globalThemeState from './globalThemeState';
 
-function setUpEventHandlers(disableDefaultFocusRing) {
+function setUpEventHandlers(disableFocusRingByDefault) {
   let hadKeyboardEvent = false;
   const keyboardModalityWhitelist = [
     'input:not([type])',
@@ -54,7 +54,7 @@ function setUpEventHandlers(disableDefaultFocusRing) {
     return triggers;
   };
 
-  if (disableDefaultFocusRing) {
+  if (disableFocusRingByDefault) {
     const css = 'body :focus { outline: none; }';
     const head = document.head || document.getElementsByTagName('head')[0];
     const style = document.createElement('style');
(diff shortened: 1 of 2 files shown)


lib/styles/trackInputModality.js
```diff
@@ -45,7 +45,7 @@ function setUpEventHandlers() {
       return el.msMatchesSelector;
     }
 
-    logging.error("Couldn't find any matchesSelector method on document.body.");
+    // logging.error("Couldn't find any matchesSelector method on document.body.");
```
MisRob, line 48: I temporarily removed logging in the last commit (and pushed this change so that the demo server can be built). The commit will be removed, and I will change this PR from draft to ready to review after I solve the following problem:

After importing `trackInputModality` [here](https://github.com/learningequality/kolibri-design-system/pull/200/files#diff-ab51280bfd8e8d29145c3345af86ef5ef6a52d0b6d676984adbb047a6d43486bR3), where building documentation, I am getting:

```
This dependency was not found
* kolibri.lib.logging in ./lib/styles/trackInputModality.js
```

I tried to fix it by moving `kolibri-tools` to `dependencies` instead of `devDependencies` in *package.json*, without success. Also, I am not sure if it's desirable to bundle `kolibri-tools` with documentation.

Any ideas on how to deal with this.

lib/styles/trackInputModality.js
```diff
@@ -45,7 +45,7 @@ function setUpEventHandlers() {
...
   })();
 
   const disableFocusRingByDefault = function() {
```
MisRob, line 51: I think we could remove `":not([modality=keyboard])"` from the line 52. It seems that [it was copied from alice/modality](https://github.com/alice/modality/blob/master/src/keyboard-modality.js#L29) which [updates this attribute](https://github.com/alice/modality/blob/master/src/keyboard-modality.js#L65) but we don't do that. Instead, we [update `globalThemeState.inputModality`](https://github.com/learningequality/kolibri-design-system/blob/v0.2.x/lib/styles/trackInputModality.js#L101) and use that information in our components.

Is it okay to remove it, or were there any plans to use it?

indirectlylit: yes - should be fine to remote it

lib/styles/trackInputModality.js
indirectlylit: The easiest thing to do is to just delete these logging-related statements. They are a hold-over from when the file was in the Kolibri codebase

For logging in docs code we provide the `consola` library, but that's separate from lib code.

lib/styles/trackInputModality.js
indirectlylit: yes - should be fine to remote it

lib/styles/trackInputModality.js
```diff
@@ -76,7 +76,9 @@ function setUpEventHandlers() {
     return triggers;
   };
 
-  disableFocusRingByDefault();
+  if (disableDefaultFocusRing) {
+    disableFocusRingByDefault();
+  }
```
indirectlylit, line 81: I'm a bit confused by the naming and intent here: "disable default focus ring" and "focus ring disabled by default" seem like two different concepts.

indirectlylit (commented):
This looks great! I am a little confused about the implementation and added a question below, but I don't think it's super important as long as things work as expected.

I definitely agree that moving to `:focus-visible` would be preferable.

lib/styles/trackInputModality.js
MisRob: Thank you, that's a language nuance I haven't recognized. I just need to find a name for the newly added parameter that would say:
If `<parameter>` is `true`, call `disableFocusRingByDefault`
If `<parameter>` is `false`, don't call `disableFocusRingByDefault`

Do you have any ideas for naming that's easy to understand?

indirectlylit: I actually don't understand why this has a separate function defined inline. I think it could just be:

```js
function setUpEventHandlers(disableFocusRingByDefault) {

  // ...

  if (disableFocusRingByDefault) {
    const css = 'body:not([modality=keyboard]) :focus { outline: none; }';
    const head = document.head || document.getElementsByTagName('head')[0];
    const style = document.createElement('style');

    style.type = 'text/css';
    style.id = 'disable-focus-ring';

    if (style.styleSheet) {
      style.styleSheet.cssText = css;
    } else {
      style.appendChild(document.createTextNode(css));
    }

    head.insertBefore(style, head.firstChild);
  }

  // ...
}
```

and the

MisRob: That's a good point, I think the only reason might be that it was copied from the original library in this way.

MisRob:
Thank you @indirectlylit, I've pushed all updates and tested it on Kolibri again

indirectlylit (approved):
looks good!

Filed a (pre-existing) follow-up issue here: https://github.com/learningequality/kolibri-design-system/issues/206

sairina:
> I definitely agree that moving to `:focus-visible` would be preferable.

Just a note - something I noticed while reading the [docs](https://github.com/WICG/focus-visible) that may affect our decision to move to `:focus-visible` in the future:

> NOTE: this means that HTML elements like <input type={text|email|password|...}> or <textarea> will always match the :focus-visible selector, regardless of whether they are focused via a keyboard or a mouse.

indirectlylit:
Interesting – that's a good thing to watch out for. It is probably not our desired behavior but would require investigation.

Issue #182, opened 2021-02-26
Would be nice if sidebar stayed static during navigation


nucleogenesis:
## Product

KDS

## Desired behavior

When you click something in the sidebar you lose its state on navigation so if you scrolled way down and are specifically interested in the K*Components you have to scroll every time you want to change pages.

## (Optional) The Value Add

Fairly minor improvement to UX for team implementing KDS.

indirectlylit:
yup - same as https://github.com/learningequality/kolibri-design-system/issues/17

> Fairly minor improvement to UX

major IMO! :)  taking a quick look now

PR #209, opened 2021-03-26
persist nav state

indirectlylit:
## Description

Persist side nav bar scroll position and filter state using `Window.sessionStorage`, which should be per-tab.

#### Issue addressed

fixes https://github.com/learningequality/kolibri-design-system/issues/182

### Before/after screenshots

| before | after |
|--|--|
|  |  |

## Steps to test

Experiment with navigation through the design system and see how it feels

* Persists changes to state
* Loads state on `mounted`

* Tries to keep the nav hidden until state has been loaded but doesn't always work as intended
* There's a brief flash during page load

## Reviewer guidance

### DIFF
--- a/docs/common/DocsPageTemplate/SideNav/index.vue
+++ b/docs/common/DocsPageTemplate/SideNav/index.vue
@@ -6,20 +6,22 @@
       class="sidenav"
       @scroll="throttleHandleScroll"
     >
-      <h1 class="header">
-        <file-svg src="./kolibri-logo.svg" class="logo" />
-        <span class="header-text">Design System</span>
-      </h1>
+      <template if="loaded">
+        <h1 class="header">
+          <file-svg src="./kolibri-logo.svg" class="logo" />
+          <span class="header-text">Design System</span>
+        </h1>
 
-      <DocsFilter v-model="filterText" />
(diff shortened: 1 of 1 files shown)


indirectlylit:
Thanks for testing!

> One thing I noticed was I found the search filter persisting after I hit the back button to be a bit surprising. I don't even know why this stood out to me - I just expected the filter to clear.

That's interesting, thanks. I debated doing it both ways and this seemed convenient for navigating within a group of related items ('button' was the filter term I used a lot.)

Let's raise it as a question/tension in the next tactical.

indirectlylit:
One idea: could persist [trailing debounced changes](https://lodash.com/docs/#debounce) (wait time ~2s) to the filter as query string changes and push them onto the browser history stack.

For example:

| URL | action |
|--|--|
| `/` | first URL|
| `/?filter=button` | after enter 'button' into ToC filter |
| `/kbutton?filter=button` | after clicking `KButton` entry |
| `/?filter=button` | browser back |
| `/` | browser back |

nucleogenesis:
<3

indirectlylit:
created follow-up issue here: https://github.com/learningequality/kolibri-design-system/issues/213

Issue #86, opened 2020-07-29
KIconButton - `tooltip` prop became required in a minor release


nucleogenesis:
@indirectlylit made this prop required in a minor release, which broke existing users.

Let's follow up in a major release in https://github.com/learningequality/kolibri-design-system/issues/168

jonboiser:
Right now, it _is_ required, and causing a lot prop validation errors on Kolibri. Should this be closed?

https://github.com/learningequality/kolibri-design-system/blob/v0.2.x/lib/buttons-and-links/KIconButton.vue#L88-L90

indirectlylit:
I broke this - will look at it tomorrow

ref: https://learningequality.slack.com/archives/CTG8XDTCL/p1616021252011600

PR #165, opened 2021-02-17
Button docs

indirectlylit:
# Component documentation

Adds documentation for library components:

* `KButton`
* `KRouterLink`
* `KExternalLink`
* `KButtonGroup`
* `KIconButton`

# Filter keywords

Adds a new 'keywords' property to the table of contents to allow filtering by related terms. For example:

```js
      new Page({
        path: '/KButton',
        title: 'KButton',
        isCode: true,
        keywords: ['button', 'link'],
      }),
```

# Cross-links to components

Adds a new `DocsLibraryLink` component to allow easy cross-linking. For example:

```vue
<DocsLibraryLink page="KExternalLink" /> is used to create standard <code>&lt;a&gt;</code> tags with an <code>href</code> attribute.
```

PR #218, opened 2021-04-12
revert prop change

indirectlylit:
## Description

reverts one change in https://github.com/learningequality/kolibri-design-system/pull/165 which caused many prop warnings

#### Issue addressed

fixes https://github.com/learningequality/kolibri-design-system/issues/86

## Steps to test

upgrade kolibri and verify that prop validation errors are gone

### DIFF
--- a/lib/buttons-and-links/KIconButton.vue
+++ b/lib/buttons-and-links/KIconButton.vue
@@ -80,14 +80,14 @@
        */
       ariaLabel: {
         type: String,
-        default: null,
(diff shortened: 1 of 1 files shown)


MisRob (approved):
Thank you @indirectlylit.
The code looks good and there is no tooltip error in Kolibri anymore.

indirectlylit:
follow-up: ttps://github.com/learningequality/kolibri-design-system/issues/168

Issue #214, opened 2021-04-07
Update KContentRenderer mixin to better support custom completion criteria


rtibbles:
* Add `timeSpent` as an optional prop for the KContentRenderer mixin
* Add `duration` as nullable prop. If prop is null, then renderer must define 'durationDefault' data item to set a default value.
* Add reusable progress by time spent computed prop to KContentRenderer mixin
  * Simplifying audio and video progress tracking by just using clock time instead of media track time
  * Allowing for unified time based progress tracking for any content type, based on `duration` or `durationDefault`
* Add prop that presents a Boolean for 'always use duration based progress tracking' - `forceTimeBasedProgress` - which reads from the contentnode options `force_time_based_progress` flag.

rtibbles:
Fixed in #224

PR #223, opened 2021-04-21
work on documenting component api

indirectlylit:
* Removes or enables all disabled ToC items
* Adds many component pages
* Improves component jsdocs

PR #224, opened 2021-04-21
Adds props/computed props to mixin for KContentRenderer

sairina:
## Description

Adds `timeSpent`, `duration` props and `forceTimeBasedProgress`, `durationBased` computed props so that we can refactor Kolibri content renderers to use unified time-tracking.

#### Issue addressed

Addresses #214 
## Steps to test

Currently, to test this PR, you'll have to use learningequality/kolibri#8037 in order to make sure the new props and computed props are working (see tech-debt items below for details).

Added props and computed props to the mixin used by `KContentRenderer`.

There are currently no tests in KDS to address the use of these props and computed props; however, tests exist on the individual renderers in Kolibri.

## Reviewer guidance


lib/content/mixin.js
```diff
@@ -130,8 +130,24 @@ export default {
       type: Boolean,
       default: false,
     },
+    timeSpent: {
+      type: Number,
+      default: null,
```
rtibbles, line 135: Looking at the implementation, maybe for simplicity we should make this default to `0`? That way the `durationBasedProgress` will just be `0` when `timeSpent` is not specified.

rtibbles (commented):
Looks good - thinking it will be useful to implement the solution to https://github.com/learningequality/kolibri/issues/7970 in parallel to this to make sure it has everything it needs?

indirectlylit:
FYI in my PR #223 I added support for extracting JSDocs from the `KContentRenderer` and displaying them as a normal component:

image

I did not try documenting the props and method yet

lib/content/mixin.js
```diff
@@ -130,8 +130,28 @@ export default {
...
+      default: 0,
+    },
+    duration: {
+      type: Number,
+      default: null,
+    },
   },
   computed: {
+    // For when we want to force a renderer to use time-based progress (e.g. instead of % completed)
+    forceTimeBasedProgress() {
```
rtibbles, line 144: A thought that occurred to me while reviewing your other PR - maybe we should make `forceDurationBasedProgress` here, so that the names match?

lib/content/mixin.js
```diff
@@ -130,8 +130,28 @@ export default {
...
+      return this.options.force_time_based_progress || false;
```
rtibbles, line 145: Likewise `this.options.force_duration_based_progress`

rtibbles (commented):
Realized a possible mistake in the original spec while reviewing!

rtibbles (approved):

indirectlylit:
looks great!

Just a reminder to file an issue for writing prop jsdocs

PR #225, opened 2021-04-23
Don't try to load non-existent css files

rtibbles:
## Description
* Stops trying to load css files that don't exist
* This can happen in hot module replacement mode when CSS files are bundled into Javascript

## Steps to test

1. Run Kolibri with this patch on KDS, with `devserver-hot` and check that the EpubRenderer now properly loads.

## Reviewer guidance


sairina (approved):
LGTM

PR #227, opened 2021-05-05
Fix keyboard outline in Kolibri - no longer showing unless keyboard navigating

nucleogenesis:
#### Issue addressed

Fixes https://github.com/learningequality/kolibri/issues/7974

## Steps to test

*Kolibri*

- `yarn link` the design system (see [KDS Readme](https://github.com/learningequality/kolibri-design-system/) for help there)
- Run Kolibri and test that the outline comes up when you're keyboard navigating but does not come up when you're hovering with a mouse.

*KDS*

- `yarn dev` the KDS documentation (or just use the Preview link below at the GIthub Actions)
- Make sure that you see the keyboard navigation box when navigating by keyboard there.

[This commit is where the bug was introduced.](https://github.com/learningequality/kolibri-design-system/commit/e26893fcc6d970c11ca86b771cdc24658892c569) @MisRob undoing this fixes the issue - was there some other effect you expected to see when making this change?

[Reverts this commit.](https://github.com/learningequality/kolibri-design-system/commit/e26893fcc6d970c11ca86b771cdc24658892c569)

### DIFF
--- a/lib/styles/trackInputModality.js
+++ b/lib/styles/trackInputModality.js
@@ -55,7 +55,7 @@ function setUpEventHandlers(disableFocusRingByDefault) {
   };
 
   if (disableFocusRingByDefault) {
-    const css = 'body :focus { outline: none; }';
+    const css = 'body :focus:not([modality=keyboard]) { outline: none; }';
     const head = document.head || document.getElementsByTagName('head')[0];
     const style = document.createElement('style');
 


indirectlylit:
> was there some other effect you expected to see when making this change?

Looks like this was discussed and I OK'd the change here: https://github.com/learningequality/kolibri-design-system/pull/200#discussion_r597752633

Clearly I didn't (don't) fully understand how keyboard modality state is tracked and referenced. However it seems that we do in fact need both a JS and CSS-based state

MisRob:
Thank you for fixing this @nucleogenesis, I've had to misunderstand something. I still don't get how come that we rely on the attribute that I had removed but if the commit revert helps with the bug, there will be some dependency in Kolibri. Yes, I wouldn't expect any regressions.

MisRob:
Wonderin the problem might be rather some CSS specificity issue that has shown up after the selector update rather than the need to have the attribute there, because I think we actually don't update or use the attribute itself.

PR #239, opened 2021-06-02
Add navigateTo listener in KContentRenderer mixin for HTML5 hashi work

marcellamaki:
## Description

Add navigateTo listener in KContentRenderer mixin for HTML5 hashi work.

## Steps to test

Please see this [kolibri PR for extensions to the navigateTo work](https://github.com/learningequality/kolibri/pull/8134).

### DIFF
--- a/lib/content/mixin.js
+++ b/lib/content/mixin.js
@@ -8,7 +8,7 @@ const interactionEvents = [
   'updateProgress',
   'updateContentState',
   'startTracking',
-  'navigateto'
+  'navigateto',
 ];
 
 const fileFieldMap = {
--- a/lib/content/mixin.js
+++ b/lib/content/mixin.js
@@ -8,6 +8,7 @@ const interactionEvents = [
   'updateProgress',
   'updateContentState',
   'startTracking',
+  'navigateto'
 ];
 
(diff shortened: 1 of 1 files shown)


sairina (approved):
LGTM. Seems to be working in the associated [Kolibri PR](https://github.com/learningequality/kolibri/pull/8134).

lib/content/mixin.js
```diff
@@ -8,6 +8,7 @@ const interactionEvents = [
   'updateProgress',
   'updateContentState',
   'startTracking',
+  'navigateto',
```
rtibbles, line 11: You called your hashi event `navigateto`, but the event you are emitting from the content renderer (which is relevant to this) was called `navigateToRegularContext` https://github.com/learningequality/kolibri/pull/8134/files#diff-cf6ea11127dacf9dddc24383c075d4e76547563841df759b5bbf0962c9293719R127

I certainly prefer `navigateTo` but perhaps with camelCase to be consistent.

rtibbles (changes requested):
Mismatch between this PR and the Kolibri one.

rtibbles (approved):

PR #248, opened 2021-06-23
Update PR template for KDS release process

sairina:
## Description

Updates the PR template with a section for the PR-writer to add in the `changelog` and also to manually point the most recent version of KDS to the Kolibri `develop` branch.

#### Issue addressed

Addresses #247

We will need to make sure that there is clear documentation about how to add to the `changelog` (this may be something we can add to the PR template in the future so that anyone can just access it). For now, we can just tell folks how to do it (internally). Same goes with how to point Kolibri to the right branch of KDS. 

## Reviewer guidance

## Comments

Doesn't need to be said, but all feedback appreciated, of course.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -33,7 +33,6 @@ Addresses #*PR# HERE*
 - [ ] If there are any front-end changes, before/after screenshots are included
 - [ ] Critical and brittle code paths are covered by unit tests
 - [ ] The change has been added to the `changelog`
-- [ ] A separate PR has been submitted to point the `unstable` branch of the affected Kolibri product to the most recent version of KDS (see post-merge updates for details)
 
 ## Reviewer guidance
 <!-- Delete anything that doesn't apply so your reviewer knows what to check for -->
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -33,7 +33,9 @@ Addresses #*PR# HERE*
 - [ ] If there are any front-end changes, before/after screenshots are included
 - [ ] Critical and brittle code paths are covered by unit tests
 - [ ] The change has been added to the `changelog`
-- [ ] A PR has been submitted to point the Kolibri `develop` branch to the most recent version of KDS
+
+## Updating KDS in Kolibri
+- [ ] After this PR has been merged into KDS, a separate PR has been submitted to point the Kolibri `develop` branch to the most recent version of KDS
 
 ## Reviewer guidance
 <!-- Delete anything that doesn't apply so your reviewer knows what to check for -->
@@ -41,6 +43,7 @@ Addresses #*PR# HERE*
 - [ ] Is the code clean and well-commented?
 - [ ] Are there tests for this change?
 - [ ] Are all UI components LTR and RTL compliant (if applicable)?
+- [ ] Is the PR ready to be merged in KDS and is a PR ready to be submitted to Kolibri to point it to the most recent KDS hash?
 - [ ] _Add other things to check for here_
 
 ## Comments
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -33,9 +33,7 @@ Addresses #*PR# HERE*
 - [ ] If there are any front-end changes, before/after screenshots are included
 - [ ] Critical and brittle code paths are covered by unit tests
 - [ ] The change has been added to the `changelog`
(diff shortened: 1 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -23,16 +21,21 @@ Addresses #*PR# HERE*
...
 
+## Testing checklist
+<!-- Complete the checklist before submitting a PR; delete anything that doesn't apply -->
 
-## Reviewer guidance
+- [ ] Contributor has fully tested the PR manually
+- [ ] If there are any front-end changes, before/after screenshots are included
+- [ ] Critical and brittle code paths are covered by unit tests
+- [ ] The change has been added to the `changelog`
+- [ ] A PR has been submitted to point the Kolibri `develop` branch to the most recent version of KDS
```
indirectlylit, line 36: Should we require similar PRs for Studio and KDP?

If that's too onerous, maybe we should actually be looking into auto-generated PRs based on github actions?

sairina: Hmm. Good point. Should we ask the folks in charge of those projects (or at least the release managers)?

indirectlylit: posted in Slack here: https://learningequality.slack.com/archives/C01PBEMPZC7/p1625172143079600

sairina: That makes a lot of sense! I'll update it to reflect that.

indirectlylit: I'm a little confused on one detail:

If some KDS PR has not been merged yet, what commit would the product update PRs point at? I was expecting that the update PRs would point to the merge commit of the KDS PR.

MisRob: That makes sense but then the question is if it should be in "Reviewer guidance" of the PR template when it's not possible to check it during the review process itself

MisRob: (I wrote my first note assuming that it's something to be done during the review)

indirectlylit: Ah, that's a good point.

Maybe we should add a post-merge checklist? e.g. something like...

> # Post-merge updates
> 
> After merging, unstabled branches of Kolibri products (Learning Platform, Studio, and Data Portal) should be updated to point at the merge commit resulting from this PR. This process should be led by the submitter of the Design System PR in collaboration with other LE team members working on the other product repos.
>  
> * [ ] Learning Platform update PR is submitted
> * [ ] Studio update PR is submitted
> * [ ] Data Portal update PR is submitted
> * [ ] Learning Platform update PR is merged
> * [ ] Studio update PR is merged
> * [ ] Data Portal update PR is merged


thoughts?

MisRob: Makes much sense to me

MisRob: I appreciate that each product has its own item on the list, could be helpful to keep track of what still needs to be done.

indirectlylit: FYI, after discussing in Product Tactical, @rtibbles decided that we should not consider the "merged" items mandatory, at least in the shorter-term. We can still include the full list, but the most important things are

1. keeping the changelog up-to-date
2. opening PRs against target products for visibility

Newer updates should supersede older ones in the case where multiple KDS have happened before a product is updated with the latest.

(Longer-term, we can automate opening of PRs with github actions, similar to dependabot)

indirectlylit: We may want to update the template text to clarify the above

.github/pull_request_template.md
```diff
@@ -33,18 +33,24 @@ Addresses #*PR# HERE*
 - [ ] If there are any front-end changes, before/after screenshots are included
 - [ ] Critical and brittle code paths are covered by unit tests
 - [ ] The change has been added to the `changelog`
-
-## Updating KDS in Kolibri
-- [ ] After this PR has been merged into KDS, a separate PR has been submitted to point the Kolibri `develop` branch to the most recent version of KDS
+- [ ] A separate PR has been submitted to point the `unstable` branch of the affected Kolibri product to the most recent version of KDS (see post-merge updates for details)
```
indirectlylit: Is this item intended to be checked before or after the post-merge updates? It seems to describe similar tasks as

```
```

sairina: Ah yes. Let's get rid of this.

indirectlylit (approved):
awesome, thank you!

PR #251, opened 2021-07-02
Upgrade Google Material Icons

MisRob:
## Description

This PR upgrades `@material-icons/svg` package (Google Material Icons fork) from `0.0.1` to its latest version `1.0.12` so that we have access to some new icons that are needed for new products features and are missing in the older version (I will expose them in a different PR). It also contains updated precompiled Vue svg icons files and reStructuredText replacement strings.

#### 1. The filled `copy`/`duplicate` and the outlined `copyToClipboard` is merged into one outlined `copy` (CHANGELOG)

| Before | After |
| --------- | ------ |
| Screenshot from 2021-07-02 13-35-03 | Screenshot from 2021-07-02 13-35-15 |
| Screenshot from 2021-07-02 13-35-09 | Screenshot from 2021-07-02 13-35-15 |

The former filled `copy`/`duplicate` icon wasn't filled anymore after the `@material-icons/svg` package upgrade and looked exactly the same as the outlined `copyToClipboard` icon, which means that the outlined icon newly represented all three aliases. @jtamiace confirmed that it's fine to deprecate the filled icon in KDS public icons set too. In addition to that, we will only keep `copy` alias for this icon which is the most commonly used alias in our products.

@radinamatic This also means that `duplicate` and `copyToClipboard` icons won't be available in reStructuredText replacement strings anymore and should be replaced by `copy` icon.

#### 2. `domain` and `facility` icons have been merged into one icon with `domain` and `facility` aliases

It seems that previously, there were some, hard to perceive, differences in svgs that are not there anymore, so the same icon is indeed the same icon now :)

| Before | After |
| --------- | ------ |
| Screenshot from 2021-07-02 14-08-32 | Screenshot from 2021-07-02 14-08-48 |
| Screenshot from 2021-07-02 14-08-40 | Screenshot from 2021-07-02 14-08-48 |

#### 3. Minor updates in svg content of `email`, `settings`, `indeterminateCheck`, `lesson`, and `hint` icons

I haven't noticed any visual differences for these.

### Updated KDS icons documentation page

kds-icons
:

## Reviewer guidance


MisRob:
@indirectlylit Where would be a good place to save a note about "1. The filled copy/duplicate and the outlined copyToClipboard is merged into one outlined copy"? I assume that at the end of the day this will come to the GitHub Releases page and KDS version history page but I am not sure where to log updates before we have a new release.

indirectlylit:
> domain and facility icons have been merged into one icon with domain and facility aliases

What kind of 'domain' is this referring to? This seems like a mistake, if we're using the facility icon for other reasons.

indirectlylit:
>  Where would be a good place to save a note about "1. The filled copy/duplicate and the outlined copyToClipboard is merged into one outlined copy"? 

please add these notes to the changelog page:

* https://github.com/learningequality/kolibri-design-system/blob/v0.2.x/docs/pages/versions.vue
* https://design-system.learningequality.org/versions/

MisRob:
@indirectlylit  What kind of 'domain' is this referring to? This seems like a mistake, if we're using the facility icon for other reasons.

I don't know how we use all icons, but taking into account that we need to define them in `iconsDefinitions` file manually, my guess would be that it's been exposed on purpose.

MisRob:
> please add these notes to the changelog page:

@indirectlylit 
Okay, where should it go? In the first section 0.2.1?

indirectlylit:
> I don't know how we use all icons, but taking into account that we need to define them in iconsDefinitions file manually, my guess would be that it's been exposed on purpose.

I don't think that it was exposed on purpose with the token `domain`, and I believe this token is no longer used anywhere.

In Material Icons, 'business' and 'domain' are aliases to the same icon:

image

The `domain` version in icon definitions has the comment `//FacilityTopNav`:

https://github.com/learningequality/kolibri-design-system/blob/efbbf8abd81b86b487f8f1fdd239fb4d821e7c9b/lib/KIcon/iconDefinitions.js#L258-L260

However, the token is not used in there anymore. It was removed when we transitioned this icon from representing 'classes' to its current meaning of 'facility':

https://github.com/learningequality/kolibri/commit/219e74e27820c79a11154a27b964ae22188c2de7

Long story short, I think we can safely drop the `domain` alias and always just use `facility`.


>  Okay, where should it go? In the first section 0.2.1?

That's fine. I need to work with @sairina at some point to get the changelog fully up-to-date with all the changes we haven't logged.

MisRob:
Thanks @indirectlylit , I removed `domain` alias and updated the versions page.

docs/pages/versions.vue
```diff
@@ -22,6 +22,15 @@
         <li>Added 'email', 'sidebar', and 'add' icons - <DocsGithubLink pull="110" /></li>
         <li>Updated 'clipboard' icon - <DocsGithubLink pull="121" /></li>
         <li>Fixed z-order bug in icon button - <DocsGithubLink pull="95" />, <DocsGithubLink pull="122" /></li>
+        <li>
+          The filled icon with <code>copy</code> and <code>duplicate</code> aliases
+          and the outlined <code>copyToClipboard</code> merged into one outlined icon
+          with <code>copy</code> alias - <DocsGithubLink pull="251" />
```
indirectlylit, line 28:
```suggestion
          The filled icon with <code>copy</code> and <code>duplicate</code> aliases and the outlined <code>copyToClipboard</code> merged into one outlined icon with <code>copy</code> alias - <DocsGithubLink pull="251" />
```

docs/pages/versions.vue
```diff
@@ -22,6 +22,15 @@
...
+        </li>
+        <li>
+          Removed <code>domain</code> alias from the icon that had previously had two
+          <code>domain</code> and <code>facility</code> aliases - <DocsGithubLink pull="251" />
```
indirectlylit, line 32:
```suggestion
          Removed <code>domain</code> alias from the icon that had previously had two <code>domain</code> and <code>facility</code> aliases - <DocsGithubLink pull="251" />
```

indirectlylit (commented):
Very minor, but for paragraphs in `/pages` I disabled the text length rule and use soft wrapping in the text editor.

This makes editing long-form documentation easier because we don't need to manage endline placement.

indirectlylit (approved):
Tested locally and looks good!

The final step (if necessary) will be to create any PRs against Studio/Kolibri/KDP with the updates necessary to make sure the copy alias changes don't cause breakage.

thank you!

MisRob:
@indirectlylit I've checked KDP, KDS, and Studio, and no updates are needed there in regards to this PR.

PR #252, opened 2021-07-13
Add new icons

MisRob:
## Description

**1. Expose the following Google Material Icons**

Screenshot from 2021-07-13 15-54-00

Screenshot from 2021-07-13 15-55-04

Screenshot from 2021-07-13 15-55-41

Screenshot from 2021-07-13 15-56-22

**2. Add the a11y custom icon**

Screenshot from 2021-07-13 15-57-57

---

As this PR only adds new icons, no updates in Kolibri, KDP, and Studio are needed.

#### Issue addressed

We need the material icons in Kolibri new features and the a11y icon has been added on @radinamatic's request.

## Steps to test

Check [the icons documentation](https://deploy-preview-252--kolibri-design-system.netlify.app/icons/#icons)

## Reviewer guidance

- [ ] Do icon aliases make sense?


custom-icons/a11y.svg
```diff
@@ -0,0 +1,11 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
MisRob, line 1: @radinamatic I minified the SVG file you provided because it contained a lot of Inkscape-specific information.

MisRob:
@khangmach Could you please check if these icon aliases make sense for designers too?

khangmach:
> @khangmach Could you please check if these icon aliases make sense for designers too?

They all seem fine to me. My only request is to change "forwardRound" to "forwardRounded" to be more specific about the styling difference between rounded vs nonrounded

indirectlylit:
For the Kolibri Design System token names we've been making the names refer to their usage in the LE ecosystem rather than mirroring the Material Design tokens. This helps us use them consistently for similar concepts in the future.

What meaning do the alternative route and fat rounded forward arrow have in Kolibri?

MisRob:
I'd say that the meaning of an `alternativeRoute` and `forwardRound` (to become `forwardRounded`) that I'd like to use on [the completion modal](https://github.com/learningequality/kolibri/issues/8109) is in this case, what it says. Regarding `rounded` part of `forwardRounded`, we need to distinguish it from another icon, called `forward`.

Screenshot from 2021-07-14 07-12-41

indirectlylit:
cool, sounds good.

We have some subset of icons ("facility", "user", etc) which roughly map to [concepts in the glossary](https://design-system.learningequality.org/glossary/).

These icons seem a bit more decorative – they don't map to a specific Kolibri-specific concept that we expect users to remember and learn, so the more generic tokens work.

MisRob:
Good to know, thanks for sharing your thoughts on this @indirectlylit

MisRob:
I've just renamed `forwardRound` to `forwardRounded`. If there are no more suggestions, this could be ready to go.

khangmach (approved):

PR #258, opened 2021-09-01
Bump kolibri-tools version to 0.15.0-dev.4

nucleogenesis:
## New description from @indirectlylit 

I've updated this PR as follows:

* Removed references in _package.json_ to dependencies that are also brought in by `kolibri-tools` in order to minimize conflicting versions
* Pinned `nuxt` to an older version that has dependencies (such as babel) which are relatively closely aligned with `kolibri-tools`. When `kolibri-tools` upgrades we can also upgrade `nuxt`
* In some cases, `nuxt` and `kolibri-tools` still specify conflicting versions of dependencies, and this was causing errors with webpack build configuration. To resolve this, added a new `resolutions` section to _package.json_ in order to force usage of a compatible dependency:
    ```json
      "resolutions": {
        "@babel/preset-env": "7.12.17",
        "postcss-loader": "3.0.0",
        "css-loader": "3.6.0"
      }
    ```
* In an effort to resolve an issues where `vue` was complaining incompatibility with `vue-template-compiler`:
  *  Deleted and re-generated _yarn.lock_ 
  * Updated to a new `dev.4` version of `kolibri-tools` which also upgrades Vue to 2.6.14, again attempting to resolve

----

## Old description from @nucleogenesis 

In order for https://github.com/learningequality/kolibri/pull/8085 and everything after it is merged to pass license checks, this version needs to be updated to no longer depend on the old version of `kolibri-tools` which depended on  `kolibri-components`.

I'm not clear where the release cycle is @sairina @indirectlylit  -  when this is merged will we just point directly to the new HEAD commit on KDS/main?

This is a draft PR until https://github.com/learningequality/kolibri/pull/8085#issuecomment-910683837 is addressed.


indirectlylit:
@sairina @nucleogenesis it looks like we need to merge this before tagging a final 1.0

That said, it certainly is a bit weird that we have KDS depending on Kolibri _and_ Kolibri depending on KDS.

nucleogenesis:
@indirectlylit KDS depends on `kolibri-tools` - possibly only for the linting? I guess the way in which I consider `kolibri-tools` independent of Kolibri is the same way I consider KDS independent of Kolibri which makes me wonder if `kolibri-design-system` should live in Kolibri/packages.

I have finally gotten everything correct and published - so now KDS will point at kolibri-tools@0.15-dev.3 which works as far as I can see.

I ran the linter and fixed everything it told me to.

nucleogenesis:
Ah - dang it - will need to come back to this to review the errors:

```
ModuleBuildError: Module build failed (from ./node_modules/postcss-loader/dist/cjs.js):
ValidationError: Invalid options object. PostCSS Loader has been initialized using an options object that does not match the API schema.
 - options has an unknown property 'order'. These properties are valid:
   object { postcssOptions?, execute?, sourceMap?, implementation? }
    at validate (/home/jacob/Code/LE/design-system/node_modules/postcss-loader/node_modules/schema-utils/dist/validate.js:105:11)
    at Object.loader (/home/jacob/Code/LE/design-system/node_modules/postcss-loader/dist/index.js:43:29)
    at runLoaders (/home/jacob/Code/LE/design-system/node_modules/webpack/lib/NormalModule.js:316:20)
    at /home/jacob/Code/LE/design-system/node_modules/loader-runner/lib/LoaderRunner.js:367:11
    at /home/jacob/Code/LE/design-system/node_modules/loader-runner/lib/LoaderRunner.js:233:18
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

indirectlylit:
I'll take a quick look and see if I can build off your work so far – thank you!

> @indirectlylit KDS depends on kolibri-tools - possibly only for the linting? I guess the way in which I consider kolibri-tools independent of Kolibri is the same way I consider KDS independent of Kolibri which makes me wonder if kolibri-design-system should live in Kolibri/packages.

yeah I think there's some truth there. I'll following up on this separately see if we can't plan out a longer-term solution that is a bit easier to grok

rtibbles:
> Ah - dang - will need to come back to this to review the errors later on

Note that we upgraded some _very old_ versions of CSS libraries recently that will have made its way into kolibri-tools now. See here for how we changed specification of postcss options in Kolibri to account for this: https://github.com/learningequality/kolibri/blob/release-v0.15.x/packages/kolibri-tools/lib/webpack.config.base.js#L90

indirectlylit:
Since KDS only depends on `kolibri-tools` for linting and it's correctly specified as a dev dependency, does this PR actually matter for downstream users of KDS? (KDP, Kolibri proper, Studio)

Or is this more of a cleanup task that can wait until after KDS publishes a 1.0

nucleogenesis:
@rtibbles @indirectlylit - I'm not seeing why/how the `kolibri-tools` webpack config has anything to do with why running `yarn dev` on KDS gets the errors noted in the GH Actions w/ PostCSS or `yarn _dev-only` on KDS - the latter doesn't even use the kolibri-tools linting.

This PR is necessary for https://github.com/learningequality/kolibri/pull/8085 to be merged.

That PR has this problem:

- Kolibri depends on KDS HEAD
- KDS HEAD depends on `kolibri-tools@0.13.0-dev.9`
- `kolibri-tools@0.13.0-dev.9` depends on `kolibri-components`
- `kolibri-components` [doesn't exist anymore](https://github.com/learningequality/kolibri/pull/8085/checks?check_run_id=3488504785)

So - once this is merged (and working properly) then I can point Kolibri to the right KDS which no longer will point to a package that depends on a dead package.

indirectlylit:
The part that surprises me is

> KDS HEAD depends on kolibri-tools@0.13.0-dev.9

because it only depends on it as a dev dependency, so I wouldn't have expected that to be installed.


FYI I just pushed a commit updating a bunch of KDS dependencies, trying to get them more in line with the current state of Kolibri 0.15. Locally I'm still seeing this error and looking into it still


```
ERROR in ./.nuxt/client.js
Module build failed (from ./node_modules/babel-loader/lib/index.js):
Error: [BABEL] /Users/d/Projects/le/design-system/.nuxt/client.js: @babel/preset-env: The plugins/built-ins '@babel/plugin-proposal-private-methods, @babel/plugin-proposal-private-property-in-object' passed to the 'exclude' option are not
    valid. Please check data/[plugin-features|built-in-features].js in babel-preset-env (While processing: "/Users/d/Projects/le/design-system/node_modules/@nuxt/babel-preset-app/src/index.js$0")
    at OptionValidator.invariant (/Users/d/Projects/le/design-system/node_modules/@babel/helper-validator-option/lib/validator.js:48:13)
    at expandIncludesAndExcludes (/Users/d/Projects/le/design-system/node_modules/@babel/preset-env/lib/normalize-options.js:49:5)
    at normalizeOptions (/Users/d/Projects/le/design-system/node_modules/@babel/preset-env/lib/normalize-options.js:128:19)
    at _default (/Users/d/Projects/le/design-system/node_modules/@babel/preset-env/lib/index.js:271:37)
    at /Users/d/Projects/le/design-system/node_modules/@babel/preset-env/node_modules/@babel/helper-plugin-utils/lib/index.js:22:12
    at /Users/d/Projects/le/design-system/node_modules/babel-loader/node_modules/@babel/core/lib/config/full.js:199:14
    at Generator.next (<anonymous>)
    at Function.<anonymous> (/Users/d/Projects/le/design-system/node_modules/babel-loader/node_modules/@babel/core/lib/gensync-utils/async.js:26:3)
    at Generator.next (<anonymous>)
    at step (/Users/d/Projects/le/design-system/node_modules/gensync/index.js:254:32)
    at evaluateAsync (/Users/d/Projects/le/design-system/node_modules/gensync/index.js:284:5)
    at Function.errback (/Users/d/Projects/le/design-system/node_modules/gensync/index.js:108:7)
    at errback (/Users/d/Projects/le/design-system/node_modules/babel-loader/node_modules/@babel/core/lib/gensync-utils/async.js:70:18)
    at async (/Users/d/Projects/le/design-system/node_modules/gensync/index.js:183:31)
    at onFirstPause (/Users/d/Projects/le/design-system/node_modules/gensync/index.js:209:13)
    at Generator.next (<anonymous>)
 @ multi ./.nuxt/client.js app[0]

 FATAL  Nuxt build error   
```

rtibbles:
> because it only depends on it as a dev dependency

Possible that installing it from source has different behaviour to installing it as a package for dev dependencies?

> updating a bunch of KDS dependencies

@nucleogenesis this is what I was assuming was the issue, was that there were now duelling versions of postcss (or things that depended on different versions of postcss) that were causing the issue you saw.

rtibbles:
@indirectlylit I think the previous version of kolibri-tools did not depend on babel-preset-env, which is why nuxt is now giving issues.

The dependency collisions you are seeing here definitely make for a compelling case for 'breaking up' kolibri-tools into smaller packages.

indirectlylit:
As a side experiment, I thought if `kolibri-tools` is only being used for linting then why not maybe remove it altogether and just duplicate config files and linter tooling in both repos?

I had to manually install some additional build dependencies that would normally come along for the ride (e.g. `node-sass`), but eventually got things building. However, this does not seem like the correct direction to go in.

rtibbles:
Yeah, it would work - but across kolibri, Studio, KDS, and KDP we have sufficient places where systematically sharing our linting makes sense.

indirectlylit:
Initially I thought 4cc69aa addressed the issue, but now it's still giving the same error remotely and for some reason I can't reproduce locally anymore at all (even after reverting the change).

I'm going to step away from this for a bit if anyone else wants to take a swing at it.

indirectlylit:
Updated description to reflect changes. This seems to be working now. I will either open a follow-up PR on the Kolibri side or take over https://github.com/learningequality/kolibri/pull/8085

indirectlylit:
one good sign is the change in _yarn.lock_ indicating a vast reduction in duplicated and conflicting dependencies:

image

image

Issue #238, opened 2021-05-29
Outer `div` shifts slightly when focus is on `KTextbox`


sairina:
## Product

Kolibri

## Expected behavior

The size of the outer `div` containing the `KTextbox` component should not move.

## Actual behavior

All of the contents in outer `div` containing `KTextbox` shift _slightly_ down every time a focus appears.

## Steps to reproduce the issue
1. Go to any `KTextbox` element
2. Click on it (or use the keyboard to tab into it) and notice the slight shift. It is more noticeable when there is error validation.

## Environment
- OS: Mac
- Browser version: Chrome

## Additional information

@indirectlylit From [here](https://github.com/learningequality/kolibri-design-system/issues/131#issuecomment-790770612):

> One thing I'm seeing is that the vertical layout is shifting slightly on focus – though this might be due to the validation and not the new focus outline. Preferably, neither of these would cause a shift in height.

I double-checked with both clicking by mouse or tabbing with the keyboard and it seems to shift with focus, which I think means that it is validation-related. I will open a separate issue for this problem.

![2021-05-24 15 02 52](https://user-images.githubusercontent.com/13563002/119413392-59aa7800-bca2-11eb-8987-af96a00347ab.gif)

_Originally posted by @sairina in https://github.com/learningequality/kolibri-design-system/issues/191#issuecomment-847385833_

PR #279, opened 2021-11-19
Fix shifting `div` from `KTextbox`

sairina:
## Description

Changes the `v-if` to the `v-show` in the KeenUI textbox so that the element is already in the DOM and we can toggle its `display`.

This is useful because it will keep that `div` below `KTextbox` for error validation, and prevent the UI from "shifting" whether or not there is an error validation for the field. 

#### Issue addressed
- Addresses #238 in KDS
- Addresses https://github.com/learningequality/kolibri/issues/8675
- Must work with: https://github.com/learningequality/kolibri/pull/8746 

### Before/after screenshots
See https://github.com/learningequality/kolibri/pull/8746 

## Steps to test

After linking KDS to Kolibri, see the following for testing: https://github.com/learningequality/kolibri/pull/8746

- We will want to double check that the places in Kolibri that this affects (see: https://github.com/learningequality/kolibri/pull/8746) are the only places where the shifting needs to be fixed.
- How can we document this, if this is how we will continue to deal with this issue in this way? Or, what can be done to be a more permanent fix?

## Testing checklist

## Reviewer guidance

- [ ] Is the code clean and well-commented (commented via `git commit`)


lib/keen/UiTextbox.vue
```diff
@@ -90,7 +90,7 @@
         </div>
       </label>
 
-      <div v-if="showHelp || hasErrorFeedback || maxlength" class="ui-textbox-feedback">
+      <div v-show="showHelp || hasErrorFeedback || maxlength" class="ui-textbox-feedback">
```
indirectlylit, line 93: Was curious if we could simply remove the conditional on this altogether?

```suggestion
      <div class="ui-textbox-feedback">
```

It's not entirely clear to me why it exists, and removing it would consistently preserve space for validation messages.

indirectlylit (commented):
possibly a simpler fix?

sairina:
@indirectlylit That was definitely the simpler fix. Seems to fix the issues that were found in https://github.com/learningequality/kolibri/issues/8675

indirectlylit (approved):
cool, thank you.

before merging, would you mind adding an entry to the changelog?

indirectlylit:
Darn, I forgot about this one before I tagged 1.2.0! I've pushed updates and will make a 1.2.1 now.

PR #280, opened 2021-11-23
clarify usage intent

indirectlylit:
## Description

Clarify that KDS is not intended for stand-alone use, and that plugins should not import it directly.

If we start publishing to NPM, this will hopefully help dissuade casual users from trying to use it in their own work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,9 +3,11 @@
 
 # Kolibri Design System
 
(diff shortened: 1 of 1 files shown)


sairina (approved):
Thank you for adding this! LGTM!

PR #278, opened 2021-11-18
Add timer icon

marcellamaki:
## Description
Makes the "timer" icon available in KDS

#### Issue addressed

Needed for [Kolibri Practice Quiz renderer and timer](https://github.com/learningequality/kolibri/pull/8645)

## Testing checklist

- [ ] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical and brittle code paths are covered by unit tests
- [ ] The change has been added to the `changelog`

## Reviewer guidance

- [ ] Is the code clean and well-commented?
- [ ] Are there tests for this change?
- [ ] Are all UI components LTR and RTL compliant (if applicable)?
- [ ] _Add other things to check for here_

## Post-merge updates and tracking

- [ ] Learning Platform update PR is submitted
- [ ] Learning Platform update PR is merged
- [ ] Studio update PR is submitted
- [ ] Studio update PR is merged
- [ ] Data Portal update PR is submitted
- [ ] Data Portal update PR is merged

PR #281, opened 2021-11-23
allow KModal to take a precise width in pixels

indirectlylit:
## Description

* Updates `KModal`'s `size` prop to take a precise width in pixels, in addition to existing `'small'`, `'medium'`, and `'large'` strings.
* Updates changelog, including reference to https://github.com/learningequality/kolibri-design-system/pull/278

#### Issue addressed

Helps address https://github.com/learningequality/kolibri/issues/8636

In this case, 'medium' was still too small but 'large' was too big.

### Before/after screenshots

| before | after |
|--|--|
| image | image |
| image | image |

## Steps to test

* Check that the new docstrings and logic make sense
* Testing PR in Kolibri: https://github.com/learningequality/kolibri/pull/8775

Because this change expands the API (hopefully without breaking anything) it should be released as a _minor_ (not major or patch) version bump.

Adds support for `size` to take a `Number`, interpreted as pixels, in addition to a `String`.

An alternative would have been to keep only strings as allowed, and instead require a unit. However this seemed to possibly introduce "too much" flexibility in the API, which we generally try to keep as constrained as possible while still allowing sufficient flexibility?

* By allowing arbitrary widths, there is a risk of encouraging UI inconsistency
* This change permanently interprets numbers as `px` values. If we ever wanted to allow another unit (e.g. `%`) we would need to update the strings to allow numbers with units.

## Testing checklist

## Reviewer guidance

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 Releases are recorded as git tags in the [Github releases](https://github.com/learningequality/kolibri-design-system/releases) page.
 
+## Version 1.2.0
+
+- [#281] - Allow `KModal` to take a `size` in pixels
+- [#278] - Adds timer icon
+
+<!-- Referenced PRs -->
+
+[#281]: https://github.com/learningequality/kolibri-design-system/pull/281
+[#278]: https://github.com/learningequality/kolibri-design-system/pull/278
+
+
 ## Version 1.1.0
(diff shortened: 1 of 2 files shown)


sairina (approved):
Works as expected when linked to Kolibri for both strings and numbers. lgtm!

PR #286, opened 2021-12-14
Add link to KDS figma file

sairina:
## Description
Adds a link to the docs for the [Figma file](https://www.figma.com/file/JPL94qI0O1s9FOHV92OeCmpS/Kolibri-Design-System?node-id=0%3A1) containing the patterns, colors, branding, etc. for KDS.

Screen Shot 2021-12-13 at 10 42 43 PM

#### Issue addressed
N/A

## Steps to test

1. Start the documentation server
2. See change in docs on home page

## Testing checklist

## Reviewer guidance
- [ ] Is this the right place for the link?

### DIFF
--- a/docs/pages/index.vue
+++ b/docs/pages/index.vue
@@ -15,7 +15,7 @@
         The design system includes both documentation design patterns and a shared front-end UI library. Together, these provide solutions to common design and implementation needs in Kolibri products.
       </p>
       <p>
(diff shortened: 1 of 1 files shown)


indirectlylit (commented):
Just curious, who is the target audience?

If it is mainly for designers who want to make edits then this makes sense because the framing of this paragraph is "If the design system is failing to achieve the objectives above, it needs to be updated..."

On the other hand, if we're providing the file as a reference then we might want to put it in a separate paragraph with a bit more context about what the Figma is and how to find useful information in it.

indirectlylit:
Oh, also a heads-up that I was editing the same paragraphs in my PR here:

https://github.com/learningequality/kolibri-design-system/pull/287/files#diff-793f96bb5957aaee533f3869859e46b550361bb67431ac83fde3aacfb0f8cd4d

So we'll need to integrate both of our work

sairina:
> Just curious, who is the target audience?

Good question! This is from one of our tasks from our OKR: "The Figma Team Library is linked to the live design system site", so it may be something we want to hash out before we merge.

indirectlylit (approved):
also happy to just merge in and update later if we want to?

up to you!

sairina:
@indirectlylit Let's merge and sort out what specifics we want to include later! Thanks!

Issue #84, opened 2020-07-28
Add option to KButtonGroup to prevent wrapping white space + layout shifting


jonboiser:
To fix a problem where two buttons in a Button Group would wrap if their container got too narrow, I added a `white-space: nowrap` rule.

https://github.com/learningequality/kolibri/pull/7393/files#diff-127f6559e90cdff95ed013563555933d

I think this will be a pretty common usage where we want buttons to stay in a side-by-side layout, so it would be helpful if this was either a default or can be controlled via a prop.

indirectlylit:
some related usage scenarios: https://github.com/learningequality/kolibri/issues/7337

PR #290, opened 2021-12-17
render function for handling vertical button wrap

indirectlylit:
## Description

adds vertical padding when buttons wrap to the next line. This is partly for example or future reference unless we decide to try and squeeze it in to 0.15

#### Issue addressed

fixes https://github.com/learningequality/kolibri-design-system/issues/84

refs:

* https://github.com/learningequality/kolibri/pull/8946#pullrequestreview-835383387
* https://learningequality.slack.com/archives/CB37UM23A/p1639690847424500

### Before/after screenshots

| before | after |
|--|--|
|  |  |

## Steps to test

update kolibri with this change and check for regressions

* Adds a wrapper div to each element added in the lost using a render function
* Gives wrapper divs a fixed height, and sets a negative bottom margin to compensate for the last row

not that I can think of

## Testing checklist

## Reviewer guidance

### DIFF
--- a/lib/buttons-and-links/KButtonGroup.vue
+++ b/lib/buttons-and-links/KButtonGroup.vue
@@ -1,13 +1,3 @@
-<template>
-
-  <div class="button-group">
-    <!-- @slot Slot to wrap buttons and provide consistent spacing -->
-    <slot></slot>
-  </div>
-
-</template>
-
-
 <script>
 
   /**
@@ -19,6 +9,16 @@
 
   export default {
     name: 'KButtonGroup',
+    render(createElement) {
+      var children = [];
+      this.$slots.default.forEach(element => {
+        if (element.tag) {
+          children.push(createElement('span', { class: 'button-group-item' }, [element]));
+        }
+      });
+
+      return createElement('div', { class: 'button-group' }, children);
+    },
   };
 
 </script>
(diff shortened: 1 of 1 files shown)


sairina (approved):
lgtm! Manual testing in browser shows that the wrapping is working appropriately.

lib/buttons-and-links/KButtonGroup.vue
```diff
@@ -19,6 +9,16 @@
...
+    render(createElement) {
+      var children = [];
+      this.$slots.default.forEach(element => {
+        if (element.tag) {
+          children.push(createElement('span', { class: 'button-group-item' }, [element]));
+        }
+      });
+
+      return createElement('div', { class: 'button-group' }, children);
+    },
```
sairina, line 21: Does doing this (as compared to the `div` in the template) make it more dynamic?

sairina (approved):

lib/buttons-and-links/KButtonGroup.vue
indirectlylit: Using a slot in a template, all items that get passed in are wrapped by whatever is around the template, for example:

```html
  <div class="button-group">
    <button />
    <button />
    <!-- ... -->
    <button />
  </div>
```

I wasn't able to figure out a way to increase the vertical margins of the buttons while still keeping them as `inline-block` so they'd wrap to the next line without putting each one in a wrapper node.

Using the render function one can iterate over the elements passed into the slot. Here I used that to wrap each one in an inline-block span with a fixed larger height like this:

```html
  <div class="button-group">
    <span class="button-group-item"><button /></span>
    <span class="button-group-item"><button /></span>
    <!-- ... -->
    <span class="button-group-item"><button /></span>
  </div>
```

PR #291, opened 2021-12-30
Set modality to keyboard only when TAB is pressed

MisRob:
## Description

Set modality to keyboard only when TAB key is pressed when tracking input modality.

#### Issue addressed

Fixes https://github.com/learningequality/kolibri/issues/8786

## Steps to test

Follow test instructions in https://github.com/learningequality/kolibri/pull/8971

I added a whitelist of keys that should change the modality to keyboard. Currently, only TAB key is in the whitelist. Any other keys don't change the modality to keyboard.

One thing to be cautious about is that many keys that used to change the input modality to keyboard won't change it anymore. I haven't noticed any problems so far. Can you think of any situations where we might need to display the focus ring for keyboard navigation when NOT using TAB/SHIFT+TAB?

> Can't think of any use case on the top of my head, but will do some searching and chime it if I find it!
> ~ @radinamatic

Development-wise, this limits ways that `trackInputModality` might be used in the future for purposes different than displaying focus rings since it won't inform us more generally about any usage of the keyboard anymore. An alternative solution could be to continue changing the input modality to keyboard on any key press and start reporting what key was pressed as the last one to `globalThemeState`. However, that would also require some more conditional logic around displaying focus rings. Since I'm not aware about any other ways that this script is useful for us other than for focus rings, I've decided to try the less complex solution first. I welcome any feedback on this decision.

## Testing checklist

## Reviewer guidance

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 Releases are recorded as git tags in the [Github releases](https://github.com/learningequality/kolibri-design-system/releases) page.
 
+## Version 1.2.2
+
+- [#291] - When tracking input modality with `trackInputModality`, modality is set to keyboard only when the TAB key is pressed
+
+<!-- Referenced PRs -->
+
+[#291]: https://github.com/learningequality/kolibri-design-system/pull/291
+
 ## Version 1.2.1
 
 - [#279] - Fix vertical shifting of `KTextbox`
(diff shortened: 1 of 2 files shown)


MisRob:
I will update the changelog and create a new release tag as soon as reviews are done

nucleogenesis (approved):
Everything looks good. I pulled it locally with that Kolibri PR and couldn't get into keyboard modality without hitting tab.

marcellamaki (approved):

PR #308, opened 2022-02-02
Adding Tests

nucleogenesis:
## Description

- Adds some package.json scripts to call `kolibri-tools` and jest from it.
- Tried copying Kolibri's jest.conf.js file - the bits I figured we'd need.

Get this error when running (it finds the spec file before failing):

```
 FAIL  tests/smoke.spec.js
  ● Test suite failed to run

    /home/jacob/Code/LE/design-system/node_modules/kolibri-tools/jest.conf/setup.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import 'regenerator-runtime/runtime';
                                                                                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

    SyntaxError: Unexpected string

      at ScriptTransformer._transformAndBuildScript (node_modules/@jest/transform/build/ScriptTransformer.js:537:17)
      at ScriptTransformer.transform (node_modules/@jest/transform/build/ScriptTransformer.js:579:25)
          at Array.forEach (<anonymous>)
```

I'm blaming babel whether it deserves it or not.

NOTE: There is already a test command that calls `jest` bare - should we just use that? I'm not sure if pulling in from `kolibri-tools` is necessary and honestly only noticed the `yarn _test` command as I was writing this PR... :)

### DIFF
--- a/jest.conf.js
+++ b/jest.conf.js
@@ -1,0 +1,12 @@
+const baseConfig = require('kolibri-tools/jest.conf');
+
+module.exports = Object.assign(baseConfig, {
+  transformIgnorePatterns: [
+    '/node_modules/(?!(keen-ui|kolibri-tools)/).*/',
+  ],
+  collectCoverageFrom: [
+    'lib/**/*.{js|vue}',
+    '!**/node_modules/**',
+    'utils/*.{js|vue}',
+  ],
+});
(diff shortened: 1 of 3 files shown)


nucleogenesis:
Closed because redundant

PR #310, opened 2022-02-04
Exercise icon flipped in rtl

### LINKED ISSUE #301: `exercise` icon should be mirrored in RTL languages
## Product

Kolibri

## Expected behavior

`exercise` icon

exercise icon

should be mirrored in RTL languages to correctly represent text direction.

## Actual behavior

It is currently not mirrored. We need to set `rtlFlip` to `true` here for that to happen:
https://github.com/learningequality/kolibri-design-system/blob/c2acc3fd19cee52ccd2a17fa8d2b10cddb2b3d63/lib/KIcon/iconDefinitions.js#L90

When this update is introduced to Kolibri, we need to remove `{ 'rtl-icon': isRtl }` from [Kolibri's `ContentIcon`](https://github.com/learningequality/kolibri/blob/2f93cb01fe9de910843bbaf462ae82c5b7a088ea/kolibri/core/assets/src/views/ContentIcon.vue#L40-L46) here:

```vue
<KIcon
  v-if="is(ContentNodeKinds.EXERCISE)"
  :icon="ContentNodeKinds.EXERCISE"
  :class="[colorClass, { 'rtl-icon': isRtl }]"
  :color="color"
  style="top:0;"
/>
```

to make sure it won't be mirrored two times resulting in not being mirrored at all.

nucleogenesis:
Updates the configuration file so that the Exercise icon flips when in RTL.

## Reviewer Guidance

- Checkout this branch locally
- Link it to your Kolibr
- Check that the icon flips in RTL

Fixes #301

### DIFF
--- a/lib/KIcon/iconDefinitions.js
+++ b/lib/KIcon/iconDefinitions.js
@@ -87,7 +87,10 @@ const KolibriIcons = {
   audio: { icon: require('./precompiled-icons/material-icons/audiotrack/baseline.vue').default },
   channel: { icon: require('./precompiled-icons/material-icons/apps/baseline.vue').default },
   document: { icon: require('./precompiled-icons/material-icons/book/baseline.vue').default },
-  exercise: { icon: require('./precompiled-icons/material-icons/assignment/baseline.vue').default },
+  exercise: {
+    icon: require('./precompiled-icons/material-icons/assignment/baseline.vue').default,
+    rtlFlip: true,
+  },
   topic: { icon: require('./precompiled-icons/material-icons/folder/baseline.vue').default },
   video: {
     icon: require('./precompiled-icons/material-icons/ondemand_video/baseline.vue').default,


lib/buttons-and-links/__tests__/k-button.spec.js
```diff
@@ -0,0 +1,5 @@
+describe("KButton", () => {
```
sairina, line 1: Is this part of this issue?

lib/KIcon/iconDefinitions.js
```diff
@@ -87,7 +87,10 @@ const KolibriIcons = {
   audio: { icon: require('./precompiled-icons/material-icons/audiotrack/baseline.vue').default },
   channel: { icon: require('./precompiled-icons/material-icons/apps/baseline.vue').default },
   document: { icon: require('./precompiled-icons/material-icons/book/baseline.vue').default },
-  exercise: { icon: require('./precompiled-icons/material-icons/assignment/baseline.vue').default },
+  exercise: { 
+    icon: require('./precompiled-icons/material-icons/assignment/baseline.vue').default,
+    rtlFlip: true
```
sairina, line 92: Ahh - so now, if this is `true`, and the language is also rtl, then in the `KIcon` file itself, the class of `'rtl-flip-icon'` will also be true, and the icon will be flipped?

sairina (changes requested):
LGTM with manual testing.

lib/buttons-and-links/__tests__/k-button.spec.js
nucleogenesis: :man_facepalming: no will remove and push again later on - thanks for catching that!

lib/KIcon/iconDefinitions.js
nucleogenesis: Yep!

sairina (approved):
LGTM! 🚀

PR #313, opened 2022-02-10
KButton Tests

nucleogenesis:
## Description
Adds initial unit tests for some KButton behavior.

Does not test `buttonMixin` behavior as that should be tested separately. This impacts other components and having tests for buttonMixin might be a better next step than any other items in the `buttons-and-links` directory.

## Review

- What behavior in KButton that is not defined in the buttonMixin should I test that isn't covered here?

#### Issue addressed

Addresses #303

### DIFF
--- a/lib/buttons-and-links/__tests__/KButton.spec.js
+++ b/lib/buttons-and-links/__tests__/KButton.spec.js
@@ -1,0 +1,67 @@
+import { shallowMount } from '@vue/test-utils';
+import KButton from '../KButton.vue';
+
+describe('KButton', () => {
+  describe('icon related props', () => {
+    it('should render an icon before the text with this.icon', () => {
+      const wrapper = shallowMount(KButton, {
+        propsData: {
+          icon: 'add',
+        },
+      });
+      expect(wrapper.find('[data-test="iconBefore"]').exists()).toBe(true);
+    });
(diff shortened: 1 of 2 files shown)


lib/buttons-and-links/__tests__/KButton.spec.js
```diff
@@ -0,0 +1,67 @@
+import { shallowMount } from '@vue/test-utils';
+import KButton from '../KButton.vue';
+
+describe('KButton', () => {
+  describe('icon related props', () => {
+    it('should render an icon before the text with this.icon', () => {
```
MisRob, line 6: It took me a while to realize that you mean `icon` property with `this.icon`.

Might be a bit better readable if we used the same style like in one of the tests below that doesn't use `this` (`should render a dropdown icon when hasDropdown is true`), changing this description into `should render an icon before the text when icon is true`? Similarly in other tests that use `this`.

MisRob (approved):
Thank you, @nucleogenesis, the test suite makes sense to me and is well organized and readable.  I've left one minor comment that's not blocking.

lib/buttons-and-links/__tests__/KButton.spec.js
nucleogenesis: I pushed updated strings

PR #352, opened 2022-08-22
Update lint action to use yarn lockfile.

rtibbles:
## Description
Updates the linting action to use yarn to ensure dependencies are pinned.


MisRob (approved):
Thank you, @rtibbles

PR #357, opened 2022-08-31
add attrs/aria-label to navigation element within KBreadcrumbs

marcellamaki:
## Description
Per conversation on Slack, I've added attributes binding and an aria-label option to `KBreadcrumbs` so that it is easier for us to use a label on the <nav> element, which is where it should be.

## Steps to test

1. See [corresponding kolibri PR](https://github.com/learningequality/kolibri/pull/9663) for testing. Successful implementation should have the aria label display on the breadcrumb elements

Is this clear enough? would props be better?

## Testing checklist

## Reviewer guidance

### DIFF
--- a/lib/KBreadcrumbs.vue
+++ b/lib/KBreadcrumbs.vue
@@ -4,7 +4,7 @@
     v-show="showSingleItem || crumbs.length > 1"
     :class="{ 'breadcrumbs-collapsed': collapsedCrumbs.length }"
   >
-    <nav class="breadcrumbs">
+    <nav class="breadcrumbs" v-bind="$attrs" :aria-label="$attrs.ariaLabel">
       <div
         v-show="collapsedCrumbs.length"
         class="breadcrumbs-dropdown-wrapper"
--- a/lib/KBreadcrumbs.vue
+++ b/lib/KBreadcrumbs.vue
@@ -138,6 +138,7 @@
   export default {
     name: 'KBreadcrumbs',
     mixins: [KResponsiveElementMixin],
+    inheritAttrs: false,
     props: {
       /**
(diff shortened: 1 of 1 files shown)


lib/KBreadcrumbs.vue
```diff
@@ -4,7 +4,7 @@
     v-show="showSingleItem || crumbs.length > 1"
     :class="{ 'breadcrumbs-collapsed': collapsedCrumbs.length }"
   >
-    <nav class="breadcrumbs">
+    <nav class="breadcrumbs" v-bind="$attrs" :aria-label="$attrs.ariaLabel">
```
rtibbles, line 7: I think we should also add `inheritAttrs: false` to the component as well? https://v2.vuejs.org/v2/api/#inheritAttrs

marcellamaki: Like we have in the `TextTruncatorCss` component? https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/views/TextTruncatorCss.vue#L42

rtibbles: Yes, it's possible it's not actually required, but this avoids the `$attrs` being erroneously applied to the containing div.

marcellamaki: @rtibbles added

rtibbles (approved):

MisRob:
@marcellamaki Gentle reminder that it'd be good to mention this in the changelog

MisRob:
@marcellamaki Regarding the changelog, I was just opening a PR to `develop`, so I thought I could sneak the update in: https://github.com/learningequality/kolibri-design-system/pull/363

marcellamaki:
@MisRob thank you!

PR #377, opened 2022-10-06
Implement k responsive window composable

### LINKED ISSUE #288: update mixins to allow composition API
## Product

All

## Desired behavior

The Vue composition API makes it possible to explicitly add functionality from modules to components using standard import mechanisms. This is preferable to using mixins because it makes it explicit which elements are being added and where it's coming from.

## Current behavior

`KResponsiveWindowMixin` for example currently provides the following reactive properties:

* `windowIsSmall`: boolean
* `windowIsMedium`: boolean
* `windowIsLarge`: boolean
* `windowHeight`: integer height in pixels
* `windowWidth`: integer width in pixels
* `windowBreakpoint`: integer breakpoint level

However the usual mechanism for adding the mixin doesn't provide any hint as to what's being added:

```js
  import KResponsiveWindowMixin from 'kolibri-design-system/lib/KResponsiveWindowMixin';

  export default {
    name: 'SomeComponent',
    mixins: [KResponsiveWindowMixin],
```

## (Optional) The Value Add

Easier discoverability and documentation

## (Optional) Possible Tradeoffs

Might be more verbose in some cases, and lots of refactoring effort with not much user-facing benefit

akolson:
## Description

This pr ports the `KResponsiveWindowMixin` to `useKResponsiveWindow` composable. Logic has also been refactored to use the `Window.matchMedia` interface

#### Issue addressed

Closes #288

### Before/after screenshots
- No visual changes in products
- `KResponsiveWindow` documentation
  - before: https://design-system.learningequality.org/kresponsivewindow/
  - after: https://deploy-preview-377--kolibri-design-system.netlify.app/kresponsivewindow/

## Steps to test
- Please follow the steps to test in Kolibri's corresponding PR https://github.com/learningequality/kolibri/pull/9763
- Read the updated documentation page and preview `KResponsiveWindow` for usage and examples

useKResponsiveWindow uses `Window.matchMedia()` to create reactive window properties including width, height, orientation etc

## Testing checklist

## Reviewer guidance

### DIFF
--- a/docs/pages/kresponsivewindow.vue
+++ b/docs/pages/kresponsivewindow.vue
@@ -154,13 +154,13 @@
         This results in two boxes that stack vertically on small screens and otherwise display side-by-side:
       </p>
       <DocsShow>
-        <div>Breakpoint level: {{ _windowBreakpoint }}</div>
-        <div>Window is large: {{ _windowIsLarge }}</div>
+        <div>Breakpoint level: {{ composableWindowBreakpoint }}</div>
+        <div>Window is large: {{ composableWindowIsLarge }}</div>
         <div>
-          <div class="box" :style="_boxStyle">
+          <div class="box" :style="composableBoxStyle">
             Box 1
           </div>
-          <div class="box" :style="_boxStyle">
+          <div class="box" :style="composableBoxStyle">
             Box 2
           </div>
         </div>
@@ -189,21 +189,21 @@
 
   import { computed } from '@vue/composition-api';
   import responsiveWindowMixin from '~~/lib/KResponsiveWindowMixin.js';
-  import useKResponsiveWindow from '~~/lib/useKResponsiveWindow.js';
+  import useKResponsiveWindow from '~~/lib/useKResponsiveWindow';
 
   export default {
     mixins: [responsiveWindowMixin],
     setup() {
       const {
-        windowBreakpoint: _windowBreakpoint,
-        windowIsLarge: _windowIsLarge,
+        windowBreakpoint: composableWindowBreakpoint,
+        windowIsLarge: composableWindowIsLarge,
       } = useKResponsiveWindow();
 
-      const _boxStyle = computed(() => {
-        return { display: _windowIsLarge.value ? 'inline-block' : 'block' };
+      const composableBoxStyle = computed(() => {
+        return { display: composableWindowIsLarge.value ? 'inline-block' : 'block' };
       });
 
-      return { _windowBreakpoint, _windowIsLarge, _boxStyle };
+      return { composableWindowBreakpoint, composableWindowIsLarge, composableBoxStyle };
     },
     computed: {
       boxStyle() {
--- a/docs/pages/kresponsivewindow.vue
+++ b/docs/pages/kresponsivewindow.vue
@@ -42,9 +42,41 @@
       <p>
         Provided reactive properties can then be accessed on the component instance via <code>this</code>.
       </p>
+
+      <h3>As composable</h3>
+      <p>It can be imported as <code>useKResponsiveWindow</code>. Register it in the script part of a component file:</p>
+
+      <!-- eslint-disable -->
+      <!-- prevent prettier from changing indentation -->
+      <DocsShowCode language="javascript">
+        import useKResponsiveWindow from 'kolibri-design-system/lib/useKResponsiveWindow';
+
+        export default {
+          setup() {
+            const {
+              windowWidth,
+              windowIsLarge,
+              windowBreakpoint,
+              ...
+            } = useKResponsiveWindow();
+
+            //Peform some operations with the properties
+
+            return {
+              windowWidth,
+              windowIsLarge,
+              windowBreakpoint,
+              ...
+            };
+          }
+        };
+      </DocsShowCode>
+      <!-- eslint-enable -->
+
+      <p><code>useKResponsiveWindow()</code> returns an object of reactive properties, that you could destruture.</p>
     </DocsPageSection>
 
-    <DocsPageSection title="Example" anchor="#example">
+    <DocsPageSection title="Mixin Example" anchor="#mixin-example">
       <p>
         Consider a Vue file with this in its template and script:
       </p>
@@ -87,6 +119,55 @@
       <p>
         Try adjusting your browser window size to see the example in action.
       </p>
+    </DocsPageSection>ßßßß
+
+    <DocsPageSection title="Composable Example" anchor="#composable-example">
+      <p>
+        Consider a Vue file with this in its template and script:
+      </p>
+      <!-- eslint-disable -->
+        <!-- prevent prettier from changing indentation -->
+        <DocsShowCode language="html">
+          <div class="box" :style="boxStyle">
+            Box 1
+          </div>
+          <div class="box" :style="boxStyle">
+            Box 2
+          </div>
+        </DocsShowCode>
+        <DocsShowCode language="javascript">
+          setup() {
+            ...
+
+            const boxStyle = computed(() => {
+              return { display: windowIsLarge.value ? 'inline-block' : 'block' };
+            });
+
+            return { 
+              ..., 
+              boxStyle,
+            };
+          }
+        </DocsShowCode>
+        <!-- eslint-enable -->
+      <p>
+        This results in two boxes that stack vertically on small screens and otherwise display side-by-side:
+      </p>
+      <DocsShow>
+        <div>Breakpoint level: {{ _windowBreakpoint }}</div>
+        <div>Window is large: {{ _windowIsLarge }}</div>
+        <div>
+          <div class="box" :style="_boxStyle">
+            Box 1
+          </div>
+          <div class="box" :style="_boxStyle">
+            Box 2
+          </div>
+        </div>
+      </DocsShow>
+      <p>
+        Try adjusting your browser window size to see the example in action.
+      </p>
     </DocsPageSection>
 
     <DocsPageSection title="Related" anchor="#related">
@@ -106,10 +187,24 @@
 
 <script>
 
+  import { computed } from '@vue/composition-api';
   import responsiveWindowMixin from '~~/lib/KResponsiveWindowMixin.js';
+  import useKResponsiveWindow from '~~/lib/useKResponsiveWindow.js';
 
   export default {
     mixins: [responsiveWindowMixin],
+    setup() {
+      const {
+        windowBreakpoint: _windowBreakpoint,
+        windowIsLarge: _windowIsLarge,
+      } = useKResponsiveWindow();
+
+      const _boxStyle = computed(() => {
+        return { display: _windowIsLarge.value ? 'inline-block' : 'block' };
+      });
+
+      return { _windowBreakpoint, _windowIsLarge, _boxStyle };
+    },
     computed: {
       boxStyle() {
         return { display: this.windowIsLarge ? 'inline-block' : 'block' };
--- a/docs/pages/kresponsivewindow.vue
+++ b/docs/pages/kresponsivewindow.vue
@@ -81,26 +81,26 @@
         Consider a Vue file with this in its template and script:
       </p>
       <!-- eslint-disable -->
-        <!-- prevent prettier from changing indentation -->
-        <DocsShowCode language="html">
-          <div class="box" :style="boxStyle">
-            Box 1
-          </div>
-          <div class="box" :style="boxStyle">
-            Box 2
-          </div>
-        </DocsShowCode>
-        <DocsShowCode language="javascript">
-          computed: {
-            boxStyle() {
-              if (this.windowIsLarge) {
-                return { display: 'inline-block' };
-              }
-              return { display: 'block' };
-            },
+      <!-- prevent prettier from changing indentation -->
+      <DocsShowCode language="html">
+        <div class="box" :style="boxStyle">
+          Box 1
+        </div>
+        <div class="box" :style="boxStyle">
+          Box 2
+        </div>
+      </DocsShowCode>
+      <DocsShowCode language="javascript">
+        computed: {
+          boxStyle() {
+            if (this.windowIsLarge) {
+              return { display: 'inline-block' };
+            }
+            return { display: 'block' };
           },
-        </DocsShowCode>
-        <!-- eslint-enable -->
+        },
+      </DocsShowCode>
+      <!-- eslint-enable -->
       <p>
         This results in two boxes that stack vertically on small screens and otherwise display side-by-side:
       </p>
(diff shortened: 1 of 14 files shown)


akolson:
Hi @jtamiace, I am tagging you here so you could have a look at the documentation. Would appreciate your feedback on this. Thanks

lib/useKResponsiveWindow.js
```diff
@@ -0,0 +1,199 @@
...
+    const queries = [
+      '(max-width: 480px)',
+      '(max-width: 600px)',
+      '(max-width: 840px)',
+      `(max-width: ${960 - SCROLL_BAR}px)`,
+      `(max-width: ${1280 - SCROLL_BAR}px)`,
+      `(max-width: ${1440 - SCROLL_BAR}px)`,
+      `(max-width: ${1600 - SCROLL_BAR}px)`,
+      `(min-width: 1601px)`,
+    ];
```
bjester, line 97: The above 2 variables don't seem to change during calls to `useKResponsiveWindow` or `widthQueries`, so they could be moved further up in scope. Although I don't think it's a big deal.

lib/useKResponsiveWindow.js
```diff
@@ -0,0 +1,199 @@
...
+    mediaQueryList.removeEventListener('change', eventHandler);
+  };
+  mediaQueryList.addEventListener('change', eventHandler);
+
+  return {
+    mediaQueryList,
+    eventHandler,
+    stopListening,
+  };
+}
```
bjester, line 47: This function is essentially a constructor. I think it's worth considering whether it would make sense to use a class here.

The benefits are fairly small for using a class here, but generally speaking they're more memory efficient. For instance, `stopListening` is a method for the returned object, and could take advantage of the classes prototype to avoid redeclaring it for every object. So whenever we define functions on returned objects, it's worth considering a class structure.

Perhaps it could also lead to some slight clarification, like for the `return;` when it's server side rendering. It isn't clear to me if this would end up raising errors in the server-side rendering since it returns an undefined value in this case. Perhaps that should be moved further up, if `useKResponsiveWindow` gets invoked in server-side rendering, leading to the invocation of `addMediaQuery`? Or is this check just a defensive measure and this code shouldn't be invoked?

lib/useKResponsiveWindow.js
```diff
@@ -0,0 +1,199 @@
...
+   */
+  const initProps = () => {
+    orientationQuery.eventHandler(orientationQuery.mediaQueryList);
+    heightQuery.eventHandler(heightQuery.mediaQueryList);
+    widthQueries.every(widthQuery => {
+      if (widthQuery.mediaQueryList.matches) {
+        widthQuery.eventHandler(widthQuery.mediaQueryList);
+        return false;
+      }
+      return true;
```
bjester, line 122: It took me moment to understand your tricks here. This is pretty cool! A comment might be nice too :)

lib/useKResponsiveWindow.js
```diff
@@ -0,0 +1,199 @@
...
+  };
+
+  /**
+   * Update window width and height
+   */
+  const updateWindow = () => {
+    const metrics = windowMetrics();
+    windowWidth.value = metrics.width;
+    windowHeight.value = metrics.height;
+  };
```
bjester, line 178: The only potential issue that I see moving towards the media query approach here is that there may be usages relying on width and height being accurate, regardless of breakpoints. From looking through your code, it seems the width and height will only update when transitioning between breakpoints and during orientation changes. Is that correct? I guess this could be something to audit in Kolibri when you integrate this work. Perhaps those should leverage something more specifically tailored to their behavior instead of this composable.

docs/pages/kresponsivewindow.vue
```diff
@@ -106,10 +187,24 @@
...
+      } = useKResponsiveWindow();
+
+      const _boxStyle = computed(() => {
+        return { display: _windowIsLarge.value ? 'inline-block' : 'block' };
+      });
+
+      return { _windowBreakpoint, _windowIsLarge, _boxStyle };
+    },
     computed: {
       boxStyle() {
```
bjester, line 209: Should you remove this computed value?

docs/pages/kresponsivewindow.vue
```diff
@@ -106,10 +187,24 @@
...
+  import { computed } from '@vue/composition-api';
   import responsiveWindowMixin from '~~/lib/KResponsiveWindowMixin.js';
+  import useKResponsiveWindow from '~~/lib/useKResponsiveWindow.js';
 
   export default {
     mixins: [responsiveWindowMixin],
+    setup() {
+      const {
+        windowBreakpoint: _windowBreakpoint,
+        windowIsLarge: _windowIsLarge,
```
bjester, line 199: If you wanted, I think you can destructure these fields without prefixing them with underscores. Just noting this because it isn't clear if there was a reason for this.

lib/useKResponsiveWindow.js
bjester: This could mean that this composable doesn't even return `windowWidth` and `windowHeight`, and instead there's an explicit composable that could be used for that to avoid unnecessary listeners on window resizing.

bjester (commented):
Left some comments, but this looks awesome!

lib/useKResponsiveWindow.js
rtibbles: It might be more explicit to just do a `for ... of` loop here and then `break;` in the match case.

docs/pages/kresponsivewindow.vue
akolson: The reason for prefixing these with underscores was to differentiate the mixin variables from the composable's(with _). For clarity, the composable variables will be renamed `composableWindowBreakPoint` and `composableWindowIsLarge` respectively.

docs/pages/kresponsivewindow.vue
akolson: This is used for the mixin. I will rename this so its more clear

lib/useKResponsiveWindow.js
rtibbles: This is following the composable function style that VueJS suggests for using with the composition API: https://vuejs.org/guide/reusability/composables.html#what-is-a-composable

I've not seen uses of classes in this way, so there may be some gotchas around binding of methods when they are returned from the setup function in a Vue component.

lib/useKResponsiveWindow.js
```diff
@@ -0,0 +1,199 @@
...
+
+/**
+ * Add a media query
+ * @param {String} query
+ * @param {requestCallback} callback
+ * @returns {Object} The mediaQueryList, eventHandler, and stopListening
+ */
+function addMediaQuery(query, callback) {
+  if (isNuxtServerSideRendering()) {
+    return;
```
rtibbles, line 30: Is it up to the consumer of this function to know when to handle an `undefined` return here, or should we be giving a consistent return type at least? If we return an empty object, the value can still be destructured, but will give undefined for each of the destructured values.

lib/useKResponsiveWindow.js
```diff
@@ -0,0 +1,199 @@
...
+
+    const mediaQueries = [];
+    queries.forEach((query, index) => {
+      const widthQuery = addMediaQuery(query, () => {
+        windowBreakpoint.value = index;
+        updateWindow();
+        updateGutter();
+      });
+      mediaQueries.push(widthQuery);
+    });
```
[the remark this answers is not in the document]
rtibbles: I also generally prefer not to use `forEach` as multiple benchmarks seem to suggest that it is less performant than other Array iteration methods, e.g. https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/#arrayforeach-vs-for-and-forof

lib/useKResponsiveWindow.js
rtibbles: One way to retain previous behaviour (but still providing an advantage over the mixin approach) is to have a single module scoped event listener listening to the window resize event, that is also updating a set of module scoped reactive data references for tracking the window width and height.

This way we only ever need one global listener, and a single reactive source of truth for the current window dimensions.

@MisRob and I have used this approach of module scoped reactives combined with function scoped reactives in order to provide some global elements and some component specific elements, e.g. https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/learn/assets/src/composables/useLearnerResources.js#L22

rtibbles: The other advantage here, is then the existing mixin could just transparently use the logic in this module in its `setup` method, and then the mixin just becomes a compatibility layer so that we don't have to migrate everything at once.

lib/useKResponsiveWindow.js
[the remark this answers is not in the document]
akolson: Interesting benchmarks here. A good one to keep in mind for future imlementations. Thanks

lib/useKResponsiveWindow.js
akolson: I think the latter makes more sense

jtamiace:
@akolson the layout and format of the documentation additions look good to me! Would there be any value to engineers in making a short statement about which scenarios composable is preferable over mixin and vice versa? (I'm not an engineer, so if this is N/A or not helpful, please disregard)

akolson:
@jtamiace you raise a valid question. More generally, composables are recommended. I will add a section to capture these thoughts. Thanks

lib/useKResponsiveWindow.js
bjester: I don't think there are any gotchas. This composable format is an old JS style prior to the days of pure JS classes just repurposed for this. This function isn't a composable either!

rtibbles: Ah sorry - yeah, this is just a utility function within the module. Carry on!

lib/useKResponsiveWindow.js
akolson: @bjester your interpretation is correct. As suggested by @rtibbles, we need something better tailored for dimension changes. The onresize listeners on the mixin seem to do the trick. Will proceed as suggested

bjester: @rtibbles The existing [mixin](https://github.com/learningequality/kolibri-design-system/blob/develop/lib/KResponsiveWindowMixin.js#L79) already had a single global listener on the window's `resize`. Ideally that should also be throttled through `requestAnimationFrame` too.

I think a separate composable makes sense for the resize event handling, tracking width and height, since that  functionality that should be deprecated going forward. And we definitely shouldn't be using that for recreating the wheel that already exists in `matchMedia`!

lib/useKResponsiveWindow.js
akolson: Thinking this through again, the code was intended to be defence measure against execution of the function during ssr as suggested by @bjester

lib/useKResponsiveWindow.js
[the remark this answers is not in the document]
bjester: As I've commented in the past, the performance degradation of `forEach` comes from the redefinition of the callback passed to it, as that link notes. It isn't that the looping behavior is faster at all.
> Since .forEach isn’t that different from the for..of loop, we don’t see much sense in using it over the traditional loop in most cases. It’s worth using only when you already have a function to invoke on every array element. In this case, it’s a one-liner, with zero performance degradation

lib/useKResponsiveWindow.js
rtibbles: Right, the main thing we could do here is only have a single function that gets invoked on the resize event, as we only need to update a single module level reactive, rather needing to update the `height` and `width` on every single active mixin, so the number of handler functions invoked no longer scales with the number of uses of the mixin.

I agree that implementing this as a separate composable as a compatibility layer would be useful, so we can deprecate and then remove it, once we migrate all uses of the mixin to no longer directly rely on `height` and `width` (I think this shouldn't be too many, as most should be referring to the breakpoints).

lib/useKResponsiveWindow.js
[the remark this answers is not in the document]
rtibbles: Yes - it would be very curious if it were otherwise!

lib/useKResponsiveWindow.js
akolson: Code has been refactored to factor this in. This file has been moved to the `lib/useKResponsiveWindow/index.js`

lib/useKResponsiveWindow.js
akolson: a new composable `useWindowDimensions` has been created and is being called in `lib/useKResponsiveWindow/index.js`

akolson:
@bjester  @rtibbles & @nucleogenesis, based on Friday feedback, some files were moved around, so incase you need to follow along based on old review, [this link](https://github.com/learningequality/kolibri-design-system/pull/377/files/b0fc2051119373ba1ce05e977c59d4504e55b998) captures the previous code including your comments.

lib/useKWindowDimensions.js
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Check if Nuxt is server side rendering
+ * @returns {Boolean}
+ */
+function isNuxtServerSideRendering() {
+  return process && process.server;
+}
+
+/**
+ * Add window listener
```
nucleogenesis, line 37: Non-blocking just some thoughts on style.

On some of the comments which describe a function like "addWindowListener" with "Add window listener" -- I think it's okay to just leave off the "Add window listener" bit altogether since it's not giving any info the function name doesn't already give. I think it would be helpful in this particular case to share some more context (ie, why do we do two different events depending on whether we're in Nuxt or not?).

On the other hand, the comment description for `removeWindowListener` seems unnecessary.

I love the JSDoc `@param` and type declarations throughout <3

nucleogenesis (commented):
I gave the code a look and this looks great. I haven't tested it in Kolibri yet but will do so tomorrow and will give the docs a closer look

lib/useKWindowDimensions.js
```diff
@@ -0,0 +1,72 @@
...
+}
+
+export default function useKWindowDimensions() {
+  onMounted(() => {
+    addWindowListener(updateWindow);
+  });
+
+  onBeforeUnmount(() => {
+    removeWindowListener(updateWindow);
+  });
```
bjester, line 66: Since `windowWidth` and `windowHeight` are global variables, it seems that this really only needs to `addWindowListener` once and likewise only `removeWindowListener` if there aren't any more active usages of `useKWindowDimensions`. Invoking `removeWindowListener` on `onBeforeUnmount` will likely stop listening all together if this composable is used more than once, meaning one instance will stop working if another is unmounted.

bjester (changes requested):
This looks really great! The only blocker, and requested change, is that I think there's likely an issue with `useKWindowDimensions`-- please see my comment on that

lib/useKResponsiveWindow/MediaQuery.js
```diff
@@ -0,0 +1,57 @@
...
+   * Stop listening for media query events
+   */
+  stopListening() {
+    if (this.mediaQueryList().removeEventListener) {
+      this.mediaQueryList().removeEventListener('change', this.eventHandler);
+    } else {
+      this.mediaQueryList().removeListener(this.eventHandler);
+    }
+  }
+}
```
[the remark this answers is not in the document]
akolson: I think so too! Love it:)

lib/useKResponsiveWindow/MediaQuery.js
```diff
@@ -0,0 +1,57 @@
...
+   */
+  constructor(query, eventHandler) {
+    this.query = query;
+    this.eventHandler = eventHandler;
+  }
+
+  /**
+   * @returns {Object} Media query list
+   */
+  mediaQueryList() {
```
rtibbles, line 18: Not needed, but could turn this into a [getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) as its only purpose is to return a value. (Would then need to update the references to it to no longer be invocations of a method).

lib/useKWindowDimensions.js
akolson: Code has been modified to add a listener only if it doesn't exist. Also, A the only added listener can only be removed only if its not in active use else where.

lib/useKResponsiveWindow/MediaQuery.js
```diff
@@ -0,0 +1,57 @@
...
+    return window.matchMedia(this.query);
+  }
```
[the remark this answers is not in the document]
akolson: function has been changed into a getter instead of as a class property

nucleogenesis (approved):
With QA before me, Blaine & Richard's thoughts in mind; I've looked through the latest additions in the test suits and they're excellent - this should be good to merge!

bjester (approved):

PR #384, opened 2022-11-16
KDateComponents

### LINKED ISSUE #360: Create KDatePicker and KDateRange components
## Product

- The components will be used in Kolibri in selecting date ranges for exporting logs
- The components will be created by [migrating](https://github.com/learningequality/kolibri-data-portal/blob/staging/kolibri_data_portal/assets/src_home/views/DataPage/CustomDateFilter.vue) functionality from Kolibri Data Portal
- [Designs](https://www.figma.com/file/SgTZugaKkES68aipU5dCk8SR/Github-issues-%26-random-fixes) that will be implemented in Kolibri

Example:
Screenshot from 2022-09-09 08-50-13

## Desired behavior

- The PR should target the `develop` branch for integration into Kolibri's `develop` branch; note we won't be able to update KDP until KDP is updated to use the latest Kolibri from its `develop` branch
- The `KDatePicker` component:
  - Should accept a required `selectionLevel` prop that will only allow `month` level to begin with, allowing future enhancements to possibly allow `day` level selection
  - Should allow use with `v-model` which will allow setting a default value
  - Should accept a `firstAllowedDate` prop that constrains the selection to after the date, disabling dates prior
  - Should accept a `lastAllowedDate` prop that defaults to the current date and constrains the selection to before that date, disabling dates after
  - Should implement 2 `KSelect`'s, one for month and one for year, when the `selectionLevel` is `month`, and wrap them in a `<fieldset>`
  - Should accept a `legendText` prop that will show in a `<legend>` tag, e.g. "Start" or "End", prepended to `<fieldset>`
  - Should have a page for it added to the documentation
- The `KDateRange` component:
  - Should have no default for the `firstAllowedDate` prop
  - Should accept a prop for a default start date selection
  - Should implement two `<KDatePicker>` instances for start and end date selection, sending the default start date to `v-model` of the start `<KDatePicker>`
  - Should define a slot to allow a subtitle/description field to appear below the modal's title
  - Should define a `title` string prop to customize the modal's title
  - Should define a `submitText` string prop to customize the submit button
  - Should replace `Start of` and `End of`, in the existing component, with translatable strings removing `of`
  - Should ensure that the <kbd>TAB</kbd> order between `KDatePicker` components is correct
  - Should have a page for it added to the documentation

## (Optional) The Value Add

Adds a rich date range selection for reuse across the Kolibri ecosystem, and will be immediately implemented in Kolibri and KDP

## (Optional) Possible Tradeoffs

To start this component would only allow selection down to the month level, but it could be further customized later to support full date selection. To prevent unnecessary refactors, we added a `selectionLevel` to `<KDatePicker>` in the event we add `day` level selection in the future.

LianaHarris360:
## Description

This PR introduces the `KDateRange` modal component that contains two input components and a calendar component to select a date range.  For translation flexibility, all strings used within the `KDateRange` can be provided as properties, and months are set using the `dateLocale` property.

#### Issue addressed

Closes #360 

### Before/after screenshots

View if there are no pre-selected start and end dates. The date input fields will be set to the placeholder.

HasNotExportedLogs

View if there is a pre-selected start date given. The end date is set to the current day.

HasExportedLogs

- The `dateLocale` property is a locale string used to set the language of the date strings and formatting. The month names shown on the calendar are automatically updated to reflect the language provided by this property.

- The date objects are zero-indexed. January is month zero and December is month eleven. For example, if the `firstAllowedDate` is set to January 1, 2022, using JavaScript's date constructor, it should be created like this: 
`new Date(2022, 0, 1)`

The KDateRange uses the `KModal` component and also contains two `KDateInput` and `KDateCalendar` components that allow for a start and end date selection. The dates can be set by typing into the inputs or selecting a date from the calendar. The date inputs are validated with the use of a state machine that validates that the entered date is:
- In the correct format
- A valid date 
- After the first allowed date
- Before the last allowed date
- An end date that cannot be before the start date

On submission, the returned values are date objects.

## Testing checklist

## Reviewer guidance


lib/KDateRange/KDateCalendar.vue
```diff
@@ -0,0 +1,475 @@
...
+        if (!this.isFirstChoice) {
+          key = 'end';
+        } else {
+          newData['end'] = null;
+        }
+        const resultDate = new Date(activeYear, activeMonth, result);
+        if (!this.isFirstChoice && resultDate < this.dateRange.start) {
+          this.isFirstChoice = false;
+          return { start: resultDate };
+        }
```
bjester, line 228: Since this `if` block returns early, and doesn't depend on the code above (that I can see), would be good to move it to the beginning of the function.

lib/KDateRange/KDateCalendar.vue
```diff
@@ -0,0 +1,475 @@
...
+      /**
+       * update end and start date of dateRange Object
+       */
+      getNewDateRange(result, activeMonth, activeYear) {
+        const newData = {};
+        let key = 'start';
+        if (!this.isFirstChoice) {
+          key = 'end';
+        } else {
+          newData['end'] = null;
```
bjester, line 222: `newData.end` is okay here

lib/KDateRange/KDateCalendar.vue
```diff
@@ -0,0 +1,475 @@
...
+          dayInWeekIndex,
+          nextActiveMonthDay,
+          nextActiveMonthDate
+        );
+        let lastDay = null;
+        if (key === 'first') {
+          lastDay = new Date(this.activeYearStart, this.activeMonth + 1, 0).getDate();
+        } else {
+          lastDay = new Date(this.activeYearEnd, this.nextActiveMonth + 1, 0).getDate();
+        }
```
bjester, line 349: Can be a ternary statement:
```js
const lastDay = (key === 'first')
  ? new Date(this.activeYearStart, this.activeMonth + 1, 0).getDate()
  : new Date(this.activeYearEnd, this.nextActiveMonth + 1, 0).getDate();
```

lib/KDateRange/KDateCalendar.vue
```diff
@@ -0,0 +1,475 @@
...
+      isValidDate(date) {
+        return new Date(date) !== 'Invalid Date' && !isNaN(new Date(date));
+      },
+      getDate(key, day) {
+        let currDate = null;
+        if (key === 'first') {
+          currDate = new Date(this.activeYearStart, this.activeMonth, day);
+        } else {
+          currDate = new Date(this.activeYearEnd, this.nextActiveMonth, day);
+        }
```
bjester, line 373: Ternary here too

lib/KDateRange/KDateCalendar.vue
```diff
@@ -0,0 +1,475 @@
...
+        this.activeYearEnd = prevMonth.getFullYear();
+      },
+      goNextMonth() {
+        const nextMonth = new Date(this.activeYearEnd, this.nextActiveMonth, 2);
+        this.activeMonth = nextMonth.getMonth();
+        this.activeYearStart = nextMonth.getFullYear();
+        this.activeYearEnd = nextMonth.getFullYear();
+      },
+      isValidDate(date) {
+        return new Date(date) !== 'Invalid Date' && !isNaN(new Date(date));
```
bjester, line 365: Is `'Invalid Date'` browser specific? Also, what if they have their browser in a different language?

lib/KDateRange/KDateCalendar.vue
```diff
@@ -0,0 +1,475 @@
...
+  }
+
+  .calendar-days li {
+    display: inline-block;
+    width: 14%;
+    margin-top: 6px;
+  }
+
+  .calendar-days li.calendar-days--disabled {
+    color: var(--disabled-button-color);
```
bjester, line 457: Do we use CSS vars elsewhere in KDS? Seems there [isn't support for them in IE 11](https://caniuse.com/?search=css%20vars)

lib/KDateRange/KDateInput.vue
```diff
@@ -0,0 +1,118 @@
...
+        :label="legendText"
+        autoComplete="off"
+        :invalid="Boolean(errorMessage)"
+        :showInvalidText="Boolean(errorMessage)"
+        :invalidText="errorMessage"
+        @input="handleInput"
+      />  
+      <input type="hidden" name="date" :value="valueAsDate" data-test="valueAsDate">
+      <span class="k-date-vhidden">
+        <span v-if="valueAsDate" id="date-desc">
```
bjester, line 18: `id="date-desc"` :eyes: -- What happens if there are multiple `KDateInput` components on the same page?

lib/KDateRange/KDateInput.vue
```diff
@@ -0,0 +1,118 @@
...
+        }
+        return '';
+      },
+      dateDescription() {
+        if (
+          Object.prototype.toString.call(this.valueAsDate) === '[object Date]' &&
+          isNaN(this.valueAsDate)
+        ) {
+          return ' ';
+        } else {
```
bjester, line 74: Since the prior `if` block returns, you can safely extract this from the `else`, like you did with `valueAsDate`. Makes the code a little cleaner

lib/KDateRange/ValidationMachine.js
```diff
@@ -0,0 +1,140 @@
...
+};
+
+/** Returns if the given date string is after the last allowed date */
+const isDateAfterLastAllowed = (dateStr, lastAllowedDate) => {
+  const [day, month, year] = dateStr.split('/');
+  const newDate = startOfDay(new Date(year, month - 1, day));
+  return isAfter(newDate, lastAllowedDate);
+};
+
+/** Returns if the given date string is before the first allowed date */
```
bjester, line 32: For these functions above, any reason not to also document the arguments, like you've done for `validate`?

bjester (commented):
Left some code review comments. Overall, this looks really awesome! 

An overview of my comments:
- making sure all user facing strings will be translated
- cross browser support and intricacies
- ensuring HTML `id`'s will be unique on a page

LianaHarris360:
> Left some code review comments. Overall, this looks really awesome!
> 
> An overview of my comments:
> 
> * making sure all user facing strings will be translated
> * cross browser support and intricacies
> * ensuring HTML `id`'s will be unique on a page

Thank you for the review, I will work on clarifying the code and making updates with these suggestions!

marcellamaki (commented):
Hi @LianaHarris360 - I've added just a couple of comments here to follow up on some of what @bjester said regarding css variables. Let me know if you have any questions!

LianaHarris360:
> Hi @LianaHarris360 - I've added just a couple of comments here to follow up on some of what @bjester said regarding css variables. Let me know if you have any questions!

@marcellamaki - I will let you know if I have any questions for sure. I am currently looking into those potential solutions, thank you!

lib/KDateRange/KDateCalendar.vue
rtibbles: Just to affirm here that we can't use CSS vars in KDS, because of the lack of IE11 support. We've investigated a ponyfill, but not implemented it yet, due to the technical complexity (and because it is not a complete polyfill it comes with a bunch of caveats).

In general, using computed `:style` attributes is preferable, as it is a little simpler to reason about, but (especially for pseudo-selectors, which cannot be applied via `style` attributes) the `$computedClass` helper function lets you generate arbitrary CSS in JS: https://design-system.learningequality.org/colors/#usage

bjester: Thank you @marcellamaki and @rtibbles!

lib/KDateRange/KDateCalendar.vue
```diff
@@ -0,0 +1,475 @@
...
+      />
+      <KIconButton
+        aria-label="Next Month"
+        tooltip="Next Month"
+        icon="chevronRight"
+        appearance="flat-button"
+        class="right"
+        size="mini"
+        @click="goNextMonth"
+      />
```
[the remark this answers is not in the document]
LianaHarris360: Good catch, I missed these somehow! They should also be provided as properties for translation like the rest of the user-facing strings

lib/KDateRange/KDateCalendar.vue
```diff
@@ -0,0 +1,475 @@
...
+      buttonStyles() {
+        return {
+          '--selected-button-bg-color': this.$themeBrand.secondary.v_50,
+          '--disabled-button-color': this.$themePalette.grey.v_300,
+        };
+      },
+    },
+    watch: {
+      nextActiveMonth(value) {
+        if (value === 0) this.activeYearEnd = this.activeYearStart + 1;
```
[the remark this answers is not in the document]
LianaHarris360: In this case it is because a computed value should not be mutated, referenced [here](https://vuejs.org/guide/essentials/computed.html#best-practices:~:text=a%20computed%20return%20value%20should%20be%20treated%20as%20read%2Donly%20and%20never%20be%20mutated%20%2D%20instead%2C%20update%20the%20source%20state%20it%20depends%20on%20to%20trigger%20new%20computations):

> a computed return value should be treated as read-only and never be mutated - instead, update the source state it depends on to trigger new computations

So when the Previous/Next Month buttons are clicked, the `activeYearStart` and `activeYearEnd` values will be updated directly and that will trigger the computed properties that depend on these values to update as well.

lib/KDateRange/KDateCalendar.vue
LianaHarris360: This is a good point, maybe the `!isNaN(new Date(date))` should be enough to validate the `date` argument.

lib/KDateRange/KDateInput.vue
LianaHarris360: Good point, I remember now that we did talk about a solution that generates unique ids. I will go back and add that.

lib/KDateRange/ValidationMachine.js
LianaHarris360: There is no reason not to. I can see how documenting the arguments makes things clearer, so I'll update these functions.

lib/KDateRange/KDateCalendar.vue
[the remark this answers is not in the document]
bjester: Yes, you shouldn't modify data in a computed prop, but that wasn't quite what I was referring to. So why did you choose a watcher instead of adding this code to the handlers for when the Previous/Next Month buttons are clicked? I do see code in those handlers that are modifying `activeYearEnd`, so it isn't clear why both are needed?

With regards to this specific code-- what happens if the picker is on the month of December, and I click Next, then Previous, then Next, then Previous, and so on? Does `this.activeYearEnd` just keep on incrementing? I suspect the code in the button handlers is possibly covering up a potential issue?

lib/KDateRange/KDateInput.vue
```diff
@@ -0,0 +1,118 @@
+<template>
+
+  <div>
+    <fieldset :aria-labelledby="legendText" class="date-input-fieldset" aria-describedby="date-desc" aria-live="polite">
```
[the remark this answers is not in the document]
bjester: Taking another look here, I believe `aria-labelledby` is supposed to accept element ID list. Is this passing a text string?

> The aria-labelledby property takes as value a space-separated id reference list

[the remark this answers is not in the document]
LianaHarris360: Yes this is correct, in this case, the `aria-labelledby` should be `aria-label` or should maybe point to the KTextBox element. There was previously a legend and the `aria-labelledby` was set to that, but when testing with the chrome extension screen reader, the label would be announced twice so I removed the legend and did not update `aria-labelledby` accordingly. Thank you for pointing this out!

lib/KDateRange/KDateCalendar.vue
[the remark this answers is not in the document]
LianaHarris360: Okay, I misunderstood your question at first. The watcher was chosen as a convenient way of keeping tabs on when the `nextActiveMonth` becomes 0 (January) because that is when `activeYearEnd` should be updated to the next year. 

When `goNextMonth` is pressed, the `activeMonth` is what is actually being updated, not `activeYearEnd`, so you are right, changing `activeYearEnd` inside of the `goNextMonth` method is  redundant.

`activeYearEnd` only needs to update when `nextActiveMonth` becomes 0 or when `goPrevMonth` is pressed and `activeYearStart` isn't the same value as `activeYearEnd`, which indicates that the right half of the calendar is in a different year. So pressing `goPrevMonth` would put both sides of the calendar in the same year when it updates the month, this is what decrements the `activeYearEnd`.

I thought the watcher was appropriate because `nextActiveMonth` is only changed after `activeMonth` is updated (which happens inside `goNextMonth`), so the watcher allows `activeYearEnd` to be updated only when the computed property `nextActiveMonth` is a certain number. 

But, based on what you're suggesting, I should be able to make `activeYearEnd` and `activeYearStart` computed properties where the original values are set using `activeMonth`, so they should both update if `activeMonth` is updated and I can still check the value of `nextActiveMonth` when updating `activeYearEnd`. I can implement it this way and update you if I have any issues!

marcellamaki (approved):
This looks great, @LianaHarris360 - thank you so much for your hard work on this! I think all of the PR feedback has been addressed now, and I feel good about merging this into KDS develop and then you moving forward with your work using it within Kolibri. There's always a chance an edge case or bug might pop up, but that's completely fine. I think we're most likely to surface that as you and the rest of the team begin to use it, and we can continue to build on the component as needed. 

🎉 Amazing, amazing work! 🎉

PR #391, opened 2022-12-01
Robustly handle missing icons.

rtibbles:
## Description
* This makes the `KIcon` component properly handle when it is passed an icon name that does not exist, preventing `TypeError`s
* Defaults to a brokenImage icon in the cases that the icon is not recognised

#### Issue addressed
Fixes https://github.com/learningequality/kolibri-design-system/issues/390

### Before/after screenshots
Added documentation of this behaviour to the main `KIcon` docs.

Screenshot from 2022-12-01 09-49-54

## Testing checklist

## Reviewer guidance


docs/pages/kicon.vue
```diff
@@ -3,6 +3,8 @@
   <DocsPageTemplate apiDocs>
     <DocsPageSection title="Overview" anchor="#overview">
       See <DocsInternalLink href="icons" text="the page on icons" /> for design guidance and a list of available icons.
+      If an invalid icon is used, in development Vue.js validation will warn about the error.
+      The icon will display as a broken image <KIcon icon="brokenImage" /> icon.
```
indirectlylit, line 7:
```suggestion
      See <DocsInternalLink href="icons" text="the page on icons" /> for design guidance and a list of available icons. If an invalid icon is used, in development Vue.js validation will warn about the error. The icon will display as a broken image <KIcon icon="brokenImage" /> icon.
```

indirectlylit: FYI, it's generally easier if content (particularly in `docs/pages`) is soft-wrapped by editors rather than hard-wrapped in the file itself:

* easier editing experience because writers don't need to worry about endline placement
* better diffs in git history over time

Kolibri Tools max-length rules are overwritten here:

https://github.com/learningequality/kolibri-design-system/blob/d24d3d590c6fe82ca8c475c722511ed9410a1da4/.eslintrc.js#L13-L15

However, it looks like I never actually wrote this guidance anywhere visible. Sorry!

marcellamaki (approved):
This looks good to me. I like the approach. And thanks, @indirectlylit for the notes on the docs!

Issue #342, opened 2022-06-28
Kselect should provide the select event


jredrejo:
## Product
This has been needed in KDP but it can be needed anywhere if the UI must know the user has selected again the same option (in KDP it should show a modal whenever the `custom` option is selected, now it's not possible because only the `change` event is available in KSelect)

## Desired behavior
Don't stop the `select` event provided KeenUISelect and let it be available in KSelect

## Current behavior
Currently KSelect only provides a `change` event

This issue should be done after #/339

AlexVelezLl:
Hi! I was reviewing this, but I noticed that the `select` event emitted by `KeenUISelect` expose an options object with the `selected` property. But this property only changes for multi-selects, for the rest it is always set as true, however multi-select is not supported by `KSelect` 🤔 . So, should we change this property to return whether the selected option was previously selected, regardless of whether it is multi-select or not?

rtibbles:
Fixed in #399 - it now produces a select event where the first value is the item selected (and in the case of multiselect, which I don't think we currently allow within KSelect - the second value is an object with a `selected` property saying whether it was selected or not).

PR #399, opened 2022-12-11
[Kselect] Expose select event

AlexVelezLl:
## Description

Expose `select` event provided by KeenUISelect.

#### Issue addressed

Addresses #342

## Steps to test

Use KSelect component and provide a `@select` event handler. The handler should execute on any select option and should receive the selectedOption, and an options object as first and second argument.

`KeenUISelect` emits an `select` event whenever an option is selected. So it just add a handler to the `KSelect` wrapper to emit the `select` event emitted by `KeenUISelect`.

## Implementation notes

The `options` argument emitted by the `select` event is just an object with just one boolean property `selected` that indicates whether the selected option was previous selected.

```JSX
<KSelect
  ...
  @select="onSelectHandler"
/>
...
onSelectHandler(newSelection, options) {
  console.log(newSelection); // New selected option
  console.log(options.selected); // True if the selected option was previously selected
}
```

## Testing checklist

## Reviewer guidance

### DIFF
--- a/lib/KSelect/KeenUiSelect.vue
+++ b/lib/KSelect/KeenUiSelect.vue
@@ -666,7 +666,7 @@
         }
 
         this.$emit('select', option, {
-          selected: this.multiple ? shouldSelect : true,
+          selected: this.isOptionSelected(option),
         });
 
         this.clearQuery();
--- a/lib/KSelect/KeenUiSelect.vue
+++ b/lib/KSelect/KeenUiSelect.vue
@@ -666,7 +666,7 @@
         }
 
         this.$emit('select', option, {
-          selected: this.isOptionSelected(option),
+          selected: !this.isOptionSelected(option),
         });
 
         this.clearQuery();
--- a/lib/KSelect/KeenUiSelectOption.vue
+++ b/lib/KSelect/KeenUiSelectOption.vue
@@ -123,8 +123,9 @@
   $ui-select-option-checkbox-color: rgba(black, 0.38) !default;
 
   .ui-select-option {
-    @include font-family-noto display: flex;
+    @include font-family-noto;
 
+    display: flex;
     align-items: center;
     font-size: $ui-dropdown-item-font-size;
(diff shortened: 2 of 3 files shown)


lib/KSelect/KeenUiSelectOption.vue
```diff
@@ -123,8 +123,9 @@
   $ui-select-option-checkbox-color: rgba(black, 0.38) !default;
 
   .ui-select-option {
-    @include font-family-noto display: flex;
+    @include font-family-noto;
```
AlexVelezLl, line 126: Fix syntax error

lib/KSelect/KeenUiSelect.vue
```diff
@@ -666,7 +666,7 @@
         }
 
         this.$emit('select', option, {
-          selected: this.multiple ? shouldSelect : true,
+          selected: this.isOptionSelected(option),
```
rtibbles, line 669: This change means that the `selected` property is always false for a newly selected item (it's possible to test this using the vue dev tools to inspect the select event on the documentation).

I think `isOptionSelected` might be reporting the state prior to the select, so I think it might be safe to just report the negation:
```suggestion
          selected: !this.isOptionSelected(option),
```

rtibbles (changes requested):
One quibble with the `selected` property, otherwise it seems to be working as intended.

lib/KSelect/KeenUiSelect.vue
AlexVelezLl: Oooh, yes, I was seeing the `selected` property as "previously selected", "This item that you have selected was selected before?", but now I see that should mean "Has the new item been selected?" and say true if it has not been selected before and should be selected now , and false otherwise

AlexVelezLl: Thanks @rtibbles! Changes made :)

rtibbles (approved):

Issue #285, opened 2021-12-14
change KEmptyPlaceholder to KOptionalText


indirectlylit:
## Product

Kolibri

## Desired behavior

`KEmptyPlaceholder` is always used with a `v-else` directive next to another text block.

We should have a `KOptionalText` component will either display the text or the existing `KEmptyPlacehlder` if it's an empty string, undefined, or null. 

## (Optional) The Value Add

Simplify product code

rtibbles:
This has been addressed by #403 and updated in Kolibri in https://github.com/learningequality/kolibri/pull/10050

PR #403, opened 2022-12-17
[KOptionalText] New component

AlexVelezLl:
## Description
Add a new component to KDS `KOptionalText` that will either display the text or `KEmptyPlacehlder` if it's an empty string, undefined, or null.

#### Issue addressed

Addresses #285

## Steps to test

1. Link this KDS branch to any product.
2. Use `KOptionalText` component as described in Implementation notes.

`KOptionalText` will recieve a `text` prop or slot that renders text nodes, if there is text to show, the text or the slot will be shown whithin a `span` element, if not, the `KEmptyPlacehlder` component will be shown. Also a `textStyle` prop can be passed which will be set to the `span` element.

Examples:

All of them will render "Hello world" in red color:

```vue
<KOptionalText text="Hello world" appearanceOverrides="color: red" />

<KOptionalText appearanceOverrides="color: red">
  Hello World
<KOptionalText/>

<KOptionalText>
  <span style="color: red">Hello </span>
  <span style="color: red">World</span>
</KOptionalText>
```

If `name` and `lastname` are nonempty strings name and lastname will be shown. If not it will render a "-":
```vue
<KOptionalText :text="`${name} ${lastname}`" />

<KOptionalText>
  <span>{{ name }}</span>
  <span>{{ lastname }}</span>
</KOptionalText>
```

## Testing checklist

## Reviewer guidance

### DIFF
--- a/docs/pages/koptionaltext.vue
+++ b/docs/pages/koptionaltext.vue
@@ -9,7 +9,7 @@
         Using the KOptionalText component is preferred to leaving a empty or blank space because it communicates to the user that a value <em>could</em> possibly exist there, and aids with visual scanning in tables, grids, and lists of data.
       </p>
     </DocsPageSection>
-    <DocsPageSection title="Example" anchor="#example">
+    <DocsPageSection title="Example using props" anchor="#example-props">
       <p>
         In the example below, an average score cannot be calculated until the user has attempted at least one exercise:
       </p>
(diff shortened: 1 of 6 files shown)


lib/KOptionalText.vue
```diff
@@ -0,0 +1,76 @@
...
+       * Text to display
+       */
+      text: {
+        type: String,
+        default: null,
+      },
+      /**
+       * If provided, sets the styles of the text
+       */
+      textStyle: {
```
akolson, line 37: for consistency, its probably better to use the `appearanceOverrides`, please see the [`KButton`](https://design-system.learningequality.org/kbutton/) as a usecase

lib/KOptionalText.vue
```diff
@@ -0,0 +1,76 @@
...
+        return !slotsText.trim();
+      },
+    },
+    methods: {
+      getSlotsText() {
+        if (!this.$slots.default || !this.$slots.default.length) {
+          return '';
+        }
+
+        return this.$slots.default
```
akolson, line 58: I think it would be better to extract this code into an independent method and instead return the new method so that it is more clearer what the code is doing. At the moment the `getSlotsText`. method is doing one thing too many.

lib/KOptionalText.vue
```diff
@@ -0,0 +1,76 @@
...
+          .map(node => {
+            if (node.text) {
+              return node.text;
+            }
+            if (node.children) {
+              return node.children.map(child => child.text).join('');
```
akolson, line 64: Refactoring the above into a method(as mentioned in my previous comment) would allow you to call the method here as well, without having to explicitly use the `map` as i think the same checks still to the children nodes as well, though i could be overthinking it :)

akolson (changes requested):
Hi @AlexVelezLl code generally LGTM! Please see herein a few recommendation

AlexVelezLl:
Hi @akolson! Thank you for the recommendations! I have updated the branch with them :)

rtibbles (changes requested):
I think it would be good to add this new component to the documentation, especially as you have already written out examples. See here on how to do this: https://github.com/learningequality/kolibri-design-system#development-of-the-documentation-site

The other slight concern is that the Kolibri frontend tests for this integration are failing - showing empty string instead of the placeholder text. Possible that this is a result of stubbing of child components, but would be good to verify what is happening there.

AlexVelezLl:
Hi @rtibbles! tests on Kolibri were fixed, and docs for this component were added! Let me know if everything is ok with the wording of the docs. Once this PR is merged, I can proceed with updating https://github.com/learningequality/kolibri/pull/10050 to bring the reference back to the main KDS branch.

docs/pages/koptionaltext.vue
```diff
@@ -0,0 +1,90 @@
...
+                <KOptionalText text="5" />
+              </td>
+              <td class="table-number">
+                <KOptionalText text="100%" appearanceOverrides="color: green" />
+              </td>
+            </tr>
+          </tbody>
+        </table>
+      </DocsShow>
+    </DocsPageSection>
```
rtibbles, line 68: Could you add a second example showing how it works with the default slot as well?

AlexVelezLl: Sure! Could it be a new row on the example table? Or should it be a new different example?

rtibbles: I think maybe do it as a separate example, and then you can add some explanatory text about how it is different to the example above. Can even show the difference in a codeblock like you did in your PR description. For an example of showing code instead of rendering the code, see here: https://github.com/learningequality/kolibri-design-system/blob/develop/docs/pages/kcheckbox.vue#L67

AlexVelezLl: Got it! I have added one more example, with code examples 😃

rtibbles (approved):

Issue #369, opened 2022-09-22
KImg: Implement the basic component


MisRob:
## Summary

Implement the KImg component with the following props
- `src`
- `altText`
- `isDecorative`
- `height`
- `width`
- `maxHeight`
- `minHeight`
- `maxWidth`
- `minWidth`

and the default slot.

See the epic issue for props and slot definitions.

## Epic

- https://github.com/learningequality/kolibri-design-system/issues/368

## Acceptance criteria

- [ ] The component contains all the props specified above and the default slot
- [ ] Setting the `src` prop renders an image in the component
- [ ] The image is rendered as `<img>` (not as the background image)
- [ ] Setting the `altText` is mandatory and throws a warning when not provided(empty) unless `isDecorative` is `true`
- [ ] Setting `isDecorative` to `true` specifies an empty string to the `alt` image property and `altText` becomes optional so no warning is thrown and `altText` is overridden to the empty string when provided.
- [ ] Setting `height`, `width`, `maxHeight`, `minHeight`, `maxWidth`, `minWidth` alters the dimensions of the component

## Linked issues

- **Blocking**
  - https://github.com/learningequality/kolibri-design-system/issues/370
  - https://github.com/learningequality/kolibri-design-system/issues/371
  - https://github.com/learningequality/kolibri-design-system/issues/372
  - https://github.com/learningequality/kolibri-design-system/issues/373
  - https://github.com/learningequality/kolibri-design-system/issues/374

nucleogenesis:
Closing as @MisRob has reviewed

Issue #368, opened 2022-09-22
KImg Component


akolson:
# Overview
The `KImg` component displays images and provides additional functionality to manipulate the images such as
- Image aspect ratio
- Image scaling

# Motivation

Inspired by thumbnails to a large extent so it should satisfy our needs for working with them, but should be usable as a general image component as well.

# Interface

### Props

| Name | Description | Type | Default | Required |
| - | - | - | - | - |
| `src` | Sets the image path | `String` | `null` | `false` |
| `altText` | Alternate text for the image. This is required and will throw a warning when its not provided(empty) unless `isDecorative` is `true` | `String` |  | `true` |
| `isDecorative` | Sets the image as decorative. This sets the `alt` image property to an empty string. | `Boolean` | `false` | `false` |
| `height` | Sets the height for the component. | `Number`, `String` | `undefined` | `false` |
| `width` | Sets the width for the component. | `Number`, `String` | `undefined` | `false` |
| `maxHeight` | Sets the maximum height for the component. | `Number`, `String` | `undefined` | `false` |
| `minHeight` | Sets the minimum height for the component. | `Number`, `String` | `undefined` | `false` |
| `maxWidth` | Sets the maximum width for the component. | `Number`, `String` | `undefined` | `false` |
| `minWidth` | Sets the minimum width for the component. | `Number`, `String` | `undefined` | `false` |
| `aspectRatio` | Sets the ratio of the width(w) to the height(h) of the image container. The required format is `w:h`, otherwise `<KImg>` will show an error. | `String` | `undefined` | `false` |
| `scaleType` | Specifies how the image should be scaled within the container. Can be one of `centerInside`, `contain` or, `fitXY`. See below the scale types to be supported.  | `String` | `contain` | `false` |
| `placeholderColor` | Sets color for the image placeholder | `String` | `$themePalette.grey.v_200` | `false` |
| `borderRadius` | Sets tne boarder radius of the image or placeholder | `Number` | `undefined` | `false` |
| `appearanceOverrides` | Overrides styling of the component | `Object` | `null` | `false` |
| `kolibriFly` | Renders the Kolibri fly logo into the `KImg`. This overrides the `src` specified | `Boolean` | `false` | `false` | 

#### `scaleType`

| Type | Description | Illustration |
| - | - | - |
[...]

PR #404, opened 2023-01-11
Initial implementation of KImg component

thanksameeelian:
## Description
This is the initial implementation of the KImg component, which will ultimately fulfill all the specifications of [the KImg epic](https://github.com/learningequality/kolibri-design-system/issues/368), but for this round only includes `src`, `altText`, `isDecorative`, `height`, `width`, `min-width`, `max-width`, `min-height`, and `max-height`.

#### Issue addressed

Addresses #369 

## Steps to test

1. `yarn dev`
2. `KImg` doc page should be accessible at http://localhost:4000/kimg
3. can edit any of the examples in docs/pages/kimg.vue to test functionality and see results reflected on the KImg doc page

## Testing checklist

## Reviewer guidance


MisRob:
@jtamiace if you'd like to read through the documentation, we'd appreciate your input

MisRob:
Documentation preview https://deploy-preview-404--kolibri-design-system.netlify.app/kimg/

docs/tableOfContents.js
```diff
@@ -232,6 +232,11 @@ export default [
         isCode: true,
         keywords: buttonRelatedKeywords,
       }),
+      new Page({
+        path: '/kimg',
+        title: 'KImg',
+        isCode: true,
```
akolson, line 238: It would be nice to add the `keywords` property to allow for better searching within the documentation. For example, currently when you try to search an image component using the "image" keyword, the search would yield no results. Please see usages within this file for implementation details

docs/pages/kimg.vue
```diff
@@ -0,0 +1,113 @@
...
+              :height="'250px'"
+              :width="225"
+            />
+          </DocsShow>
+
+          <pre><samp>
+            :src="require('../assets/img_sample_for_kimg.png')"
+            altText="Computers are run by bees"
+            :height="'250px'"
+            :width="225"
```
akolson, line 30: It would be good to add the entire snippet(including the `KImg` tag as well) and not just the properties for better readability and usability from the developers' perspective. Also, wrapping `<DocsShowCode>{Your snippet}</DocShowCode>` around the snippet would give it a better formatting.

docs/pages/kimg.vue
```diff
@@ -0,0 +1,113 @@
...
+                :height="50"
+                :width="50"
+              />
+            </DocsShow>
+          </div>
+          <pre><samp>
+            :src="require('../assets/img_sample_for_kimg.png')"
+            isDecorative
+            :height="50"
+            :width="50"
```
akolson, line 54: Please see the comment above

lib/KImg.vue
```diff
@@ -0,0 +1,159 @@
+<template>
+
+  <div>
+    <img
+      id="kimg"
```
akolson, line 5: Using id could prove problematic especially when more than one `KImg` is used on a single page. It is on this basis that css class referencing is recommended instead

lib/KImg.vue
```diff
@@ -0,0 +1,159 @@
...
+      },
+      cssVars() {
+        return {
+          '--height': this.imgHeight,
+          '--width': this.imgWidth,
+          '--max-height': this.imgMaxHeight,
+          '--min-height': this.imgMinHeight,
+          '--max-width': this.imgMaxWidth,
+          '--min-width': this.imgMinWidth,
+        };
```
akolson, line 115: This seems like neat way to achieve the styling, however it looks like [css variables are not supported in IE 11](https://caniuse.com/css-variables) and this would definitely break compatibility.  [Class & style binding](https://vuejs.org/guide/essentials/class-and-style.html) might be a better approach to this

lib/KImg.vue
```diff
@@ -0,0 +1,159 @@
...
+      }
+    },
+    methods: {
+      validateAndFormatUnits(propValue) {
+        if (propValue) {
+          if (Number.isInteger(propValue)) {
+            return `${propValue}px`;
+          } else {
+            // split numbers apart from units
+            const [, ...arr] = propValue.match(/(\d*)([\s\S]*)/);
```
akolson, line 130: @thanksameeelian Awesome approach 💯  :) This was so much fun and exciting to review. It felt like a sprinkling of magic dust!

lib/KImg.vue
```diff
@@ -0,0 +1,159 @@
...
+<style scoped>
+
+  #kimg {
+    height: var(--height);
+    width: var(--width);
+    max-height: var(--max-height);
+    min-height: var(--min-height);
+    max-width: var(--max-width);
+    min-width: var(--min-width);
+  }
```
akolson, line 156: Please see `cssVars` comment

lib/KImg.vue
```diff
@@ -0,0 +1,159 @@
...
+    },
+    created() {
+      if (!this.isDecorative && !this.altText) {
+        throw new Error('Missing required prop - provide altText or indicate isDecorative');
+      }
+    },
+    methods: {
+      validateAndFormatUnits(propValue) {
+        if (propValue) {
+          if (Number.isInteger(propValue)) {
```
akolson, line 126: This check would fail for a `propValue` that is a floating point number(eg 30.5).  The same `propValue` but with a unit(eg 30.5rem) would pass the validation. This not only causes an inconsistency that needs to be streamlined but also means that the regex fails for floating point numbers with units(eg 30.5rem). Maybe using the `isNaN` is a better option. This would also mean updating the regex to capture floating point numbers as well. I tweaked your regex to cater for the decimal points part and the final result is this `(\d*\.\d+)([\s\S]*)`

akolson:
Hi @thanksameeelian I reviewed your code alongside @MisRob and we think your code generally looks good! Great Job :). We found few issues for you to look into. Also, to give us better design perspective especially on the documentation, we've added @jtamiace to give her thoughts on  https://deploy-preview-404--kolibri-design-system.netlify.app/

akolson:
Something that we pondered about was whether it was necessary to widen the scope of units used. At the moment, only `px`, `em` `rem`, `vh` are being used. @marcellamaki We are happy to hear your thoughts

akolson (changes requested):
Please see comments

thanksameeelian:
> Hi @thanksameeelian I reviewed your code alongside @MisRob and we think your code generally looks good! Great Job :). We found few issues for you to look into. Also, to give us better design perspective especially on the documentation, we've added @jtamiace to give her thoughts on https://deploy-preview-404--kolibri-design-system.netlify.app/kimg/

thanks so much @akolson & @MisRob! i've read through your comments & i don't think i have any questions about them, but will be sure to ask if i end up needing clarification. i'll get on these changes but also keep my eyes open for feedback from the design team!

thanksameeelian:
> Something that we pondered about was whether it was necessary to widen the scope of units used. At the moment, only `px`, `em` `rem`, `vh` are being used. @marcellamaki We are happy to hear your thoughts

it was just an oversight that i didn't include a more exhaustive list of units initially, so i'm def happy to update and try to catch 'em all! 😀

lib/KImg.vue
akolson: This is another update to my earlier recommended regex. using `(\d*\.?\d+)([a-zA-Z,\s]*)`  instead will ensure that values like 30px, 12.5rem would pass validation. However values like 10.em would fail validation as such an input would not work when applied in css.

marcellamaki:
> Something that we pondered about was whether it was necessary to widen the scope of units used. At the moment, only `px`, `em` `rem`, `vh` are being used. @marcellamaki We are happy to hear your thoughts

@akolson @MisRob @thanksameeelian - I think having more units would be great idea! At the same time, in my opinion it isn't necessarily a **blocker**, unless we can think of a specific use case in one of our products right now where we would likely be needing/using some other unit. If so, that should be prioritized - otherwise I see no problem with expanding this on an as-needed basis.

Thanks Amelia for this great work, and Misha and Samson for the very thoughtful and comprehensive review! Very exciting to see this happening :)

jtamiace:
This looks great!  In terms of additional content, I'm not 100% sure if this suggestion is appropriate for this page which is focused on technical guidance and documentation, but I'm wondering if it makes sense to record any general patterns we use for missing images or error states if an image is broken, or if this is something that's out of the scope of KImg and handled elsewhere. For example in Studio and the Learning Platform we use a light background gray along with a centered icon of the resource type or a generic "image" icon.

MisRob:
One unit I'd suggest adding would be `%` since it's used commonly

thanksameeelian:
> This looks great! In terms of additional content, I'm not 100% sure if this suggestion is appropriate for this page which is focused on technical guidance and documentation, but I'm wondering if it makes sense to record any general patterns we use for missing images or error states if an image is broken, or if this is something that's out of the scope of KImg and handled elsewhere. For example in Studio and the Learning Platform we use a light background gray along with a centered icon of the resource type or a generic "image" icon.

thank you so much for your feedback jessica!! we've split this project up over several prs ([here's the epic!](https://github.com/learningequality/kolibri-design-system/issues/368)) so creating a placeholder image will be handled in the next pr, and additional error states potentially after that. i'd really appreciate your feedback on those once they're ready if you have the time, so i'll be sure to add examples of them in the doc & tag you/request your review!

docs/tableOfContents.js
akolson: I think the changes made should suffice. Thanks @thanksameeelian. Non-blocking but would be good to follow existing patterns i.e. pass variable to keywords as opposed to the actual array object

thanksameeelian:
thanks so much for the initial round of review, @akolson, @MisRob, @marcellamaki, & @jtamiace!! i made all the changes indicated and would love to hear your thoughts on this new version! in terms of CSS units, i added `%` (and adjusted the regex) as well as all of the units i think i've encountered in the wild, based on [lists provided here](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units) - i'd be happy to list all of them if you think that's appropriate, but just didn't want to create anything overwhelming at the moment 🙂 really interested to hear your thoughts!

docs/tableOfContents.js
thanksameeelian: thanks! i checked out that pattern when i was adding the keywords here, but only saw the pattern of using a variable when multiple items had the same keywords, so the variable was being used in at least two places. 

i don't think that pattern is applicable here, as `KImg` is the only item using those specific keywords. i am following the pattern i saw for each item that had a unique list of variables, as in those cases the array itself was passed. you can see this patten in `KSelect` and `KSwitch` in the table of contents. i think this makes sense because it avoids unnecessary abstraction. 

but if another item were to share the same keywords at kimg, i would totally agree that a variable would be the right call!

akolson (approved):
Hi @thanksameeelian,  Thanks for the changes, they LGTM! @MisRob @jtamiace, any other thoughts? Also @marcellamaki , at what point do we do manual QA on this? Should we wait for the component to develop further or have QA take a first look?

marcellamaki:
While I think some manual QA would be ideal in most situations, since we don't an immediate github issue in Kolibri that requires this as a fix, and we don't have plans to refactor immediately, I think it's okay to proceed without manual QA for now. There have been a lot of eyes on this, and thoughtful work and feedback, so I feel okay about moving forward without that testing. And, we should remain open to the possibility that some tweaks may be needed as we start integrating this into our products, and that's okay 😄

MisRob (approved):
Thank you @thanksameeelian, good work.

I too feel fine about doing manual QA as soon as the component is complete, as some upcoming issues may perhaps require tweaking code here and there a bit in areas that were implemented before.

Issue #324, opened 2022-03-21
Kmodal should be able to show dropdowns select without making scrollbars appear


jredrejo:
## Product
KDP, but applicable to Kolibri as well

## Expected behavior
When a dropdown is clicked in a modal window, the option items should open over the modal window, covering the actions div. That's what vuetify does by default: https://codesandbox.io/s/serverless-meadow-8z3qu?file=/src/components/Playground.vue

## Actual behavior
It seems [this code](https://github.com/learningequality/kolibri-design-system/blob/main/lib/KModal.vue#L267-L269) is recalculating the size of the modal on the fly, making the scrollbars appear causing an ugly effect:
![problem](https://user-images.githubusercontent.com/1008178/159321138-74eb22d3-6800-4f9d-879d-d2c6cd6751dd.gif)

## Steps to reproduce the issue
Insert a KSelect in a KModal with a few items, and click on it

jtamiace:
related: https://github.com/learningequality/kolibri/issues/4442

rtibbles:
The difference between Vuetify VSelect behaviour and the KSelect behaviour seems to be that VSelect attaches the menu to the DOM outside of the modal, and then positions it to be in the correct place visually.

This doesn't seem to do anything bad for keyboard navigation, but we might need to take a look at the vuetify source to get a robust solution for this if we wanted to emulate it.

MisRob:
May be related https://github.com/learningequality/kolibri-design-system/issues/210

rtibbles:
Fixed in #429

akolson:
Reopening this issue after experiencing a [similar issue](https://github.com/learningequality/studio/pull/4666#issuecomment-2312999234) while working with KSelects in Modals in Studio.

MisRob:
Some of the work introduced to KDS in https://github.com/learningequality/kolibri-design-system/pull/722 may help to resolve this

lokesh-sagi125:
hey @jredrejo i would like to try solving this bug? i saw no 'help wanted' tag can i still work on this?

AlexVelezLl:
Hey @lokesh-sagi125! Right now you have other 2 issues assigned, lest solve them before assigning you another issue! :) Thanks for your willingness to contribute to Kolibri 💓

lokesh-sagi125:
yes @AlexVelezLl  ill make sure to solve them before getting assigned here, thanks:)

PR #422, opened 2023-02-08
DRAFT - WIP: kselect in kmodal, existing vendored UiPopover edition

thanksameeelian:
# TODO: Fix highlighting functionality & ability to arrow up/down through options
i've left some notes about my struggles in `KeenUISelect.vue`, lines 793-818 & replicated below:

```
      // i have attempted to take what we need from KeenUI's update to UiSelect and accomplish the rest through
      // additional functions & CSS targeting. the dropdown now extends beyond the modal as desired but you cannot tab
      // through it and it does not highlight upon mouseover (though that specifically could be solved with CSS :hover
      // targeting, it doesn't address the underlying loss of functionality)
      //
      // each time i end up in a tangle...some highlights (lowlights haha):

      // - with unintended side effects leading to behavior of dropdown closing upon highlight;

      // - unsuccessfully attempted to adapt functionality from KDropdownMenu where UiPopover is ostensibly working with
      // highlight functions [rather than CSS targeting of :hover] & tabbing through options;
      //
      // - this onMouseover function is "working" and mouseovers are registering, still not providing desired
      // highlight-upon-hover functionality;
      //
      // - CSS targeting of each element on :hover in KeenUiSelectOption solves highlight-upon-hover issue on surface
      // level but does not address inability to tab through options and simply disguises broken functionality underneath).
      //

      // have tried to clean up most attempts since they were muddying the waters 🫠
```

resources i tried to use to inform my work: 
- we have implemented `UiPopover` with functioning highlighting & arrow up/down through options in [KDropdownMenu.vue](https://github.com/learningequality/kolibri-design-system/blob/6bae8277e0dd61d88122a1a85b63bc79a65be22e/lib/KDropdownMenu.vue)
- [the specific changes made in KeenUI](https://github.com/JosephusPaye/Keen-UI/commit/e593d10729d62e25ad042d9a7164b4ff44531c48) that fixed the issue of the dropdown being constrained to modal (however there were also changes to UiPopover that we have not made)

## Description
Previously, if `KSelect` was within a modal its height would be constrained to that of the modal and unnecessary scrollbars would sometimes appear. By nesting `KeenUiSelect` within a `UiPopover` component,  `KSelect` is no longer limited to the dimensions of the modal and is rendered in a separate part of the DOM.

#### Issue addressed

Addresses #324: Kmodal should be able to show dropdowns select without making scrollbars appear 

### Before/after screenshots

problem

Screen Shot 2023-02-08 at 12 25 17 PM

## Steps to test

- In this branch I've included a temporary test page within the KDS docs that features `KSelect` within a `KModal`. Once you're running this branch of KDS locally, you can locate it at http://localhost:4000/kselectinkmodal and do any testing or code experimentation that you want there
- You can also link your local Kolibri to this branch and test it out on existing `KSelect`s within `KModal`s, of course!

## Testing checklist

## Reviewer guidance

### DIFF
--- a/docs/pages/kselectinkmodal.vue
+++ b/docs/pages/kselectinkmodal.vue
@@ -1,0 +1,75 @@
+<template>
+
+  <DocsPageTemplate>
+    <DocsPageSection title="Overview" anchor="#overview">
+      <DocsShow>
+        <KModal
+          :title="'KSelect inside KModal'"
+        >
+          <KSelect
+            v-model="exampleData"
+            :options="options"
+            label="Who are you?"
+            placeholder="Nobody"
+          />
+          <p>Value: {{ exampleData }}</p>
+        </KModal>
+      </DocsShow>
+    </DocsPageSection>
+  </DocsPageTemplate>
+
+</template>
+
+<script>
+
+  export default {
+    data() {
+      return {
+        exampleData: null,
+        options: [
+          {
+            label: 'Somebody',
+            value: 'somebody',
+          },
+          {
+            label: 'You used to know me',
+            value: 'usedto',
+          },
+          {
+            label: 'friend',
+            value: 'friend',
+          },
+          {
+            label: 'roman',
+            value: 'roman',
+          },
+          {
+            label: 'countryman',
+            value: 'countryman',
+          },
+          {
+            label: 'abraham lincoln',
+            value: 'lincoln',
+          },
+          {
+            label: 'albert einstein',
+            value: 'einstein',
+          },
+          {
+            label: 'bob barker',
+            value: 'barker',
+          },
+          {
+            label: 'billie jean',
+            value: 'billiejean',
+          },
+        ],
+      };
+    },
+  };
+
+</script>
+
+<style scoped>
+
+</style>
--- a/docs/tableOfContents.js
+++ b/docs/tableOfContents.js
@@ -255,6 +255,11 @@ export default [
         keywords: ['field', 'box'],
       }),
       new Page({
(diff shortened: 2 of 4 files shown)


marcellamaki (commented):
Hi @thanksameeelian - thanks for your work on this. 

I have started taking a look, and I do think there are some pieces that just aren't fitting together with the components as they exist right now and the focus especially. 

I saw your note regarding the changes to UiPopover and not wanting to change the scope of this PR, and I totally understand that, and I also think that UiPopover is not really used in that many places - just our dropdown menus and buttons. So, doing regression testing there would be very manageable. 

I just wonder if that might solve some of your problems. I started playing around with it a little bit, by pulling in the new UIPopover code and it does seem to still work the previous version of the KSelect/index.vue file, although there were some errors in the modal version. I am curious if this would be the easier path. 

I'd suggest maybe trying it out and making another branch off of develop (i.e. don't edit your work here) and then just give yourself a time-boxed amount of time of 90 minutes on Monday morning, to pull in the changed code, use the existing Kselect/index.vue on `develop`, and just see what happens. Don't worry about regressions yet. As for the additional dependencies - from first I don't think we need to worry about `tippy` and I don't think we'd need the functionality for anything. I'd try just deleting the import and any references to it to start. 

Let me know if you have questions - we can talk through it more next week and dig into things if that is a wild goose chase and we need to come back to this approach... I'm just curious if in the end, an alternative might actually be more straightforward. 

ETA: not sure if this was already a part of your discussions with @rtibbles about the vendoring and he suggested some reason why this would not be a good approach. In which case, ignore me!

PR #423, opened 2023-02-15
Update icon page link to ensure proper navigation

rtibbles:
## Description
Currently clicking on the `icons` page link in the KIcon page takes you to a 404 - this does not happen when running locally though.

Updates the href to be an absolute path to ensure it navigates to the correct place.


docs/pages/kicon.vue
```diff
@@ -2,7 +2,7 @@
 
   <DocsPageTemplate apiDocs>
     <DocsPageSection title="Overview" anchor="#overview">
-      See <DocsInternalLink href="icons" text="the page on icons" /> for design guidance and a list of available icons. If an invalid icon is used, in development Vue.js validation will warn about the error. The icon will display as a broken image <KIcon icon="brokenImage" /> icon.
+      See <DocsInternalLink href="/icons" text="the page on icons" /> for design guidance and a list of available icons. If an invalid icon is used, in development Vue.js validation will warn about the error. The icon will display as a broken image <KIcon icon="brokenImage" /> icon.
```
indirectlylit, line 5: There are (at least) a couple other places where this issue seems to exist:

https://github.com/learningequality/kolibri-design-system/blob/37dcf4f8db24d0afc5a72586bb35771dd29c8b1d/docs/pages/ktextbox.vue#L5

https://github.com/learningequality/kolibri-design-system/blob/37dcf4f8db24d0afc5a72586bb35771dd29c8b1d/docs/pages/kmodal.vue#L5
 

Could possibly add some validation and/or fallback behavior to the `DocsInternalLink` component? Looks like most `href` attributes include a preceding `/` and this is the technically correct behavior.

MisRob:
Will need to be rebased on top of the latest develop to get the new changelog. Also, please update it (the check doesn't work here yet, but perhaps it will start after the rebase)

rtibbles:
Is there a reason this is targeted to develop rather than release? Bad handling of missing icons is one of the most common errors we see emanating from KDS on Sentry.

rtibbles:
Sorry, scratch that - the extra commits here are the result of unmerged work from 1.4.x, attempting to resolve here: https://github.com/learningequality/kolibri-design-system/pull/456

MisRob:
> Is there a reason this is targeted to develop rather than release?

No, it was just uninformed best guess based on the age of this PR.

> the extra commits here are the result of unmerged work from 1.4.x, attempting to resolve here

Ah, thank you. Should we close this one then?

MisRob:
Merged https://github.com/learningequality/kolibri-design-system/pull/456

docs/pages/kicon.vue
indirectlylit: This change was lost. Added a PR here: https://github.com/learningequality/kolibri-design-system/pull/462

PR #425, opened 2023-03-04
Adds 'pinned' and 'unpinned' icons. Updates 'cloud' icon to outline

akolson:
## Description

This PR adds the `pinned` and `unpinned` icons. It also updates the `cloud` icon to an outline icons to match that in the [remote content browsing spec](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=194%3A2922).

#### Issue addressed

NA

### Before/after screenshots

NA

## Reviewer guidance


akolson:
Hi @jtamiace  please confirm that the naming of the icons is okay. 
image

radinamatic:
Maybe for the second would be better `not-pinned`...? 🤔

akolson:
@radinamatic It looks like hyphens are not used in naming icons. from the convention I see only `camelCase` and `lowercase` naming. So maybe `notPinned`? Though still leaning towards maintaining the `unpinned` name :)

radinamatic:
Thank you, I was forgetting that convention...  `notPinned` sounds good 👍🏽 

The reason I'm favoring it instead of `unpinned` is because the latter may signify that something previously has been _pinned_ and then got _unpinned_, which may not correspond to the actual user case. **`notPinned`** just declares the current state, does not give information about previous ones.

akolson:
Thanks for this insight. Let me make this change

PR #426, opened 2023-03-06
KTabsList: Fix incorrect focus state + add a new event

MisRob:
## Description

- Removes setting `focusedTabIdx` data in `mounted` to the active tab idx as the active tab doesn't necessarily need to be focused
- Adds a new `'click'` event in addition to the already existing `'activate'` where the former is emitted only when a user interacts with tabs, and the latter can be emitted programmatically in addition to that. It is motivated by some Kolibri use cases where I need to have more nuanced information available about user interaction.

## Steps to test

- You can visit[ the `KTabsList` documentation page preview](https://deploy-preview-426--kolibri-design-system.netlify.app/ktabslist/) and check that the live example is not broken
- As these are just little follow-ups to a bigger whole that was reviewed and tested in Kolibri thoroughly recently, perhaps for now just focusing on if the code changes make sense. Also, you will soon see it in action in Kolibri's related final PR. 

## Testing checklist

## Reviewer guidance

- [ ] ~Are all UI components LTR and RTL compliant (if applicable)?~


marcellamaki (approved):
looks good, thanks @MisRob!

PR #433, opened 2023-04-21
Add an option to freeze loader

MisRob:
## Description

Accompanying PR to https://github.com/learningequality/kolibri/pull/10558 which, by means of two new props `show` and `freeze`, adds an option to prevent the loader from hiding for the chosen amount of time even when the condition for displaying it (e.g. "is the request still processing?") is not truthy anymore. Example:

```vue
<KCircularLoader
  :show="shouldShow"
  :freeze="3000"
/>
```

means that the loader component will be displayed for 3 seconds even when `shouldShow` is falsy sooner than 3s.

This is to prevent the loader from appearing and disappearing very fast (see `props` comments to understand how this is different from the `delay` prop)

## Steps to test

It's used in https://github.com/learningequality/kolibri/pull/10558. When you click the download button on a remote content page, the loader will stay there for 3 seconds even though the mocked download takes only 1 second to complete. 

## Comments

I am not sure if `freeze` is a good name because it may suggest that we freeze it in the sense of stopping loading animation which is not the case. Thoughts welcome.

## Testing checklist

## Reviewer guidance

### DIFF
--- a/lib/loaders/KCircularLoader.vue
+++ b/lib/loaders/KCircularLoader.vue
@@ -11,6 +11,7 @@
 
   <transition name="ui-progress-circular--transition-fade">
     <div
+      v-if="showLoader"
       class="ui-progress-circular"
       :class="[classes, { delay }]"
       :style="{ 'width': size + 'px', 'height': size + 'px' }"
@@ -92,6 +93,13 @@
         default: 'indeterminate', // 'indeterminate' or 'determinate'
       },
       /**
+       * Whether the loader should be displayed or not. It needs to be used instead of `v-if/v-show` for `freeze` to work.
+       */
+      show: {
+        type: Boolean,
+        default: true,
+      },
+      /**
        * For determinate loaders, value between 0 and 100 representing percentage completion
        */
       progress: {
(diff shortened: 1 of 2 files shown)


lib/loaders/KCircularLoader.vue
```diff
@@ -105,6 +113,13 @@
         type: Boolean,
         default: false,
       },
+      /**
+       * Do not hide the loader until `freeze` in milliseconds. Useful to avoid jarring UX when the actions finishes very fast. In comparison to `delay`, `freeze` emhpasises that an action associated with the loader is indeed taking place. `show` needs to be used instead of `v-if/v-show` for this to work.
```
marcellamaki, line 117: this comment explaining the difference from delay 

> emhpasises that an action associated with the loader is indeed taking place

is excellent!

Very small spelling nit-pick: emhpasises should be emphasizes

marcellamaki (changes requested):
Overall, the code changes here look fine, and thank you so much for implementing this based on my comment :) I appreciate that you built this in even though it really was not strictly necessary for the Kolibri PR you were working on. 

As for calling it `freeze` ... with the explanation, it makes sense to me, but if I were to just come across the prop and hadn't known the background, I don't know that I would have fully understood what the prop is trying to convey. I think something a bit more descriptive would be helpful, even if it is longer to read. Perhaps `minDisplayDuration` or `minVisibleTime` or something along these lines? This and the small spelling correction I think are the only two improvements. The rest looks great to me

MisRob:
Thank you, @marcellamaki, I think it's a valuable idea that could be extended to other UI elements.

`minVisibleTime` communicates it well, thank you, renamed. Also fixed the typo.

PR #443, opened 2023-07-17
Update KDateRange to use input[type=date]

### LINKED ISSUE #442: Translate KDateRange date input placeholder text
## Product
Kolibri

## Expected behavior
`KDateRange` component start and end date input placeholder text should be translated to match Kolibri/the user's system language.

## Actual behavior
The `KDateRange` component input placeholder text is set as `DD/MM/YYYY` no matter what language Kolibri is set to.

1. In Kolibri, navigate to Facility > Data page
2. Make sure there are no previously generated session or summary logs
3. Change language in Kolibri to something other than English.
4. Every word is translated except the start and end date input placeholder text.

## Additional information
Language: Italian
Expected Start/End date placeholder text: gg/mm/aaaa

Screenshot 2023-07-17 at 12 45 17 PM

LianaHarris360:
## Description

This pull request updates the component KDateRage to use input[type=date] to allow for the use of browser placeholders for language translation in Kolibri.

#### Issue addressed
Closes #442 

### Before/after screenshots
Before; Italian
image
After:
Italian
italian

Arabic
arabic

English (UK)
english-uk

## Steps to test
See this [Kolibri pull request](https://github.com/learningequality/kolibri/pull/10987) for testing.
1. In Kolibri, navigate to Facility > Data page
2. Make sure there are no previously generated session or summary logs
3. Change language in Kolibri to something other than English
4. *May have to change language in system settings as well*

- Removed validation constant `DATE_FMT` (DD/MM/YYYY)
- Set validation machine start and end date values to null instead of `DATE_FMT`
- Updated validation machine error handling
- Changed input type to date inside `KDateInput` component 
- Added event listener on mount to disable native date pickers once input is clicked or has focus
- Replaced instances of `DATE_FMT` with null
- Updated `KDateRange` tests to work with input date format

This update uses an `input` of type `date` which automatically sets the placeholder text to the user's system language date format.

## Testing checklist

## Reviewer guidance

- [ ] Is this change functional across multiple browsers?

## Comments

With Safari browser, the placeholder is set to the current date rather than DD/MM/YYYY, but this date is not automatically selected on the calendar.

### DIFF
--- a/lib/KDateRange/KDateInput.vue
+++ b/lib/KDateRange/KDateInput.vue
@@ -4,7 +4,6 @@
     <fieldset :aria-label="legendText" class="date-input-fieldset" :aria-describedby="inputId" aria-live="polite">
       <KTextBox
         :ref="inputRef"
-        :class="langDir"
         :value="value"
         type="date"
         pattern="\d{4}-\d{2}-\d{2}" 
@@ -81,12 +80,6 @@
         }
         return '';
       },
-      langDir() {
-        if (this.isRtl) {
-          return 'is-rtl';
-        }
-        return 'not-rtl';
-      },
     },
     mounted() {
       this.$nextTick(() => {
@@ -133,23 +126,21 @@
 
   /* HIDES BROWSER NATIVE DATEPICKER */
   /deep/ input[type='date'] {
-    width: 150px;
-    text-transform: uppercase !important;
     border: 0;
+    text-align: left;
+    text-transform: uppercase !important;
     &::-webkit-inner-spin-button,
     &::-webkit-calendar-picker-indicator {
       display: none;
       appearance: none;
       visibility: hidden !important;
     }
+    &:dir(rtl) {
+      clip-path: inset(0 0 0 25px);
+      text-align: left;
+    }
+    &:dir(ltr) {
+      clip-path: inset(0 25px 0 0);
+    }
   }
-
-  /deep/ div.is-rtl input[type='date'] {
-    clip-path: inset(0 0 0 25px);
-  }
-
-  /deep/ div.not-rtl input[type='date'] {
-    clip-path: inset(0 25px 0 0);
-  }
-
 </style>
--- a/lib/KDateRange/KDateInput.vue
+++ b/lib/KDateRange/KDateInput.vue
@@ -4,6 +4,7 @@
     <fieldset :aria-label="legendText" class="date-input-fieldset" :aria-describedby="inputId" aria-live="polite">
       <KTextBox
         :ref="inputRef"
+        :class="langDir"
         :value="value"
         type="date"
         :label="legendText"
@@ -79,6 +80,12 @@
         }
         return '';
       },
+      langDir() {
+        if (this.isRtl) {
+          return 'is-rtl';
+        }
+        return 'not-rtl';
+      },
     },
     mounted() {
       this.$nextTick(() => {
@@ -124,18 +131,24 @@
   }
 
   /* HIDES BROWSER NATIVE DATEPICKER */
-  /deep/ input[type='date']::-webkit-inner-spin-button,
-  /deep/ input[type='date']::-webkit-calendar-picker-indicator {
-    display: none;
-    appearance: none;
-    visibility: hidden !important;
-  }
-
   /deep/ input[type='date'] {
     width: 150px;
-    clip-path: inset(0 25px 0 0);
     text-transform: uppercase !important;
     border: 0;
+    &::-webkit-inner-spin-button,
+    &::-webkit-calendar-picker-indicator {
+      display: none;
+      appearance: none;
+      visibility: hidden !important;
+    }
+  }
+
+  /deep/ div.is-rtl input[type='date'] {
+    clip-path: inset(0 0 0 25px);
+  }
+
+  /deep/ div.not-rtl input[type='date'] {
+    clip-path: inset(0 25px 0 0);
   }
 
 </style>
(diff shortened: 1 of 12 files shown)


rtibbles:
I don't think this is a blocker, but because of the way that Android date pickers are implemented (using the OS level date picker widget), the date input doesn't fill on Android. I think this is fine, but it does mean that manually entering a date doesn't work - but I think that's also the case for the Android standard date picker, because you use the OS level picker, so it doesn't seem any different.

image

Definitely not a blocker!

lib/KDateRange/KDateInput.vue
```diff
@@ -5,17 +5,17 @@
...
         :value="value"
-        type="text"
+        type="date"
         :label="legendText"
         autoComplete="off"
         :invalid="Boolean(errorMessage)"
         :showInvalidText="Boolean(errorMessage)"
         :invalidText="errorMessage"
+        :floatingLabel="false"
         @input="handleInput"
```
rtibbles, line 15: Hrm, it would be preferable, rather than manually assigning event listeners here, to do the prevent default using vue event binding:
```
@click.prevent="() => {}"
@focus.prevent="() => {}"
```

See here for details: https://v2.vuejs.org/v2/guide/events.html#Event-Modifiers

But, as it seems that KTextBox is wrapping UiTextBox and neither is passing through `$listeners` to the `input` tag, then this will not work. So, nothing we can do here without a bigger refactor which is not desirable right now!

rtibbles (commented):
Changes make sense - and I think using the built in date input has made this a bit cleaner in terms of the data handling, which is nice.

One comment about some necessary tech debt this is introducing - but that we can't resolve until we unpick some of our keen-ui vendoring.

LianaHarris360:
@pcenov Thanks for pointing this out. Support for RTL languages within the input fields has been added, the selected dates should be fully displayed now. As for the date format not being displayed in Internet Explorer 11 or Chrome on Android, Richard is correct about this behavior being set on Android by default, which certainly isn't ideal.

rtibbles:
Yeah, I think this is not implemented in Chrome for Android as it always defers to the UI date pickers, so it's not actually possible to manually input the date with the keyboard. I think for that reason, this is acceptable, because it's parallel to the default implementation.

IE11 doesn't actually support the date input type, I see: https://caniuse.com/?search=input%20date so this might require a little bit more finesse. I think the only mediation we might be able to provide here is to use the fallback text input 'pattern' attribute to at least validate the value as described here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date#handling_browser_support

That said, as we are deprecating IE11 with the upcoming Kolibri release, I think we can entertain a somewhat degraded experience if needed - may be good to just validate that the `input` properly parses the date in IE11 and make sure that there aren't bugs at least.

MisRob:
I'm just confirming that we're waiting for a new round of QA, or do you need anything else @LianaHarris360?

pcenov:
Hi @LianaHarris360, I can see that the build artefacts in https://github.com/learningequality/kolibri/pull/10987 are still the ones I tested the last time. Could you take a look and update those so that we can test the latest version? Thanks!

LianaHarris360:
Hi @pcenov I've re-ran the job for the Kolibri build artifacts and it should be updated now, my apologies!

MisRob:
@LianaHarris360 @pcenov If it still doesn't work after re-running the build, it might be due to the yarn cache. Including commit hash here https://github.com/learningequality/kolibri/pull/10987/files#diff-38850ae5a41a5ee8591a1bf8815381974e716d122e241e3337e4755bbf016c70R24 and also committing updated `yarn.lock` (after running `yarn install`) is usually how we do it to avoid problems.

pcenov:
Hi @LianaHarris360 retested with the latest build artefacts from https://github.com/learningequality/kolibri/pull/10987 and unfortunately the truncation both of the date format and the selected date is still present in Chrome on both Ubuntu and Windows when in RTL mode (regardless if the operating system locale is in RTL language or LTR):

ChromeAR

It's working correctly in Firefox though:

FirefoxAR

MisRob:
@LianaHarris360 I can see you updated `yarn.lock`. From our previous experience, that might not be sufficient and is most likely the reason why @pcenov can't see updates.

I'd recommend including the commit hash in `package.json` so instead of `kolibri-design-system": "https://github.com/LianaHarris360/kolibri-design-system#kdaterange-date-input` it will be `kolibri-design-system": "https://github.com/LianaHarris360/kolibri-design-system#107036b42d6e507c6ad29ca515414b1728fba860`. Then you can again run `yarn install` and commit `yarn.lock` together with updated `package.json`. Also please note there are two `package.json` with KDS dependency in Kolibri. I don't know if it will have an impact but would recommend updating both to be sure.

You can see https://www.notion.so/learningequality/KDS-release-process-361de41012a742f394c636d0aecf08b4 for detailed guidance and example PRs.

LianaHarris360:
Thank you @MisRob for the detailed help! I updated both files that have the KDS dependency to include the correct commit hash @pcenov the clipped date inputs should display correctly within Chrome for Windows and Ubantu now (whether the operating system locale is in a RTL language or LTR), there was an oversight on my part, but it is corrected now, thanks for catching that!

pcenov (approved):
Hi @LianaHarris360, I confirm that the date input is no longer being truncated so I am approving this PR.
While regression testing I noticed the following issue in IE11:
 - if it's not caused by any of the code changes made here I'll report it separately.

LianaHarris360:
Hi @pcenov The date input had a fixed width that was previously removed because it seemed like it had no effect, but I've added it back. Can you see if the extra long input issue persists?

pcenov:
Hi @LianaHarris360, yes I confirm that the issue is fixed now and the date fields are displayed correctly in IE11.

Issue #395, opened 2022-12-08
Audit icons used in KIcons to see if any need to be updated so their color can be overridden


marcellamaki:
## Product
Kolibri

There have been ongoing occasional instances where KIcons are not exactly the color they are expected. Either the `:color` prop doesn't work, because the icon is not exported in a format that is overrideable, or the KIcon is being set with a theme token when perhaps it shouldn't.

We may need to adjust KIcon itself to accept default or override colors, adjust how we are using theme tokens, or change the actual icons themselves so that the svg is in a different format, or some combination. 

The purpose of this issue is to further investigate and plan, and come up with a proposed solution (which will be a separate issue). 

(cc @jtamiace @radinamatic)

laxmikanta415:
@marcellamaki the default value of the color if no prop is passed is set to null that may be causing the unexpected behaviour. 

 Can you add screenshot of what the problem scenario exactly is as i am unable to reproduce if a value is set for color prop ?

PR #446, opened 2023-08-22
Remove hardcoded fill color within icon svg files

LianaHarris360:
## Description

Co-authored with @nucleogenesis 
This pr addresses the issue where some `KIcon` colors not being overridden. Instances where the icon SVG files contained `fill="black"` were removed so that the `:color` property could work.
 
#### Issue addressed

Addresses #395 

### Before/after screenshots
For example, with the color property set to pink:
```
      <KIcon
        v-if="aliasList && aliasList.length"
        :icon="aliasList[0]"
        color="pink"
        class="icon"
      />
```
The icon color is overridden as expected:
Screenshot 2023-08-22 at 1 04 50 PM

Removed hardcoded fill color within some SVG files

## Testing checklist

## Reviewer guidance

## Comments
Selecting the Netlify preview link and navigating to the Icon page will show the display of icons with the color overridden.


LianaHarris360:
@tomiwaoLE @jtamiace @marcellamaki 
Are there some cases where the color should not be overridden? Such as with the a11y icon and the Learning activities icons:
Screenshot 2023-08-22 at 1 20 56 PM

Screenshot 2023-08-22 at 1 04 35 PM

MisRob:
> Are there some cases where the color should not be overridden?

If this shows to be indeed the case, I'd suggest mentioning it in the documentation at relevant places, one of them being `KIcon`'s `color` prop documentation, so that developers know it's not intended work in some cases

LianaHarris360:
`Note:` This PR is currently a draft only to demonstrate that setting the`:color` prop to pink correctly displays the icons to be pink. Decisions regarding the icons that do not change color with the `:color` property need to be made to determine if that is the intended presentation. If all color changes are confirmed to be correct, the line `:color="pink"` will be removed from the `KIcon` documentation page, otherwise appropriate changes will be made. Other than that, this PR is ready for review.

indirectlylit:
Hi folks! I'm still lurking around in KDS :)

>  icons that do not change color

One option would be to add a new boolean property like `fixedColor: true` to these icon definitions:

https://github.com/learningequality/kolibri-design-system/blob/8e5916b9f56324f18814c3510e141d1772114797/lib/KIcon/iconDefinitions.js#L13-L17

When `fixedColor: true`, trying to use the `color` prop of `KIcon` would show an error, helping prevent inadvertent color overrides.

Just an idea!

LianaHarris360:
Hi @indirectlylit, thank you for lurking around in KDS,  I like this suggestion! Displaying an error in the console/logs when the color prop is used on icons where the `fixedColor` is `true` would be a great way to inform developers of icons with unchanging colors in addition to referencing it in the documentation.

LianaHarris360:
This PR has been updated to reflect agreed upon changes such as:

- `fixedColor` property that determines if the icon color is customizable or not has been added to `lib/KIcon/iconDefinitions.js`
- Icons where the property `fixedColor` is `true` will provide an error if the `KIcon` color property is set. 
    - For example with the icon `a11y`:
Screenshot 2023-09-05 at 1 24 34 PM


`Note:` A follow-up issue will be created for icons that use multiple colors but should still be customizable.

marcellamaki (approved):
This looks great @LianaHarris360 and @nucleogenesis . As we discussed during KDS, there will probably be some follow up work around our icons after this, but this is an awesome start and a really helpful fix. Thank you 🎉

PR #451, opened 2023-09-25
New changelog, adds changelog check GH action

MisRob:
## Description

TBD

#### Issue addressed

Addresses #*PR# HERE*

## Steps to test

1. Step 1
2. Step 2
3. ...

## Testing checklist

## Reviewer guidance


nucleogenesis (approved):
I absolutely love this thank you so much for this Misha!

MisRob:
@rtibbles I'm merging it as I got some approvals already and would like to rebase some PRs on top of that after creating a new release branch. I'd still appreciate if you could have a look at the action. Can update as follow-up. Thank you.

rtibbles:
The action seems fine - I am always a little wary of third party actions, but you have version pinned it, some it should be fine.

I do think it would be helpful to update the PR template to prompt for the information requested in the changelog, so that there's a place to encode these things when firsts opening a PR (can't actually update the changelog when first opening a PR unless you correctly predict your PR number).

MisRob:
@rtibbles Thank you.

> I am always a little wary of third party actions, but you have version pinned it, some it should be fine

Is your concern about security or rather about an action unexpectedly breaking after a 3rd party update? 

> I do think it would be helpful to update the PR template to prompt for the information requested in the changelog, so that there's a place to encode these things when first opening a PR (can't actually update the changelog when first opening a PR unless you correctly predict your PR number).

Overall this approach sounds helpful to me, I think it's something similar that @jredrejo suggested too.

Specifically, do you mean having the guidelines https://github.com/learningequality/kolibri-design-system/pull/451/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR1-R38, or at least this part https://github.com/learningequality/kolibri-design-system/pull/451/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR23-R35 in the PR template so that an author would fill it in when opening a PR and then anybody could just paste it into the changelog?

rtibbles:
> Is your concern about security or rather about an action unexpectedly breaking after a 3rd party update?

Both, but more the former, as the pinning gives assurance about the latter. 

Yes, just having it in the PR template would mean it would only need to be copy pasted into the changelog (or in theory we could automate updating the changelog once a PR is merged to avoid the inevitable merge conflicts that duelling changes will produce for PRs). 

The other thing that occurred to me is that we should probably not run this check when merging from release branch to release branch or when merging into develop - but that's probably a pain to encode, so just being aware that we should ignore this check in cascade merges is probably sufficient.

MisRob:
No problem to build on top of that action over time or we could even remove the 3rd party dependency if that'd be better. If you see value in it, I can open an issue, @rtibbles (just tried to come with something now because building a reliable changelog in retrospect was time demanding so I concluded it'd be best to enforce this for all future updates asap)

Issue #237, opened 2021-05-28
Document KResponsiveElementMixin performance


MisRob:
Due to `ResizeSensor`, `KResponsiveElementMixin` causes page reflows which can cause significant performance issues, especially when many instances are present on a page, as can be seen in https://github.com/learningequality/kolibri/issues/7525 and https://github.com/learningequality/kolibri/issues/8124.

It would be nice to add performance warnings and recommended usage for `KResponsiveElementMixin` to its documentation. Discussion and outcomes of Kolibri issues referenced above might be useful for getting more information.

MisRob:
Closed by https://github.com/learningequality/kolibri-design-system/pull/362

PR #453, opened 2023-09-27
useKResponsiveWindow update fixing resizing behavior in Learn > Library

thanksameeelian:
## Description
this PR fixes a bug observed during QA review of Kolibri in `Learn` > `Library` - in certain cases, rapidly resizing the browser window would cause the sidepanel to automatically display when the screen was small, behavior we did not expect nor want at that screen size. beneath the surface, the components were receiving conflicting information about the current screen size because `useKResponsiveWindow` was not updating in the way we expected. 

to address those issues, this PR introduces changes to the scope of variables within `useKResponsiveWindow` and to the way the composable uses `useKWindowDimensions`. 

#### Issue addressed
[reported as a Kolibri issue ](https://github.com/learningequality/kolibri/issues/11212)

### Before/after screenshots
BEFORE, courtesy of @pcenov:

https://github.com/learningequality/kolibri/assets/79847249/0b63d243-e140-4830-b9b9-0a915bddfd61

AFTER (just a boring video of me resizing successfully):

https://github.com/learningequality/kolibri-design-system/assets/43046460/b67220e5-2fcb-45b1-94bf-8cbe29866df3

## Steps to test

Please see reviewer guidance provided in https://github.com/learningequality/kolibri/pull/11315 & test behavior of this change on that kolibri branch

this doesn't necessarily introduce tech debt, but has revealed some existing underlying issues regarding the functioning of `KResponsiveWindowMixin` (which is still in use in many places), as it is likely attaching an excessive amount of listeners which could have a performance impact. there is an issue forthcoming for retrofitting the mixin to use the composable updated here to reduce the amount of unnecessary and sometimes detrimental extra listeners.

## Testing checklist

## Reviewer guidance

- [ ] Are there tests?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-<!-- 
+<!--
 
 Changelog Guidelines
 
@@ -42,6 +42,15 @@ Changelog Guidelines
 Changelog is rather internal in nature. See release notes for the public overview and guidelines. Releases are recorded as git tags in the [Github releases](https://github.com/learningequality/kolibri-design-system/releases) page.
 
 ## Version 1.5.x
+
+- [#453]
+  - **Description:** Fix sidepanel opening in Kolibri Library page after resizing window
+  - **Products impact:** bugfix
+  - **Addresses:** https://github.com/learningequality/kolibri/issues/11212
+  - **Components:** `useKResponsiveWindow` (composable)
+  - **Breaking:** no
+  - **Impacts a11y:** no
+  - **Guidance:** -
 
 <!-- Release notes prepared for all items below -->
 
@@ -214,7 +223,7 @@ Changelog is rather internal in nature. See release notes for the public overvie
 [#426]: https://github.com/learningequality/kolibri-design-system/pull/426
 
 - [#426]
-  - **Description:** Fix `KTabsList` focus state 
+  - **Description:** Fix `KTabsList` focus state
   - **Products impact:** bugfix
   - **Addresses:** -
   - **Components:** `KTabsList`, `KTabs`
@@ -338,7 +347,7 @@ Changelog is rather internal in nature. See release notes for the public overvie
   - **Description:** Fix `KDropdownMenu` not showing after its refactor in [#346] by adding missing template tags to `KButton`
   - **Products impact:** bugfix
   - **Addresses:** https://github.com/learningequality/kolibri/issues/9754
-  - **Components:** `KDropdownMenu`, `KButton` 
+  - **Components:** `KDropdownMenu`, `KButton`
   - **Breaking:** no
   - **Impacts a11y:** no
   - **Guidance:** -
--- a/lib/useKResponsiveWindow/index.js
+++ b/lib/useKResponsiveWindow/index.js
@@ -1,132 +1,139 @@
-import { computed, onBeforeUnmount, onMounted, ref } from '@vue/composition-api';
+import { computed, onBeforeUnmount, onMounted, ref, watch } from '@vue/composition-api';
 import useKWindowDimensions from '../useKWindowDimensions';
 import MediaQuery from './MediaQuery';
+
+/** Global variables */
+const { windowWidth, windowHeight } = useKWindowDimensions();
+
+/*
+  Implementing breakpoint controlled by watchers to work around
+  optimization issue: https://github.com/vuejs/vue/issues/10344
+  If that issue ever gets addressed, we should make them computed props.
+*/
+const windowBreakpoint = ref(null);
+const windowGutter = ref(16);
+
+const windowIsPortrait = ref(false);
+const windowIsLandscape = ref(false);
+const windowIsShort = ref(false);
+
+const windowIsLarge = computed(() => windowBreakpoint.value > 2);
+const windowIsMedium = computed(() => windowBreakpoint.value === 2);
+const windowIsSmall = computed(() => windowBreakpoint.value < 2);
+
+const usageCount = ref(0);
+
+const orientationQuery = new MediaQuery('screen and (orientation: portrait)', event => {
+  updateOrientation(event.matches);
+});
+const heightQuery = new MediaQuery('screen and (max-height: 600px)', event => {
+  windowIsShort.value = event.matches;
+});
+
+/**
+ * Initialize media query window properties
+ */
+function initProps() {
+  orientationQuery.eventHandler(orientationQuery.mediaQueryList);
+  heightQuery.eventHandler(heightQuery.mediaQueryList);
+};
+
+initProps();
+
+/**
+ * Start listening to media query changes
+ */
+function startListening() {
+  orientationQuery.startListening();
+  heightQuery.startListening();
+};
+
+/**
+ * Stop listening to media query changes
+ */
+function stopListening() {
+  orientationQuery.stopListening();
+  heightQuery.stopListening();
+};
+
+/**
+ * Update window breakpoint
+ */
+function updateBreakPoint() {
+  const SCROLL_BAR = 16;
+  const widthBreakpoints = [
+    480, 600, 840, 960 - SCROLL_BAR, 1280 - SCROLL_BAR, 1440 - SCROLL_BAR, 1600 - SCROLL_BAR,
+  ];
+
+  // if windowWidth is null (upon initialization), breakpoint = 0
+  const firstMatchingWidthIndex = windowWidth.value
+    ? widthBreakpoints.findIndex(matchingWidth => windowWidth.value < matchingWidth)
+    : 0;
+
+  // if windowWidth > 1585, breakpoint = 7
+  const matchingBreakpoint = firstMatchingWidthIndex >= 0
+    ? firstMatchingWidthIndex
+    : 7;
+
+  windowBreakpoint.value = matchingBreakpoint;
+  updateGutter();
+};
+
+watch([windowWidth, windowHeight], updateBreakPoint);
+
+/**
+ * Update window gutter
+ */
+function updateGutter() {
+  if (windowIsSmall.value || smallestWindowDimensionIsLessThan(600)) {
+    windowGutter.value = 16;
+  } else {
+    windowGutter.value = 24;
+  }
+};
+
+/**
+ * Check if the smallest window dimension(width or height) is smaller than the specified dimension
+ * @param {Number} dimension in px
+ * @returns {Boolean}
+ */
+function smallestWindowDimensionIsLessThan(dimension) {
+  return (
+    windowBreakpoint.value < 4 && Math.min(windowWidth.value, windowHeight.value) < dimension
+  );
+};
+
+/**
+ * Update window orientation
+ * @param {Boolean} portrait
+ */
+function updateOrientation(portrait) {
+  windowIsPortrait.value = portrait;
+  windowIsLandscape.value = !windowIsPortrait.value;
+};
 
 /**
  * Export window properties
  * @returns {Object} Object with window properties
  */
 export default function useKResponsiveWindow() {
-  /** Window properties */
-  const { windowWidth, windowHeight } = useKWindowDimensions();
-  const windowBreakpoint = ref(null);
-  const windowIsPortrait = ref(false);
-  const windowIsLandscape = ref(false);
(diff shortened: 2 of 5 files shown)


lib/useKResponsiveWindow/index.js
```diff
@@ -2,125 +2,119 @@ import { computed, onBeforeUnmount, onMounted, ref } from '@vue/composition-api'
...
+  windowIsLandscape.value = !windowIsPortrait.value;
+};
+
+
+/**
+ * Export window properties
+ * @returns {Object} Object with window properties
+ */
+export default function useKResponsiveWindow() {
   onMounted(() => {
```
rtibbles, line 118: The simplest way to achieve this, with everything defined in the global scope would be to just call `initProps()` and then `startListening()` in the module scope, so that as soon as this is parsed, we start listening to things.

The other way of doing this, would be to follow the useKWindowDimensions route, and keep track of usage: https://github.com/learningequality/kolibri-design-system/blob/develop/lib/useKWindowDimensions.js#L74

So you would only call initProps and startListening if it wasn't already in use, and only call stopListening if the usage count has dropped to 0.

lib/useKResponsiveWindow/index.js
```diff
@@ -1,133 +1,152 @@
...
+    ? widthBreakpoints.findIndex(matchingWidth => windowWidth.value < matchingWidth)
+    : 0;
+
+  // want windowWidth > 1600 ===> result 7
+  const matchingBreakpoint = firstMatchingWidthIndex >= 0
+    ? firstMatchingWidthIndex
+    : 7;
+
+  windowBreakpoint.value = matchingBreakpoint;
+  updateGutter(); // incl here? or in watch?
```
rtibbles, line 55: I think this is the right place for it!

lib/useKResponsiveWindow/index.js
```diff
@@ -1,133 +1,152 @@
...
+const windowIsShort = ref(false);
+
+// should these be .value or just plain const? (if windowBreakpoint is no longer a ref)
+const windowIsLarge = computed(() => windowBreakpoint.value > 2);
+const windowIsMedium = computed(() => windowBreakpoint.value === 2);
+const windowIsSmall = computed(() => windowBreakpoint.value < 2);
+
+const orientationQuery = new MediaQuery('screen and (orientation: portrait)', event => {
+  updateOrientation(event.matches);
+  updateGutter();
```
rtibbles, line 31: I think you can actually remove this from here, as this is only dependent on the window dimensions, right, not the orientation?

lib/useKResponsiveWindow/index.js
```diff
@@ -1,132 +1,134 @@
...
+function updateOrientation(portrait) {
+  windowIsPortrait.value = portrait;
+  windowIsLandscape.value = !windowIsPortrait.value;
+};
+
+/**
+ * Export window properties
+ * @returns {Object} Object with window properties
+ */
+export default function useKResponsiveWindow() {
```
rtibbles, line 117: So, the other thing that is happening here, is that the onMounted and onBeforeUnMount hooks for useKWindowDimension are never being called, because we've moved the invocation into the module scope.

So, to make sure they do get initialized, we can just make a dummy call to these in the function scope as well:

```
// Do this to ensure that we call the useKWindowDimension lifecycle hooks
useKWindowDimension();
```

MisRob:
@thanksameeelian I just created the release branch and retargeted this PR into it since KDS 1.5. is going to Kolibri's 0.16. Could you please rebase on top the lates `release-v1.5.x` branch to get the latest changelog updates into this PR? Thank you.

CHANGELOG.md
```diff
@@ -45,6 +45,16 @@ Changelog is rather internal in nature. See release notes for the public overvie
 
 <!-- Release notes prepared for all items below -->
 
+- [#453]
```
MisRob, line 48: Just a very minor note, could you please move it above `<!-- Release notes prepared for all items below -->` line? That's just so that we know at what point release notes ends. This should be temporary way of doing so, as soon as we have the first release ready.

MisRob: In any case, nice first item :) Looking great

thanksameeelian: ah great point about the placement, @MisRob, thank you!

pcenov:
Hi @thanksameeelian I confirm that the issue described in https://github.com/learningequality/kolibri/issues/11212 is fixed now.

nucleogenesis (approved):
Code looks great - love the full test coverage of the changes here. I tested the associated Kolibri PR as well and tried out several areas of the app resizing windows without issue.

MisRob (changes requested):
Thanks for all the debugging and reworking the composable @thanksameeelian, this will be an impactful update for all the products. The PR as a whole looks like a high quality work.

I can see one important performance regression. Previously, we calculated breakpoints and all derived info, such as`windowIsSmall`etc., by using media queries but now we rely heavily on `useKWindowDimensions` instead. I understand that the current implementation using local scopes is problematic and agree that moving them to global scope and making sure we don’t initialize same listeners many times is a good, robust solution.

However, I’m concerned about turning away from media queries as radically as removing them completely. Building everything around `useKWindowDimensions` which reads `window.innerWidth/Height` on every `resize` event has some [performance issues that `matchMedia` solves](https://webdevetc.com/blog/matchmedia-events-for-window-resizes/). Relatedly, [reading `innerWidth/Height` causes layout/reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a#getting-window-dimensions) which can be a significant performance bottleneck.

In Kolibri, we currently rarely reach to `window.innerWidth/Height` as in the majority of use cases, we only access rough breakpoints that don’t use that logic. For illustration, see [`windowWidth` usages](https://github.com/search?q=repo%3Alearningequality%2Fkolibri%20windowWidth&type=code) (reads `innerWidth`, used 3x)  vs [`windowIsSmall` usages](https://github.com/search?q=repo%3Alearningequality%2Fkolibri%20windowIsSmall&type=code) (doesn’t read `innerWidth`, used 58x and there's much more similar props). Building calculations for rough breakpoints around `useKWindowDimensions`would significantly change this (after we use it more widely instead of the mixin, strictly speaking). For that reason, I suggest that we don't use it whenever possible.

In this PR, I imagine we could keep the best of both worlds - using global props and making sure we don’t initialize listeners often than needed like you already did while still updating breakpoints by using media queries.

If this was something you already tried out and experienced some technical blockers I missed, let me know any time and we can have a look at it.

MisRob:
Also opened related issue https://github.com/learningequality/kolibri-design-system/issues/461. This won't solve the above but should help in situations we absolutely need to read `window.innerWidth/Height`.

MisRob:
For reference, you may also briefly look at
- https://develop--kolibri-design-system.netlify.app/kresponsiveelement/#guidelines
- https://github.com/learningequality/kolibri/issues/8124
- https://github.com/learningequality/kolibri/issues/7525

where it's illustrated that we already experience pretty serious performance issues. It's not the same situation, but similar issues were one contributing factor.

MisRob:
For reference, you may also briefly look at

- https://github.com/learningequality/kolibri/issues/7525
- https://develop--kolibri-design-system.netlify.app/kresponsiveelement/#guidelines
- https://github.com/learningequality/kolibri/issues/8124
- https://github.com/learningequality/kolibri-design-system/issues/237

where it's illustrated that we already experienced pretty serious performance problems. It's not the same situation, but similar issues were one of the contributing factors.

rtibbles:
> In this PR, I imagine we could keep the best of both worlds - using global props and making sure we don’t initialize listeners often than needed like you already did while still updating breakpoints by using media queries.

> If this was something you already tried out and experienced some technical blockers I missed, let me know any time and we can have a look at it.


The use of mediaQueries was the cause of the main bug here, the move to globally registered listeners was to fix this while not causing a huge performance regression (by having a global variable to track, rather than multiple listeners for window resize). 

Because the Media Query API requires a specific query, we were obliged to have a listener registered to the media query for _every_ breakpoint - this meant that listeners were being triggered during a resize event for each media query as it changed, and then frequently updating the breakpoint value out of order. Because the media query listeners are triggered independently of each other, the only safe way to use them is when there is a 1:1 correlation between the media query and the value we are tracking (so using it to detect potrait mode is safe, because we are only using 1 media query listener to track it).

I don't see that @thanksameeelian's changes will cause a performance regression here, as we were _already_ attaching the global window resize event listener to update the windowHeight and windowWidth by using the `useKWindowDimensions` composable from within `useKResponsiveWindow` - so any performance degradation caused by that resize listener is pre-existing.

MisRob:
> we were already attaching the global window resize event listener to update the windowHeight and windowWidth by using the useKWindowDimensions composable from within useKResponsiveWindow

I assumed we'd move away from attaching it completely eventually

> Because the Media Query API requires a specific query, we were obliged to have a listener registered to the media query for every breakpoint - this meant that listeners were being triggered during a resize event for each media query as it changed, and then frequently updating the breakpoint value out of order. Because the media query listeners are triggered independently of each other, the only safe way to use them is when there is a 1:1 correlation between the media query and the value we are tracking (so using it to detect potrait mode is safe, because we are only using 1 media query listener to track it).

I had to misunderstood the motivation for this bugfix then, it seemed to me that the main issue were local props + I didn't realize that independent working of more media queries is the main culprit. Consider my comment irrelevant, apologies. Thanks for clarification.

MisRob:
@rtibbles @thanksameeelian For my better understanding of the issue and to know whether there's some space to do some improvements in the future or no, can I understand right that you see no way of using media queries at all for similar use cases? I wonder if reformulating the queries
 
from

```
 const queries = [
    '(max-width: 480px)',
    '(max-width: 600px)',
    '(max-width: 840px)',
    ...
  ];
```

to


```
 const queries = [
    '(max-width: 480px)',
    '(min-width: 481px) and (max-width: 600px)',
    '(min-width: 600px) and (max-width: 840px)',
    ...
  ];
```

would help to resolve some of the problems, particularly the 1:1 correlation. Or was it about something else?

That said, I don't know though how the number of around 8 listeners would compare to reading `innerWidth/Height` frequently performance wise, that might need some research. For now, I'd only like to understand better media queries issues to see if there are some ways still open towards planning some optimizations in this area eventually, taking into account wide usage.

rtibbles:
It's possible that would ameliorate it, but the issue is more down to the fact that multiple callbacks can trigger a value update for the breakpoint - whereas a single callback that comes from watching both the width and height prevents this.

MisRob:
> the issue is more down to the fact that multiple callbacks can trigger a value update for the breakpoint

This means that more media queries, in general, should never update the same value, is that right? Here we'd need to rather have more variables instead of one, e.g. `isBreakpoint0` that would map to the first query, `isBreakpoint1` that would map to the second, etc. for them to work as expected? That's not a suggestion at this point, just an example to confirm I get their limitations as I wouldn't expect this before diving deeper here.

rtibbles:
Yes, I think that would work - the only concern would be having multiple sources of truth - but if the media queries are working appropriately, they should all agree.

I do think the performance concerns will be mostly covered by #461 though, rather than using media queries.

MisRob:
@rtibbles My curiosity is satisfied :) Thanks for going into detail.

MisRob:
> I do think the performance concerns will be mostly covered by https://github.com/learningequality/kolibri-design-system/issues/461 though, rather than using media queries.

I hope so, yes, especially if we make sure we keep using it globally.

rtibbles:
One other possible follow up would be to retrofit the mixin to use the now globally scoped composable, which would remove all current performance concerns as well.

MisRob:
> One other possible follow up would be to retrofit the mixin to use the now globally scoped composable, which would remove all current performance concerns as well.

That'd be helpful for the current state of things for sure.  I don't know whether that's high priority at this point, my concern was originally rather about not removing something that was meant to improve situation, however I didn't get that it was a deliberate step rather than just a side effect of refactoring of variables scope.

We also have https://github.com/learningequality/kolibri/issues/11321 and https://github.com/learningequality/kolibri-design-system/issues/459 which will help us to move away from the mixin completely eventually. I imagine that refactor wouldn't be complicated, perhaps it's rather QA that would take time due to this logic being used frequently.

PR #456, opened 2023-09-28
Release v1.4.x in Release v1.5.x

rtibbles:
It seems this was never merged up.


MisRob (approved):
Thanks @rtibbles. I am skipping the changelog check here as we're starting the new changelog since 1.15.

Issue #373, opened 2022-09-22
KImg: Kolibri Fly!


MisRob:
## Summary

Truthy `kolibriFly` prop renders the Kolibri logo into the `KImg`. This overrides the `src` specified

## Epic

- https://github.com/learningequality/kolibri-design-system/issues/368

## Acceptance criteria

- [ ] Setting `kolibriFly` to `true` renders the Kolibri logo into the component
- [ ] Any set `src` is overridden

## Linked issues
  
- **Blocked by**
  - https://github.com/learningequality/kolibri-design-system/issues/369
  - https://github.com/learningequality/kolibri-design-system/issues/370

MisRob:
Some KDS and products source files with the logo
- https://github.com/learningequality/kolibri-design-system/blob/develop/docs/common/DocsPageTemplate/SideNav/kolibri-logo.svg
- https://github.com/learningequality/studio/blob/unstable/contentcuration/contentcuration/frontend/shared/views/KolibriLogo.vue
- https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/views/CoreLogo/index.vue

ShivangRawat30:
I would like to work on this issue.

MisRob:
Hi @ShivangRawat30, good choice, this is IMHO the most fun issue in KDS. I'm sure that the team will love it :)

MisRob:
Note that the specification here was changed to creating a new component, `KLogo` instead. https://github.com/learningequality/kolibri-design-system/pull/460#pullrequestreview-1651978076

MisRob:
Closed by https://github.com/learningequality/kolibri-design-system/pull/460

PR #460, opened 2023-10-01
Made the KLogo component

ShivangRawat30:
## Description

This PR Adds `KLogo`

#### Issue addressed

Addresses #373 

## Steps to test

1. Navigate to http://localhost:4000/klogo

## Testing checklist

## Reviewer guidance


lib/KImg.vue
```diff
@@ -14,6 +14,8 @@
 
 <script>
 
+  import kolibriLogo from '../docs/common/DocsPageTemplate/SideNav/kolibri-logo.svg';
```
nucleogenesis, line 17: I'm not sure if we should be linking to an SVG in the docs directory here. Perhaps it'd be worth copy-pasting it directly into the same directory as this file?

lib/KImg.vue
```diff
@@ -115,6 +124,9 @@
       imgMinWidth() {
         return this.validateAndFormatUnits(this.minWidth);
       },
+      logo() {
```
nucleogenesis, line 127: Since the item being shown won't always be a "logo" maybe we can call this something like `imgSrc`

nucleogenesis (commented):
I'll leave approval to @thanksameeelian to be sure here, but I do question whether or not this should be a prop in this component or if we should instead make a separate `KLogo` or something which uses `KImg` without these changes rather than making the otherwise generic `KImg` component configurable for one particular special purpose by a prop.

MisRob:
@nucleogenesis 

> but I do question whether or not this should be a prop in this component or if we should instead make a separate KLogo or something which uses KImg without these changes rather than making the otherwise generic KImg component configurable for one particular special purpose by a prop

It was just a light-hearted issue we made with @akolson to conclude his project in a little silly way utilizing our LE slogan "Kolibri Fly!" :) I admit I didn't put much technical thinking into it. This PR follows the specification exactly https://github.com/learningequality/kolibri-design-system/issues/373, so the fault is on me.

You're right that `KLogo` would make more sense. So, perhaps we could
- If @ShivangRawat30 wants to refactor it, you're welcome, but please know we realize that the problem is the specification and not your work on it
-  If you can't to make these updates @ShivangRawat30, I would still vote for accepting the PR @nucleogenesis if it passes the review and then I would make sure to address this concern myself in a follow-up.

Would that be okay?

lib/KImg.vue
MisRob: Definitely would be good to copy it to `KImg`/`KLogo` directory

MisRob: I doubt that someone re-organizing docs code and pages would get an idea to check whether it broke something in library. So good to have strict separation.

ShivangRawat30:
@MisRob I would appreciate the opportunity to improve it. Should I create a new file named `Klogo.vue` and create a separate tag for the logo image?

MisRob:
Great, @ShivangRawat30, thanks for being open. Yes, it'd be a new component named `KLogo.vue`. You can find some guidance on creating a new component in the dev docs (https://github.com/learningequality/kolibri-design-system/blob/release-v1.5.x/dev_docs/03_how_to_update_library.md#addupdate-a-component). Optionally, you could create a new simple documentation page for it if you'd like to give it a try https://github.com/learningequality/kolibri-design-system/blob/release-v1.5.x/dev_docs/04_how_to_update_docs.md#create-a-documentation-page-for-a-new-component but we can also open a follow-up issue. It's up to you.

@nucleogenesis Could you please provide some guidance to @ShivangRawat30 if he has some more questions?

ShivangRawat30:
@MisRob @nucleogenesis Should I create a new folder specifically for images to house the `kolibrilogo.svg`, or is it more appropriate to retain it within the lib folder?

thanksameeelian:
> @MisRob @nucleogenesis Should I create a new folder specifically for images to house the `kolibrilogo.svg`, or is it more appropriate to retain it within the lib folder?

@ShivangRawat30 you can make your new `KLogo` component a directory that houses an `index.vue` and the `svg` file, like [the file structure used in this example from Kolibri](https://github.com/learningequality/kolibri/tree/release-v0.16.x/kolibri/core/assets/src/views/CoreLogo). this is mostly a question of preference about organization rather than about function, but it would be nice to uphold that same pattern 🙂

ShivangRawat30:
@nucleogenesis Please review the PR with the changes I've implemented.

ShivangRawat30:
Screenshot 2023-10-04 at 12 00 15 AM
screenshot of the implemented change.

MisRob:
Wonderful @ShivangRawat30, and thanks for drafting the documentation page! I will leave the final review of this work on @thanksameeelian or @nucleogenesis. For now, only one high-level request - instead of ``, could you use `<KImg>`? Then, we wouldn't need to have the same logic two times.

You can just pass `KLogo` props down to `KImg`, for example

```
<div>
  <KImg
    :height="height"
    :width="width"
    ...
  />
</div>
```

lib/KLogo/index.vue
```diff
@@ -0,0 +1,157 @@
+<template>
+
+  <div>
+    <img
+      :src="logo"
+      :alt="alternateText"
+      :style="styleObject"
+    >
+    <slot></slot>
```
MisRob, line 9: I think we don't need the slot here

MisRob:
Also, @radinamatic, when displaying our logo https://raw.githubusercontent.com/learningequality/kolibri-design-system/ff15d1a2d8a3ae82c61de82ede74caf8177a1c9d/lib/KLogo/kolibri-logo.svg, does it need to have alternative text or should it be empty?

radinamatic:
@MisRob Yes, alternative text for the logo image would be required. 

If the logo is used just as a flat image, the string `Kolibri logo` should suffice. If, however, the logo image is also a link, than the alternative text should give the context where the link is leading (for example `Go to home page`, or similar).

MisRob:
@radinamatic Thank you.

@ShivangRawat30 based on @radinamatic's comment, I'd then suggest different API for `KLogo` than for `KImg` in regards to alternative text:

- no `isDecorative` prop (as logo will never be decorative)
- `altText` prop will stay (as it may need to be configured based on context) and it will have the default value "Kolibri logo"

You will then just send `KLogo`'s `altText` to `KImg`es `altText`

ShivangRawat30:
@MisRob I've made the suggested changes, please review it.

MisRob:
Please note that `release-v1.5.x` branch was renamed to `main`

lib/KLogo/index.vue
```diff
@@ -0,0 +1,150 @@
...
+       * 	Sets the minimum width for the component
+       */
+      minWidth: {
+        type: [Number, String],
+        default: undefined,
+      },
+    },
+    data() {
+      return {
+        styleObject: {
```
MisRob, line 73: `styleObject` and all related calculations are not needed here because `KImg` will take care of it after max/min/height/width/ props are sent there:

```
<KImg
      :src="logo"
      :altText="alternateText"
      :height="height"
      :width="width"
      ....
/>
```

MisRob (changes requested):
Thank you, @ShivangRawat30, this is starting to look great.

I added one note related to removing the logic that's duplicate of `KImg`'es logic

lib/KLogo/index.vue
ShivangRawat30: Done!!

ShivangRawat30:
@MisRob Done!!

MisRob (changes requested):
Almost there @ShivangRawat30. Left you one comment. In `KLogo`, there is no need for any duplicate of `KImg` logic. We only need to keep things specific particular to `KLogo` there (e.g. different alt text requirements)

lib/KLogo/index.vue
```diff
@@ -0,0 +1,143 @@
+<template>
+
+  <div>
+    <KImg
+      :src="logo"
+      :altText="alternateText"
+      :height="imgHeight"
+      :width="imgWidth"
+      :maxHeight="imgMaxHeight"
```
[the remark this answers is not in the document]
MisRob: This will allow you to remove `validateAndFormatUnits` from `KLogo`. Is the motivation for this change clear?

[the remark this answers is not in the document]
ShivangRawat30: Done!!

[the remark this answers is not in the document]
MisRob: Hi @ShivangRawat30, thanks for the latest changes. Not everything from the feedback has been resolved yet as I can still see `validateAndFormatUnits` present in `KLogo`. This is also the reason why the linter check is failing (`Unused method found: "validateAndFormatUnits"`).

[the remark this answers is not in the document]
MisRob: See lint commands

https://github.com/learningequality/kolibri-design-system/blob/main/dev_docs/01_getting_started.md#development-environment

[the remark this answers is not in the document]
ShivangRawat30: done!!

MisRob (approved):
@ShivangRawat30 Thanks for all work here!

I pushed a commit to finalize the documentation page - finishing docs is something we don't require from contributors, so I organized it myself rather than writing review. It was primarily adjustments related to organization and conventions we use. Your draft was very helpful and saved us time, thank you.

On that opportunity I also pushed a small cleanup to `KLogo` itself, primarily using `required` instead of throwing a custom warning (it's a different situation from `KImg`'s warning).

You're welcome to ask any questions and I can elaborate more if there's something you're interested in.

I will update changelog next week and merge.

docs/pages/klogo.vue
```diff
@@ -0,0 +1,51 @@
...
+          width="100%"
+          :maxWidth="150"
+        />
+      </DocsShowCode>
+    </DocsPageSection>
+
+    <DocsPageSection title="Usage" anchor="#usage">
+      <h3>Alternative text</h3>
+
+      <p>Alternative text (<code>altText</code>) is always required for the logo image. When creating it, consider the following:</p>
```
radinamatic, line 28: Looking good @MisRob, I'd just suggest to remove 'always', as in the wide world logos are not absolutely required to have alternative text. They can also be used for decorative purposes, but I'm unaware we are using ours that way anywhere in Kolibri. In our case just 'required' should suffice'.

ShivangRawat30:
@MisRob I greatly appreciate your help related to the PR, and I'm eager to make further contributions to the codebase. Please consider assigning me more issues as I'm enthusiastic about resolving them.

docs/pages/klogo.vue
MisRob: Alright, will do before merging. Thank you

Issue #459, opened 2023-09-29
Make KResponsiveWindow mixin show deprecation warning


MisRob:
We recently implemented `useKResponsiveWindow` composable that should replace [`KResponsiveWindow` mixin](https://github.com/learningequality/kolibri-design-system/blob/release-v1.5.x/lib/KResponsiveWindowMixin.js) (that will be removed one day).

To make sure that we don't introduce `KResponsiveWindow` mixin to new places in products consuming Kolibri Design System, it should show a deprecation console warning with a message guiding developers to use `useKResponsiveWindow` composable instead.

ShivangRawat30:
can I work on this issue ?

MisRob:
Thanks @ShivangRawat30. assigning you.

MisRob:
Closed by https://github.com/learningequality/kolibri-design-system/pull/463

PR #463, opened 2023-10-06
Warning when using `KResponsiveWindow`

ShivangRawat30:
## Description

This PR consoles a warning when `KResponsiveWindow` mixin is used instead of `useKResponsive`
#### Issue addressed

Addresses #459 

### Before/after screenshots

Screenshot 2023-10-07 at 12 52 08 AM

## Reviewer guidance

### DIFF
--- a/lib/KResponsiveWindowMixin.js
+++ b/lib/KResponsiveWindowMixin.js
@@ -186,7 +186,7 @@ export default {
     },
   },
   mounted() {
-    console.warn('Please Use useKResponsiveWindow instead of KResponsiveWindow Mixin');
+    console.warn(`Please use 'useKResponsiveWindow' composable instead of 'KResponsiveWindow' mixin`);
     addWindowListener(this._updateWindow);
   },
   beforeDestroy() {
--- a/lib/KResponsiveWindowMixin.js
+++ b/lib/KResponsiveWindowMixin.js
@@ -186,6 +186,7 @@ export default {
     },
   },
   mounted() {
+    // prettier-ignore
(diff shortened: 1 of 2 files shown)


lib/KResponsiveWindowMixin.js
```diff
@@ -65,6 +65,7 @@ const windowListeners = [];
 /* methods */
 
 function windowMetrics() {
+  console.warn('Please Use useKResponsiveWindow instead of KResponsiveWindow Mixin');
```
MisRob, line 68: `windowMetrics` is called on each window resize so this warning might get printed repeatedly which would be unnecessary. I'd suggest placing it to `mounted` lifecycle hook instead. Then, it will get printed only when a component using this mixin is created in DOM for the first time.

MisRob (changes requested):
Thanks for the contribution @ShivangRawat30. Just one update is needed.

ShivangRawat30:
@MisRob Done with the changes, please take a look.

MisRob:
Please note that `release-v1.5.x` branch was renamed back to `main`

lib/KResponsiveWindowMixin.js
```diff
@@ -186,6 +186,7 @@ export default {
     },
   },
   mounted() {
+    console.warn('Please Use useKResponsiveWindow instead of KResponsiveWindow Mixin');
```
MisRob, line 189:
```suggestion
    console.warn('Please use 'useKResponsiveWindow' composable instead of 'KResponsiveWindow' mixin');
```

MisRob (approved):
All good @ShivangRawat30, thank you. I made a tiny tweak and updated changelog.

Issue #450, opened 2023-09-22
Add KTextTruncator component


MisRob:
## Product

Kolibri, Studio

## Summary

In Kolibri, we have `TextTruncatorCss` component. It would be useful in Studio as well (https://github.com/learningequality/studio/issues/4086, https://github.com/learningequality/studio/issues/4166).

Move `TextTruncatorCss` to this repository and rename it to `KTextTruncator`. Create a draft documentation page with some basic guidance on using it. Additionally, make sure that `KTextTruncator` works in RTL (right-to-left) languages (you can see https://github.com/learningequality/studio/pull/4288 for how to achieve it)

## Acceptance criteria

- [ ] It supports RTL
- [ ] It is sucessfuly used in at least one place in Kolibri
- [ ] (can be done as follow-up) It has a draft documentation page

## Outside of scope

There is no need to be replacing all `TextTruncatorCss` occurrences in Kolibri by `KTextTruncator` in the scope of this issue, only one is needed though to test that everything is okay. Neither it needs to be used in Studio. These could be follow-up tasks.

MisRob:
See developers documentation
- Creating a component: https://github.com/learningequality/kolibri-design-system/blob/develop/dev_docs/03_how_to_update_library.md
- Creating a documentation page for a component https://github.com/learningequality/kolibri-design-system/blob/develop/dev_docs/04_how_to_update_docs.md#create-a-documentation-page-for-a-new-component

MisRob:
This is reserved for contribution (was open as a follow-up to already merged contribution)

muditchoudhary:
@MisRob Okay I understand the issue. Could you please assign it to me?

MisRob:
Hi @muditchoudhary, great, it's all yours, thank you

MisRob:
@muditchoudhary It'd be best if you started working in this repository and as soon as you feel you're ready for

> _It is sucessfuly used in at least one place in Kolibri_

please let me know and I will provide a bit of guidance as it might be tricky for the first time due to some problems with `yarn link` we recently discovered.

muditchoudhary:
> It'd be best if you started working in this repository and as soon as you feel you're ready for

Yeah, sure! I have set up the local env. no issue with it. 

> please let me know and I will provide a bit of guidance as it might be tricky for the first time due to some problems with `yarn link` we recently discovered.

Sure, I'll let you know if I get into some issue that I cannot resolve. Thanks for telling about the yarn issue.

[...]

PR #464, opened 2023-10-10
Issue-450: Made KTextTruncator 

muditchoudhary:
## Description

This PR creates a KTextTruncator component that will be used to truncate texts and this PR creates a ktexttruncator.vue for the docs of the KTextTruncator component.

#### Issue addressed

Addresses #450 

## Steps to test

1. Go to docs/pages/playground.vue
2. Add the KTextTruncator
```js
<KTextTruncator text="Option (1) is simpler and sufficient as the first step, so we recommend starting that way. On certain occasions, it may be necessary to test an update in a product. If you're a contributor, you are welcome to try (2) as well, but typically we do not require contributors to do so, or we will provide support if needed." :maxLines="1" />
```
3. Run the server and visit `http://localhost:4000/playground`
`yarn dev`

## Testing checklist

## Reviewer guidance


MisRob:
Thank you @muditchoudhary! We will have a look. Could you please rebase on top of the `main` branch and then retarget the PR there? Alternatively, you can let me know and can do so.

muditchoudhary:
> Thank you @muditchoudhary! We will have a look. Could you please rebase on top of the `main` branch and then retarget the PR there? Alternatively, you can let me know and can do so.

@MisRob I rebase successfully to the main. Hope it would be good to test.

MisRob:
Great, thank you @muditchoudhary

docs/pages/ktexttruncator.vue
```diff
@@ -0,0 +1,20 @@
+
+import KTextTruncator from '../../lib/KTextTruncator.vue';
```
MisRob, line 2: This import is not needed, I believe (and if it was, it would need to be done in the `<script>` part of the Vue file

docs/pages/ktexttruncator.vue
MisRob, line 1: Nice page, thanks for working on it!

MisRob: One possible improvement would be adding documentation strings above the `text` and `maxLines` props in `KTextTruncator.vue` file (similarly to how it's done for `lineHeightIE`). Then, these two descriptions that are currently empty will get auto-generated from them:
review-1

lib/KTextTruncator.vue
```diff
@@ -0,0 +1,138 @@
...
+        if (this.maxLines === 1) {
+          return {
+            overflow: 'hidden',
+            textOverflow: 'ellipsis',
+            whiteSpace: 'nowrap',
+          };
+        }
+
+        // 54 is random number only to be able to define `supports` test condition
+        if (process.client && 'CSS' in window && CSS.supports && CSS.supports('-webkit-line-clamp: 54')) {
```
MisRob, line 79: This is not necessary, just nice to have.

In other places in KDS, we usually define this constant `const nuxtServerSideRendering = process && process.server;` so that by its name it's clear immediately what the check is about. It could be nice to do something similar here.

MisRob (commented):
Thanks @muditchoudhary, code is looking good overall. I left few smaller notes. Also, could you please fix linting check? See package.json for linting commands.

As soon as this is resolved, I will try it out in Kolibri and see if I can observe the issue with the space between ellipsis and text that you mentioned some time ago. In any case, I think this PR won't be a culprit and should be good to go soon.

MisRob:
I'll take care of the changelog check myself before we merge this, so no worries about that one

lib/KTextTruncator.vue
```diff
@@ -0,0 +1,151 @@
...
+          return {
+            overflow: 'hidden',
+            textOverflow: 'ellipsis',
+            whiteSpace: 'nowrap',
+          };
+        }
+
+        // 54 is random number only to be able to define `supports` test condition
+        if (
+          nuxtClientSideRendering &&
```
MisRob, line 88: @muditchoudhary Thanks for the naming change and with apology I missed this during the first round of review - what was actually the reason for adding the `process.client` check at this particular place? 

Now when I look at all conditions in `truncate` method as a whole, originally it said something like:

if one line - use ellipsis
if multiple lines - if a browser supports `webkit-line-clamp` then use it, else use IE fallback

Now with your addition it says something along the lines

if one line - use ellipsis
if multiple lines - if a browser supports `webkit-line-clamp` **and is not server side rendering** then use it, else use IE fallback

which doesn't seem to make sense high-level, right?

MisRob: Ah was it about accessing the `window`?

MisRob: If so, I think we could rather place this check at the beginning of the `truncate` method just to make it a bit more clear. I will wait for your reply though to be sure what was the motivation for having it here.

MisRob: @muditchoudhary If that's the case, I'd suggest to rather put this at the very beginning of `truncate` method:

```
const nuxtServerSideRendering = process && process.server;
if (nuxtServerSideRendering) {
  return;
}
```

just to make the placement of that condition a bit more clear. This is similar to other places in KDS.

MisRob: It's rather a minor note, nice to have. Nothing blocking.

muditchoudhary:
@MisRob I added the condition you told me and tested it out, and there is no issue.

Also thanks for being helpful in the review.

PR #469, opened 2023-10-12
debounced updateWindow

### LINKED ISSUE #461: Optimize `useKWindowDimensions` performance
## Summary

In `useKWindowDimensions`, we read `window.innerWidth/Height` on each `resize` event

https://github.com/learningequality/kolibri-design-system/blob/9ea4bae765fb3f0270c3c764c88cb44206147cc7/lib/useKWindowDimensions.js#L45

https://github.com/learningequality/kolibri-design-system/blob/9ea4bae765fb3f0270c3c764c88cb44206147cc7/lib/useKWindowDimensions.js#L25-L26

This is not good performance-wise since [reading innerWidth/Height causes layout/reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a#getting-window-dimensions).

It'd be good to debounce the event handler calls so that `innerWidth/Height` is not read so often.

`lodash/debounce` is available in this repository for that purpose.

## References

- > If the resize event is triggered too many times for your application, see [Optimizing window.onresize](https://web.archive.org/web/20220714020647/https://bencentra.com/code/2015/02/27/optimizing-window-resize.html) to control the time after which the event fires. (https://developer.mozilla.org/en-US/docs/Web/API/Window/resize_event)

- https://medium.com/@vaibhavar/ui-performance-window-resize-handler-debouncing-2ec5f7432165

Jaspreet-singh-1032:
fixes #461

I also linked this with Kolibri and tested this out, and It seems to be working fine for me.

## Reviewer guidance

### DIFF
--- a/lib/useKResponsiveWindow/__tests__/useKResponsiveWindow.spec.js
+++ b/lib/useKResponsiveWindow/__tests__/useKResponsiveWindow.spec.js
@@ -36,16 +36,19 @@ describe('useKResponsiveWindow composable', () => {
     resizeWindow(1700);
     const wrapper = mount(TestComponent());
     await wrapper.vm.$nextTick();
-    expect(wrapper.vm.windowWidth).toEqual(1700);
-    expect(wrapper.vm.windowHeight).toEqual(768);
-    expect(wrapper.vm.windowBreakpoint).toEqual(7);
-    expect(wrapper.vm.windowIsPortrait).toEqual(false);
-    expect(wrapper.vm.windowIsLandscape).toEqual(true);
-    expect(wrapper.vm.windowGutter).toEqual(24);
-    expect(wrapper.vm.windowIsShort).toEqual(false);
-    expect(wrapper.vm.windowIsLarge).toEqual(true);
-    expect(wrapper.vm.windowIsMedium).toEqual(false);
-    expect(wrapper.vm.windowIsSmall).toEqual(false);
+    // Wait for the throttling delay
+    setTimeout(() => {
+      expect(wrapper.vm.windowWidth).toEqual(1700);
+      expect(wrapper.vm.windowHeight).toEqual(768);
+      expect(wrapper.vm.windowBreakpoint).toEqual(7);
+      expect(wrapper.vm.windowIsPortrait).toEqual(false);
+      expect(wrapper.vm.windowIsLandscape).toEqual(true);
+      expect(wrapper.vm.windowGutter).toEqual(24);
+      expect(wrapper.vm.windowIsShort).toEqual(false);
+      expect(wrapper.vm.windowIsLarge).toEqual(true);
+      expect(wrapper.vm.windowIsMedium).toEqual(false);
+      expect(wrapper.vm.windowIsSmall).toEqual(false);
+    }, 20);
   });
 
   describe('check if windowBreakpoint is set on width change', () => {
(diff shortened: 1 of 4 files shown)


lib/useKWindowDimensions.js
```diff
@@ -1,5 +1,6 @@
 import './composition-api'; //Due to @vue/composition-api shortcomings, add plugin prior to use in kolibri, studio and tests
 import { onMounted, onUnmounted, ref } from '@vue/composition-api';
+import debounce from 'lodash/debounce';
```
indirectlylit, line 3: Just FYI, [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) might be appropriate here because it implements a dynamic throttle. The old responsive window mixin used the `frame-throttle` wrapper library for this:

https://github.com/learningequality/kolibri-design-system/blob/e578edef93471f78c051599a8162043710ad8c28/lib/KResponsiveWindowMixin.js#L59

https://github.com/learningequality/kolibri-design-system/blob/e578edef93471f78c051599a8162043710ad8c28/lib/KResponsiveWindowMixin.js#L74-L77

Jaspreet-singh-1032: Hello @indirectlylit, Thanks for your review. Just to confirm, do I need to implement it similar to how it is implemented in `KResponsiveWindowMixin.js` using `throttle`?

indirectlylit: Hi @Jaspreet-singh-1032 - that's for you (and @MisRob) to decide. This is a _very_ minor note and should not block the PR.

That said if you're interested, some links:

* https://css-tricks.com/using-requestanimationframe/
* https://css-tricks.com/debouncing-throttling-explained-examples/

Personally, I prefer using this form of throttling for anything related to fluid UI changes because the system can choose an appropriate refresh rate rather than hard-coding a somewhat arbitrary delay period.

Jaspreet-singh-1032: Thank you, @indirectlylit, I was previously unaware of this function. I will read about it and If it is a better option then I will definitely use it here.

rtibbles: For most browsers this will use requestAnimationFrame, which will invoke the callback just before the next repaint - which means that it should not be the cause of a reflow by reading the inner width and height of the window (which is a strong part of the motivation for this issue), so I do think this is a good change to use here.

Jaspreet-singh-1032: Thanks, @rtibbles, I read some blogs about `requestAnimationFrame` and I also think this is a better option to use here.

MisRob: Thanks @indirectlylit, that's a good idea

akolson (commented):
Hi @Jaspreet-singh-1032, with exception build errors that I think are resolvable(Some linting issues at the very least), the changes made, make sense me. (cc @MisRob on the build failures )

MisRob:
I'll push update that will resolve the changelog check after all remaining checks are working and the PR is approved (@akolson this will be done by reviewers eventually for majority of PRs from our contributors, I will follow-up soon)

akolson:
Hi @MisRob! My comment was meant to specifically address the build failure. I see these have been addressed(or to be addressed). I am more confident approving it now. Thanks

MisRob:
Okay @akolson, thank you. Changelog is updated.

MisRob:
Checks should be okay now. Would recommend confirming it in one or two places in Kolibri. It's up to you how you want to proceed.

MisRob:
Seems all is done here so merging

lib/useKResponsiveWindow/__tests__/useKResponsiveWindow.spec.js
MisRob, line 1: @Jaspreet-singh-1032 One of my colleagues noticed an issue with using `setTimeout` after we merged it, so I opened a follow-up https://github.com/learningequality/kolibri-design-system/issues/480. You don't need to work on it, of course! It's rather FYI in case you'd find the note interesting. I didn't realize that either.

PR #470, opened 2023-10-12
Match media testing guard and bug fix

rtibbles:
## Description
* Adds an additional guard to prevent attempting to access the window.matchMedia API when it is in an environment that it is not supported
* Fixes a hitherto unnoticed bug, whereby the use of a JS getter to return the value of window.matchMedia meant that removing event listeners was doing nothing, as the getter was returning a new instance each time (and so was not unbinding the listener from the instance that it had actually been bound to).
* Does this by caching the window.matchMedia result call, to minimize code changes and to continue to allow delay of instantiation of the mediaQueryList object.

#### Issue addressed
Fixes failing tests seen in develop on Kolibri
Fixes unreported bug that would stop listeners from being unbound

## Testing checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,17 @@ Changelog is rather internal in nature. See release notes for the public overvie
 
 <!-- All new changelog items should come here -->
 
+- [#470]
+  - **Description:** Fix bug and add test guard in MediaQuery implementation
+  - **Products impact:** none
+  - **Addresses:** -
+  - **Components:** none
+  - **Breaking:** no
+  - **Impacts a11y:** no
+  - **Guidance:** -
+
+[#470]: https://github.com/learningequality/kolibri-design-system/pull/470
+
 - [#469]
   - **Description:** Throttle the resize listener handler
   - **Products impact:** updated API
--- a/lib/useKResponsiveWindow/MediaQuery.js
+++ b/lib/useKResponsiveWindow/MediaQuery.js
@@ -10,13 +10,17 @@ export default class MediaQuery {
   constructor(query, eventHandler) {
     this.query = query;
     this.eventHandler = eventHandler;
+    this._mediaQueryList = null;
   }
 
   /**
    * @returns {Object} Media query list
    */
   get mediaQueryList() {
-    return window.matchMedia(this.query);
+    if (!this._mediaQueryList) {
+      this._mediaQueryList = window.matchMedia(this.query);
+    }
+    return this._mediaQueryList;
   }
 
   /**
--- a/lib/useKResponsiveWindow/MediaQuery.js
+++ b/lib/useKResponsiveWindow/MediaQuery.js
@@ -32,8 +32,8 @@ export default class MediaQuery {
    * @returns {Object} Containing mediaQueryList, eventHandler, and stopListening
    */
   startListening() {
-    //Prevent function execution if Nuxt is server side rendering
-    if (this.isNuxtServerSideRendering()) {
+    // Prevent function execution if Nuxt is server side rendering
+    if (this.isNuxtServerSideRendering() || !window.matchMedia) {
       return;
     }
 
@@ -48,6 +48,10 @@ export default class MediaQuery {
    * Stop listening for media query events
    */
   stopListening() {
+    // Prevent function execution if Nuxt is server side rendering
+    if (this.isNuxtServerSideRendering() || !window.matchMedia) {
+      return;
+    }
     if (this.mediaQueryList.removeEventListener) {
       this.mediaQueryList.removeEventListener('change', this.eventHandler);
     } else {


CHANGELOG.md
```diff
@@ -47,6 +47,17 @@ Changelog is rather internal in nature. See release notes for the public overvie
 
 ## Version 2.0.0
 
+- [#470]
```
MisRob, line 50: Could you please move this istem to the "Upcoming version"?

The current items in the "Version 2.0.0" are already released. I will rename this section to "Version 2.0.0-beta0 (released)" to make it clearer. If you have some others ideas of what may help, let me know.

MisRob: For now I updated the changelog in this way
Screenshot from 2023-10-13 12-32-27

rtibbles: From my experience looking through dependency updates for many different JS packages - I have found the style of having changes only mentioned in alphas, betas, and rcs and not in the final release much more difficult to parse, especially when looking for breaking changes. I'd recommend not separating the changelog in this way, and see the beta releases as try outs for what the final release will be.

MisRob: That makes sense and will be ideal in the long term.

This is temporary only to help us now when we're in the transition towards re-introducing a stable release process. We already have beta with release notes so that we can start introducing and regression testing. Without them, it'd be difficult because we were postponing releases and accumulated lots of breaking changes. So release notes were created in advance and with that we will need to keep track for some time what other changes need to be eventually added.

This should get better with more frequent releases. With @marcellamaki and @thanksameeelian, we decided to cleanup the release notes from all betas and merge them for the final release as soon as it's ready so that should address it. On that occasion we can do one last changelog cleanup and merge all sections for v2.

cc @thanksameeelian @marcellamaki

MisRob: Alternatively, we could just add a comment to indicate to the changelog where the release notes end. I tried something like that already, but it seems it's hard to notice for devs. In any case, I'd be okay with any way of making it easier for us to track where the existing release notes end in the changelog, if you have any other ides.

akolson (commented):
Code additions look correct to me. Thanks @rtibbles

CHANGELOG.md
rtibbles: Will rebase now to move this.

rtibbles:
Updated!

MisRob:
I wanted to note that I only jumped in to notify about changelog, but didn't review myself. If there is anything needed from my side, let me know.

rtibbles:
Rebased again to resolve the CHANGELOG merge conflict.

At the risk of sounding needy, I do need approval - the changes are fairly straight forward so would be a quick review.

MisRob:
I assume that @akolson review was meant as approval so marking it green. Please reach out if anything @akolson

MisRob (approved):

CHANGELOG.md
MisRob: Thanks

MisRob:
I skimmed it too now and haven't noticed anything blocking. Leaving only one minor note.

lib/useKResponsiveWindow/MediaQuery.js
```diff
@@ -32,8 +36,8 @@ export default class MediaQuery {
    * @returns {Object} Containing mediaQueryList, eventHandler, and stopListening
    */
   startListening() {
-    //Prevent function execution if Nuxt is server side rendering
-    if (this.isNuxtServerSideRendering()) {
+    // Prevent function execution if Nuxt is server side rendering
+    if (this.isNuxtServerSideRendering() || !window.matchMedia) {
```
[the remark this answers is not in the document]
MisRob: Not sure in what browsers, just assuming it may happen when I see this condition.

CHANGELOG.md
```diff
@@ -6,6 +6,17 @@ Changelog is rather internal in nature. See release notes for the public overvie
 
 <!-- All new changelog items should come here -->
 
+- [#470]
+  - **Description:** Fix bug and add test guard in MediaQuery implementation
+  - **Products impact:** none
+  - **Addresses:** -
+  - **Components:** none
```
MisRob, line 13: Just a note that here, all affected components would ideally be included here even though it doesn't modify them directly, it's something that's used under the hood (this will help with the release notes).

MisRob: No need to do anything about it, I already prepared the release notes for this item too. Just jumping into PRs and using them as spreading the word.

CHANGELOG.md
```diff
@@ -6,6 +6,17 @@ Changelog is rather internal in nature. See release notes for the public overvie
 
 <!-- All new changelog items should come here -->
 
+- [#470]
+  - **Description:** Fix bug and add test guard in MediaQuery implementation
+  - **Products impact:** none
```
MisRob, line 11: Here, the product impact would be bugfix. Here's an example of release notes for `useKResponsiveWindow`. We'd ideally be able to read this from the changelog rather than opening a PR. As mentioned below, no action needed.

Screenshot from 2023-10-23 13-34-29

PR #465, opened 2023-10-11
KDropdown and KIconButton tweaks

nucleogenesis:
## Description

- Extend **KDropdownButton** with a new slot to customize the dropdown options' rendering
- Drill the `containFocus` prop through **KDropdownButton** to `UiPopover`
- Add `v-bind="$attrs"` to **KIconButton**

#### Issue addressed

This was necessary to accommodate needs for the updates to Kolibri Quiz.

### Before/after screenshots

 
n/a

## Steps to test

I suggest testing my [Kolibri PR](https://github.com/learningequality/kolibri/pull/11382)  which puts these changes to use. 

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [ ] The change has been added to the `changelog`

## Reviewer guidance

- [ ] Is the code clean and well-commented?
- [ ] Are there tests for this change?
- [ ] Are all UI components LTR and RTL compliant (if applicable)?

PR #497, opened 2023-11-28
Support for Quiz Management (Kolibri 0.17)

nucleogenesis:
## Description

- Adds new "Expand All" icon
- Updates `KDropdownMenu` to emit a `@tab` event when the `[[Tab]]` key is pressed.
- Updates `KDropdownMenu` to emit the `@close` event when the menu is programmatically closed. 

## Changelog

- [#497]
  - **Description:** KDropdownMenu now emits a @tab event when the user hits the [Tab] key and a @close event when the menu is closed programmatically. Additionally, a new icon for Expand All was added and can be used just like any other icon with the "expandAll" name.
  - **Addresses:** -
  - **Breaking:** No
  - **Guidance:**  The @tab event can be used for more precise focus management as the popover by default could end up sending focus to the root HTML element by default. Note that the browser event is passed to the handler function, so you may need/want to call `preventDefault()` on that event depending on your use case.

[497]: https://github.com/learningequality/kolibri-design-system/pull/497

## After review

## Comments

Originally this work was in #465 but this new PR is specifically the things needed for the work on Quiz Management.


marcellamaki (approved):
thank you @nucleogenesis

PR #502, opened 2023-12-07
feat: add `onError` handler for `KImg`

### LINKED ISSUE #372: KImg: Add error event
## Summary

`KImg` emits the `'error'` event when an error is encountered while loading the image specified in `src`

## Epic

- https://github.com/learningequality/kolibri-design-system/issues/368

## Acceptance criteria

- [ ] `KImg` emits the `'error'` event as described above

## Linked issues
 
- **Blocked by**
  - https://github.com/learningequality/kolibri-design-system/issues/369

EshaanAgg:
## Description
`KImg` now emits the 'error' event when an error is encountered while loading the image specified in src with the `event` parameter that can be used by the clients. 

#### Issue addressed
Fixes #372 

## Changelog

- [#502]
  - **Description:** Add dispatching of 'error' event when failed to load image for 'KImg'
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/372
  - **Components:** KImg
  - **Breaking:** no
  - **Impacts a11y:** yes  
  - **Guidance:** Allows the consumer to hook into the lifecycle of 'KImg' and handle the cases when the image fails to load

[#502]: https://github.com/learningequality/kolibri-design-system/pull/502

Used the `@error` provided by `Vue` and declared a new event `error`

No

## Testing checklist

## Reviewer guidance

## After review

### DIFF
--- a/lib/KImg.vue
+++ b/lib/KImg.vue
@@ -155,8 +155,7 @@
       },
       onError(event) {
         /**
-         * Emitted when the image fails to load
-         * @payload {Event} event The DOM event that triggered the error
+         * Emitted when the image fails to load. The DOM event that triggered the error is available in the payload.
          */
         this.$emit('error', event);
       },
--- a/lib/KImg.vue
+++ b/lib/KImg.vue
@@ -5,6 +5,7 @@
       :src="src"
       :alt="alternateText"
       :style="styleObject"
+      @error="onError"
     >
     <slot></slot>
   </div>
@@ -81,6 +82,14 @@
         default: undefined,
       },
     },
+    emits: {
+      /**
+       * Emitted when the image fails to load
+       * @param {Event} event
+       * @returns {true} Denotes that the validation has passed
+       */
+      error: (event) => true,
+    },
     data() {
       return {
         styleObject: {
@@ -152,6 +161,9 @@
           }
         }
       },
+      onError (event) {
+        this.$emit('error', event);
+      }
     },
   };
 
--- a/lib/KImg.vue
+++ b/lib/KImg.vue
@@ -87,7 +87,8 @@
        * Emitted when the image fails to load
        * @param {Event} event
        * @returns {true} Denotes that the validation has passed
-       */
+      */
+      // eslint-disable-next-line no-unused-vars
       error: (event) => true,
     },
     data() {
--- a/lib/KImg.vue
+++ b/lib/KImg.vue
@@ -87,9 +87,9 @@
        * Emitted when the image fails to load
        * @param {Event} event
        * @returns {true} Denotes that the validation has passed
-      */
+       */
       // eslint-disable-next-line no-unused-vars
-      error: (event) => true,
+      error: event => true,
     },
     data() {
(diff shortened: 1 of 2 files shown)


MisRob:
Hi @EshaanAgg, thanks for the contribution! I will review some time next week.

lib/KImg.vue
```diff
@@ -81,6 +82,15 @@
         default: undefined,
       },
     },
+    emits: {
```
MisRob, line 85: This would be great, but it seems it's only available for Vue 3 https://v3-migration.vuejs.org/breaking-changes/emits-option.html? Doesn't seem to be breaking anything but I think it'd be best to remove it.

When you place the documentation string directly above the emit like this
```
/**
 * Emitted when the image fails to load
 * /
this.$emit('error', event);
```

I think our documentation generator will pick it up. It may also be nice to document that `event` is available in the payload (just in case consumers needed it)

MisRob (commented):
Thanks for the contribution @EshaanAgg. Overall it's looking good and works. Just a bit of cleanup needed and then I think we can merge.

MisRob:
@EshaanAgg I also wanted to mention I won't be able to follow-up here this year as this is my last day before few weeks of time off. Also, the whole team's availability will too soon be limited until the second week of January. Thank you and hopefully see you next year :)!

MisRob:
@LianaHarris360 There's nothing needed at this point. If you could just kindly merge in my absence after the feedback is resolved, I'd appreciate it. Otherwise, I'll have a look next year. Thank you.

lib/KImg.vue
EshaanAgg: Got it! Will make the necessary changes!

EshaanAgg:
@LianaHarris360, I have made the suggested changes by @MisRob! Would love it if you could review the same :)
Merry holidays!

lib/KImg.vue
```diff
@@ -152,6 +153,13 @@
           }
         }
       },
+      onError(event) {
+        /**
+         * Emitted when the image fails to load
+         * @payload {Event} event The DOM event that triggered the error
```
MisRob, line 159: One last little note - in general it's really good to use JSDoc syntax, however in this case our documentation auto-generation doesn't handle it:

Screenshot from 2023-12-15 19-10-50

(from http://localhost:4000/kimg#events)

So I think it'd be best to just communicate this as a plain text to ensure it gets propagated to the docs. Something along the lines of

```
/**
 * Emitted when the image fails to load. The DOM event that triggered the error is available in the payload.
 * /
 ```

MisRob:
Thank you @EshaanAgg and happy holidays to you too. I left one last note before signing off. Be well.

lib/KImg.vue
EshaanAgg: Ahh okay! Didn't know about this limitation of Kolibri. Pushing the changes in a minute!

LianaHarris360 (approved):
The changes look good! Thank you @EshaanAgg

PR #521, opened 2024-01-16
Move date-fns dependency to depencies rather than dev-dependencies.

rtibbles:
## Description
* Fixes bug that would happen when installing KDS in a non-Kolibri environment because of an unspecified runtime dependency

## Changelog

- [#521]
  - **Description:** Move date-fns dependency to depencies rather than dev-dependencies.
  - **Products impact:** bugfix
  - **Addresses:** N/A
  - **Components:** KDateRange/KDateInput
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** ~

[#521]: [https://github.com/learningequality/kolibri-design-system/pull/521]

## Steps to test

1. Install in Studio.
2. Run devserver.
3. See it not break.


MisRob (approved):
Thank you for helping with this, @rtibbles. The code change makes sense and I confirm I could run Studio and see it not break.

Issue #296, opened 2022-01-21
Full theming support for toolbars 


MisRob:
## Product

Kolibri

## Desired behavior

`UiToolbar` component is fully themed internally and can also accept theme tokens through the `textColor` prop.

## Current behavior

`UiToolbar` accepts `black` or `white` values as text color

https://github.com/learningequality/kolibri-design-system/blob/b41e4f61ea05bd4f01b08f28f4cc10abb63d2d0f/lib/keen/UiToolbar.vue#L67-L69

These values are then translated into CSS classes

https://github.com/learningequality/kolibri-design-system/blob/b41e4f61ea05bd4f01b08f28f4cc10abb63d2d0f/lib/keen/UiToolbar.vue#L103

that have hardcoded colors

https://github.com/learningequality/kolibri-design-system/blob/b41e4f61ea05bd4f01b08f28f4cc10abb63d2d0f/lib/keen/UiToolbar.vue#L232

There is no straightforward way of setting the text color so that we could use `tokens.text` or `tokens.textInverted`.

Another hardcoded color is the `white` color used for the default background:

https://github.com/learningequality/kolibri-design-system/blob/b41e4f61ea05bd4f01b08f28f4cc10abb63d2d0f/lib/keen/UiToolbar.vue#L206

## The Value Add

It is important to be able to use text tokens to ensure that themes have sufficient contrast of text against the background. This will make it possible in the following Kolibri toolbars:

- [AppBar](https://github.com/learningequality/kolibri/blob/9048d77029e622b57d95009ab8b968f12dd5d99a/kolibri/core/assets/src/views/AppBar.vue#L7)
- [ImmersiveToolbar](https://github.com/learningequality/kolibri/blob/9048d77029e622b57d95009ab8b968f12dd5d99a/kolibri/core/assets/src/views/ImmersiveToolbar.vue#L3)
- [LearningActivityBar](https://github.com/learningequality/kolibri/blob/9048d77029e622b57d95009ab8b968f12dd5d99a/kolibri/plugins/learn/assets/src/views/LearningActivityBar.vue#L3)
- [the toolbar in the custom channel content modal](https://github.com/learningequality/kolibri/blob/9048d77029e622b57d95009ab8b968f12dd5d99a/kolibri/plugins/learn/assets/src/views/ChannelRenderer/ContentModal.vue#L10)
[...]

PR #536, opened 2024-01-26
Add internal theming support

nick2432:
## Description
This pull request enhances the UiToolbar component by integrating theme tokens for background and text colors. Previously, the component used hardcoded colors, and now it dynamically fetches the appropriate color values from the theme tokens based on the component's properties, textColor. This modification ensures that the component adheres to theming conventions and allows for easier customization through theme tokens.

#### Related Issues
Closes https://github.com/learningequality/kolibri-design-system/issues/296

Addresses #*PR# HERE*

### Screenshot
Screenshot from 2024-01-26 19-05-54
Screenshot from 2024-01-26 19-05-40


lib/keen/UiToolbar.vue
```diff
@@ -23,16 +21,16 @@
           </div>
         </slot>
       </div>
+    </div>
+
+    <div class="ui-toolbar__body" :class="{ 'has-brand-divider': hasBrandDivider }">
```
MisRob: I don't understand how this change relates to the theming issue being resolved in this pull request, could you explain what's its motivation?

lib/keen/UiToolbar.vue
```diff
@@ -23,16 +21,16 @@
...
       <slot>
         <div v-if="title" class="ui-toolbar__title">
           {{ title }}
         </div>
       </slot>
-      <slot name="navigation" class="ui-toolbar__nav"></slot>
```
MisRob, line 31: Why is this slot removed?

lib/keen/UiToolbar.vue
```diff
@@ -66,7 +64,7 @@
       },
       textColor: {
         type: String,
-        default: 'black', // 'black' or 'white'
+        default: 'default', // 'black' or 'white'
```
MisRob, line 65: There are still places in this component that expect `textColor` be either `'black'` or `'white'`, for example  `progressColor` function or it seems that even some `Ui` sub-components.

The most straightforward way to deal with it I can think of, while preserving the approach you've used, would be to keep

```
textColor: {
  type: String,
  default: 'black', // 'black' or 'white'
}
```

and then internally map `'black'` to `tokens.text` and `'white'` to `tokens.textInverted` when setting the text color (similarly to your `getColor`). This won't break other places that still expect these two strings.

lib/keen/UiToolbar.vue
```diff
@@ -116,6 +118,14 @@
     },
 
     methods: {
+      getColor() {
+        if (this.textColor === 'default') {
+            return this.$themeTokens.surface;
+        } else {
+        const tokens = this.textColor.split('.');
+        return this.$themeTokens[tokens[1]] || '';
```
MisRob, line 124: I don't understand what you intended to do here, can you explain?

lib/keen/UiToolbar.vue
```diff
@@ -134,12 +144,9 @@
   .ui-toolbar {
     position: relative;
     display: flex;
-    align-content: center;
     align-items: center;
-    justify-content: space-between;
     height: $ui-toolbar-height;
     padding-left: rem(16px);
-    max-width: 100%;
```
MisRob: I don't think we need changes of styles other than those related to theming

MisRob:
Thanks @nick2432, this is heading in a good direction. I left some first pointers.

lib/keen/UiToolbar.vue
MisRob: This is just one example. I'd recommend cleaning up other places as well. If you saw a bug that you wanted to implement an extra fix for, please always explain the motivation for changes.

nick2432:
@MisRob , I accidentally made a lot of changes. I apologize for that.

lib/keen/UiToolbar.vue
nick2432: This is for passing any of the theme tokens listed in the KDS documentation. For example, in the above-given screenshot, I used the themetoken.document token, and it worked.
Screenshot from 2024-01-26 19-05-54

MisRob: `textColor="$themeTokens.document"` means that you're passing string `"$themeTokens.document"` rather than variable `$themeTokens.document` that will evaluate into a color code.

If we were going to pass theme tokens via the `textColor` prop, then we'd want to accept them like `:textColor="$themeTokens.document"` which is a shortcut for `v-bind` that passes variables https://v2.vuejs.org/v2/guide/components-props#Passing-Static-or-Dynamic-Props. Then you wouldn't need

```
const tokens = this.textColor.
return this.$themeTokens[tokens[1]] || '';
```

**However, looking at the approach you took overall, I'd recommend you accept only `'black'` and `'white'` strings as `textColor` (same as previously) but translated them into theme colors as explained here https://github.com/learningequality/kolibri-design-system/pull/536#discussion_r1476049877.**

MisRob:
No problem, @nick2432. Please (1) look at look at guidance here https://github.com/learningequality/kolibri-design-system/pull/536#discussion_r1476049877 and (2) clean up all changes that are not related to theming.

MisRob:
I marked this as draft since there's lots of work in progress and it seems we're figuring out a suitable approach. For your work, that doesn't change anything. Please mention me when you finished your updates and I will have a look again. Thank you!

MisRob:
Hi @nick2432, please note that the `main` branch is now renamed to `develop. You may need to update your local environment.

MisRob:
Hi @nick2432 I'm sorry but I will close this since meanwhile we needed to do some work around theming and will need to rethink our approach to this issue.

MisRob:
Thank you!

Issue #269, opened 2021-10-26
deprecate 'tune' alias


indirectlylit:
## Product

Need to check for all instances

## Desired behavior

We should rename all instances of `tune` to `settings` and deprecate the old alias:

https://design-system.learningequality.org/icons/#token-tune

## Current behavior

The icon we use for 'settings' is currently called 'tune' based on the Material name.

## (Optional) The Value Add

We should use KDS-specific naming to help engineers and designers reuse icons correctly.

nucleogenesis:
General guidance for this is to grep the whole code base of Kolibri, Studio and KDP to ensure that there is no use of the `tune` icon and to replace it w/ the `settings` icon as noted here.

nucleogenesis:
@nikkuAg or @sivangbagri -- if one of you'd like to take this on it seems that the only instance of this being used in Kolibri is in the epub_viewer's "SettingsButton.vue" file.

That said, fixing this will require two parts:

1: Update Kolibri to use `settings` there instead of `tune` and make a PR against Kolibri.
2: Update the Kolibri Design System `iconDefinitions.js` file to no longer include the `tune` icon, as it will no longer be used. Then make a PR on Kolibri Design System for that change

MisRob:
Closing because we have one another `settings` icon so `tune` can't be renamed to `settings`.

PR #537, opened 2024-01-29
Issue 269, removed usage of tune icon

RaviKumarAgurla:
Closes https://github.com/learningequality/kolibri-design-system/issues/269

### DIFF
--- a/lib/KIcon/iconDefinitions.js
+++ b/lib/KIcon/iconDefinitions.js
@@ -261,10 +261,8 @@ const KolibriIcons = {
     icon: require('./precompiled-icons/material-icons/perm_device_information/baseline.vue')
       .default,
   },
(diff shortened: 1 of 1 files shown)


MisRob:
Hello @RaviKumarAgurla, thanks for volunteering.

It seems there's misunderstanding caused by naming conflicts. Your changes remove `tune`/`filter` icon completely (to be replaced by the different `settings` icon) whereas the original intent of https://github.com/learningequality/kolibri-design-system/issues/269 was to keep the `tune/filter` icon but only rename it to `settings`. It's quite old issue so I assume it was created before we introduced yet another `settings` icon. So I don't think we want to remove `tune` completely, even though the suggested new name for it is taken now. I will also close the issue since it's obsolete. Thank you for looking into it, even though I can't merge, it helped.

Issue #532, opened 2024-01-25
Establish the new and automated release process


MisRob:
## Kolibri-Studio-KDS workflow

- KDS releases will be installed to
  - Kolibri's `develop` and `release-v*`
  - Studio's `unstable` 
- **Kolibri's `develop` and `release-v*`, and Studio's `unstable` always track the latest KDS release. They will never reference a KDS commit directly.**
- **Working branches targeting `develop/unstable` can reference an unreleased KDS commit (typically new KDS work in support of the Kolibri/Studio feature being developed), but it is not allowed to merge a working branch into `develop/unstable` as long as it references an unreleased KDS commit.** Example:
  - (1) A developer works on refactoring card components in Kolibri on a working branch from `develop`
  - (2) They need to make some new updates to KDS card component, so they open a KDS pull request. It is reviewed and merged.
  - (3) In their Kolibri working branch, they reference the latest KDS commit directly.
  - (4) KDS release manager prepares a new KDS release that contains the change
  - (5) This KDS release is installed into `develop`
  - (6) The developer rebases their working branch on top of the latest `develop` before merging their PR and can therefore remove their direct KDS commit reference 
- **Kolibri's `release-v*` works in the same way, with the exception that at some point we will freeze the KDS version and won't continue updating it**.

## Automation

To be able to support the workflow above efficiently, we will need to release KDS frequently (perhaps bi-weekly or even weekly, to be determined) and be always ready to release upon request so that it doesn't block merging PRs in the products. This will be supported by the automated releases utilizing `npm` and `dependabot`. The automated KDS release part of the workflow above will work in the following way:

- **(1) After a new KDS version is tagged, it will be automatically published to npm**
  - We can re-use https://github.com/learningequality/le-utils/blob/main/.github/workflows/npm-publish.yml

- **(2) dependabot will open PRs with the new KDS version installed**
  - PRs are opened to Kolibri's `develop` and `release-v*`, and Studio's `unstable`
  - PRs are then treated as for any other dependency - we test and merge them, push some updates if needed, or reject and tell dependabot to not update the branch anymore (this will be the case for Kolibri's `release-v*` some time before the release is out)

```[tasklist]
### Tasks
- [ ] Automate publishing KDS to npm triggered by tagging a new KDS version
- [ ] Configure and turn on dependabot for Kolibri's `develop` and `release-v*`, and Studio's `unstable` 
- [ ] Present the new workflow to the team (@MisRob)
[...]

PR #541, opened 2024-02-07
GitHub Actions workflow to publish a new release on npm

MisRob:
## Description

Adds a GitHub Actions workflow that publishes a new release or pre-release on npm.

Also fixes package.json version.

#### Issue addressed

Part of https://github.com/learningequality/kolibri-design-system/issues/532

## Reviewer guidance

- [ ] Does code makes sense?
- [ ] Are there some other steps that'd be useful to have implemented?

We will need to merge it and then I will make a new release to test it for this repository. I tested the same script in my playground repository here https://github.com/MisRob/hello-npm/blob/main/.github/workflows/npm-publish.yml and it went well https://www.npmjs.com/package/@robosovam/hello-npm?activeTab=versions.

## Comments

I considered exploring having a package.json version increment guard, but actually the action will fail because npm doesn't allow publishing the same version twice as can be seen here https://github.com/MisRob/hello-npm/actions/runs/7799031860/job/21268937399, where I released v1.0.2, but didn't update package.json's v1.0.1.

Screenshot from 2024-02-07 08-33-51

So if I forgot, then I will get notified by this action failure and create a new patch with the correct package.json version. We have the version increment in the checklist now though, so hopefully this won't happen often. 

## Changelog

- [#541]
  - **Description:** Add a GitHub Actions workflow to publish a new release on npm
  - **Products impact:** none
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/532
  - **Components:** -
  - **Breaking:** -
  - **Impacts a11y:** -
  - **Guidance:** -

[#541]: https://github.com/learningequality/kolibri-design-system/pull/541

## Testing checklist

## Reviewer guidance

- [ ] ~Are there tests for this change?~
- [ ] ~Are all UI components LTR and RTL compliant (if applicable)?~

## After review


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "kolibri-design-system",
-  "version": "v1.3.0",
+  "version": "3.0.0",
```
MisRob, line 3: I also removed `v` because I had an impression it shouldn't be there, now that we're going to install KDS as

```
"dependencies": {
  "kolibri-design-system": "^3.0.0",
}
```

https://docs.npmjs.com/about-semantic-versioning#using-semantic-versioning-to-specify-update-types-your-package-can-accept

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,20 @@
+name: Publish (pre)release to npm
+
+on:
+  release:
+    # include release and pre-release, exclude release draft
+    types: [released,prereleased]
```
rtibbles: I think it would be better to just subscribe to the published type, rather than these two, as noted here: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

The release and prereleased types will not trigger if we draft a release and then later publish it.

package.json
rtibbles: Yes, this seems correct!

Might be good to create a small utility at some point to check that the tag matches the version, like we do for Kolibri python versions.

rtibbles (changes requested):
I think we should be subscribing to the published type. Everything else seems correct.

.github/workflows/npm-publish.yml
MisRob: Ah I misread this part for `published`

> "or draft of a release was published"

It sounded to me that saving a draft would trigger the action, which wasn't meant like that. Thank you, will fix.

MisRob:
Thank you @rtibbles, updated

rtibbles (approved):

PR #549, opened 2024-02-18
kselect refactor

### LINKED ISSUE #523: Refactor KSelect: remove unused code from KeenUI components
**Refactoring [KSelect](https://github.com/learningequality/kolibri-design-system/tree/main/lib/KSelect) by removing the unused code from KeenUI components.**

Currently, there are two files `KeenUiSelect` and `KeenUiSelectOption` that have vendored Keen code. The `KSelect` component is built using both of these components. It has more customization and limits `KeenUiSelect` API to what we really need.

We want to get rid of `KeenUiSelect` and `KeenUiSelectOption` components by moving their code right into `KSelect` or still have these components by renaming them i.e. `KeenUiSelect` to `KUiSelect` if needed. The main goal of this task is to locate and remove the piece of code from both these `KeenUi` components that we don't use. We could remove all the behavior related to props that we don't expose on the `KSelect` component, for example.

As a result future debugging, fixing, or adding new features will be much simpler because we won't need to dig through many lines of code and APIs that are not used at all.

---

Refer to this [conversation](https://learningequality.slack.com/archives/C03S2EN192A/p1705079469489539?thread_ts=1704638375.616619&cid=C03S2EN192A) for more details.

Jaspreet-singh-1032:
closes #523 

- moved `KeenUiSelect` to `KSelect` and removed unused code
- renamed `KeenUiSelectOption` to `KSelectOption`

Tested it by linking `kds` with Kolibri
https://www.loom.com/share/66f72fc4572641a6a46c09373d319ad8?sid=6034a517-8088-4b4d-9645-8200761b5cb2

## Testing checklist

## Reviewer guidance

## After review


MisRob:
Thanks @Jaspreet-singh-1032. I will review. It may take a while - lots of work here!

MisRob:
@Jaspreet-singh-1032 The conflicts here will be caused by our rebranding work and I think it'd be simplest not to do much about it right now because there may be some further updates. I think we will just prefer updates from this PR after the review and then before merging I will make sure it's rebranded, so in this regard there won't be any action necessary from your part. Thank you for your patience.

lib/KSelect/KeenUiSelect.vue
```diff
@@ -1,1207 +0,0 @@
...
-      <div v-if="hasFeedback" class="ui-select-feedback">
-        <div v-if="showError" class="ui-select-feedback-text">
-          <slot name="error">
-            {{ error }}
-          </slot>
-        </div>
-
-        <div v-else-if="showHelp" class="ui-select-feedback-text">
-          <slot name="help">
-            {{ help }}
```
ozer550, line 183: Even tho we are not passing in the help prop, I think it would be good to have it in for the future! Also, error prop requires this piece so would be good to have it I think.

lib/KSelect/index.vue
```diff
@@ -60,123 +167,606 @@
...
     },
     model: {
       event: 'change',
     },
     props: {
-      /**
-       * Object currently selected
-       */
       value: {
         type: Object,
```
ozer550: This can be type: [String, Number, Object, Array], any specific reason we are only allowing object?

lib/KSelect/index.vue
```diff
@@ -60,123 +167,606 @@
...
+      hasLabel() {
+        return Boolean(this.label) || Boolean(this.$slots.default);
+      },
+
+      hasFloatingLabel() {
+        return this.hasLabel && this.floatingLabel;
+      },
+
+      isLabelInline() {
+        return this.selection.length === 0 && !this.isActive;
```
ozer550, line 367: I see we are using this.selection instead of prop this.value, even tho they are the same thing would recommend not replacing this.value for clarity. Any reason for doing this?

lib/KSelect/index.vue
```diff
@@ -60,123 +167,606 @@
...
+        return this.invalid && (Boolean(this.invalidText) || Boolean(this.$slots.error));
+      },
+
+      filteredOptions() {
+        return this.options.filter((option, index) => {
+          return this.defaultFilter(option, index);
+        });
+      },
+
+      displayText() {
```
ozer550, line 388: I see some of the computed properties are extremely reduced like this one. Just a note that we should be cautious while doing this as it may break some unknown things unknowingly!

lib/KSelect/index.vue
```diff
@@ -60,123 +167,606 @@
...
+        // need to computed the submittedValue
+        if (!this.name || !this.selection) {
+          return;
+        }
+
+        if (Array.isArray(this.selection)) {
+          return this.selection.map(option => option[this.keys.value] || option).join(',');
+        }
+
+        return this.selection[this.keys.value] || this.selection;
```
ozer550, line 424: same thing here, using this.selection and this.value should be consistent! probably  ` return this.value[this.keys.value] || this.value;` would be better. I see more of those cases in this computed property.

lib/KSelect/index.vue
```diff
@@ -60,123 +167,606 @@
...
+        }
+      },
+
+      // Checks if option is highlighted
+      isOptionHighlighted(option) {
+        return looseEqual(this.highlightedOption, option);
+      },
+
+      isOptionSelected(option) {
+        return looseEqual(this.selection, option);
```
ozer550, line 714: Logic reduced warning!

ozer550 (commented):
While validating one of the props value, I want to share some of the concerns:
1. Logic of computed properties are reduced which may be dangerous and may cause bugs unknowingly.
2. Inconsistent use of this.selection and this.value, even tho they are same thing we should prioritize using what was already used naming convention, if there's some specific reason we are doing this would love to know!

lib/KSelect/index.vue
Jaspreet-singh-1032: Hi @ozer550, In the current [KSelect](https://github.com/learningequality/kolibri-design-system/blob/develop/lib/KSelect/index.vue#L79), It is defined as object. So, I keep it same. And wherever this component is used, the value is passed as object as per I checked.

lib/KSelect/index.vue
Jaspreet-singh-1032: `multiple` prop is not exposed from `KSelect`. So, I removed the logic handling of this prop. Because I think it was never used. [here](https://github.com/learningequality/kolibri-design-system/blob/develop/lib/KSelect/KeenUiSelect.vue#L396)

lib/KSelect/index.vue
Jaspreet-singh-1032: The same reason of `multiple` prop is here.

lib/KSelect/KeenUiSelect.vue
Jaspreet-singh-1032: Yes, makes sense. :+1:

lib/KSelect/index.vue
Jaspreet-singh-1032: I will get back to you on this. I think there was a reason. maybe for the initial value. But I will test it again and let you know.

lib/KSelect/index.vue
ozer550: yes! and this is my concern cause even tho we are not passing in the `multiple` prop directly from `Kselect` its still being used interconnectedly with other props and computed functions. from looking into the code it seems like this prop is used to indicate whether the select component can accept multiple selected options or not. Even tho this is not our current use case there is a high probability it might be needed in near future! looking it from the perspective if we keep this prop we not only make other computed properties secure but also become a little more future proof according to me! maybe @MisRob can guide us a bit more over this?

lib/KSelect/index.vue
ozer550: This makes sense! but we need to be sure that the places where this prop is being used its expected to be an Object!

lib/KSelect/index.vue
MisRob: Overall the goal was to get rid of unused logic and I think lots of APIs can be removed but it's possible that some important APIs will stay even though we don't use them right now. I would feel fine about keeping `multiple` if it feels safer. It's true this is one of the core features that may be handy.

For me, this is learning that as we continue with similar refactors, it would be good to spend some time with auditing APIs in more detail and agree beforehand what should stay and what should go. Another observation is that it'd be good to (1) move the code untouched from file to file in one commit and then (2) have a separate commit for each API removed so we can easily track dependencies and motivations for a particular API removal. I will make sure to incorporate this as suggestions to issues to come.

Thanks both, @ozer550 for your thoughtful observations and helping with review, and @Jaspreet-singh-1032 for your patience with the process and jumping with courage into a project with quite entangled tasks. This is a pioneer PR that will also help greatly with figuring out how to best approach other components.

lib/KSelect/index.vue
```diff
@@ -60,123 +167,606 @@
...
+
+      hasFeedback() {
+        return Boolean(this.invalidText) || Boolean(this.$slots.error);
+      },
+
+      showError() {
+        return this.invalid && (Boolean(this.invalidText) || Boolean(this.$slots.error));
+      },
+
+      filteredOptions() {
```
ozer550, line 382: Seems like the logic for returning all the options without filtering them is missing? Do we always intend to do filtering using default method?

lib/KSelect/index.vue
```diff
@@ -60,123 +167,606 @@
...
+      displayText() {
+        return this.selection ? this.selection[this.keys.label] || this.selection : '';
+      },
+
+      hasDisplayText() {
+        return Boolean(this.displayText.length);
+      },
+
+      hasNoResults() {
+        if (this.query.length === 0) {
```
ozer550, line 406: loading prop may be required here to know if things are still loading or not? Is it ok to omit this?

lib/KSelect/index.vue
Jaspreet-singh-1032: Hi @ozer550, The default value of `disableFilter` and `filter` props are `false` and `null`. And they are not exposed from `KSelect`. That's why I removed this logic.

lib/KSelect/index.vue
Jaspreet-singh-1032: Its default value is also `false` and not exposed from `KSelect`. So, It also does not have any effect.

Jaspreet-singh-1032:
Hi @ozer550 @MisRob, thanks for the review. I am writing down the things that need to be fixed.
- Add the `help` prop back
- Remove `this.selection` and use `this.value`

Let me know if you want to add anything else.

MisRob:
Hi @Jaspreet-singh-1032, thank you. I think @ozer550 was going to do some testing in Kolibri. After that, I will give the PR the final look. I think it'd be fine to resolve @ozer550's first feedback already if you'd like to, since it seems there's agreement on it.

MisRob:
@Jaspreet-singh-1032 We have started QA. Now it would be good to resolve the feedback from @ozer550, if you can, because after the QA, I'd like to rebase this on top of the latest work and resolve conflicts related to the rebrand.

lib/KSelect/index.vue
Jaspreet-singh-1032: Hi @ozer550, I remembered the reason for using `this.selection` here. This variable is currently available in the [Kselect/index.vue](https://github.com/learningequality/kolibri-design-system/blob/develop/lib/KSelect/index.vue#L155) file. This holds the currently selected dropdown option and passes it to the `value` prop of `KeenUiSelect` component. So, whenever the dropdown option is selected, we emit the event and then the value of `selection` gets updated. The latest `selection` value is passed to the `KeenUiSelect` component and it displays it on the frontend. 

Since we have merged the code of `KeenUiSelect` and `KSelect` components we don't have a parent component to provide the currently selected option as `value` prop. We can't change the prop value that's why I added the `selection` state and am using it to display the currently selected option. `selection` initially holds the value of `value` prop.

Let me know if you have any other ideas.

lib/KSelect/index.vue
```diff
@@ -60,123 +173,614 @@
...
     components: {
-      UiSelect,
+      UiIcon,
+      KSelectOption,
     },
     model: {
       event: 'change',
     },
     props: {
-      /**
```
MisRob: Many props docstrings were lost in the process, could you add them back? They will be used for auto-generation of the bottom section of the KDS documentation page for `KSelect`, such as for the currently released version

Screenshot from 2024-04-15 14-44-07

MisRob (commented):
Hi @Jaspreet-singh-1032, apologies it took me such a long time and thanks one more time @ozer550 for providing first round of feedback.

I've just did my best to review as well, focusing on going through the previous `KSelect` API and checking that everything that needed to be moved there from `KeenUiSelect` is indeed there. I haven't noticed any issues, as far as I can say, and for the rest of it I think we can rely on linters and thorough QA :-)

I left one note in regards to docstrings, and I also think that `multiple` and `multipleDelimiter` logic needs to be returned as @ozer550 asked for some time ago.

---

I believe the work on this has been pretty demanding and thanks again. I learned a lot and I think the strategy for all upcoming refactors like this could be to (1) just rename and move the logic while not deleting any APIs, (2) determine and list APIs to be removed, if any (or alternatively document them if we decide to keep them), (3) do whatever was planned in the previous step gradually. If you had some reflections on how to make work and review smooth for issues like this, we'd be very glad to hear from you.

MisRob:
@Jaspreet-singh-1032 Good news :) Our QA team hasn't noticed any problems with selects in Kolibri, except of just one that is not related to this work. Great work! I believe we'll be close to merge after the feedback I posted earlier is resolved.

Jaspreet-singh-1032:
The `multiple` prop functionality is now working.

image

MisRob:
Great, thanks so much, @Jaspreet-singh-1032. I will take it from here and resolve the conflict related to the rebranding because I think it's best to have some more background for that.

MisRob (approved):
I resolved conflicts in style and tested one more time on the documentation page as well as at a few places in Kolibri. All looking good to me, thanks everyone.

Jaspreet-singh-1032:
Thanks @MisRob

PR #552, opened 2024-02-20
Klistwithoverflow component

AlexVelezLl:
## Description
Introduces a new component `Klistwithoverflow` that renders an horizontal list that automatically renders a "show more" component if any items overflow the container.

#### Issue addressed
Addresses #556 

### Before/after screenshots

https://github.com/learningequality/kolibri-design-system/assets/51239030/b7b3031f-489f-496e-bb04-2093fd6a3cbe

## Changelog

- [#552]
  - **Description:** New `KListWithOverflow` component.
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/556, https://github.com/learningequality/studio/issues/3423, https://github.com/learningequality/kolibri/issues/11923.
  - **Breaking:** no.
  - **Impacts a11y:** no.
  - **Guidance:** -.

[#552]: https://github.com/learningequality/kolibri-design-system/pull/552

- [#552]
  - **Description:** New `useKResponsiveElement` private composable, `KResponsiveElementMixin` translated to this composable.
  - **Products impact:** -.
  - **Addresses:** -.
  - **Components:** -.
  - **Breaking:** no.
  - **Impacts a11y:** no.
  - **Guidance:** -.

[#552]: https://github.com/learningequality/kolibri-design-system/pull/552

## Steps to test

1. Run the development server.
2. Go to http://localhost:4000/klistwithoverflow.
3. Resize the window or change examples containers widths, and check that the list is responsive.

1. Leave the list of items as `flex-wrap: nowrap` and `overflow: visible` so that all the items are on a single line without taking more height and without showing the overflow bar.
2. To control which items are seen and which are not, we set its  visibility to `hidden` or `visible`, this allows us to know the width of all the items without the user seeing them.
3. Based on the widths of the items, I fill the available space, and set its visibility depending on whether they fit or not.
4. We accumulate the widths of the visible items and set a max-width to the list so that it takes only the required space and does not affect the positioning of the more button.

## Testing checklist

## Reviewer guidance

## After review


lib/composables/_useKResponsiveElement.js
```diff
@@ -0,0 +1,34 @@
...
+    elementWidth.value = clientWidth || 0;
+    elementHeight.value = clientHeight || 0;
+  }
+
+  onMounted(() => {
+    updateEl();
+    _resizeSensor = new ResizeSensor(instance.proxy.$el, updateEl);
+  });
+
+  onUnmounted(() => {
```
MisRob: Since we're accessing the element in the reactive logic, I think that using `onBeforeUnmount` could be more suitable

docs/pages/klistwithoverflow.vue
```diff
@@ -0,0 +1,177 @@
+<template>
+
+  <DocsPageTemplate apiDocs>
+    <DocsPageSection title="Overview" anchor="#overview">
+      <p>
+        The <code>KListWithOverflow</code> component is useful when we want a horizontal list that is responsive and adaptable to the container size. The list only shows the items that fit on the container element. When items exceed the available space, the component seamlessly integrates a "see more" button to show additional items.
+      </p>
+      <p>
+        This component is renderless, so it needs an array of items, slots to render the visible items and separators on the list, and a slot to render the "see more" button.
```
MisRob: I think there's no need for this paragraph with technical details in the overview part and rather just leave examples to demonstrate this.

marcellamaki (commented):
I have only done an initial read through of this, @AlexVelezLl, and will do a more thorough review tomorrow with more comments and notes, but this looks like really great work! 🎉 I'm excited :)

MisRob:
Hi @AlexVelezLl, overall the interface is looking great and flexible, wonderful work on research and implementation. This will be very helpful for many components across all products.

I will submit a full review later on.

One thing I'd like to suggest at the outset is that we don't use `ResizeSensor` in `useKResponsiveElement`. `KResponsiveElement`'s usage is discouraged due to performance issues https://design-system.learningequality.org/kresponsiveelement/#guidelines. From what I know, I think that copying the same logic to `useKResponsiveElement` (and subsequently to each instance of `KListWithOverflow` which will eventually be used very frequently, e.g. in many cards on one page) would likely cause performance bottlenecks.

Since we won't need to support IE anymore, I think we could try to use [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver)? Or you're welcome to research other alternatives. If you were going to use the `ResizeObserver`, please note that using a single `ResizeObserver` instance is recommended to observe multiple elements ([source](https://github.com/WICG/resize-observer/issues/59#issuecomment-408098151)).

AlexVelezLl:
Thank you @MisRob! Yes, I've worked before with `ResizeObserver` but it has 96.7% browser support, so I saw the way `KBreadcrumbs` accomplished its responsiveness and realized about the `ResizeSensor` and thought it was the standard way to do it.

So, I'll use the `ResizeObserver` but will add a check to add backwards compatibility for older browsers and avoid it if its undefined. In the end in the first render we would still have the responsiveness, we just lose the reactivity when some element is resized (which are fewer use cases).

A solution if issues arise due to this behavior (not recalculating the overflowItems when item resizing in older browser) could be to hide the entire list and only render the "more button", which would allow it to remain functional.

MisRob:
@AlexVelezLl We don't need to support all browsers. There are actually plans to remove IE11 specific logic from KDS, so it'd be best not to be adding support for it or other browsers that are not needed. Please see [the Kolibri browserlist](https://github.com/learningequality/kolibri/blob/develop/packages/browserslist-config-kolibri/index.js) and if all these browsers support `ResizeObserver`, let's not use `ResizeSensor` at all (and rather plan to remove it completely from the codebase at some point)

MisRob:
> I saw the way KBreadcrumbs accomplished its responsiveness and realized about the ResizeSensor and thought it was the standard way to do it.

It used to be, yes. The removal of IE11 support is a pretty recent update that will take effect in Kolibri 0.17. Generally, Kolibri has the most strict requirements in our ecosystem so its browserlist is the thing to watch for.

MisRob:
One thing to watch for in this period of transition is not to use this logic in Kolibri 0.16, for example not to refactor any components to use it. I don't think we will do that :) But will make sure to mention it in the release notes.

AlexVelezLl:
Thanks @MisRob, I just updated the PR with the ResizeObserver.

> Please see [the Kolibri browserlist](https://github.com/learningequality/kolibri/blob/develop/packages/browserslist-config-kolibri/index.js) and if all these browsers support ResizeObserver, let's not use ResizeSensor at all (and rather plan to remove it completely from the codebase at some point).

Yes, it seems like not all Kolibri's supported browsers support `ResizeObserver`. I just added [this check](https://github.com/learningequality/kolibri-design-system/pull/552/commits/501e37c63aa182ed538d5ffcad93f036f04695e6#diff-49138948e2456f25d84040434c546535a347a7db98100dc0e3971abea469bfd4R5) to avoid any null exception. This way we just dont provide the reactivity when the element is resized if the browser dont have support for `ResizeObserver`.

AlexVelezLl:
> One thing to watch for in this period of transition is not to use this logic in Kolibri 0.16

I was planning to use it for the languageSwitcher in the Kolibri's setup wizard https://github.com/learningequality/kolibri/issues/11923. Do you mean not including this in future 0.16 patches?

MisRob:
> I was planning to use it for the languageSwitcher in the Kolibri's setup wizard https://github.com/learningequality/kolibri/issues/11923. Do you mean not including this in future 0.16 patches?

I don't see a milestone yet for https://github.com/learningequality/kolibri/issues/11923 yet. So to sum it up:

(1) If the issue is planned for 0.16 => we need IE11 support in this component now and will remove it later
(2) If the issue is planned for 0.17 => no need for IE11 support

(2) will is simpler, but if it's important to have it in 0.16, we can add IE fallback now as you suggested and remove it later.

I'd leave you decide together with @marcellamaki about the milestone for the issue.

lib/KListWithOverflow.vue
```diff
@@ -0,0 +1,236 @@
...
+      ></slot>
+    </div>
+  </div>
+
+</template>
+
+
+<script>
+
+  import throttle from 'lodash/throttle';
```
MisRob, line 46: There's `frame-throttle` available in KDS and already used in a few components that have something to do with resizing. Since reading client width of elements [forces layout/reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a#element-apis) and we do it in this component frequently, do you think that `frame-throttle` would be useful instead of `throttle` in the context of this component? I'm not entirely sure, let's discuss. I also welcome any ideas how to further optimize in general.

docs/pages/klistwithoverflow.vue
```diff
@@ -0,0 +1,177 @@
...
+      </p>
+    </DocsPageSection>
+    <DocsPageSection title="Example" anchor="#example">
```
MisRob, line 9: Could you add code previews via `DocsShowCode` for both the basic usage example as well the dividers example? This component has a bit complex API so I'd be good to demonstrate how to use it. Examples would ideally have both template part and a script part with items in `data()`. Also, I'd like to ask you don't copy `getItems` to `DocsShowCode` but rather demonstrate with objects `{ label, icon }` literally so the usage is apparent at first glance.

AlexVelezLl:
Got it! I dont think its super urgent tho

MisRob (commented):
@AlexVelezLl Overall, this is great work, very clear and flexible component design and I appreciate your comments. I am not going to pretend I fathomed all the calculations in detail but as a whole it makes sense to me and I didn't notice anything suspicious. Posting a few comments and a question or two.

MisRob:
And one changelog update request, please let's communicate it's a private composable here and that products impact is none rather than new API since it's not public.

Screenshot from 2024-02-28 21-26-02

MisRob:
I think that's all from my first review round

lib/composables/_useKResponsiveElement.js
AlexVelezLl: Oooh, that makes a lot of sense!

lib/KListWithOverflow.vue
AlexVelezLl: I ran some tests with `frame-throttle` and it seems that it excecute the method much more times than if we throttled it with 100ms, in both scenarios, when I manually resize the window, and when there is an animation:

using  `frame-throttle`:

https://github.com/learningequality/kolibri-design-system/assets/51239030/425ef33e-2869-405d-8211-e7d9a58b558a


using  `lodash/throttle` with 100ms

https://github.com/learningequality/kolibri-design-system/assets/51239030/bf8cfbac-62b6-40e0-85e2-57e876eae29d

MisRob: @AlexVelezLl The main benefit of `frame-throttle` is not lesser frequency but rather that it can choose the moment to repaint in a more clever way, see https://github.com/learningequality/kolibri-design-system/pull/469#discussion_r1358711535. The metrics here would be 'how many force repaints' rather than 'how often'. It seems particularly useful to prevent from extra repaints that will be happening to some extent via this component. I don't know exactly how these two techniques compare in this particular scenario and think we'd need to measure performance for larger amounts of instances to be certain. So it's a minor note based largely on intuition. You're welcome to take any direction ;) Thanks for looking into it.

AlexVelezLl: Thanks! I find this topic very interesting, and I have run some other tests isolating the component and testing with multiple instances, and I don't see much difference in the repainting of both. Here are some tests:

Using `frame-throttle`:

https://github.com/learningequality/kolibri-design-system/assets/51239030/dbe2d1e6-d63e-41b0-a2cb-bc825f37a9fc

Using `lodash/throttle` with 100ms:

https://github.com/learningequality/kolibri-design-system/assets/51239030/884be1f0-d280-45cb-8eb6-ffb6ba7d65ca

But the script time taken for the same example differs drastically, and using `frame-throttle` is 3x expensive:

Using `frame-throttle`:

image

Using `lodash/throttle` with 100ms:

image

I see that, indeed, the painting takes less time, but compared to the difference in the script time, I think is ok preferring the `lodash/throttle` for this example.

AlexVelezLl: (At the end I just ended adding the `frame-throttle` in the `_useKResponsiveElement` as I think that makes sense and we follow the pattern of the `useKWindowDimensions` (as both do almost the same thing), but kept the regular throttle with 100ms in the elementWidth watch in KListWithOverflow) :open_hands:.

AlexVelezLl:
Thank you @MisRob! Comments addressed :).

lib/KListWithOverflow.vue
MisRob: @AlexVelezLl thank you for looking into it in detail

MisRob (approved):
Thanks @AlexVelezLl, good work here.

PR #554, opened 2024-02-22
issue-#10491 Allow moving with arrow keys over radio buttons in firefox

muditchoudhary:
## Description

The PR fixed the accessibility issue of arrow keys not working for moving over radio buttons in Firefox.

#### Issue addressed

Addresses #*PR# HERE*

### Before/after screenshots

https://github.com/learningequality/kolibri-design-system/assets/74391865/c6777cba-6a4e-4808-9428-8eb9d95b3df5

## Changelog

- [#PR no]

[#PR no]: PR link

## Steps to test

1. Step 1
2. Step 2
3. ...

## Testing checklist

## Reviewer guidance

## After review

### DIFF
--- a/docs/pages/playground.vue
+++ b/docs/pages/playground.vue
@@ -1,32 +1,140 @@
 <template>
 
-  <!--
-    Playground: A private page for components development
-    *****************************************************
-    Place a component you're working on here and see it
-    live on http://localhost:4000/playground
-
-    Please do not commit your local updates of this file.
-  -->
   <div style="padding: 24px">
-    <!-- 
-      Example: Uncomment lines bellow, change something
-      in lib/KBreadcrumbs.vue and see the updates reflected
-      on this page
-    -->
-    <!-- <KBreadcrumbs
-      :items="[
-        { text: 'Global Digital Library', link: { path: '#' } },
-        { text: 'English', link: { path: '#' } },
-        { text: 'Level 2 ', link: { path: '#' } },
-      ]"
-    /> -->
+    <div style="border: 2px solid blue;">
+      <button>
+        Before
+      </button>
+      <div class="fieldset" style="border: 2px solid red;">
+        <label class="fieldset-label">Default landing page</label>
+        <KRadioButton
+          ref="landingPageRadio0"
+          label="Learn page"
+          :value="landingPageChoices.LEARN"
+          :currentValue="landingPage"
+          @input="(option) => handleLandingPageChange(option, 0)"
+        />
+        <KRadioButton
+          ref="landingPageRadio1"
+          label="Sign-in page"
+          :value="landingPageChoices.SIGN_IN"
+          :currentValue="landingPage"
+          @input="(option) => handleLandingPageChange(option, 1)"
+        />
+        <div class="fieldset" style="margin-left: 32px; border: 2px solid green;">
+          <KRadioButton
+            ref="landingPageRadio3"
+            label="Allow users to explore resources without signing in"
+            :value="SignInPageOptions.ALLOW_GUEST_ACCESS"
+            :currentValue="signInPageOption"
+            :disabled="disableSignInPageOptions"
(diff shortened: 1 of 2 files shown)


muditchoudhary:
I have used the same logic as is in KTabsList component for arrow keys movement.

The arrow keys are working fine. 

But I cannot do the tab logic. I have thought about this logic:
- For tab keys, as the user presses the tab on the first radio button on a group I transfer the focus to the next component with the help of a $ref which I have given to the next component.
- Even if we have more than one radio group on the page, like I have in this draft PR. We can just give different refs to the different next and previous components after each radio group.
- **Issue with this logic** But I stuck with radio buttons which are in the middle. How can I disable the tab on them? If the user presses the tab on a radio button in the middle with a mouse and then presses the tab, it should not work. But currently, it works because I could not disable the tab on the middle radio buttons.
- **Preview Below**

https://github.com/learningequality/kolibri-design-system/assets/74391865/908f147c-1732-4702-a7db-f0f8118beb20

- I need your help with the tab logic.

## My Thoughts
In the original issue which occurred in Devices/Settings, We can add this logic on the setting file. But I think there are many different places where Radio Buttons are used we have to add the same code to each file. Which is not a good idea. I think we can create a different component that contains all the radio buttons of a single group.

What do you think?

About the tabs, I think as the arrow keys are working. The users now do not need to use the tabs in Firefox too and even if they use it it will not break the code or logic. So I think we can skip it if we don't have a proper logic for the tab after discussion. Let's see.

So currently as per https://www.w3.org/WAI/ARIA/apg/patterns/radio/#keyboardinteraction guidelines the
- Arrow keys are working
- Space is working
- Tabs not working.

MisRob:
Hi @muditchoudhary, thank you for looking into this!

In regards to troubling TAB key, can I understand right that the problem is that under certain circumstances, when you click TAB, the focus moves to the next radio button rather than outside the group, right?

muditchoudhary:
@MisRob Yes, that's correct.

MisRob:
Some general references that came to mind when I read your comment
  - In `KTabsList`, I used `tabindex` to control TAB sequence. It's value was dynamically set to 0 or -1.
  - I tried some issues you described, as far as I could understand, on this example https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio/ Combination of mouse and TAB key seems to work well. They too use `tabindex`.

Could some of this information help?

MisRob:
And also in regards to

> But I think there are many different places where Radio Buttons are used we have to add the same code to each file. Which is not a good idea. I think we can create a different component that contains all the radio buttons of a single group.

I agree we wouldn't want to implement this over and over again and it'd be best to have something like 
`KRadioButtonGroup` component. Let me think about it and after the issues you mentioned are resolved, let's chat.

muditchoudhary:
> Could some of this information help?

I'll check it out in the morning.

muditchoudhary:
Hi @MisRob The tab issue has been fixed now. Thanks for telling about `tabindex`


https://github.com/learningequality/kolibri-design-system/assets/74391865/d6af887e-a715-484a-bd1a-88bab8eee093

muditchoudhary:
I will try to create `KRadioButtonGroup` component next. Do you have any suggestions or docs that tell what things to keep in mind when creating a KDS component?

MisRob:
Thanks @muditchoudhary. I think that `KRadioButton`'s usage would ideally look something like

```
<KRadioButtonGroup>
  <KRadioButton>
  <KRadioButton>
</KRadioButtonGroup>
```

It would be good to check few places in Kolibri to see if this will do for our use cases.

[This page in KDS docs](https://github.com/learningequality/kolibri-design-system/blob/develop/dev_docs/03_how_to_update_library.md) will help with creating a new component, even though you probably already know most of that.

It'd be good to have some tests for it.

muditchoudhary:
Okay understood!

MisRob:
Closing in favor of https://github.com/learningequality/kolibri-design-system/pull/650

PR #572, opened 2024-03-07
Add `aria-checked` property to KCheckbox

nucleogenesis:
As I was working in Kolibri I went to double check a11y stuff on the checkboxes and discovered that we should have this property applied because it can be used to indicate the same three states our KCheckbox does. What we call "indeterminate" they call "mixed", so we pass that value to the parameter easily because we have the props with all of the information we need.

- [x] Update changelog


MisRob (approved):
Thanks @nucleogenesis

MisRob:
I'll take care of the changelog conflict and merge it in a moment @nucleogenesis so that we can have this available in the release that I will create later today :)

PR #573, opened 2024-03-12
Update client width to use getBoundingClientRect

AlexVelezLl:
## Description

This PR replaces the use of `clientWidth` to use `getBoundingClientRect().width` to have a more accurate measurement of the width of the KListWithOverflow items. It also recalculates the width of the  moreButton element on each overflow calculation to make sure this value hasnt changed.

### Before/after screenshots
In very specific cases, using `clientWith` can lead to cases like this due to the lack of decimals:

image

Now this is fixed:

image

## Changelog

- [#573]
  - **Description:** More precise calculation of list with in KListWithOverflow.
  - **Addresses:** -.
  - **Breaking:** no.
  - **Impacts a11y:** no.
  - **Guidance:** -.

[#573]: https://github.com/learningequality/kolibri-design-system/pull/573

## Testing checklist

- [ x If there are any front-end changes, before/after screenshots are included

## Reviewer guidance

## After review

### DIFF
--- a/lib/KListWithOverflow.vue
+++ b/lib/KListWithOverflow.vue
@@ -103,17 +103,12 @@
       this.$watch('elementWidth', this.throttledSetOverflowItems);
     },
     methods: {
-      getWidth(element) {
+      getSize(element) {
         if (!element) {
-          return 0;
+          return { width: 0, height: 0 };
         }
-        return element.getBoundingClientRect().width;
-      },
-      getHeigth(element) {
-        if (!element) {
-          return 0;
-        }
-        return element.getBoundingClientRect().height;
+        const { width, height } = element.getBoundingClientRect();
+        return { width, height };
       },
       /**
        * Sets the items that overflow the list, the visibility of the more button,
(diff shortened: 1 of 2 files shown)


lib/KListWithOverflow.vue
```diff
@@ -103,26 +103,43 @@
       this.$watch('elementWidth', this.throttledSetOverflowItems);
     },
     methods: {
+      getWidth(element) {
+        if (!element) {
+          return 0;
+        }
+        return element.getBoundingClientRect().width;
```
rtibbles: One thing to flag, getBoundingClientRect forces a reflow, so this can cause performance issues. More info here: https://toruskit.com/blog/how-to-get-element-bounds-without-reflow/

AlexVelezLl: Thanks Richard, I will batch all getWidth calls before changing anything in the DOM

AlexVelezLl:
Hi @rtibbles I have batched all calls to getBoundingClientRect before setting the items visibilities. Hope this helps to optimize the performance and avoids reflows.

rtibbles:
It will still be quite a few, as we are calling it for each element in the list, it seems?

I wonder if it might be better to take the Intersection Observer approach here, and require that Kolibri integrate the Intersection Observer polyfill for unsupported browsers.

I think we can merge this as is, but maybe we file a follow up issue for a potential performance improvement.

AlexVelezLl:
Oh I had read [here](https://kellegous.com/j/2013/01/26/layout-performance/) that if there is no changes to the DOM between calls to properties that needs the layout to be recalculated, then the layout is no recalculated as is already up to date. What causes reflow is when we invalidate the layout by changing the DOM nodes, so I think batching all these calls at the beginning would do the work.

This is a similar approach they do in [their example](https://gist.github.com/kellegous/4646014#file-layout-js-L52) with multiple different DOM nodes in a list.

rtibbles:
Oh, great - thanks for doing the due dilligence here! No need for the follow up, which would add some complexity.

rtibbles (approved):

Issue #570, opened 2024-03-06
KModal does not shrink when the content has a smaller height


AlexVelezLl:
## Product
Studio

## Expected behavior

The height of the KModal should dynamically adjust to the height of the content it renders.

## Actual behavior

If for some reason the content of the KModal is expanded after its first render, the KModal is expanded accordingly, but if it is then shrinked, the KModal remains with the height of the expanded content, and this can lead to cases like this:

If initially we have a componente with a description like this: 

![image](https://github.com/learningequality/kolibri-design-system/assets/51239030/ffb5421c-2211-46b2-9288-f83bee0833c7)

But then we want to delete it, we get this view: 

![image](https://github.com/learningequality/kolibri-design-system/assets/51239030/b9ad3ef4-35d0-4f82-80c3-de359994fa5a)

Cases like this occur in several places where the height of some element varies while the modal is displayed.

## Steps to reproduce the issue
1. Go to Studio and checkout the branch `studio-usability-enhancements`
2. Edit the content nodes of a channel
3. Edit the title and description of a node

## Environment
- OS: Ubuntu
- Browser version: Chrome Version 122.0.6261.94 (Official Build) (64-bit)

shrudge:
hey @AlexVelezLl great notice, but I think you listed this issue in the wrong Repo, correct me if am i wrong tho, I am really new to open source so..

AlexVelezLl:
Hi @shrudge! No, the issue is in the correct repository. KDS is responsible for providing the components that will be used in other Kolibri products, you can learn more about KDS in the [repo readme](https://github.com/learningequality/kolibri-design-system?tab=readme-ov-file#what-is-kolibri-design-system).

The issue is about the [KModal component](https://github.com/learningequality/kolibri-design-system/blob/d653942b300b48f60fef37783e84c8f6057c1a61/lib/KModal.vue), here in KDS, probably it has something to do with the [updateContentSectionStyle](https://github.com/learningequality/kolibri-design-system/blob/d653942b300b48f60fef37783e84c8f6057c1a61/lib/KModal.vue#L285) method.

galovdev:
@AlexVelezLl Can I get this issue assigned?

AlexVelezLl:
Hi @galovdev! Sure! Let me know if you have any question :)

MisRob:
Closed by https://github.com/learningequality/kolibri-design-system/pull/590

PR #585, opened 2024-03-20
Implement dynamic modal resizing based on content changes

galovdev:
## Description

The modal now resizes like the description input every time the input size changes.

#### Issue addressed

Addresses #570

### Before/after screenshots

Before:

image

image

After:

https://github.com/learningequality/kolibri-design-system/assets/107520089/30326279-724d-495e-8197-f781d9086a7b

## Changelog

- [#585]
  - **Description:** Modal now resizes dynamically to adjust size based on content changes.
  - **Addresses:**  https://github.com/learningequality/kolibri-design-system/issues/570
  - **Components:** KModal.vue.
  - **Breaking:** Will this change break something in a consumer? No
  - **Impacts a11y:** Does this change improve a11y or adds new features that can be used to improve it? N/A
  - **Guidance:** Why and how to introduce this update to a consumer? Consumers need to ensure the modal content is wrapped correctly for the resizing logic to work effectively. No additional steps required for integration.

[#585]: [PR link](https://github.com/learningequality/kolibri-design-system/pull/585)

## Steps to test

1. Go to studio 
2. Studio channels
3. Edit title and description

## Testing checklist

- [ X] Contributor has fully tested the PR manually
- [ X] If there are any front-end changes, before/after screenshots are included

## After review


lib/KModal.vue
```diff
@@ -261,6 +261,8 @@
       // if modal contains KSelect, special classes & styles will be applied
       const kSelectCheck = document.querySelector('div.modal div.ui-select');
       this.containsKSelect = !!kSelectCheck;
+      this.observeContentChanges();
```
AlexVelezLl: Hmm I dont think we need to apply more listeners, since with the current implementation we already catch the changes, and run the `updateContentSectionStyle` method on every update. If we remove these listeners the solution should continue working.

lib/KModal.vue
```diff
@@ -310,12 +312,35 @@
...
             // add a vertical scrollbar if content doesn't fit
-            if (this.$refs.content.scrollHeight > this.$refs.content.clientHeight) {
-              this.$refs.content.style.overflowY = 'auto';
-            }
+            this.$refs.content.style.overflowY =
+              this.$refs.content.scrollHeight > this.$refs.content.clientHeight ? 'auto' : 'hidden';
           }
         }
+        if (this.$refs.content) {
+          this.$refs.content.style.height = 'auto';
```
AlexVelezLl: What does the trick here is that we are updating the height of the content to `auto`, and after that we are reading its `scrollHeight`. The problem is that this causes a reflow, and it can cause performance issues considering how often we excecute this method (you can read more about this [here](https://kellegous.com/j/2013/01/26/layout-performance/)).

But this has made me question why do we need to set the height of the content programatically?, I think that a `height: auto` attribute should do the work, which would be basically the same thing we are doing here (setting the heiht to auto and after set its scrollHeight), but without causing reflow.

~~Let me check with the team if there is a reason why we need to set the height programatically.~~

Reviewing the code it seems that this is to avoid these little jumps on the modal height when there are super minor changes in the height of the content (You can see these jumps in your example video when you focus the textArea). Soo another solution that comes to my mind could be to calculate the content childrens height (all but the absolute and sticky positioned) and have this meassure to compare and update the contenstHeight [here](https://github.com/learningequality/kolibri-design-system/blob/59a56e025175d0effda591d1e122408a678711d3/lib/KModal.vue#L287).

If any other solution comes to your mind, you can suggest it too.

PR #590, opened 2024-03-25
Modal Adjust to Input Size

galovdev:
## Description

The modal now resizes like the description input every time the input size changes.

#### Issue addressed

Addresses #570

### Before/after screenshots

Before:

image

image

After:

https://github.com/learningequality/kolibri-design-system/assets/107520089/551fcaa9-8961-4c3b-8605-80105c684d94

## Changelog

- [#590]
  - **Description:** Modal now resizes dynamically to adjust size based on content changes.
  - **Addresses:**  https://github.com/learningequality/kolibri-design-system/issues/570
  - **Components:** KModal.
  - **Breaking:** No
  - **Impacts a11y:** -
  - **Guidance:** Consumers need to ensure the modal content is wrapped correctly for the resizing logic to work effectively. No additional steps required for integration.

[#590]: https://github.com/learningequality/kolibri-design-system/pull/590/

## Steps to test

1. Go to studio 
2. Studio channels
3. Edit title and description

## Testing checklist

- [ X] Contributor has fully tested the PR manually
- [ X] If there are any front-end changes, before/after screenshots are included

## After review

### DIFF
--- a/lib/KModal.vue
+++ b/lib/KModal.vue
@@ -99,7 +99,7 @@
 </template>
   
 
-  <script>
+<script>
 
   import debounce from 'lodash/debounce';
   import useKResponsiveWindow from './composables/useKResponsiveWindow';
@@ -349,7 +349,7 @@
 </script>
 
 
-  <style lang="scss" scoped>
+<style lang="scss" scoped>
 
   @import './styles/definitions';
 
--- a/lib/KModal.vue
+++ b/lib/KModal.vue
@@ -299,17 +299,6 @@
         }
       }, 50),
 
-      attachInputListeners() {
-        const textInputs = this.$refs.content.querySelectorAll('input[type="text"], textarea');
-        const handleInput = () => {
-          // Call adjustContentHeight to manage height changes
-          this.adjustContentHeight();
-        };
-        textInputs.forEach(input => {
-          input.removeEventListener('input', handleInput); // Ensure no duplicates
-          input.addEventListener('input', handleInput);
-        });
-      },
       adjustContentHeight() {
         // Manage content height adjustment here
         this.$refs.content.style.height = 'auto';
@@ -449,4 +438,5 @@
   }
 
 </style>
+
   
--- a/lib/KModal.vue
+++ b/lib/KModal.vue
@@ -206,9 +206,8 @@
         };
       },
       modalWidth() {
-        if (this.size === 'small') return '300px';
-        if (this.size === 'medium') return '450px';
-        if (this.size === 'large') return '100%';
+        if (this.size === SIZE_SM) return '300px';
+        if (this.size === SIZE_MD) return '450px';
         return `${this.size}px`;
       },
       maxModalWidth() {
--- a/lib/KModal.vue
+++ b/lib/KModal.vue
@@ -97,7 +97,7 @@
   </transition>
 
 </template>
-  
+
 
 <script>
 
@@ -208,6 +208,7 @@
       modalWidth() {
         if (this.size === SIZE_SM) return '300px';
         if (this.size === SIZE_MD) return '450px';
+        if (this.size === SIZE_LG) return '100%';
         return `${this.size}px`;
       },
       maxModalWidth() {
@@ -216,7 +217,6 @@
       contentSectionMaxHeight() {
         return {
           'max-height': `${this.maxContentHeight}px`,
-          height: 'auto',
         };
       },
     },
@@ -292,8 +292,9 @@
           // (otherwise KSelect will be trapped inside modal if KSelect is opened a second time)
           if (this.$refs.content.clientHeight !== 0 && !this.containsKSelect) {
             // add a vertical scrollbar if content doesn't fit
-            this.$refs.content.style.overflowY =
-              this.$refs.content.scrollHeight > this.$refs.content.clientHeight ? 'auto' : 'hidden';
+            if (this.$refs.content.scrollHeight > this.$refs.content.clientHeight) {
+              this.$refs.content.style.overflowY = 'auto';
+            }
           }
         }
       }, 50),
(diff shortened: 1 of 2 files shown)


lib/KModal.vue
```diff
@@ -95,23 +95,25 @@
...
-
+  
+  </template>
+  
+  
+  <script>
+  
   import debounce from 'lodash/debounce';
+
+  import KResponsiveWindowMixin from './KResponsiveWindowMixin';
```
AlexVelezLl: It seems that we have included this dependency, but we are not using it. We can remove it

lib/KModal.vue
```diff
@@ -285,40 +288,54 @@
...
+          if (this.contentHeight !== this.$refs.content.clientHeight) {
+              this.$refs.content.style.height = `${this.contentHeight}px`;
+          }
+      }
+  }, 50),
+      attachInputListeners() {
+        const textInputs = this.$refs.content.querySelectorAll('input[type="text"], textarea');
+        textInputs.forEach(input => {
+          input.addEventListener('input', () => {
+            this.$refs.content.style.height = 'auto';
```
AlexVelezLl: The reason we should not add listeners to the inputs is that inputs are not just the ones that will grow or shrink, this can be anything that are conditionally rendered within the modal, so this solution will not work for modals that have a form that could (conditionally) have warnings and errors, just to give an example. 

Check this video I recorded running this PR:

https://github.com/learningequality/kolibri-design-system/assets/51239030/ff67e929-c573-4291-a05e-0bf4ac78e9a3

This is because this solution is not general for all use cases. Also, we should try to avoid setting the content height to auto, as this could cause too many reflows.

We need to think a way to track the actual height of the content and update accordingly only when there are more than 8px of difference.

lib/KModal.vue
```diff
@@ -285,40 +288,54 @@
...
+              // add a vertical scrollbar if content doesn't fit
+              this.$refs.content.style.overflowY =
+                  this.$refs.content.scrollHeight > this.$refs.content.clientHeight ? 'auto' : 'hidden';
           }
-        }
-      }, 50),
+  
+          // we only update when necessary
+          if (this.contentHeight !== this.$refs.content.clientHeight) {
+              this.$refs.content.style.height = `${this.contentHeight}px`;
```
AlexVelezLl: We should avoid setting the height directly to the content element, when we set the contentHeight attribute in [line 298](https://github.com/learningequality/kolibri-design-system/pull/590/files#diff-11808f59cbc23e186d8a068a6b633df2d71c7319b7d93597a03eaedaa9597b51R298) we already are setting the height of the template.

AlexVelezLl (commented):
Hi! @galovdev This is looking better! But we still need to find a way how to update the height of the content, I will check with the team for further Ideas. Also, I see there are some linting errors, you can fix them with `yarn run lint-fix` and also we need to update the [CHANGELOG.md](https://github.com/LearningEquality/kolibri-design-system/blob/develop/CHANGELOG.md) file with the update of this PR.

AlexVelezLl:
Hi @galovdev I talked with the team, and we are going to take the approach of keep it simple, and better set an `auto` height, and try to avoid the pixels toggling like the one on KTexbox hover. You are free to try to fix the KTextbox pixel toggling on hover if you want, but for now just setting the content height to auto works fine.

Also, we will need to retarget the PR to the `develop` branch. You can find how to retarget PRs [here](https://kolibri-dev.readthedocs.io/en/develop/howtos/rebasing_a_pull_request.html).

galovdev:
> We should avoid setting the height directly to the content element, when we set the contentHeight attribute in [line 298](https://github.com/learningequality/kolibri-design-system/pull/590/files#diff-11808f59cbc23e186d8a068a6b633df2d71c7319b7d93597a03eaedaa9597b51R298) we already are setting the height of the template.

Give me till monday to make the changes and re'target the branch, ill be out this weekend.

AlexVelezLl:
Hi! @galovdev  How are you? Have you been able to advance in the PR? I'm not sure if maybe someway I misdirected you, but we are expecting in this PR that we clean up the JS code related to the `this.contentHeight` calculation, so it will be better to delete [this line](https://github.com/learningequality/kolibri-design-system/pull/590/files#diff-11808f59cbc23e186d8a068a6b633df2d71c7319b7d93597a03eaedaa9597b51R231), [this method](https://github.com/learningequality/kolibri-design-system/pull/590/files#diff-11808f59cbc23e186d8a068a6b633df2d71c7319b7d93597a03eaedaa9597b51R324), and [this logic](https://github.com/learningequality/kolibri-design-system/pull/590/files#diff-11808f59cbc23e186d8a068a6b633df2d71c7319b7d93597a03eaedaa9597b51R292), and any code that became unused because of the changes.

Also we will need to retarget the PR to the `develop` branch. So first you can rebase your branch onto the develop branch as indicated [here](https://kolibri-dev.readthedocs.io/en/develop/howtos/rebasing_a_pull_request.html), and after that you can change the base branch here in github.

Let me know if you have any question :).

lib/KModal.vue
```diff
@@ -216,19 +208,15 @@
       modalWidth() {
         if (this.size === SIZE_SM) return '300px';
         if (this.size === SIZE_MD) return '450px';
-        if (this.size === SIZE_LG) return '100%';
```
AlexVelezLl: It looks like we accidentally deleted this line and we need to bring it back.

AlexVelezLl:
Hi @galovdev! How are you? Have you been able to advance in the PR? Let us know if there is something we can help with.

AlexVelezLl (approved):
Thank you @galovdev! Code looks good, and [QA has confirmed that there are no regressions](https://github.com/learningequality/kolibri/pull/12209#pullrequestreview-2086934416). Its good to go! :tada:.

Issue #584, opened 2024-03-20
KTextbox: Support a "clearable" prop to show a clear button


AlexVelezLl:
## Product

For now in Studio to achieve some specs for the Studio Usability Enhancements project.

## Desired behavior

Add a "clear" button in the KTextbox component that only appears if there is non-empty text. When the user presses the clear button, the text should be deleted.

![image](https://github.com/learningequality/kolibri-design-system/assets/51239030/8e570c36-6968-4a6a-9ca7-68cbf7cbcb71)

Vuetify currently supports this using a `clearable` prop on its `VTextField`, so we can replicate this prop name.

## The Value Add

We give the user an option to easily delete all the text they have written.

practicatto:
Hi! I'm new to this project and eager to start contributing. I've been looking for first issues and came across this issue. I believe I can provide the needed assistance to help solve it. Would it be possible for me to be assigned to this issue? :)

ozer550:
> Hi! I'm new to this project and eager to start contributing. I've been looking for first issues and came across this issue. I believe I can provide the needed assistance to help solve it. Would it be possible for me to be assigned to this issue? :)

Hi @practicatto! Welcome and thanks for volunteering for this issue, I will assign this to you! feel free to reach out for any help while working with the issue.

MisRob:
Closed by https://github.com/learningequality/kolibri-design-system/pull/591

PR #591, opened 2024-03-26
KTextbox: Support a "clearable" prop to show a clear button

practicatto:
## Description

KTextbox support to show clearable button when clearable prop to true

#### Issue addressed

Addresses #584

## Changelog

- [#591]
  - **Description:** Added a clearable prop to KTextbox
  - **Products impact:** -
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/584
  - **Components:** KTextbox
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -

[#591]: https://github.com/learningequality/kolibri-design-system/pull/591

## Testing checklist

## Reviewer guidance

## After review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ Changelog is rather internal in nature. See release notes for the public overvie
 
 ## Version 4.x.x (`release-v4` branch)
 
+- [#591]
+  - **Description:** Added a clearable prop to KTextbox
+  - **Products impact:** -
+  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/584
+  - **Components:** `KTextbox`
+  - **Breaking:** no
+  - **Impacts a11y:** no
+  - **Guidance:** -
+
+[#591]: https://github.com/learningequality/kolibri-design-system/pull/591
+
 - [#582]
   - **Description:** Upgrade popper.js from 1.14.6 to 1.16.1
   - **Products impact:** -
--- a/docs/pages/playground.vue
+++ b/docs/pages/playground.vue
@@ -24,9 +24,6 @@
 
     <!-- Play around with your component here: -->
 
-
-
-
   </div>
 
 </template>
--- a/lib/KTextbox/__tests__/KTextbox.spec.js
+++ b/lib/KTextbox/__tests__/KTextbox.spec.js
@@ -155,4 +155,60 @@ describe('KTextbox component', () => {
       expect(wrapper.emitted().input).toBeTruthy();
     });
   });
+  describe('KTextbox with clearable', () => {
+    it('should have the clear button when clearable is true and there is text in the input', async () => {
+      const wrapper = mount(KTextbox, {
+        propsData: {
+          clearable: true,
+        },
+      });
+      const input = wrapper.find('input');
+      input.element.value = 'test';
+      input.trigger('input');
+      await wrapper.vm.$nextTick();
+      const clearButton = wrapper.find('[data-test="clearIcon"]');
+      expect(clearButton.exists()).toBeTruthy();
+    });
+    it('should not show the clear button when clearable is true and there is no text in the input', async () => {
+      const wrapper = mount(KTextbox, {
+        propsData: {
+          clearable: true,
+        },
+      });
+
+      const input = wrapper.find('input');
+      input.element.value = '';
+      input.trigger('input');
+      await wrapper.vm.$nextTick();
+      expect(wrapper.find('[data-test="clearIcon"]').exists()).toBeFalsy();
+    });
+
+    it('should not show the clear button when clearable is false and there is text in the input', async () => {
+      const wrapper = mount(KTextbox, {
+        propsData: {
+          clearable: false,
+        },
+      });
+      const input = wrapper.find('input');
+      input.element.value = 'test';
+      input.trigger('input');
+      await wrapper.vm.$nextTick();
+      expect(wrapper.find('[data-test="clearIcon"]').exists()).toBeFalsy();
+    });
+    it('should clear the input when clear button is clicked', async () => {
+      const wrapper = mount(KTextbox, {
+        propsData: {
+          clearable: true,
+        },
+      });
+      const input = wrapper.find('input');
+      input.element.value = 'test';
+      input.trigger('input');
+      await wrapper.vm.$nextTick();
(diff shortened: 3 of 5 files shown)


lib/KTextbox/__tests__/KTextbox.spec.js
```diff
@@ -156,3 +156,35 @@ describe('KTextbox component', () => {
...
+    const wrapper = mount(KTextbox, {
+      propsData: {
+        clearable: true,
+      },
+    });
+    const input = wrapper.find('input');
+    input.element.value = 'test';
+    input.trigger('input');
+    await wrapper.vm.$nextTick();
+    const clearButton = wrapper.findComponent({ name: 'KIconButton' });
```
AlexVelezLl: I think we can use a data-test property here, in case that in the future we have another KIconButton within the KTextbox.

lib/KTextbox/__tests__/KTextbox.spec.js
```diff
@@ -156,3 +156,35 @@ describe('KTextbox component', () => {
     });
   });
 });
+
+describe('KTextbox with clearable', () => {
```
AlexVelezLl: We can also check that clearable button is not present when there is no text input, and when we set the `clearable` prop to false

docs/pages/playground.vue
```diff
@@ -24,16 +24,15 @@
...
-
-
   </div>
 
 </template>
 
 
 <script>
 
+  //import KTextbox from '../../lib/KTextbox/index.vue';
```
AlexVelezLl: We can remove these lines

lib/keen/UiTextbox.vue
```diff
@@ -224,6 +229,18 @@
     },
 
     computed: {
+      
+      showClearButton() {
+        return this.clearable && this.value;
```
AlexVelezLl: We should also validate that we dont show the button when the textbox is disabled: 
image

lib/keen/UiTextbox.vue
```diff
@@ -320,6 +337,20 @@
         this.$emit('input', value);
       },
 
+      clearText() {
+        this.updateValue(this.initialValue);
```
AlexVelezLl: We should update the value to `""`, because the initialValue can be any string that is first passed to the component, and we will be resetting the text to that string.

lib/keen/UiTextbox.vue
```diff
@@ -320,6 +337,20 @@
...
+        this.refreshSize()
```
AlexVelezLl: After updating the value to `""` we can refactor this to autosize the textArea:

```javascript
this.$nextTick(() => {
  this.refreshSize();
  this.focus();
});
```
The `refreshSize` before the this.$nextTick has no effect since the value change had not taken effect yet, until the next tick.
With the `refreshSize` method, we dont need to check the existence of the textarea ref, since in the `refreshSize` method only updates the size when there is a textArea, and `autosize` is true.

lib/keen/UiTextbox.vue
```diff
@@ -550,7 +601,11 @@
     transition: border 0.1s ease;
   }
 
-    .ui-textbox-textarea {
+  .ui-textbox-textarea {
```
AlexVelezLl, line 604: We need to be careful about who is rendering the border. As `.ui-textbox-label` also renders a border, and we can end up in a double-border status.

For example, look at this small diference in the border bottom on hover:
Before:
image

After:
image


And when `invalid` and `invalidText` are set:

Before:
image


After:
image

This is a small diference because `.ui-input-content` is now rendering a border that is probably not necessary

AlexVelezLl (changes requested):
Hi! @practicatto! Thanks for your work here, this is looking great! But there are some little things we can improve =)

AlexVelezLl:
Also, a check is failing because we need to update the [Changelog](https://github.com/learningequality/kolibri-design-system/blob/release-v4/CHANGELOG.md) file with a new entry that is the same as the one you put in the PR descripciton.

lib/keen/UiTextbox.vue
practicatto: I believe now it's working as expected 

disabled-example

lib/keen/UiTextbox.vue
practicatto: I believe this is now working as expected :) 

a really long invalid text

CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ Changelog is rather internal in nature. See release notes for the public overvie
...
 
+- [#591]
+  - **Description:** Added a clearable prop to KTextbox
+  - **Products impact:** -
+  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/584
+  - **Components:** `KTextbox`
+  - **Breaking:** no
+  - **Impacts a11y:** no
+  - **Guidance:** -
+
```
AlexVelezLl, line 15: Here we need to add the link of the reference:

`[#591]: https://github.com/learningequality/kolibri-design-system/pull/591`

lib/keen/UiTextbox.vue
```diff
@@ -17,68 +18,68 @@
...
+            :rows="rows"
+            :style="isActive ? { borderBottomColor: $themeTokens.primaryDark } : {}"
+            :tabindex="tabindex"
+            @blur="onBlur"
+            @change="onChange"
+            @focus="onFocus"
+            @input="updateValue($event.target.value)"
+            @keydown.enter="onKeydownEnter" @keydown="onKeydown"
+          ></textarea>
+          <KIconButton
```
AlexVelezLl, line 74: There is one last issue here, the clear button is not accessible, we should add an `aria-label` property, but we can not add it inside KDS because we dont have the i18n set up here, but the implementers are the ones responsible for that. So, a workaround for this is to pass the aria-label as prop. So we can make that the KTextbox has a new prop `clearAriaLabel` and we can pass it to the UiTextbox and set the `:ariaLabel="clearAriaLabel"` to this `KIconButton`

AlexVelezLl (commented):
THanks! @practicatto Just one last thing and we are ready to go!

AlexVelezLl (approved):
Thanks for your work @practicatto! Code makes sense and everything seems to be working ok. Thank you!

practicatto:
> Thanks for your work @practicatto! Code makes sense and everything seems to be working ok. Thank you!

Thank you for your guidance!! :)

Issue #595, opened 2024-03-28
Make KTooltip's width configurable 


BabyElias:
🌱 **Are you new to the codebase? Welcome and please see [the contributing guidelines](https://github.com/learningequality/kolibri-design-system/blob/release-v4/CONTRIBUTING.md).**

## Desired behavior

`KTooltip`'s maximum width can be configured, ideally via standard CSS `max-width`:

```
<KIcon ref="icon" icon="info" />
<KTooltip
  :style="{ 'max-width': '100px' }"
  reference="icon"
  :refs="$refs"
>
  Tooltip text tooltip text tooltip text tooltip text tooltip text
</KTooltip>
```

or alternatively (2) via a new `maxWidth` property

```
<KIcon ref="icon" icon="info" />
<KTooltip
  :maxWidth="100px"
  reference="icon"
  :refs="$refs"
>
  Tooltip text tooltip text tooltip text tooltip text tooltip text
</KTooltip>
```

where (1) is preferred and (2) acceptable (if (1) it's too much of struggle due to the use of the 3rd party library under the `KTooltip`'s hood)

Note that setting maximum width needs to be optional. `KTooltip` should maintain its current behavior of adjusting its size to fit its content if maximum width is not set.

## Current behavior

Setting max width

```
<KIcon ref="icon" icon="info" />
<KTooltip
  :style="{ 'max-width': '100px' }"
  reference="icon"
  :refs="$refs"
>
  Tooltip text tooltip text tooltip text tooltip text tooltip text
</KTooltip>
```

has no effect

![Screenshot from 2024-04-06 07-02-05](https://github.com/learningequality/kolibri-design-system/assets/13509191/107299f2-0bc6-41a4-926f-2180ce4201ef)

## Steps to reproduce

Run KDS development server and paste the code snippets from above to [the playground page](https://github.com/learningequality/kolibri-design-system/blob/release-v4/dev_docs/03_how_to_update_library.md#1-preview-updates-in-the-kolibri-design-system-environment).

## Impact

This is blocking migrating from Vuetify to KDS in [Studio](https://github.com/learningequality/studio) in some places where we need to be able to configure the maximum width, for example in [HelpTooltip.vue](https://github.com/learningequality/studio/blob/9ed0725411d013a9c08688e601a2083893f7fdfe/contentcuration/contentcuration/frontend/shared/views/HelpTooltip.vue).

| Vuetify | KDS |
| -------- | ------- |
[...]

PR #611, opened 2024-04-09
made KTooltip's width configurable

satyamkale27:
## Description
This PR configures the maxWidth of KTooltip,  If maxWidthis not passed it will maintain its current behavior of adjusting its size to fit its content

#### Issue addressed
#595

Addresses #*PR# HERE*

###  screenshots
If maxWidth is not passed, KTooltip will maintain its current behavior of adjusting its size to fit its content.
before

If maxWidth passes.
after

## Changelog

- [#PR no]
  - **Description:** This PR configures the maxWidthof KTooltip,  If maxWidth is not passed it will maintain its current behavior of adjusting its size to fit its content
  -  **Products impact:** 
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/595
  - **Components:** -
  - **Breaking:** no
  - **Impacts a11y:**  -
  - **Guidance:** -
  

[#PR no]: PR link

## Steps to test

1. Step 1 
  Run the KDS By  ``` yarn dev ```
2. Step 2
Past this code snippet to the playground.vue file in following path  docs/pages /playground.vue

```vue
<KIcon ref="icon" icon="info" />
    <KTooltip
      :maxWidth="'100px'"
      reference="icon"
      :refs="$refs"
    >
      Tooltip text tooltip text tooltip text tooltip text tooltip text
    </KTooltip>
```

3. Past this in your browser   http://localhost:4000/playground  and see the changes

##### Briefly describe how this works 
If the maxWidth is passed,the maxWidth will get applied. but if maxWidth is not passed it will continue its current  behavior of adjusting its size to fit its content.

## Testing checklist

## Reviewer guidance

- [X] Is the code clean and well-commented?

## After review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ Changelog is rather internal in nature. See release notes for the public overvie
 
 ## Version 3.x.x (`release-v3` branch)
 
+- [#611]
+  - **Description:** Adds a new `maxWidth` prop
+  -  **Products impact:** new API 
+  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/595
+  - **Components:** KTooltip
+  - **Breaking:** no
+  - **Impacts a11y:** no
+  - **Guidance:** -
+
+[#611]: https://github.com/learningequality/kolibri-design-system/pull/611
+
 - [#612]
   - **Description:** Prevent KListWithOverflow hidden elements from taking up space on the screen.
   - **Products impact:** bugfix.
--- a/lib/KTooltip/index.vue
+++ b/lib/KTooltip/index.vue
@@ -11,7 +11,7 @@
     <div
       dir="auto"
       class="k-tooltip"
-      :style="{ backgroundColor: $themeTokens.text, color: $themeTokens.textInverted, maxWidth: maxWidth ? maxWidth : 'calc(100vw - 10px)', }"
+      :style="{ backgroundColor: $themeTokens.text, color: $themeTokens.textInverted, ...styles }"
     >
       <!-- If text prop is provided, display the text -->
       <template v-if="text"> 
@@ -75,9 +75,9 @@
         type: String,
         default: null,
       },
-      maxWidth: {
-        type: String,
-        default: null,
+      styles: {
+        type: Object,
+        default: () => ({}),
       },
     },
     data() {
@@ -129,6 +129,7 @@
     z-index: 24;
     width: max-content;
     min-width: 75px;
+    max-width: calc(100vw - 10px);
     padding: 8px;
     font-size: 12px;
     font-weight: normal;
--- a/lib/KTooltip/index.vue
+++ b/lib/KTooltip/index.vue
@@ -11,7 +11,7 @@
     <div
       dir="auto"
       class="k-tooltip"
-      :style="{ backgroundColor: $themeTokens.text, color: $themeTokens.textInverted }"
+      :style="{ backgroundColor: $themeTokens.text, color: $themeTokens.textInverted, maxWidth: maxWidth ? maxWidth : 'calc(100vw - 10px)', }"
     >
       <!-- If text prop is provided, display the text -->
       <template v-if="text"> 
@@ -75,6 +75,10 @@
         type: String,
         default: null,
       },
+      maxWidth: {
+        type: String,
+        default: null,
+      },
     },
     data() {
       return {
@@ -125,7 +129,6 @@
     z-index: 24;
     width: max-content;
     min-width: 75px;
-    max-width: calc(100vw - 10px);
     padding: 8px;
     font-size: 12px;
     font-weight: normal;
--- a/lib/KTooltip/index.vue
+++ b/lib/KTooltip/index.vue
@@ -11,7 +11,7 @@
     <div
       dir="auto"
       class="k-tooltip"
-      :style="{ backgroundColor: $themeTokens.text, color: $themeTokens.textInverted, maxWidth: maxWidth ? maxWidth : 'calc(100vw - 10px)', }"
+      :style="{ backgroundColor: $themeTokens.text, color: $themeTokens.textInverted, ...styles }"
     >
       <!-- If text prop is provided, display the text -->
       <template v-if="text"> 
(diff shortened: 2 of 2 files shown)


satyamkale27:
@MisRob please take a look.

MisRob:
Thanks @satyamkale27, as mentioned in the issue I think this implementation is an option if the preferred API

```
<KTooltip
  :style="{ 'max-width': '100px' }"
  reference="icon"
  :refs="$refs"
>
  Tooltip text tooltip text tooltip text tooltip text tooltip text
</KTooltip>
```

is not feasible for some reason.

Could you elaborate if you've explored it and what were the blockers?

satyamkale27:
@MisRob I have not yet explored an alternative approach to it, but I could. correct me if I am wrong, means we don't want max-width set inline like here in this code snippet 
`<KIcon ref="icon" icon="info" /> <KTooltip :maxWidth="'100px'" reference="icon" :refs="$refs" > Tooltip text tooltip text tooltip text tooltip text tooltip text </KTooltip>`   please explain me in brief so i can work on it.

MisRob:
Thank you @satyamkale27. The difference between the two options is that in the first one I suggest we explore first, max width is set via the standard CSS `max-width`, for example with `:style="{ 'max-width': '100px' }"`. The question here is - if you copy the code from the "Current behavior" of https://github.com/learningequality/kolibri-design-system/issues/595 to the playground, is there something we can do inside `KTooltip` to make it work so that we don't even need the new `maxWidth` property?

If it doesn't seem to be feasible, we could add a new prop `maxWidth` for sure, but if it is possible, I would say that using usual style is preferable.

satyamkale27:
thanks for explaning I got it now, soon I will make new commit.

satyamkale27:
@MisRob I made a new commit which will allow max-width to run but there is a problem in it, i cant use the style prop because "style" is a reserved attribute and cannot be used as component prop.  so i used styles to pass prop.

MisRob:
Hey, @satyamkale27, thanks for following up.

> "style" is a reserved attribute and cannot be used as component prop.

Ah yes, I didn't mean you use it as a component prop, but rather as ordinary native inline `style` attribute. Vue syntax

```
<KTooltip
  :style="{ 'max-width': '100px' }"
  reference="icon"
  :refs="$refs"
>
  Tooltip text tooltip text tooltip text tooltip text tooltip text
</KTooltip>
```

will translate into native HTML inline `style="max-width:100px;"` and my suggestions was to explore whether we can simply use it to achieve the desired result rather than adding a new prop. For example by experimenting with adjusting internal styles of the `KTooltip` so that they inherit/respect the root container `style` width settings and/or perhaps with attaching `v-bind:style` on one of the children elements in `KTooltip`.

A simple parallel would be, let's say we'd have two nested divs

```
<div style="max-width:100px;">
  <div class="some-styles">
    Tooltip text tooltip text tooltip text tooltip text tooltip text
  </div>
</div>
```

and now the task is to investigate what about the styles applied by `some-styles` class cause that the inner div doesn't accept the outer div's max width settings.

Note that this doesn't necessarily reflect `KTooltip`s internals exactly, I'm just using it as a simplified example. I think that if we understood more what's happening in this regard, it will help to inform next steps.

Please let me know if this helped to clarify.

If you won't find a way to achieve so in this manner, we can return to the `maxWidth` option, I just think it'd be useful to see whether this direction is feasible or not and why.

MisRob:
@satyamkale27 

A simple parallel would be, let's say we'd have two nested divs

```
<div style="max-width:100px;">
  <div class="some-styles">
    Tooltip text tooltip text tooltip text tooltip text tooltip text
  </div>
</div>
```

and now the task is to investigate what about the styles applied by `some-styles` class cause that the internal div doesn't accept the external div's max width settings.

Note that this doesn't necessarily reflect `KTooltip`s internals exactly, I'm just using it as a simplified example. I think that if we understood more what's happening in this regard, it will help to inform next steps.

satyamkale27:
@MisRob thanks for guidance, I tried many methods to work it without passing prop but I didn't find any way to achieve it.

satyamkale27:
@MisRob thanks for your guidance, I have explored many methods to achive without passing prop , but none of them worked.

MisRob:
Alright, thanks @satyamkale27, I will take a look

MisRob:
@satyamkale27 I played around with it too a bit and like you, I haven't really discovered a way to achieve the first option in a reasonable time frame due to Popper. Thanks for your time investigating. Let's go with the `maxWidth` prop then.

satyamkale27:
@MisRob yeh sure, I will make a new commit soon.

MisRob:
Thanks @satyamkale27, I think that just reverting to https://github.com/learningequality/kolibri-design-system/pull/611/commits/d82513c6ad9de6420add5770d6035f20ad52ef40 will work

satyamkale27:
@MisRob I made a revert please take a look.  in this case we can only use **maxWidth**  not other styles, but if we pass **styles** prop as a object we can use other styles like color and so on, also including "max-width"  as i done in previous commit [611](https://github.com/learningequality/kolibri-design-system/pull/611/commits/4d2f7172887be8426ac96ada123e951c8b6d7922),
so If we dont want **styles** as prop I have made revert already which will allow to use **maxWidth**

MisRob:
Thank you @satyamkale27. You're right that `styles` prop would be more general and has a benefit of allowing for other customizations. Right now, I lean towards `maxWidth` since when native Web APIs are not possible to use, we tend to use restricted Vue component API to promote consistent experience tailored to our use-cases. Also specifically for the tooltip, it seems that over the years there haven't been much need for other customizations. So that's what behind the decision, however it's just one viewpoint, and each approach has its pros and cons. In the future, if we find out we need more flexibility, we can consider `styles` prop certainly. For now we can leave the PR as is and I will finish the review next week.

satyamkale27:
@MisRob yeh, thats good thank you.

MisRob (approved):
Thank you, @satyamkale27, all good.

Before merging, I updated the changelog file and also documented the new prop because this section of the documentation page is auto-generated from the in-code docstrings

Screenshot from 2024-04-15 05-39-54

satyamkale27:
Thankyou that's look good.

Issue #528, opened 2024-01-24
KCard Technical Specification


MisRob:
## Summary

`KCard` provides an accessible base card component. Furthermore, it offers the most frequently used card content and thumbnail layouts, as well as other customization options.

It is not meant to be the final card displayed in the Learning Equality Products. Rather, it offers a logic that is common to all of them and makes creation of final card components easier, faster, and consistent. The Products are expected to build their many card components, such as lesson card, resource card, and channel card, by importing and configuring `KCard`, and providing product specific logic to its slots. For illustration, here are some Kolibri cards that will be built on top of `KCard`: 

| Example 1 | Example 2 | Example 3 | Example 4 | Example 5 |
|-- | -- | -- | -- | -- |
| ![ex-1](https://github.com/learningequality/kolibri-design-system/assets/13509191/95786be9-de47-4527-8471-5bbfb8eda03d) |  ![ex-2](https://github.com/learningequality/kolibri-design-system/assets/13509191/0a4d4d11-7569-4d36-b309-c434b3216eb7) | ![ex-3](https://github.com/learningequality/kolibri-design-system/assets/13509191/48dbe59b-6147-430e-94a6-3b65bd9d0a80) | ![ex-4](https://github.com/learningequality/kolibri-design-system/assets/13509191/f06431e8-c696-454b-9d7c-777de450a180) | ![ex-5](https://github.com/learningequality/kolibri-design-system/assets/13509191/023d7ee1-1c12-4227-ba9a-0542f6409018) |

Here for example, `KCard` will provide the `#footer` slot and positions it. In the Kolibri codebase, we will then implement `ResourceCard.vue`, `LessonCard.vue`, etc. Each of them will use `KCard` and provide its `#footer` slot with its own logic such as progress bar, context menu, leave it empty, etc.

## Available layouts for card content and thumbnails

- These are the main layouts that `KCard` will support (labels containing `#` correspond to slots, `layout` and `thumbnailDisplay` are props)

![available-layouts](https://github.com/learningequality/kolibri-design-system/assets/13509191/b2d61748-4388-423a-973a-bb72bc3aa44d)

- Note that except of the title prop/slot, all slots are optional. Therefore, a variety of layouts can be achieved for each `layout` and `thumbnailDisplay` combination. Relatedly, slots are not fixed in their place (for example, see `#title` slot taking place of `#aboveTitle` when `#aboveTitle` is not used). Few examples:

[...]

PR #625, opened 2024-04-23
New component: Initial implementation of `KCard` component

### LINKED ISSUE #529: Implement BaseCard
## Blocks

- #530

## Summary

Implement `BaseCard`, a helper private component that will be used by the public `KCard` component. Please see #528 first. 

## Markup requirements

- The outermost element is `<li>`
- `<a>` (generated from the `<router-link>`) is placed inside the heading element and the title is inside the link
- Contains the heading element and it is placed above all its siblings that will be passed down from `KCard` via the default slot

```html
<li>
  <h[2-6]>
    <a href="/resource">Resource</a>
  </h[2-6]>
  <slot>
</li>
```

This ensures that no matter of how we refactor the parent component `KCard` in the future, during the process we won't mingle new updates with basic a11y criteria related to card list, heading placement, and link that need to be satisfied under all circumstances. In a way, one of the functions of the `BaseCard` is to 'protect' them.

## Style requirements

- Doesn't have set width/height (this will be controlled from `KCard` or card grid)
- Even though the whole card is not wrapped in the link, the whole card area is clickable and has the pointer cursor style
- Has focus ring on focus utilizing our `$coreOutline` (see ["Keyboard navigation"](https://design-system.learningequality.org/accessibility/#keyboard)). Even though the whole card is not wrapped in the link, the focus ring still needs to show around the whole card area.

- Even though the whole area is clickable, text content is still selectable
- Has shadow styles
- Has the same hover style as the one used for the current Kolibri cards

- Has rounded corners

Visit the [Figma designs](https://www.figma.com/file/1tQQ0Sw1yDjzO66hyjLVPM/Kolibri-UX-Fixes?node-id=1868%3A55351&mode=dev) to retrieve some of the related CSS styles.

### Props

| Name | Description | Type | Required | Default |
| - | - | - | - | - |
| `to` | Router-link object that enables user navigation to the intended route. | `Object` | `true` | - |
| `title` | Sets the value of the title text. A simpler alternative to the `title` slot. | `String` | `false` | `null` |
| titleLines | A number of lines a title text is truncated to. | Number | false | 2 |
| `headingLevel` | A level of the heading element (`<h2>`-`<h6>`) wrapping a title text. Supported values: `2-6`. | `Number` | `true` | - |

### Slots

| Name | Description |
| - | - |
| `default` | Contents of the card except its title. |
| `title` | The title contents. It should not contain a heading element. An alternative to the `title` prop. |

### Events

| Name | Description |
| - | - |
| `focus` | Emitted when the card element has received focus. |
| `hover` | Emitted when the mouse pointer enters or leaves the card element |

## Acceptance criteria

- [ ] Conforms fully to the specification above
- [ ] Is not exposed publicly and doesn't have a documentation page
- [ ] Contains validation to check that `title` prop or slot has been used, throwing an error if neither is provided

## Guidance

Using these particular techniques is not required, just a few tips that may help:

### Existing code

In [kolibri/kolibri/plugins/learn/assets/src/views/cards/](https://github.com/learningequality/kolibri/tree/release-v0.16.x/kolibri/plugins/learn/assets/src/views/cards), you can find lots of code for current cards in Kolibri. I wouldn't recommend it when it comes to markup and architecture since our first cards aren't very consistent and accessible (`KCard` is here to solve that). However, it may be useful for pulling out pieces of logic that would still be useful rather then re-implementing it. One example wold be existing styles for `:hover` state in `CardLink`.

### Hover and focus

`:focus-within` may be helpful for implementing focus and hover styles around the whole card area
- [MDN `:focus-within`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within)
- See an example in [Inclusive Components: Cards](https://inclusive-components.design/cards/) ("Affordance" section)

### Selectable textual content

To make textual content selectable within the whole clickable card area, "[Inclusive Components: Cards](https://inclusive-components.design/cards/) ("The redundant click event" section) offers a technique that may serve as inspiration for a Vue based solution:

> _We need to detect how long the user is taking between mousedown and mouseup and suppress the event if it's "likely to be selecting text" territory._

```javascript
const cards = document.querySelectorAll('.card');  
Array.prototype.forEach.call(cards, card => {  
    let down, up, link = card.querySelector('h2 a');
    card.onmousedown = () => down = +new Date();
    card.onmouseup = () => {
        up = +new Date();
        if ((up - down) < 200) {
            link.click();
        }
    }
});
```

### LINKED ISSUE #530: Implement KCard
## Blocked by

- #529

## Summary

Implement `KCard`, an accessible base card component offering the most frequently used card content and thumbnail layouts, as well as other customization options. Please see #528 first, and also note that `KCard` is built on top of the `BaseCard` #529.

## Markup requirements

- Passes all its content via the default slot down to `BaseCard`

_KCard.vue_

```vue
<BaseCard>
  ...
</BaseCard>
```

## Style requirements

- Doesn't have any styles affecting its placement within its parent component (e.g. outer margins)
- Has temporarily set width/height (so that we can preview it in a reasonable manner, but this will be worked on later as part of the card grid component)
- Implements spaces between slots and thumbnail area
- Doesn't assume slots content (e.g. slots have no fixed height)
- Robust content tolerance (very short/very long title, with or without thumbnail, ...) 
- Follows [Figma designs](https://www.figma.com/file/1tQQ0Sw1yDjzO66hyjLVPM/Kolibri-UX-Fixes?node-id=1868%3A55351&mode=dev) as closely as possible

## Architecture requirements

- Is built on top of `BaseCard`
- Uses [`KImg`](https://main--kolibri-design-system.netlify.app/kimg/) for the thumbnail area implementation
  - sets `isDecorative` to `true` on `KImg` (the thumbnail image is decorative)
  - `KCard`'s `thumbnailScaleType` is passed down to `KImg`'s `scaleType`
- Uses [`KTextTruncator`](https://main--kolibri-design-system.netlify.app/ktexttruncator/) for title truncation

### Props

| Name | Description | Type | Required | Default |
| - | - | - | - | - |
| `to` | Router-link object that enables user navigation to the intended route. | `Object` | `true` | - |
| `title` | Sets the value of the title text. A simpler alternative to the `title` slot. | `String` | `false` | `null` |
| `titleLines` | A number of lines a title text is truncated to. | `Number` | `false` | `2` |
| `headingLevel` | A level of the heading element (`<h2>`-`<h6>`) wrapping a title text. Supported values: `2-6`. | `Number` | true | - |
| `layout` | Determines the visual layout of card's content. Can be one of: `'horizontal'`, `'vertical'`. | `String` | `false` | `'horizontal'` |
| `thumbnailDisplay` | Determines whether a thumbnail area is displayed or not, and sets its size. Can be one of: `'none'`, `'small'`, `'large'`. When it's `'small'` or `'large'` and `thumbnailSrc` is empty or invalid, the thumbnail area will be filled with `thumbnailBackgroundColor` acting as a placeholder. | `String` | `false` | `'none'` |
| `thumbnailSrc` | Source URL of a thumbnail image | `String` | `false` | `null` |
| `thumbnailScaleType` | Specifies how an image should be scaled within the thumbnail area. Can be one of  `'centerInside'`, `'contain'`, or `'fitXY'`. See `KImg`'s `scaleType`. | `String` | `false` | `'centerInside'` |

### Slots

| Name | Description |
| - | - |
| `title` | Optional slot section containing the title contents, should not contain a heading element. An alternative to the `title` prop. |
| `aboveTitle` | Places content above the title area. |
| `belowTitle` | Places content below the title area. |
| `footer` | Places content to the footer area. |
| `thumbnailPlaceholder` | When a thumbnail placeholder area is displayed (for example when the image is not available or can't be loaded), you can use this slot to place content to the area. | 

### Events

| Name | Description |
| - | - |
| `focus` | Emitted when the card element has received focus. |
| `hover` | Emitted when the mouse pointer enters or leaves the card element |

## Acceptance criteria

- [ ] Conforms fully to the specification above
- [ ] Supports all layouts in "Available layouts for card content and thumbnails" section of https://github.com/learningequality/kolibri-design-system/issues/528
- [ ] Supports RTL
- [ ] Contains validation to check that `thumbnailScaleType` prop value is a valid scale type (see the same validator in `KImg`) 
- [ ] There is a documentation page for `KCard`
  - has examples illustrating the main content and thumbnail layouts
  - contains guidance on the `headingLevel` to notify developers that they need to choose a correct level based on context
  - emphasizes the need for a unique and descriptive `title` to aid users in searching through lists of links

### LINKED ISSUE #624: KCard Component Enhancements
```[tasklist]
### Tasks
- [x] Add consistence padding around all inner  content area of the` KCard`
- [x] Add  prop` to ` to the router-link i.e passed to `router-link`
- [x] Validate that the heading level is a value between 2 and 6
- [x] Move this validation logic to the BaseCard component if it's applicable to both BaseCard and` KCard.` If the validation is specific to` KCard,` implement it within the `KCard` component. If the validation applies to both BaseCard and` KCard,` move it to the BaseCard
- [x] Change the outermost element of the `KCard `component from `<button>` to` <li>`
- [x] Remove the duplicate link element that wraps the title text within the [`BaseCard `](https://github.com/learningequality/kolibri-design-system/issues/529)component.
- [x] Add a mechanism to allow users to select and copy text content within a clickable area (potentially the entire KCard), even though the entire area is a clickable link.
- [x] Refactor the existing test suite for the` KCard` component to ensure it adequately covers  edge case(s).
- [x] When a user tabs to a KCard and presses Enter, it should open the card's intended action (e.g., navigate to a linked page)
```

AllanOXDi:
## Description
This PR introduces the KCard component  to KDS, It will act  a foundational building block for creating various card types within our product eco system. By serving as a base component for creating different card types like lesson cards, resource cards, and channel cards. 

#### Issue addressed

https://github.com/learningequality/kolibri-design-system/issues/530
https://github.com/learningequality/kolibri-design-system/issues/528
https://github.com/learningequality/kolibri-design-system/issues/529

## Changelog

-  [#625]
  - **Description:** Initial implementation of `KCard` component
  - **Products impact:** New Component
  - **Addresses:** [#530](https://github.com/learningequality/kolibri-design-system/issues/530)
  - **Components:** KCard
  - **Breaking:** No
  - **Impacts a11y:**  Yes

## Steps to test

I suggest looking at these issues 
https://github.com/learningequality/kolibri-design-system/issues/530
https://github.com/learningequality/kolibri-design-system/issues/528
https://github.com/learningequality/kolibri-design-system/issues/529

## Testing checklist

## Reviewer guidance

## After review


AllanOXDi:
I've added a few cards to the playground for @radinamatic  to perform QA testing. Once everything is approved by the QA team, we'll remove the cards before merging the code.

radinamatic:
Overall very good implementation, you followed all the recommended accessibility guidelines 👏🏽 💯 🙂 

One thing that needs to be corrected is the use of the semantic ``header`` and ``footer`` elements inside cards. We should have been more clear in the component specification, but the fact that we call something _card **footer**_ and _card **header**_ does not imply that semantic ``header`` and ``footer`` landmark elements need to be implemented. These are page-level elements and having them (announced) for each of potentially tens of cards on the page would be confusing.

Header | Footer
-- | --
header | footer

radinamatic:
Regarding the responsiveness, I don't know if it's because of Netlify view of the _playground_ (🙃) but I'm wondering if it's how it's supposed to look and behave...? Or maybe we can only properly test this once the cards are in the grid? 🤔 


https://github.com/learningequality/kolibri-design-system/assets/1457929/faa54f4a-6f99-4643-8967-0ad6248af5a0

radinamatic:
The last question I had was regarding the other interactive elements **inside** the card (button/icon to open other options, etc.): that is not in the scope of this PR and is supposed to come in subsequent ones, correct?

AllanOXDi:
> The last question I had was regarding the other interactive elements **inside** the card (button/icon to open other options, etc.): that is not in the scope of this PR and is supposed to come in subsequent ones, correct?


I think it's something we tried and tested here https://github.com/learningequality/studio/pull/4480

AllanOXDi:
> but I'm wondering if it's how it's supposed to look and behave...? Or maybe we can only properly test this once the cards are in the grid? 🤔

I think that's how it should behave for now.

> Regarding the responsiveness, I don't know if it's because of Netlify view of the _playground_ (🙃) but I'm wondering if it's how it's supposed to look and behave...? Or maybe we can only properly test this once the cards are in the grid? 🤔
> 
>  Untitled.Project.mp4

 The Grid will control it's behavior further . I will defer to @MisRob  for more thoughts and guidance on this.

radinamatic:
> I think it's something we tried and tested here [learningequality/studio#4480](https://github.com/learningequality/studio/pull/4480)

I don't recall having tested that, is it live somewhere for me to take a look?

AllanOXDi:
> I don't recall having tested that, is it live somewhere for me to take a look?

Not- it's not yet live for testing as dev is still on going.

MisRob:
Thank you @radinamatic and @AllanOXDi.

Regarding responsiveness - that's something that a grid will control, so yes, it is to be implemented and tested later on.

Regarding buttons in the footer area - `KCard` itself only provides place for them. It is products that will be responsible for adding buttons, context menus, etc. and ensuring it doesn't break a11y. We tried it a bit in a Studio PR as @AllanOXDi mentioned, but that's work in progress and there should be more thorough QA any time we add new cards in a product. After we start seeing some patterns, we may also consider adding some guidance to the documentation or adding features to `KCard` to further support such cases.

radinamatic:
All good then, will wait for the refactor of semantic `header` and` footer` landmark elements here to approve, and looking forward to testing the PRs for follow up steps of the new KCard! 🙂

docs/tableOfContents.js
```diff
@@ -402,6 +402,12 @@ export default [
         title: 'KListWithOverflow',
         isCode: true,
       }),
+      new Page({
+        path: '/KCard',
```
AlexVelezLl: The path needs to be lowercase `/kcard`. Otherwise the nuxt generate script wont catch the page:
image

docs/pages/kcard.vue
```diff
@@ -0,0 +1,359 @@
+<template>
+
+  <DocsPageTemplate apiDocs>
+    <DocsPageSection title="Overview" anchor="#overview">
+      <p>
+        <code>KCard</code>
+        provides an accessible component for managing most frequently card content such as thumbnail,layouts and other customization options
```
AlexVelezLl: We are missing a space here after `thumbnail,`, and a period at the end of the paragraph.

docs/pages/kcard.vue
```diff
@@ -0,0 +1,359 @@
...
+      <p> <code>KCard</code>  serves as a base component for creating different card types such as lesson cards, resource cards, and channel cards. You can configure KCard by adding product-specific content and logic to create distinct card components.</p>
+
+      <p>
+        KCard supports multiple layouts for card content and thumbnails, offering flexibility in design. The available layouts include:.
+      </p>
+
+      <h3>Vertical layout</h3>
+      <p>
+        Vertical
+        <DocsInternalLink href="/KCard#prop:layout">
```
AlexVelezLl: In these props references we need to lowercase the path too.

docs/pages/kcard.vue
```diff
@@ -0,0 +1,359 @@
...
+          <code> thumbnailDisplay</code>
+        </docsinternallink> solely consists of content without any accompanying thumbnail image. This layout is suitable when visual representation is not necessary or when the content itself is sufficiently descriptive or engaging without the need for an image.
+      </p>
+      <DocsShowCode language="html">
+        <div>
+          <KCard
+            :to="{ name: '/' }"
+            title="how to get started with Kolibri"
+            headingLevel="2"
+            layout="verticle"
```
AlexVelezLl: Here we have a typo. This should be `layout="vertical"`

docs/pages/kcard.vue
```diff
@@ -0,0 +1,359 @@
...
+            thumbnailDisplay="none"
+          >
+            <template #title>
```
AlexVelezLl: I didnt understand the role of this #title slot here. If we pass a title slot we dont need to pass the `title` prop? 🤔

docs/pages/kcard.vue
```diff
@@ -0,0 +1,359 @@
...
+            </template>
+            <template #aboveTitle>
```
AlexVelezLl: I dont see this aboveTitle above the title, I see it below :smiling_face_with_tear: :
image

docs/pages/kcard.vue
```diff
@@ -0,0 +1,359 @@
...
+        Vertical
+        <DocsInternalLink href="/KCard#prop:layout">
+          <code>layout</code>
+        </DocsInternalLink> and 
+        <DocsInternalLink href="/KCard#prop:thumbnailDisplay">
+          <code> thumbnailDisplay </code> 
+        </docsinternallink> small
+      </h5>
+      <p>
+        This layout offers a combined view of content and a thumbnail image. This provides users with both textual information
```
AlexVelezLl: I think we missed finishing this paragraph, maybe we wanted to put something like `This provides users with both textual information and visual representation.` or not?

docs/pages/kcard.vue
```diff
@@ -0,0 +1,359 @@
...
+      </p>
+
+      <DocsShow block>
+        <div style="max-width: 300px;">
+          <KCard
+            :to="{ name: '/' }"
+            layout="vertical"
+            thumbnailDisplay="large"
```
AlexVelezLl: Here we are showing an example of `thubmnailDisplay` small, but we are passing the value as "large".

lib/KCard/index.vue
```diff
@@ -0,0 +1,226 @@
...
+        type: Number,
+        required: true,
+        default: 2,
+      },
+      /**
+       * An object containing the route definition for the link.
+       */
+      to: {
+        type: Object,
+        required: true,
```
AlexVelezLl, line 101: IMO `to` prop should not be required, I think we could use cards just to show information or not? 🤔

docs/pages/kcard.vue
```diff
@@ -0,0 +1,359 @@
...
+              </div>
+            </template>
+          </KCard>
+        </div>
+      </DocsShow>
+      <h3>
+        Horizontal Layout
+      </h3>
+      <p>
+        It's important to note that besides the mandatory   <DocsInternalLink href="/KCard#prop:title">
```
AlexVelezLl: I think this paragraph could go in a different section before this section, as this doesnt feel part of the "Horizontal Layout" section.

docs/pages/kcard.vue
```diff
@@ -0,0 +1,359 @@
...
+          <code>title</code>
+        </DocsInternalLink> prop/slot, all other slots are optional. 
+        This flexibility allows for creating diverse layouts by combining different options for the <DocsInternalLink href="/KCard#prop:layout">
+          <code>layout</code>
+        </DocsInternalLink>
+        and
+        <DocsInternalLink href="/KCard#prop:thumbnailDisplay">
+          <code> thumbnailDisplay</code>
+        </DocsInternalLink>  properties. Additionally, the order of slots is not strictly enforced. For instance,
```
AlexVelezLl: I didn't understand this explanation of the order of the slots and the `title`, `aboveTitle`. I think in a previous section we could indicate what our definition of Title is and the difference between `aboveTitle`, `title` prop, `title` slot, `belowTitle`.

docs/pages/kcard.vue
```diff
@@ -0,0 +1,359 @@
...
+        the  <DocsInternalLink href="/KCard#slot:title">
+          <code>title</code>
+        </DocsInternalLink> 
+        slot can take the position of the 
+        <DocsIternalLink href="/KCard#slot:aboveTitle">
+          <code>aboveTitle</code>
+        </DocsIternalLink> slot if the latter is not used.
+      </p>
+      <p>
+        The thumbnail's appearance depends on two factors: the size of the original image and the chosen
```
AlexVelezLl: I think we can introduce thumbnails and its appearance in a different section and introduce the thubmnail-related props "thumbnailDisplay", "thumbnailSrc", "thumbnailScaleType". But I dont think we need to introduce this inside the `Horizontal` section, because this is not specific to the horizontal layout and can also be used with the vertical layout

docs/pages/kcard.vue
```diff
@@ -0,0 +1,359 @@
...
+          <code>
+            thumbnailScaleType
+          </code>
+        </DocsInternalLink>
+        property.  This property controls how the image is scaled to fit the designated thumbnail area. 
+        The available scaling options are the same as those offered by the KImg component's scaleTypes property.
+      </p>
+      Here are a few examples...
+
+      <p>
```
AlexVelezLl, line 180: I think this should be a heading to keep consitency with the vertical layout examples

docs/pages/kcard.vue
```diff
@@ -0,0 +1,359 @@
...
+            <template #footer>
+              <div>
+                This is a sample footer section for the kcard
+              </div>
+            </template>
+          </KCard>
+        </div>
+      </DocsShow>
+
+      <h4>
```
AlexVelezLl: Note that in the vertical layout subheadings we are using h5, and here h4

docs/pages/kcard.vue
```diff
@@ -0,0 +1,359 @@
...
+      <p>
+        in the event where <DocsInternalLink href="/KCard#prop:thumbnailDisplay">
+          <code> thumbnailDisplay</code>
+        </DocsInternalLink> and  <DocsInternalLink href="/KImg#prop:thumnailcaletype">
+          <code>
+            thumbnailScaleType
+          </code>
+        </DocsInternalLink> is not provided we will end something like this
+      </p>
+      <DocsShowCode language="html">
```
AlexVelezLl, line 185: Here we are not showing any visual example:
image

docs/pages/kcard.vue
```diff
@@ -0,0 +1,359 @@
...
+            <template #footer>
+              <div>
+                This is a sample footer section for the kcard
+              </div>
+            </template>
+          </KCard>
+        </div>
+      </DocsShowCode>
+      <p>
+        And if the layout is set  Horizontal
```
AlexVelezLl: I see diferent layuots between this example and [this one](https://github.com/learningequality/kolibri-design-system/pull/625/files#diff-afca5902f76934ce03daa69d66c65def4f84818a26273d3dff10bd122efa56a5R301) but the description of "horizontal layout" and "small thumbnailDisplay" are the same.

docs/pages/kcard.vue
```diff
@@ -0,0 +1,359 @@
...
+        </div>
+      </DocsShow>
+      <DocsShow block>
+        <div>
+          <KCard
+            :to="{ name: '/' }"
+            :headingLevel="2"
+            title="How to get started with Kolibri"
+            layout="horizontal"
+            thumbnailDisplay="none"
```
AlexVelezLl: What is the difference between an horizontal and vertical layout if we dont have thumbnails in neither?

docs/pages/kcard.vue
```diff
@@ -0,0 +1,359 @@
...
+
+  export default {};
+
+</script>
+
+
+<style lang="scss" scoped>
+
+  em {
+    font-style: normal;
```
AlexVelezLl: With this rule we would be using the `em` tag as a `strong` tag isnt it?

lib/KCard/BaseCard.vue
```diff
@@ -0,0 +1,127 @@
...
+  >
+    <component
+      :is="headerLevel"
+    >
+      <router-link
+        tabindex="-1"
+        :to="to"
+      >
+        <KTextTruncator
+          v-if="title !== null"
```
AlexVelezLl: I think this title !== null check should be at the level of the header
```vue
    <component
      v-if="title !== null"
      :is="headerLevel"
    >
```

Because if we dont pass a Title we would be still rendering the header tag and the `router-link` even thought its invisible because we dont have text inside it.

lib/KCard/index.vue
```diff
@@ -0,0 +1,226 @@
...
+      },
+      /**
+       * The layout style of the hero banner. Required and cannot be empty.
+       *
+       * @validator
+       * @param {String} value - The layout value.
+       * @returns {Boolean} - True if the value is not empty, false otherwise.
+       */
+      layout: {
```
AlexVelezLl, line 110: I think we sould also validate the layout values, and include them in the jsdocs for implementers to know which value this prop can have.

lib/KCard/index.vue
```diff
@@ -0,0 +1,226 @@
...
+    :titleLines="titleLines"
+    :to="to"
+  >
+    <template #default>
+      <aside
+        v-if="layout === 'vertical' && thumbnailDisplay !== 'none'"
+      >
+        <KImg
+          :src="thumbnailSrc"
+          :isDecorative="true"
```
AlexVelezLl, line 25: Is this allways true? Are all the images on a card decorative? Or maybe in the future we could support a `thumbnailAlt`?I'm just asking because I find it curious.

lib/KCard/index.vue
```diff
@@ -0,0 +1,226 @@
...
+
+  <BaseCard
+    :title="title"
+    :headingLevel="headingLevel"
+    :titleLines="titleLines"
+    :to="to"
+  >
+    <template #default>
+      <aside
+        v-if="layout === 'vertical' && thumbnailDisplay !== 'none'"
```
AlexVelezLl: We should avoid burned strings in the code and instead use constants to avoid any typos and make it easier to maintain in the future if we want to change the string.

So we can maybe declare constants:

```javascript
const CardLayout = {
  VERTICAL: "vertical",
  HORIZONTAL: "horizontal",
};

const CardThumbnailDisplay = {
  NONE: "none",
  SMALL: "small",
  LARGE: "large",
};
```

lib/KCard/index.vue
```diff
@@ -0,0 +1,226 @@
...
+
+        <div>
+          <header
+            v-if="thumbnailDisplay === 'none'"
+            class="spacing"
+          >
+            <!-- @slot Places content to be placed above the title.. -->
+            <slot name="aboveTitle"></slot>
+            <!-- @slot for the title content -->
+            <slot name="title"></slot>
```
AlexVelezLl: I think this title slot is an alternative to the title prop. So I think we will probably want to make this part of the BaseCard component perhaps?

lib/KCard/index.vue
```diff
@@ -0,0 +1,226 @@
+<template>
+
+  <BaseCard
+    :title="title"
+    :headingLevel="headingLevel"
+    :titleLines="titleLines"
+    :to="to"
+  >
+    <template #default>
```
AlexVelezLl, line 15: The logic in general is a bit confusing, there are quite a few v-ifs nested in various places and there is repeated code and there is no clear flow of what happens in each case.

I think we could use `display: flex`, and `flex-direction` to play with the layout a bit. Something to keep in mind is that the content will always be the same, what changes is the layout, so we could have something like:

```Vue
<BaseCard>
  <div class="display-flex" :style="wrapperStyle">
    <aside v-if="thumbnailDisplay!='none'">
      <KImg ... :style="imageStyle">
    </aside>
    <div>
      <slot #default>
    </div>
  </div>
  <div class="footer" v-if="$slots.footer">
    <slot #footer>
  </div>
</BaseCard>

...

<script>
...
wrapperSyle() {
  if (layout === "vertical") {
    return { flex-direction: "column" }
  }
  if (layout === "horizontal" && thumbnailDisplay === "small") {
    return { flex-direction: "row-reverse" } // To kep the image on the right
  }
  if (layout === "horizontal" && thumbnailDisplay === "large") {
    return { flex-direction: "row" } // To keep the image on the left
  }
}
</script>
```

In this way we do not repeat so many KImg, nor the content and we only play with the styles of the image, of the content depending on each case.

lib/KCard/index.vue
```diff
@@ -0,0 +1,226 @@
...
+          if (!value) {
+            console.error('Error: Prop layout is required and can not be empty.');
+            return false;
+          }
+          return true;
+        },
+      },
+      /**
+       * Controls the display of the thumbnail image.
+       * Options: 'none' (default), 'block', or 'inline'.
```
AlexVelezLl: The options we say here are 'none', 'block', 'inline'. But the ones we use in the code are 'none', 'small' and 'large'.

lib/KCard/index.vue
```diff
@@ -0,0 +1,226 @@
...
+        } else {
+          return {
+            padding: '0em',
+          };
+        }
+      },
+      isVerticalLayout() {
+        return (
+          this.layout === 'vertical' ||
+          (this.layout === 'horizontal' && this.thumbnailDisplay === 'large')
```
AlexVelezLl: Does this means that an horizontal layout with large thumbnail behaves like a vertical layout?

lib/KCard/BaseCard.vue
```diff
@@ -0,0 +1,127 @@
...
+    }
+  }
+
+  .remove-list-style {
+    list-style-type: none;
+  }
+
+  .spacing {
+    padding: 1em;
+    color: black;
```
AlexVelezLl: Here we are setting a burned color, instead of the $themeTokens ones.

AlexVelezLl (changes requested):
Hi @AllanOXDi! I am glad that our KCard is already taking shape :tada:. I noticed a few things, especially in the documentation, and in the KCard code structure. Let me know any thoughts of any of the comments :open_hands:.

lib/KCard/index.vue
AllanOXDi: No- I think  prop `to` is required . Remember the entire card is clickable.  There for a prop` to ` is Router-link object that enables user navigation to the intended route. Think of the current channel cards for example,  We are trying to lean on this [guidelines ](https://inclusive-components.design/cards/)as much as we can

docs/pages/kcard.vue
AllanOXDi: That's a good point

docs/pages/kcard.vue
AllanOXDi: Good catch!

docs/pages/kcard.vue
AllanOXDi: I am thinking that's is the visual example below it- No?  I want to show few visual example.  Otherwise, the  documentation page will get cultured and I have not figured out a way to handle it.

docs/pages/kcard.vue
AllanOXDi: No, rule does not actually make the `em `tag function exactly like the `strong` tag.
I think` em ` indicate emphasis or stress within a sentence. Can be rendered as italic, bold, or underline depending on browser and user styles.

`strong ` indicate strong importance of a word or phrase, typically rendered as bold.

lib/KCard/index.vue
AllanOXDi: I think not all images on a card are always decorative.  From my understanding the `:isDecorative `property used to indicate to assistive technologies (screen readers) whether the image provides essential information or not. Screen readers might skip decorative images to improve reading efficiency for visually impaired users.

Your intuition is right. It's very common to have a descriptive text alternative (alt text) for informative thumbnails. Including a `thumbnailAlt` property would allow us to provide this description, which would be beneficial for visually impaired users as well. https://www.w3.org/WAI/tutorials/images/decorative/

lib/KCard/index.vue
AllanOXDi: I am afraid to take this approach  because  a couple of things could go wrong  here with the` CardLayout` and `CardThumbnailDisplay` typos in the enum values (e.g., "vertival" instead of "vertical") could lead to unexpected behavior or errors.
If we accidentally use a misspelled value (e.g., CardLayout.VERTICLE), it wouldn't match any defined value in the enum

Another thing I am thinking of is that that  it only cover basic layout and thumbnail display options, it only define string values.  And If we need to associate additional data with each layout or display option this  might not be sufficient.

lib/KCard/index.vue
AllanOXDi: That would not be a good idea. I think `BasCard` is to enforce consistent structure and accessibility for cards across  our product and Isolates basic accessibility requirements from potential changes in the `KCard` component,

lib/KCard/index.vue
AllanOXDi: Yes, my reasoning behind this  is that a large thumbnail in a horizontal layout might visually occupy a significant portion of the space, creating a layout that resembles a vertical one where elements are stacked.

docs/pages/kcard.vue
AllanOXDi: I thinks it's something I just copied and pasted - will fix it.

docs/pages/kcard.vue
AllanOXDi: No- I was just trying with different slots - But I think it's confusing. Beside the title prop is not required.

docs/assets/vertical-layout.png
MisRob, line 1: [required] I believe this is unused file that can be removed

lib/KCard/index.vue
MisRob: [required] I second that this would be a valuable improvement and suggest we do it right in the first iteration as it will make any further work easier

lib/styles/trackInputModality.js
```diff
@@ -25,6 +25,7 @@ function setUpEventHandlers(disableFocusRingByDefault) {
     '[role=textbox]',
     'a',
     'button',
+    '[data-focus=true]',
```
MisRob, line 29: [required] Could you add a brief comment on the motivation behind this and mention its link to `KCard`'s focus?

MisRob (commented):
Wonderful work overall @AllanOXDi, and it's great that a11y QA passes, that was the most tricky thing!

I am posting some first feedback. I haven’t reviewed yet everything - will follow up gradually.

MisRob:
[required] `title` slot is to be an alternative to the `title` prop, however now the slot now renders in addition to the `title` slot and is not showing in a correct place. For example, for

```
<KCard
    :to="{ name: '/' }"
    :headingLevel="2"
    layout="vertical"
  >
    <template #title>
      Card title
    </template>
</KCard>
```

I see the title text not being rendered as heading:

Screenshot from 2024-05-18 13-22-17

Could you adjust the behavior to follow the expected behavior in all of the following cases? Note that in one or two cases it works well already, but not in all of them - including all for clarity.

| | Expected |
| --- | --- |
| only `title` prop ```<KCard title="Card title">``` | Screenshot from 2024-05-18 13-12-14 |
| only `title` slot ```<KCard><template #title>Card title</template></KCard>``` |  Screenshot from 2024-05-18 13-12-14 |
| both `title` prop and `title` slot ```<KCard title="Card title from the prop"><template #title>Card title from the slot</template></KCard>``` - give preference to the `title` prop and ignore the slot | Screenshot from 2024-05-18 13-20-30 |

MisRob:
[required] `title` slot is to be an alternative to the `title` prop, however now the slot renders in addition to the `title` slot and is not showing in a correct place. For example, for

```
<KCard
    :to="{ name: '/' }"
    :headingLevel="2"
    layout="vertical"
  >
    <template #title>
      Card title
    </template>
</KCard>
```

I see the title text not being rendered as heading:

Screenshot from 2024-05-18 13-22-17

Could you adjust the behavior to follow the expected behavior in all of the following cases? Note that in one of them it works well already, but not for all - including all for clarity. There's also an additional issue in the screenshots below regarding the placement of the `aboveTitle` but I believe it is already being addressed.

| | Expected |
| --- | --- |
| only `title` prop ```<KCard title="Card title">``` | Screenshot from 2024-05-18 13-12-14 |
| only `title` slot ```<KCard><template #title>Card title</template></KCard>``` |  Screenshot from 2024-05-18 13-12-14 |
| both `title` prop and `title` slot ```<KCard title="Card title from the prop"><template #title>Card title from the slot</template></KCard>``` - give preference to the `title` prop and ignore the slot | Screenshot from 2024-05-18 13-20-30 |

MisRob:
[required] One important area to address here is to check the resulting UI against [the "Updated Cards Specs Metadata" designs](https://www.figma.com/design/1tQQ0Sw1yDjzO66hyjLVPM/Kolibri-UX-Fixes?node-id=1868-55352) and make sure that all the logic and styles that `KCard` is responsible for aligns to them (while leaving out areas that `KCard` is not responsible for, such as content of slots like icons, buttons, etc.) Few examples:

|Design | Implementation | 
| -------- | -------------------- |
| Screenshot from 2024-05-18 14-23-11 | expected-1 |
| Screenshot from 2024-05-18 14-35-32 | expected-2 |
| Screenshot from 2024-05-18 14-42-42 | expected-3 |

These are just a few places I noticed, and is not meant to be a full list. Could you please take some time to test all available layouts and compare with designs? Placement, font sizes, padding/margins, and pretty much anything we can do to follow designs closely. Figma will provide you with all information and useful measurement tools, such as



Note that letterboxing should be preserved though even though it's not showed on the designs. The image on the designs corresponds to the whole thumbnail area (possibly letterboxed) in the final implementation.

MisRob:
@AllanOXDi There are some acceptance criteria from the issues this PR is meant to close that appear not yet be implemented or have unclear status. Could you please read through all related issues, particularly "Markup/style requirements" and the "Acceptance criteria" sections and resolve them or mention where things are if there are any blockers? Some examples: "_Supports RTL_", "_Even though the whole area is clickable, text content is still selectable_", "[documentation] _contains guidance on the headingLevel to notify developers that they need to choose a correct level based on context_", and perhaps few more.

I know it's lots of work, but there's time to finish it and would suggest we do so in this PR since it's not draft like the first Studio PR. I think it will be easier than opening many follow ups since there's no pressure around merging. This will prepare robust ground for further work and help with implementing card grids as well. If you needed to chat around how to organize work or encountered some blocking issues, we'll have time to chat :)

lib/KCard/__test__/KCard.spec.js
```diff
@@ -0,0 +1,76 @@
...
+    wrapper.vm.$on('hover', data => {
+      hoverData = data;
+    });
+
+    expect(hoverData).toBeUndefined();
+  });
+
+  //testing slots
+  it('renders slotted content', () => {
+    expect(wrapper.find('[slot="aboveTitle"]').exists()).toBe(false);
```
MisRob: Did you mean to test for the presence of the slot? If so, you'd need to say `toBe(true)`. The test, as implemented now, confirms that the slot _doesn't exist_ rather than that it does.

I think that lines 65-68 are sufficient to confirm that slot content gets displayed.

lib/KCard/__test__/KCard.spec.js
```diff
@@ -0,0 +1,76 @@
...
+    },
+  });
+  // });
+
+  it('renders passed header level', () => {
+    expect(wrapper.find('h4').exists()).toBe(true);
+  });
+
+  //testing focus and hover events
+  it('renders data on focus', () => {
```
MisRob: Could you please elaborate a bit on what this scenario is supposed to test?

lib/KCard/__test__/KCard.spec.js
```diff
@@ -0,0 +1,76 @@
...
+    let emittedData;
+    wrapper.vm.$on('focus', data => {
+      emittedData = data;
+    });
+
+    expect(emittedData).toBeUndefined();
+  });
+
+  it('emits data on hover', () => {
```
MisRob: Could you please elaborate a bit on what this scenario is supposed to test?

MisRob:
@AllanOXDi I think I've finished the first round. At this point, I focused rather on higher-level issues. I will review one more time after this feedback as well as feedback from @radinamatic and @AlexVelezLl is addressed since there will be some important changes. Thanks a lot for this work, it's shaping up really great!

lib/KCard/index.vue
MisRob: `BaseCard` is meant to have the validation _"Contains validation to check that title prop or slot has been used, throwing an error if neither is provided"_ (see acceptance criteria here https://github.com/learningequality/kolibri-design-system/issues/529) so I think the slot will need to be present in the `BaseCard`. It will add some clarity as well. This is part of the enforcement that `BaseCard` does - it needs to ensure that title is provided, and that can happen either via the prop or the slot.

@AllanOXDi you too have made a good point and regarding that, we need to ensure that `BaseCard` shouldn't have any styles or layout logic related to how the title appears.

lib/KCard/__test__/KCard.spec.js
AllanOXDi: I was trying to make sure that the card does not emit data by default and only emits data when it receives focus.

lib/KCard/__test__/KCard.spec.js
AllanOXDi: I was trying to make sure that the card does not emit data by default and only emits data when it receives focus.

lib/KCard/__test__/KCard.spec.js
AllanOXDi: I was trying to make sure that the card does not emit data by default and only emits data when it receives focus.

MisRob: What kind of data is it expected to emit on receiving the focus?

MisRob:
For the reviewers, just a note that resolving the first PR feedback is still in progress, so the time for re-review will come later.

MisRob:
Note that RTL will be done as part of https://github.com/learningequality/kolibri-design-system/issues/665 rather than in this PR

lib/KCard/BaseCard.vue
```diff
@@ -0,0 +1,144 @@
...
+    @extend %dropshadow-2dp;
+
+    position: relative;
+    display: block;
+    text-align: left;
+    text-decoration: none;
+    cursor: pointer;
+    border-radius: 0.5em;
+    outline-offset: -1px;
+    transition: border-color $core-time ease;
```
MisRob: I was just doing KDS shadows motion audit for designers when I noticed the transition doesn't seem to work in the card. I think we'd need to use `box-shadow` instead of `border-color`? Since that's the style that dropshadows as defined here modify https://github.com/learningequality/kolibri-design-system/blob/d784b830967d810ae88a42e207b755bbe5bc2318/lib/styles/definitions.scss#L59

docs/pages/kcard.vue
```diff
@@ -0,0 +1,416 @@
...
+        The <code>KCard</code> component is a versatile and accessible base component
+        for creating various card types, including lesson cards, resource cards, and channel cards.
+        It provides slots for adding content such as titles, subtitles, and footers,
+        along with options for displaying thumbnail images and customizing the layout.
+      </p>
+
+    </DocsPageSection> <DocsPageSection title="Usage" anchor="#usage">
+      <p>
+        <code>KCard</code> is a versatile base component for creating various card types.
+        It can be configured by adding product-specific content and logic to generate distinct card components.
```
AlexVelezLl: This seems very similar to the text in the introduction section. I would suggest to keep this text as introduction instead of the current paragraph, as it seems more general.

docs/pages/kcard.vue
```diff
@@ -0,0 +1,416 @@
...
+      <p>
+        <code>KCard</code> offers two main types of layouts to accommodate different content needs: horizontal and vertical.
+      </p> 
+      <h3>Horizontal Layout</h3>
+      <p> The horizontal layout displays the thumbnail image on the left and the content on the right. </p>
+      <p> Here's an example of how to use the horizontal layout: </p>
+      <DocsShow block>
+        <div>
+          <KCard
+            :to="{ name: '/' }"
```
AlexVelezLl: I think we could add the redirect to the same kcard page here: `:to="{ name: 'kcard' }"` to avoid the redirection error in the documentation when we click on the card:
Screenshot from 2024-07-09 13-24-16

lib/KCard/index.vue
```diff
@@ -0,0 +1,314 @@
...
+        }
+        return {};
+      },
+      cardContentPartial() {
+        if (
+          this.layout === 'horizontal' &&
+          (this.thumbnailDisplay === 'small' || this.thumbnailDisplay === 'large')
+        ) {
+          return {
+            width: '50%',
```
AlexVelezLl: This value could be calculated dinamically, since because of this restriction, we can see too much padding in the card:
image

lib/KCard/index.vue
```diff
@@ -0,0 +1,314 @@
...
+  <BaseCard
+    :to="to"
+    :title="title"
+    :headingLevel="headingLevel"
+    :titleLines="titleLines"
+    :style="wrapperStyle"
+  >
+    <template #title>
+      <!-- @slot Optional slot section containing the title contents, should not contain a heading element. -->
+      <div class="title-slot">
```
AlexVelezLl: It seems like we are not using the title slot of the basecard, since we dont have the heading level set on the title text: 
image

We would need to add the <slot name="title"/> here to pass the title slot of the KCard to the title slot of the Base Card

lib/KCard/index.vue
```diff
@@ -0,0 +1,314 @@
...
+  }
+
+  .title-slot {
+    font-size: 16px; 
+    font-weight: 600;
+    line-height: 1.5;
+  }
+
+  /deep/ .base-card-heading {
+    order: 2;
```
AlexVelezLl, line 303: It seems we have a little error here. Because of this order: 2 style, we are having the title slot at the end of the card, and since we are not using the title slot of the base card, it appears as an empty space at the bottom of a vertical card:

image

lib/styles/trackInputModality.js
AlexVelezLl: Seems like this is still missing

lib/KCard/BaseCard.vue
```diff
@@ -0,0 +1,154 @@
...
+      },
+      cardClick() {
+        this.$router.push(this.to);
+      },
+      onMouseDown() {
+        this.mouseDownTime = new Date().getTime();
+      },
+      onMouseUp() {
+        const mouseUpTime = new Date().getTime();
+        if (mouseUpTime - this.mouseDownTime < 200) {
```
AlexVelezLl, line 163: I think it would be nice to have a comment here explaining that the reason for this behaviour is to allow selecting text inside the card.

lib/KCard/index.vue
```diff
@@ -0,0 +1,314 @@
...
+        />
+        <slot v-if="!thumbnailSrc" name="thumbnailPlaceholder"></slot>
+      </aside>
+      <div class="display-inline-block" :style="cardContentPartial">
+        <div
+          data-testid="aboveTitle"
+          class="above-title-style"
+          :class="{ 'above-title-horizontal': layout === 'horizontal' }"
+        >
+          <slot name="aboveTitle"></slot>
```
AlexVelezLl: @MisRob I wonder if these "aboveTitle", "belowTitle" slots should also be present on the  BaseCard component? I dont see these slots in the BaseCard specs. But we dont have a way of control these slots here in the KCard to make them appear above or below the title, since it is the BaseCard who has the responsibility of rendering the title, so probably its also the BaseCard the responsilble for rendering the belowTitle and aboveTitle.

lib/KCard/index.vue
```diff
@@ -0,0 +1,314 @@
...
+          class="thumbnail-image"
+          :style="imageRadius"
+        />
+        <slot v-if="!thumbnailSrc" name="thumbnailPlaceholder"></slot>
+      </aside>
+      <div class="display-inline-block" :style="cardContentPartial">
+        <div
+          data-testid="aboveTitle"
+          class="above-title-style"
+          :class="{ 'above-title-horizontal': layout === 'horizontal' }"
```
AlexVelezLl: I dont see any CSS rules defining none of the classes: above-title-horizontal, title-horizontal, below-title-horizontal, so we will need to remove them from the nodes also.

lib/KCard/index.vue
```diff
@@ -0,0 +1,314 @@
...
+
+        <div
+          data-testid="belowTitle"
+          class="below-title"
+          :class="{ 'below-title-horizontal': layout === 'horizontal' }"
+        >
+          <slot name="belowTitle"></slot>
+        </div>
+        <div
+          class="footer"
```
AlexVelezLl: I think we could add a check here, and render this node just if there is a `$slots.footer`

lib/KCard/index.vue
```diff
@@ -0,0 +1,314 @@
...
+  import KImg from '../KImg';
+  import BaseCard from './BaseCard.vue';
+
+  const ScaleTypes = {
+    CENTER_INSIDE: 'centerInside',
+    CONTAIN: 'contain',
+    FIT_XY: 'fitXY',
+  };
+
+  function isValidScaleType(value) {
```
AlexVelezLl: I dont think we should do this validation logic here, since we already do it inside the KImg component, and this create duplicated code, and we have to keep in mind update both places if we need to add a new ScaleType for example

lib/KCard/__test__/KCard.spec.js
```diff
@@ -0,0 +1,110 @@
...
+
+const router = new VueRouter({
+  routes: [{ path: '/some-link' }],
+});
+
+function makeWrapper({ propsData = {}, slots = {} } = {}) {
+  return mount(KCard, { router, propsData, slots, localVue });
+}
+
+describe('index.vue', () => {
```
AlexVelezLl: The name of this Test Suite should be the same as the component we are testing, to more easily identify which component each test belongs to.

lib/KCard/__test__/KCard.spec.js
```diff
@@ -0,0 +1,110 @@
...
+    await wrapper.find('li').trigger('mouseup');
+
+    expect(wrapper.vm.$router.currentRoute.path).toBe('/some-link');
+  });
+
+  describe('it renders slotted content', () => {
+    it('renders slotted content with belowTitle slot', () => {
+      const wrapper = makeWrapper({
+        propsData: { to: { path: '/some-link' }, title: 'sample title ' },
+        slots: {},
```
AlexVelezLl: I think we missed passing slots here. Here we have the intention of test slotted content, but we are not passing anything to the slots prop.

AlexVelezLl (commented):
Looks amazing!! This is already super polished and looks super good. I just left some minor things I noticed that could add value to the implementation <3

lib/KCard/__test__/KCard.spec.js
AllanOXDi: I did not get you here

AlexVelezLl: Oh I was talking about renaming the describe block to 
```js
describe('KCard', () => {
```

To see in the  tests output that it is testing the KCard component. This is the output right now, and as you can see index.vue is not very descriptive (especially because this is logged among a lot of other tests).

```
  index.vue
    ✓ renders passed header level (20 ms)
    ✓ renders the correct accessibility  structure when title slot is passed (14 ms)
    ✓ renders the correct accessibility structure when title pass as props (6 ms)
    ✓ should not navigate on long click (508 ms)
    ✓ should navigate on quick click (109 ms)
    it renders slotted content
      ✓ renders slotted content with belowTitle slot (6 ms)
      ✓ renders slotted content with footer slot (6 ms)
```

MisRob (dismissed):
Thank you for updates, @AllanOXDi, and @AlexVelezLl for re-review. I won't re-review everything right now since I can see there's lots of feedback from @AlexVelezLl. However, I will request one change - there are important regressions in regard to card title, when passed both via slot and via prop. This is important to have in place since it's why we're building the card.

This is the expected output structure markup of a card, under any circumstances:

Screenshot from 2024-07-10 17-44-07

At this point, when the title is passed via slot, the markup structure is

Screenshot from 2024-07-10 17-37-50

and when passed via prop, it is

Screenshot from 2024-07-10 17-39-01

Even though the `title-slot` div is there, it's always empty and the title is rendered somewhere else than here in `<a>` which is the main problem here. 

Likely related, the title doesn't show at all when passed as prop:

Screenshot from 2024-07-10 17-46-13

It may help you to return more examples to the playground page, for more layout modes of the card, and also for when title is passed to the card in both these ways. Please make sure that in these cases, the card satisfies markup and visual requirements.

In regard to minor feedback, I think it'd be fine to open follow-ups it if helps. Please let me know if you needed anything.

lib/KCard/index.vue
MisRob: Agreed

lib/KCard/index.vue
MisRob: @AlexVelezLl There's a way to control them via styling with flex order which is needed for more reasons anyways - primarily because the placement of slots in the code really shouldn't change the resulting markup structure. However, even though it's technically valid and possible, it creates some complexity when `KCard` needs to do some styling in the `BaseCard`, so I hear you. If the separation of `KCard` and `BaseCard` brings more cons than pros, an alternative that would resolve all this would be to merge these two components into a single `KCard` component. I'm open to this and leave to @AllanOXDi's judgement.

lib/KCard/index.vue
MisRob: Agreed

lib/KCard/index.vue
MisRob: Note that even if we merged those two components though, we'd still need to show the content of `aboveTitle` below the title in the markup and achieve placing it above the title by means of CSS rather than HTML. This is a standard a11y friendly coding technique and a core feature of the card ensuring that screen readers announce it as expected.

So the question of merging vs not would really be about cross-component vs one-component styling. In both cases we'll still be using `flex-order`.

MisRob:
Hi @AllanOXDi, I started the final round of review. Overall it feels much more robust now, layouts look great, and I really appreciate you took time to prepare cards preview similar to the cards we will implement in Kolibri. I am sure this will make all further work in products smooth. Thank you for all hard work :)

As we chatted earlier, it'd be great to have this merged before you're back online and open follow-ups if there's some more work. I will open a few, but for some changes, it's a bit easier for me to push them right away rather than explaining in new issues, and I also hope it will help with other work we have in progress in Kolibri and Studio that you can return to when you are back. I hope you don't mind - it's done with the intention of support. I'm trying my best to describe reasons for changes in the commit messages, and will be available to talk about each of them. The main areas updated I considered important

- Fixes for some last issues with the title slot
  - Motivation: User-facing 
- Consistency of the title styles no matter if title provided via prop or slot
  - Motivation: User-facing  
- Text selection feature completion
 - Motivation: User-facing  
- Some last styling touches
  - Motivation: User-facing  
- Re-organization of styles in `KCard` regarding various layouts
  - Motivation: This is really not so important for users, however I believe it will help a lot with future development, and it also helped me with some styling updates. I think we had this done already as response to @AlexVelezLl's first review, but seems to get lost with big styles overhaul after fixes to markup structure.

(You will also see some documentation and renaming mixed in my commits and that's just me nitpicking and doing little improvements when I see an opportunity as I'm in the code already, nothing really important, and in majority of cases, I wouldn't be posting to review :)

After I am done, I will ask @AlexVelezLl to review my updates, and then we will merge. Thanks both and talk to you soon.

MisRob:
Hi @AllanOXDi, I finished the final review. Overall it feels much more robust now, layouts look great, and I really appreciate you took time to prepare cards preview similar to the cards we will implement in Kolibri. I am sure this will make all further work in products smooth. Thank you for all your hard work :)

As we chatted earlier, it'd be great to have this merged before you're back online and open follow-ups if there's some more work. I will open a few, but for some changes, it's a bit easier for me to push them right away rather than explaining in new issues, and I also hope it will help with other work we have in progress in Kolibri and Studio that you can return to when you are back. I hope you don't mind - it's done with the intention of support, and hope it aligns well with all we talked about together. I'm trying my best to describe reasons for changes in the commit messages, and will be available to talk about each of them. The main areas updated I considered important (cc @AlexVelezLl for review):

**User-facing:**
 - Fixes for some last issues with the title slot
 - Consistency of the title styles no matter if title provided via prop or slot
 - Improvements to styles robustness and content tolerance,  and some last styling touches
- Ensuring that thumbnails can display placeholders via "thumbnailPlaceholder" slot when thumbnail is not available

**Other:**
- Re-organization of styles in `KCard` regarding various layouts so that styles belonging to particular layouts are grouped together
  - This is really not so important for users, however I believe it will help with future development, and it also helped me a lot with some styling updates I am mentioning above. I think we had this done already to some extent as response to @AlexVelezLl's first review, but seems to have gotten lost with big styles overhaul after fixes to markup structure. It's lots of changes but rather organizational with a few fixes applied, the foundational techniques you put in place stay the same.
 
(You will also see some documentation and renaming mixed in my commits and that's just me nitpicking and doing little improvements when I see an opportunity as I'm in the code already, nothing really important, and in majority of cases, I wouldn't be posting to review :)

After I am done, I will ask @AlexVelezLl to review my updates, and then we will merge. Thanks both and talk to you soon.

MisRob:
Would you please give the components one last look, @AlexVelezLl? Please also see the playground page that simulates more use-cases now. I will delete it before merging, but keep a copy somewhere for testing (this will be perfect first task for automated visual tests :)

Also please note that the following will be resolved in follow-ups (some already open, and I will open later remaining ones). If there's any feedback from you that's not resolved and not blocking, please mention.

- Clarify how cards will behave in a grid regarding their height with designers, e.g. should the cards height be always the same or should the row adjust to the highest card (current implementation is my best guess)? Based on that, determine if we need to be able to override cards heights/widths, remove spaces occupied by empty slots, etc. (I will do this as part of grid work)
- Some more guidelines in documentation for frequent tasks (probably me)
- Vertical/horizontal layout with no thumbnail (me or Allan)
- Adding constants for `layout` and `thumbnaiDisplay` and related prop validators (Allan)
- Implementing RTL support (Allan)

lib/KCard/BaseCard.vue
```diff
@@ -0,0 +1,204 @@
...
+      },
+      // Make textual content selectable within the whole
+      // clickable card area
+      onMouseUp() {
+        const mouseUpTime = new Date().getTime();
+        // Calculate the time difference between the mouse button press and release.
+        // If the time difference is greater than or equal to 200 milliseconds,
+        // it means that the mouse button was pressed and held for a longer time,
+        // which is not typically interpreted as a click event. Do not navigate
+        // in such case.
```
AlexVelezLl, line 162: I see value in adding that this allows us select text for example.

lib/KCard/index.vue
```diff
@@ -0,0 +1,393 @@
...
+        >
+          <!-- @slot Places content to the thumbnail placeholder area. -->
+          <slot name="thumbnailPlaceholder"></slot>
+        </span>
+      </div>
+      <div
+        data-test="aboveTitle"
+        class="above-title"
+      >
+        <slot name="aboveTitle"></slot>
```
AlexVelezLl, line 49: I think we should document these slots as well.

radinamatic:
I took a look at the latest version of the playground and tested with NVDA in Windows. 

One thing popped up as a concern, but again I'm not sure if it's determined by this base card implementation, or will it be further _injected_ by other products. Namely, the color contrast of the channel name (is that in the ``aboveTitle`` slot?) is insufficient:






Similarly, even though I understand what was mentioned earlier, that:

> products that will be responsible for adding buttons, context menus, etc. and ensuring it doesn't break a11y.

I just wanted to reiterate 🙂 , that we need to pay attention to offer full and comprehensive information about those element in the footer. For example, this is the full readout of one of the cards by NVDA:

```
clickable  
<Title>  
link    
heading    level 2  
<Below title>
PracticeShort Activity  20
```

All above looks Ok, just 2 comments regarding the last line:
* a separation is missing between the type of resource and the duration 
* [``progress``](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress#accessibility) element needs a role and a label, otherwise the screen reader outputs just the number (20) without any context

I know this might not be in the scope of the present PR, but it cannot hurt to repeat it 😉

AlexVelezLl (approved):
This LGTM! Code looks good and well documented, documentation looks very informative and easy to follow, and visual examples in the playground looks great! I think is ready to go! Just left a super minor comments about docs. 

Thank you @AllanOXDi for all your consistent hard and great work here!! and @MisRob for your great guidance! Will leave the final merge to @MisRob once playground is cleaned and QA passes.

lib/KCard/BaseCard.vue
MisRob: Ah yes, it was couple of lines above - will make sense to have it here together. Updated, thank you.

lib/KCard/index.vue
MisRob: Ah yes, thank you. Done

MisRob:
Thank you @radinamatic, I appreciate it. Yes, these are all the things that consuming apps need to ensure as they will inject it. I'd like us to have some guidance into the documentation to clarify for developers what we need to take care of when implementing cards on top of this component vs what is it already taking care of. Also, as soon as we have first few Kolibri card components implemented and a11y tested, we could use the way they look for the documentation examples to reflect this all a bit better.

MisRob:
^ We will keep working on this all gradually as part of the project, this component is a core step, and also a beginning of larger body of work in a sense.

MisRob:
Thanks for feedback @AlexVelezLl! Merging now.

Issue #623, opened 2024-04-17
Improve overflow handling in KTabsList component


nucleogenesis:
🌱 **Are you new to the codebase? Welcome and please see [the contributing guidelines](https://github.com/learningequality/kolibri-design-system/blob/release-v4/CONTRIBUTING.md).**

## Desired behavior

- [`KTabsList`](https://design-system.learningequality.org/ktabslist/), and therefore also [`KTabs`](https://design-system.learningequality.org/ktabs/) (which is built on top of it)
 
![Screenshot from 2024-04-19 12-23-59](https://github.com/learningequality/kolibri-design-system/assets/13509191/d1e39ec6-4dff-4d20-b167-9e47ca980dbc)

should always take up one row. If the tab list items can't fit into one row, the items for which there is not enough space should collapse into a dropdown menu, such as

![Screenshot from 2024-04-19 11-51-43](https://github.com/learningequality/kolibri-design-system/assets/13509191/a2d5cc49-bbab-4cd6-8610-438ddc051333)

- `KTabsList` should utilize [`KListWithOverflow`](https://design-system.learningequality.org/klistwithoverflow/) under the hood to achieve this behavior. See the "Guidance" section below for more information.

- This should be the default behavior of both `KTabsList` and `KTabs`.

## Current behavior

When there is not enough space, `KTabsList`'s items wrap into more rows:

![Screenshot from 2024-04-19 11-52-53](https://github.com/learningequality/kolibri-design-system/assets/13509191/96c8894b-b3f7-488c-b679-5f4e554626f4)

## Steps to reproduce

Run the KDS development server and paste the following into [the playground page](https://github.com/learningequality/kolibri-design-system/blob/release-v4/dev_docs/03_how_to_update_library.md#1-preview-updates-in-the-kolibri-design-system-environment):

```vue
<template>

  <div>
    <KTabsList
      v-model="activeTabId"
      tabsId="tabs-1"
      ariaLabel="Coach reports"
      :tabs="tabs"
    />
    <KTabsPanel
      tabsId="tabs-1"
      :activeTabId="activeTabId"
    />
  </div>

</template>


<script>
  export default {
    data() {
      return {
        tabs: [
          { id: 'tabLessons', label: 'Lessons' },
          { id: 'tabLearners', label: 'Learners' },
          { id: 'tabGroups', label: 'Groups' },
        ],
        activeTabId: 'tabLessons',
      };
    },
  };

</script>
```

## Impact

[...]

Issue #588, opened 2024-03-25
Issues with keyboard navigation on KDropdownMenu


AlexVelezLl:
## Product
Noted in Studio, and in the [KDS Docs](https://release-v3--kolibri-design-system.netlify.app/klistwithoverflow/) on the last example of KListWithOverflow.

## Expected behavior

Two things are not working very well right now:
* When the list has dividers, the keyboard navigation should ignore them and focus on the next or previous item in the list.
  * In the `options` list, We can include dividers with a `{ type: "Divider" }` object.
* When the user presses `tab` or `esc`, we should return the focus to the last element focused before KDropdownMenu was active. This can be done by saving the `document.activeElement` before focusing the first option of the dropdown menu.

## Actual behavior
* Right now, the focus doesn't change when there are dividers and we want to focus the next item, and we can't jump the dividers.
* As KDropdownMenu is rendered in the body element, when `esc` or `tab` is pressed, the focus returns to the first focusable element on the document after a second. But before the focus returns to the body, the next option in the menu is focused, and after a second, the menu closes, which is a bit misleading since, at first, it seems that we could navigate through the menu with a tab.

## Steps to reproduce the issue
1. Go to the [KListWithOverflow documentation](https://release-v3--kolibri-design-system.netlify.app/klistwithoverflow/).
2. Click on the box of the last list example.
3. Navigate with the keyboard through the dropdown menu and note the issues described above.

  ## Additional information

https://github.com/learningequality/kolibri-design-system/assets/51239030/427bdbd5-b02b-4b73-bec1-d54299940b50

galovdev:
Can I get this assigned?

MisRob:
Hi @galovdev, thank you for volunteering, I will assign you

nucleogenesis:
Hello @galovdev, are you still interested in contributing on this issue? If not, we will reassign it.

galovdev:
> Hello @galovdev, are you still interested in contributing on this issue? If not, we will reassign it.

Hi, please give me this weekend, Im working on the other issue

marcellamaki:
Thank you @galovdev - no problem. Please reach out if you have any questions.

MisRob:
Hi @galovdev, are you still planning to work on this?

Sahil-Sinha-11:
Hey @AlexVelezLl may I work on this issue?

AlexVelezLl:
[...]

PR #635, opened 2024-05-09
KTabLists overflow enhancements

AlexVelezLl:
## Description
This PR introduces KListWithOverflow in KTabsList, so KTabsList now shows a "more" button for the overflowed tabs if they dont fit in the available space.

#### Issue addressed
Addresses #623

### Screencasts

https://github.com/learningequality/kolibri/assets/51239030/613e416b-f966-4235-9313-15e620f4e0ee

https://github.com/learningequality/kolibri/assets/51239030/cced15c4-a9cb-4508-8f2c-9a46c223f399

## Changelog

- [#635]
  - **Description:** KListWithOverflow now takes into account the margin of items to calculate available space.
  - **Addresses:** -.
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -

- [#635]
  - **Description:** KListWithOverflow now emits an event `updateOverflowItems` when overflowItems are recalculated.
  - **Addresses:** -.
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -.

- [#635]
  - **Description:** Kbutton and KIconButton now emit @focus event.
  - **Addresses:** -.
  - **Components:** KButton, KIconButton.
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -.

- [#635]
  - **Description:** KTabsList now shows a "more" button for the overflowed tabs if they dont fit in the available space.
  - **Products impact:** updated API.
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/623.
  - **Components:** KTabsList, KTabs.
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -.

- [#635]
  - **Description:** KTabsList now receives a `tabAppearanceOverrides` prop to override styles of tab items.
  - **Addresses:** .
  - **Components:** KTabsList.
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -.  

[#635]: https://github.com/learningequality/kolibri-design-system/pull/635

## Steps to test

Play with KTabsList and KTabs in the playground and check that it handles overflow properly.

## Testing checklist

## Reviewer guidance

## After review

### DIFF
--- a/lib/tabs/KTabsList.vue
+++ b/lib/tabs/KTabsList.vue
@@ -123,10 +123,6 @@
         type: String,
         required: true,
       },
-      tabAppearanceOverrides: {
-        type: [Object, String],
-        default: null,
-      },
     },
     data() {
       return {
--- a/lib/tabs/KTabsList.vue
+++ b/lib/tabs/KTabsList.vue
@@ -70,12 +70,14 @@
         tooltip="More"
         icon="optionsHorizontal"
         :style="overflowButtonStyles()"
-        @focus="focusedTabIdx = moreButtonFocusIdx"
       >
         <template #menu>
           <KDropdownMenu
+            ref="moreButton"
             :options="overflowItems"
             @select="(tab) => onClick(tab.id)"
+            @focusBeforeFirstElement="focusBeforeMoreButton"
+            @focusAfterLastElement="focusAfterMoreButton"
           />
         </template>
       </KIconButton>
@@ -210,11 +212,23 @@
         });
       }
       this.$el.addEventListener('keyup', this.onKeyUp);
+      document.addEventListener('keydown', this.preventScroll);
     },
     beforeDestroy() {
       this.$el.removeEventListener('keyup', this.onKeyUp);
+      document.removeEventListener('keydown', this.preventScroll);
     },
     methods: {
+      preventScroll(event) {
+        if (['ArrowLeft', 'ArrowRight'].includes(event.key)) {
+          if (
+            this.$el.contains(document.activeElement) ||
+            this.$refs.moreButton.$el.contains(document.activeElement)
+          ) {
+            event.preventDefault();
+          }
+        }
+      },
       classes(tabId) {
         return [
           'tab',
@@ -258,10 +272,13 @@
       },
       focusPreviousTab() {
         let newFocusedTabIdx;
-        if (this.isMoreButtonFocused) {
-          newFocusedTabIdx = this.lastTabIdx;
-        } else if (this.focusedTabIdx === this.firstTabIdx) {
-          newFocusedTabIdx = this.lastTabIdx;
+        if (this.focusedTabIdx === this.firstTabIdx) {
+          if (this.overflowTabs.length) {
+            this.$refs.moreButton.focusItem(this.overflowTabs.length - 1);
+            return;
+          } else {
+            newFocusedTabIdx = this.lastTabIdx;
+          }
         } else {
           newFocusedTabIdx = this.focusedTabIdx - 1;
         }
@@ -269,10 +286,13 @@
       },
       focusNextTab() {
         let newFocusedTabIdx;
-        if (this.isMoreButtonFocused) {
-          newFocusedTabIdx = this.firstTabIdx;
-        } else if (this.focusedTabIdx === this.lastTabIdx) {
-          newFocusedTabIdx = this.firstTabIdx;
+        if (this.focusedTabIdx === this.lastTabIdx) {
+          if (this.overflowTabs.length) {
+            this.$refs.moreButton.focusItem(0);
+            return;
+          } else {
+            newFocusedTabIdx = this.firstTabIdx;
+          }
         } else {
           newFocusedTabIdx = this.focusedTabIdx + 1;
         }
@@ -285,6 +305,12 @@
       focusActiveTab() {
         const activeTabIdx = this.getTabIdx(this.activeTabId);
         this.focusTab(activeTabIdx);
+      },
+      focusBeforeMoreButton() {
+        requestAnimationFrame(() => this.focusLastTab());
+      },
+      focusAfterMoreButton() {
+        requestAnimationFrame(() => this.focusFirstTab());
       },
       onClick(tabId, navigate, event) {
         this.focusedTabIdx = this.getTabIdx(tabId);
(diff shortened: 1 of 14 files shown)


AlexVelezLl:
I am not completely sure if this should be retarget to `develop` since @MisRob suggested to not introduce `KListWithOverflow` in `0.16.x`, and introducing it into `KTabsList` will introduce them in 0.16.x

lib/tabs/KTabsList.vue
```diff
@@ -63,7 +65,22 @@
         </slot>
       </button>
     </template>
-  </div>
+    <template #more="{ overflowItems }">
+      <KIconButton
+        tooltip="More"
+        icon="optionsHorizontal"
+        :style="overflowButtonStyles()"
```
AlexVelezLl, line 73: An improvement we can discuss is how can we improve the visibility of the selected tab when it is overflowed. But I think this can be a follow-up issue.

MisRob:
> I am not completely sure if this should be retarget to develop since @MisRob suggested to not introduce KListWithOverflow in 0.16.x, and introducing it into KTabsList will introduce them in 0.16.x

Yes, that's right, `develop` would be the right target here, thank you

MisRob:
@AlexVelezLl could you please loop in @radinamatic for QA and give her directions for testing to confirm [these acceptance criteria](https://github.com/learningequality/kolibri-design-system/issues/623)?

 
Screenshot from 2024-05-13 17-29-29

lib/tabs/KTabsList.vue
```diff
@@ -63,7 +65,22 @@
         </slot>
       </button>
     </template>
-  </div>
+    <template #more="{ overflowItems }">
+      <KIconButton
+        tooltip="More"
```
nucleogenesis, line 71: This should be a prop -- something like `moreButtonTooltip` because we don't have any i18n machinery setup in KDS.

So in Kolibri or Studio, we would need a translation we can pass into that prop like:

`<KTabsList :moreButtonTooltip="coreStrings('moreLabel')" /* ... */>`

That way we'd pass in the translated version of the string to be used as the tooltip label.

nucleogenesis (changes requested):
One change requested related to i18n stuff. 

I'll make time tomorrow to pull this down and run Kolibri with it to give it a spin.

AlexVelezLl:
Thank you @nucleogenesis! I have added the prop, and updated the [Kolibri PR](https://github.com/learningequality/kolibri/pull/12146/commits/3f3eae69d7a9c2214965d60c5f9e04c285aec5f6) to include it.

MisRob:
@AlexVelezLl Thanks a lot, this will be such a valuable update in many places, and it also confirms that `KTabsListWithOverflow` is easy and flexible for use.

I won't manage to do a full review here but I see @nucleogenesis is already doing some code review, and @radinamatic has started a11y testing. As soon as the acceptance criteria I referenced are satisfied, I feel okay about proceeding.

I tested briefly and noticed the following issues. I used Firefox (may be present in Chrome too but I haven't tried yet). Some of them will perhaps be related.

**(1) The responsiveness doesn't work fully and also at some point focus ring strangely appears on its own, even though I didn't use the keyboard here.** Could point to some unexpected troublesome issues in the implementation.

Peek 2024-05-18 15-44

 **(2) As soon as calculation is triggered, something unexpected happens, and all tabs are hidden to the menu instead of gradually one by one** starting with "Groups" which the only one I'd expect to be moved to the context menu here:

PEEK-2

(3) **When I click TAB when in the context menu, the focus shifts to the next item and immediately after the menu gets closed by its own.** I am not entirely sure what's expected keyboard navigation for this context menu (arrow keys work well) but pretty certain that we'll need to improve this. @radinamatic will know more about what's expected behavior. In this recording I
- TAB to the context menu icon
- ENTER to open the menu
- then tried to TAB again (causing the issue I'm describing above)

Also note that somehow we have two focus rings in the menu.
 
peek-3

AlexVelezLl:
Hi @MisRob! Thanks for your review.

1. This is not a focus ring, that border indicates that the selected tab is hidden in the more button. I followed the implementation that we had in EQM for pointing that the selected tab was hidden. I commented that [here](https://github.com/learningequality/kolibri-design-system/pull/635#discussion_r1595612589) because yes, it can be confusing, and I think the communication that the selected tab is hidden can be improved.
2. Havent tested in the kds docs, and havent catch this. But this happens because currently KListWithOverflow doesnt have support for `inline-block` parent elements, because its available width can't be determined easily. I have tried a couple of options reading the parent's parent's width but it has a lot of edge cases to calculate the available space :c. So for now the use of `display: flex` and setting a `max-height` is needed if we want to achieve something similar to inline-block. I have just updated the docs examples with this, let me know what are your thoughts on this (It's also worth saying that this KListWithOverflow restriction has yet to cause problems where we've used it, and I didnt found any inline-block parents in Kolibri's KTabLists either).
3. This is a known issue we have with KDropdownMenu (https://github.com/learningequality/kolibri-design-system/issues/588), and in previous discussions [in slack](https://learningequality.slack.com/archives/C05FSV4BEJ1/p1713820270887209) it was determined that @tab should return the focus to the next focusable element. And the two focus rings is because one is the focus ring, and the other is the border describing that the selected tab is hidden.

MisRob:
Thanks for looking into it @AlexVelezLl.

(1) Ah yes, I didn't realize the connection, and agree it's fine to do it as a follow-up issue
(2) I see. I think it's fine to operate within some limits as long as it's communicated. For calculations that need width, it is very much understandable that parent needs to have the width information. Generally speaking, I haven't found a reasonable way to do width calculations without that either and agree that it's better to ensure the component gets rendered in contexts with width available. One example would be ratio calculations on `KImg` where I made a note about it in the documentation _"Note that the ratio styles calculations need to have the width information, therefore it needs to be available in some way. For example, you can set the width directly on KImg. Alternatively, you could ensure that KImg's parent element has width by setting it explicitly or by using a block element."_ Could you add a similar guidance to
- `KListWithOverflow` docs - detailed guidance
- `KTabsList` brief note and reference to `KListWithOverflow` docs
- `KTabs` brief note and reference to `KListWithOverflow` docs

(3) I see, all good then. Also perhaps the discussion you have with @radinamatic in the related Kolibri PR will help to resolve this.

MisRob:
Now I also noticed that icons (and more generally contents of the `tabs` slot) don't get transferred to the menu so we may need to look into that at some point. I think that's a further improvement that would be fine to have a follow-up issue for.

Screenshot from 2024-05-21 11-09-12

AlexVelezLl:
Closing since another approach will be implemented to handle overflow in tabs. See https://github.com/learningequality/kolibri/pull/12146#issuecomment-2127951322

PR #648, opened 2024-05-31
KSelect: add event documentation

nucleogenesis:
## Description
 

I got stuck on an issue in Kolibri by not realizing that `@change` will trigger anytime the value / v-model changes, even if done externally. `@select` on the other hand does what I wanted... which was only react to the user actually selecting something. This guidance may have helped me :)

## Changelog

- [#648]
  - **Description:** Updates KSelect to include guiding documentation for the events it emits
  - **Products impact:** none
  - **Addresses:** n/a
  - **Components:** KSelect
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** See docs

[#648]: [PR link](https://github.com/learningequality/kolibri-design-system/pull/648)

## Steps to test

See the docs page for KSelect


MisRob (approved):
Thanks a lot @nucleogenesis, I am pretty sure this will help more of us :)

PR #649, opened 2024-06-03
KTable - Week 2

BabyElias:
## Description
- Created child component KTableGridItem for use with KTable
- Separated datatype into String, numeric,date and others. Added left alignment for numeric datatype within grid.
- Added keyboard accessibility using arrow keys
- Added sorting functionality


docs/pages/playground.vue
```diff
@@ -21,6 +21,8 @@
         { text: 'Level 2 ', link: { path: '#' } },
       ]"
     /> -->
+    <KTable :headers="tableHeaders" :rows="tableRows" caption="User Information Table" />
```
MisRob, line 24: Just a note that in the final version, `caption` and any other strings will need to be received via prop from products, such as Kolibri, because KDS doesn't contain mechanism for translations.

lib/KTable/KTable.vue
MisRob, line 1: One unwritten convention for when a component has more files in a directory is that the root component file, here `KTable.vue` is called `index.vue`. The reason is consistency and also imports are then simpler, instead of `import path/KTable/KTable.vue` we can use `import path/KTable`

lib/KTable/KTable.vue
```diff
@@ -0,0 +1,234 @@
...
+          <th
+            v-for="(header, index) in headers"
+            :key="index"
+            tabindex="0"
+            @click="handleSort(index)"
+            @keydown="handleKeydown($event, -1, index)"
+          >
+            {{ header.label }}
+            <span v-if="header.dataType !== 'others'" class="sort-icon">
+              <span v-if="sortKey === index && sortOrder === 'asc'">⬆️</span>
```
MisRob, line 19: In the next iteration, could be nice to consider using icons as illustrated in the design
table

lib/KTable/KTable.vue
```diff
@@ -0,0 +1,234 @@
...
+      caption: {
+        type: String,
+        required: false,
+        default: null,
+      },
+    },
+    data() {
+      return {
+        sortKey: null,
+        sortOrder: 'asc',
```
MisRob, line 72: Could be helpful to use constants for `'asc'`, `'desc'`, `'numeric'`, `'date'` etc. strings. You could even define them in the `useTable` composable and import in the component from there. This will help to have them available in both places (and even be imported for custom sort functions in the future)

lib/KTable/KTable.vue
```diff
@@ -0,0 +1,234 @@
...
+    props: {
+      headers: {
+        type: Array,
+        required: true,
+      },
+      rows: {
+        type: Array,
+        required: true,
+      },
+      caption: {
```
MisRob, line 63: Regarding caption prop, this is good approach. In the final version, caption and any other strings will need to be received via prop from products, such as Kolibri, because KDS doesn't contain mechanism for translations.

lib/KTable/KTable.vue
```diff
@@ -0,0 +1,234 @@
...
+
+  import KTableGridItem from './KTableGridItem.vue';
+
+  export default {
+    name: 'KTable',
+    components: {
+      KTableGridItem,
+    },
+    props: {
+      headers: {
```
MisRob, line 55: No need to do it now, rather a note for when we're finalizing API (that is shortly before using the table in Kolibri) - I recommend documenting props, slots, and events with JSDocs comments.

Some examples
- https://github.com/learningequality/kolibri-design-system/blob/release-v4/lib/KImg/index.vue#L177-L201
- https://github.com/learningequality/kolibri-design-system/blob/release-v4/lib/KImg/index.vue#L15
- https://github.com/learningequality/kolibri-design-system/blob/release-v4/lib/KImg/index.vue#L364-L366

The documentation will then use these to auto-generate the bottom section of the documentation page https://design-system.learningequality.org/kimg/#props

lib/KTable/KTable.vue
```diff
@@ -0,0 +1,234 @@
...
+        type: Array,
+        required: true,
+      },
+      rows: {
+        type: Array,
+        required: true,
+      },
+      caption: {
+        type: String,
+        required: false,
```
MisRob, line 65: Just for my understanding, what was the reason for not making the caption required? Are there some use-cases when it's not needed in regard to a11y?

lib/KTable/KTable.vue
```diff
@@ -0,0 +1,234 @@
...
+          </th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr v-for="(row, rowIndex) in sortedRows" :key="rowIndex">
+          <KTableGridItem
+            v-for="(cell, cellIndex) in row"
+            :key="cellIndex"
+            :content="cell"
+            :align="headers[cellIndex].dataType === 'numeric' ? 'right' : 'left'"
```
MisRob, line 32: I recall you mentioned that one of the motivation for having `KTableGridItem` abstraction was to encapsulate the alignment logic so if that's still the case, perhaps we could just provide`KTableGridItem` with `dataType` and let it make the decision about the alignment?

lib/KTable/KTable.vue
```diff
@@ -0,0 +1,234 @@
...
+              nextRowIndex = rowIndex - 1;
+            } else {
+              return;
+            }
+            break;
+          case 'ArrowDown':
+            if (rowIndex === -1) {
+              nextRowIndex = 0;
+            } else {
+              nextRowIndex = (rowIndex + 1) % totalRows;
```
MisRob, line 130: Clever :)!

lib/KTable/KTable.vue
```diff
@@ -0,0 +1,234 @@
...
+            }
+            break;
+          case 'ArrowLeft':
+            if (rowIndex === -1) {
+              nextCellIndex = cellIndex > 0 ? cellIndex - 1 : totalCols - 1;
```
MisRob, line 135: It seems to me that `cellIndex` is holding the column index for the cell, is that right? If so, I'd recommend renaming all related variables to `columnIndex`, `nextColumnIndex`, etc. or something similar to help a bit with clarity.

lib/KTable/KTableGridItem.vue
```diff
@@ -0,0 +1,53 @@
...
+
+</template>
+
+
+<script>
+
+  export default {
+    name: 'KTableGridItem',
+    props: {
+      content: {
```
MisRob, line 15: In the next iteration, I'd recommend using slot rather than `content` prop, since we will need to work with rich HTML content. This will make transitioning towards exposing scoped slots easier.

MisRob:
Looking great, @BabyElias! I left a few notes you can have a look later and will merge it now (but if you had any questions or things to chat about in more detail, we can still discuss here)

MisRob (approved):

lib/KTable/KTable.vue
BabyElias: No, actually. If I was not providing a default value when required was set to true, it was failing the tests. So I disabled it to false then. I'll update the default value in next PR.

lib/KTable/KTable.vue
BabyElias: Okayy!

lib/KTable/KTable.vue
MisRob: Ah I see. I think you can just set `required` to `true` and then make sure you always provide the value for it from the tests - test failure when required value is not provided is expected.

PR #650, opened 2024-06-04
Made a new KRadioButtonGroup component to fix Firefox radio button movement with arrow keys issue

### LINKED ISSUE #10491: Keyboard navigation inside radio button group (Firefox)
Sep 3 update: Will be resolved by https://github.com/learningequality/kolibri/issues/12596

## Observed behavior
Proper keyboard interaction for moving through options **inside the radio button group** is with the ⬇️ ⬆️  keys, not the <kbd>Tab</kbd> key. This is currently working well in in Kolibri when browsing with Chrome, but not with Firefox, where users can only navigate through radio buttons with the <kbd>Tab</kbd> key, even after they select/change the radio option they need with the <kbd>Space</kbd> key, they still need to Tab-their way out the radio button group to the next interactive element.

### Firefox
https://user-images.githubusercontent.com/1457929/232547281-4943cd3b-0133-4c1a-a9d1-73a4b768a253.mp4

### Chrome
https://user-images.githubusercontent.com/1457929/232547284-a0fbe332-6705-4698-86e4-8ce93002c229.mp4

---

The benefit of the above recommended navigation (use arrow keys to navigate through the radio buttons within the group, and use the <kbd>Tab</kbd> key to jump out of the radio button group) is evident when there is a long list of radio buttons like in our **Change language** modal. In Firefox users need to Tab-their-way-out though the whole list of 30+ languages even after they selected the one they want, just so they can confirm and close the modal. Not optimal... 🙅🏽 

### Firefox
https://user-images.githubusercontent.com/1457929/232547280-4a457a80-0a5f-4122-b84d-6bd02a91907c.mp4 

### Chrome
https://user-images.githubusercontent.com/1457929/232547277-d1ba7faf-68ab-4d8c-9265-4b367c6c34a1.mp4

---

I was so happy to find the [Chrome extension](https://chrome.google.com/webstore/detail/show-keys/bkcpmidbgmfbbeanejhepkjakagfaaon?hl=en) and [Firefox add-on](https://addons.mozilla.org/en-US/firefox/addon/addon_screen_key/) to display the pressed keys, that I went a bit finger-happy while recording, and I only now realize that it's maybe too fast to notice the interaction... 😅 

## Errors and logs
…

## Expected behavior
https://www.w3.org/WAI/ARIA/apg/patterns/radio/#keyboardinteraction

## User-facing consequences
Keyboard interaction not optimized on Firefox as on Chrome.

## Steps to reproduce
Try changing the language using the keyboard in Firefox.

## Context
 * Kolibri version: latest develop
 * Operating system: both Windows & Ubuntu
 * Browser: Firefox 🔴  , Chrome ✔️

muditchoudhary:
## Description

- This PR, creates a new `KRadioButtonGroup` component that fixes radio button movement with arrow keys issues in Firefox.
- It also has some Unit tests for `KRadioButtonGroup`.
- It creates docs for `KRadioButtonGroup`

#### Issue addressed

Fixes: [10491](https://github.com/learningequality/kolibri/issues/10491)

Addresses #*PR# HERE*

#### Before

https://github.com/learningequality/kolibri-design-system/assets/74391865/65716ac4-9cd1-4573-b65f-1e103bf39f25

#### After

https://github.com/learningequality/kolibri-design-system/assets/74391865/23d8d6a4-d20e-44a4-a118-7c975f61432f

## Changelog

- [#PR no]

[#PR no]: PR link

## Steps to test

1. Yarn link the `Kolibri` with KDS. How? given: [here](https://github.com/learningequality/kolibri-design-system/blob/release-v4/dev_docs/03_how_to_update_library.md#2-preview-updates-in-a-product)
2. Import `KRadioButtonGroup` component and wrap `KRadioButton` components inside `KRadioButtonGroup` component, in these files:
  2.1 kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue
  2.2 kolibri/core/assets/src/views/language-switcher/LanguageSwitcherModal.vue
3. Run Server and test in Firefox by visiting these pages:
  3.1 http://127.0.0.1:8000/en/device/#/settings
  3.2 Choose Language

#### Second way
1. Paste the given playground.vue code. [here](https://gist.github.com/muditchoudhary/6a5af21ff7e8085df572706e698291fd)
2. Run the KDS server and visit playground and test.

## Testing checklist

## Reviewer guidance

## After review

## Comments

For the change language model `KGrid` is used, which creates two `KRadioButtonGroup` components as the array is split. 
Although the arrow keys are working fine but when we shift tab it puts focus on the last focus radio button on the first list.
I want to know your thoughts. Is that acceptable?


muditchoudhary:
Hi @MisRob Could you please review the PR?

MisRob:
Hi @muditchoudhary, thanks a lot! I think as first step, we will QA in Kolibri. Would you be able to open a Kolibri PR for testing that would reference the latest commit of this KDS PR and use `KRadioButton` at least at one Kolibri place? Our QA team doesn't work with code directly. If you couldn't do that, I can prepare the PR as well.

muditchoudhary:
Sure @MisRob I had tested on Kolibri and I have the changes in stash, so I will have no problem in making a PR. I will send a PR soon.

lib/__tests__/KRadioButtonGroup.spec.js
```diff
@@ -0,0 +1,96 @@
...
+            '<KRadioButton label="Option A" buttonValue="val-a" />',
+            '<KRadioButton label="Option B" buttonValue="val-b" />',
+            '<KRadioButton label="Option C" buttonValue="val-c" />',
+          ],
+        },
+      });
+      await wrapper.vm.$nextTick();
+      const radioButtons = wrapper.findAllComponents(KRadioButton);
+      await radioButtons.at(2).trigger('click');
+      expect(wrapper.vm.focusedRadioIdx).toBe(2);
```
MisRob: It'd be great if we could test that this radio button is focused without dependency on its internal implementation, that is looking at `vm.focusedRadioIdx`. Imagine a situation when we'd rename/remove `focusedRadioIdx`, but in a way that focus would still work properly for users. In such a situation, this test would give negative result, even though there would be no problem for users. So for maintenance and clarity benefits, whenever possible, we to implement tests in a way that's similar to how the user interact with elements.

I think this technique should do https://stackoverflow.com/a/53042010 (but with `attachTo` as `attachToDocument` is deprecated)

lib/__tests__/KRadioButtonGroup.spec.js
```diff
@@ -0,0 +1,96 @@
...
+      expect(radioButtons.at(2).vm.tabIndex).toBe(0);
+      expect(radioButtons.at(1).vm.tabIndex).toBe(-1);
```
MisRob: For the same reason as mentioned above, here I'd recommend looking into `attributes` of the DOM node rather than into `vm`'s private `data`. I assume it could look something like `expect(radioButtons.at(1).attributes('tab-index')).toBe(0)`.

lib/KRadioButtonGroup.vue
```diff
@@ -0,0 +1,133 @@
...
+<script>
+
+  export default {
+    name: 'KRadioButtonGroup',
+    props: {
+      /**
+       * Specifies whether the radio button group is enabled or disabled.
+       * Used to make the first radio button active when the radio group becomes enabled again after being disabled.
+       */
+      enable: {
```
MisRob: I don't quite understand yet in what case we'd need to use `enable` functionality, would you explain on an example?

lib/KRadioButtonGroup.vue
```diff
@@ -0,0 +1,133 @@
...
+        if (newVal && !oldVal && this.radioButtons.length > 0) {
+          this.setChecked(0);
+          this.focusedRadioIdx = 0;
+        }
+      },
+    },
+    mounted() {
+      this.$nextTick(() => {
+        if (this.isFirefox) {
+          if (this.$children[0].$options._componentTag === 'KGridItem') {
```
MisRob: Instead of checking specifically for `KGridItem`, could we recursively search through `$children` and their `$children` etc. until we find the level of radio buttons?
 
This would make `KRadioButtonGroup` work with any other component that wraps its `KRadioButton`s, such as `KFixedGridItem` or even HTML elements such as `div`.

The alternative would be to properly document that only this single use-case is supported, however it seems to me that this suggestion shouldn't hopefully be complicated and I believe it would also help with clarity and deduplicated some of the code.

lib/KRadioButtonGroup.vue
```diff
@@ -0,0 +1,133 @@
...
+            this.$children.forEach(gridItem => {
+              gridItem.$children.forEach(fixedGridItem => {
+                fixedGridItem.$children.forEach(radioBtn => {
+                  this.radioButtons.push(radioBtn);
+                  radioBtn.$el.addEventListener('keyup', this.onKeyUp);
```
MisRob: In larger groups of radio buttons, such as the language selection, adding an event listener to each one of them could potentially have negative performance impact, particularly in combination with looping through many elements. A general recommendation we follow in our codebases is to delegate listeners to a container element higher in the structure, so that we have as less listeners as possible.

In this case, `KRadioButtonGroup` element when `mounted` (and only in Firefox) would attach a single `keyup` event listener. This event's handler function would then determine whether `event.target` was a radio button, and if yes, it would run the logic needed.
  
I would also recommend doing the very same thing for the `input` event (line 71)

So you'd have something like

```
mounted() {
  this.$el.addEventListener('keyup', onKeyUp); 
  this.$el.addEventListener('input', onInput);
  ...
```

and these two would be the only event listeners in `KRadioButtonGroup`.

Related logic would need to be adjusted to work in this updated way.

lib/KRadioButtonGroup.vue
```diff
@@ -0,0 +1,133 @@
...
+      enable(newVal, oldVal) {
+        if (newVal && !oldVal && this.radioButtons.length > 0) {
+          this.setChecked(0);
+          this.focusedRadioIdx = 0;
+        }
+      },
+    },
+    mounted() {
+      this.$nextTick(() => {
+        if (this.isFirefox) {
```
MisRob: Just a minor suggestion that could help with code readability. This would avoid one level of nesting:

```
mounted() {
  if (!this.isFirefox) {
    return;
  }
  this.$nextTick(() => {
  ...
  ...
  ...
```

lib/KRadioButtonGroup.vue
```diff
@@ -0,0 +1,133 @@
...
+      };
+    },
+    computed: {
+      isFirefox() {
+        return navigator.userAgent.toLowerCase().indexOf('firefox') > -1;
+      },
+    },
+    watch: {
+      // To focus on the first radio button when radio group enables again
+      enable(newVal, oldVal) {
```
MisRob: I believe that before running the code below, we'd also need `isFirefox` check here

lib/KRadioButtonGroup.vue
```diff
@@ -0,0 +1,133 @@
...
+      setChecked(radioIdx) {
+        for (let i = 0; i < this.radioButtons.length; i++) {
+          const radioBtn = this.radioButtons[i];
+          radioBtn.setTabIndex(-1);
+        }
+        this.radioButtons[radioIdx].setTabIndex(0);
+      },
+      focusRadio(radioIdx, event) {
+        if (radioIdx !== undefined && radioIdx !== null) {
+          this.radioButtons[radioIdx].toggleCheck(event);
```
MisRob, line 127: Since we're calling `KRadioButton`'s (currently private) `toggleCheck` method, could you please add `@public` JSDoc comment to `toggleCheck`, similarly to how you did for `setTabIndex`? It will ensure that if refactoring `KRadioButton`, we will know that the method is being used from another component and don't remove/rename it accidentally.

lib/KRadioButton.vue
```diff
@@ -220,6 +222,13 @@
          */
         this.$emit('blur');
       },
+      /**
+       * @public
+       * Set the tabIndex value
```
MisRob: Thank you for making sure its documented. Just a little improvement for clarity:

```suggestion
       * Sets `tab-index` on input
```

MisRob (commented):
Thanks so much @muditchoudhary, firstly for researching the issue and approaches to solve it, and then in this PR for finalizing it. Also thanks for taking care of docs and tests. Overall it's great work and feels solid.

I'm leaving a couple of comments. The most important is the one related to many event listeners and it's the only change I request before proceeding to merging. I'd also like to clarify the `enable` logic.

I'd appreciate if we could have the improvement regarding the dependency to `KGridItem`, however it's also something that I could open a follow-up issue for. Same applies to the remaining comments that are rather minor. It really depends on your bandwith - let me know what works for you.

muditchoudhary:
Thanks a lot, @MisRob for giving the detailed review and suggesting changes. I will send the changes soon.

lib/KRadioButtonGroup.vue
muditchoudhary: > I don't quite understand yet in what case we'd need to use `enable` functionality, would you explain on an example?

https://github.com/user-attachments/assets/df3498f8-e6e0-40f5-858a-a0e5d114dfe7

It was a redundant logic. In the settings page earlier I used two KRadioButtonGroup, 

first to wrap **Learn-page** and **Sign-in page** options
second to wrap **below three radio buttons.**

What used to happen was after we had selected an option from those 3 radio buttons and when we chose **Learn-page** option that disables the **Sign-in page** options. Then when we again chose **Sign-in page** radio, it enables its options but instead of the first option being selected by default the last option which we have chosen before used to be selected. 
For that, I have used that **enable prop** logic. I check if the old enable value is false and the new enable value is true then select the first sign-in page option and the old option must not remain selected.

But now it does not happen because I use only a single KRadioButtonGroup component to wrap all these 5 radio buttons.

lib/KRadioButtonGroup.vue
muditchoudhary: > I would also recommend doing the very same thing for the `input` event (line 71)

For the `keyup` listener I am able to have the listener only on the parent `KRadioButtonGroup` But I have issues with putting `input` listener on the parent.

So to query all the `KRadioButton` and put all those in a list I have used recursion as you have told me.
```vue
// const KRadioButtonGroupChildren = this.$children;
// this.queryAndAddRadioBtns(KRadioButtonGroupChildren);

 queryAndAddRadioBtns(children) {
        if (children[0].$options._componentTag === 'KRadioButton') {
          for (let i = 0; i < children.length; i++) {
            const radioBtnVueComp = children[i];
            this.radioButtons.push(radioBtnVueComp);
            radioBtnVueComp.setTabIndex(-1);
          }
          return;
        }
        this.queryAndAddRadioBtns(children[0].$children);
        if (children.length > 1) this.queryAndAddRadioBtns(children[1].$children);
        return;
}
```

Now the issue is, 

1. I cannot listen for `input` on `KRadioButtonGroup` component as `input` is emitted by `KRadioButton`. 
 ```vue
 this.$on('input', () => {}); // no no won't work
 ```
I tried to find some answers if some way we can listen for the child event on the parent but did not find anything useful.
If you know how we could listen for the `input` event on parents where the children are rendered with the help of a slot. If you know it would be a great help.

2. The second thing I tried is having an `input` listener and `click` listener on the `KRadioButtonGroup`
and during `queryAndAddRadioBtns` we set a custom `btnIndex` attribute to each input element to know each radio btn index for later purposes.

But when we click on `radio icon` I thought we would get click on `input` element but **NO** we get an event on 
`radio icon svg` element or `div containing the label` (if clicks on the label). Because of this my logic failed because I checked 'k-radio-button-input` className on  `onKeyUp` method
```vue
onKeyUp(event) {
        if (event.target.className === 'k-radio-button-input') {
          const handlers = {
            ArrowLeft: this.focusPreviousRadio,
            ArrowRight: this.focusNextRadio,
            ArrowUp: this.focusPreviousRadio,
            ArrowDown: this.focusNextRadio,
          };
          const handler = handlers[event.key];
          if (handler) {
            event.preventDefault();
            event.stopPropagation();
            handler(event);
          }
        }
}
```
So could you help me here what I can do? currently what I have moved on with is having `keyup` event only on `KRadioButtonGroup` and `on input` on every `KRadioButton`.

muditchoudhary:
Hi @MisRob I have made some changes in the code as per your suggestions [NOT PUSHED THE COMMITS YET] . All the things are good except with `input` listener for which I have asked you for some help [here](https://github.com/learningequality/kolibri-design-system/pull/650#discussion_r1703135786). 

And regarding the `enable` logic I have answered [here](https://github.com/learningequality/kolibri-design-system/pull/650#discussion_r1703125699)

lib/KRadioButtonGroup.vue
MisRob: I see, thanks for explanation. I think we can remove it then. If there are some situations when we need to focus the first radio button, I think we can use `KRadioButton`s public `focus` method for the same purpose.

lib/KRadioButtonGroup.vue
MisRob: I see, thank you. 

The main part of my feedback was concerned with `keyup` and I only added `input` because I assumed it would work as well and it'd be clearer to do all events in the same way. I looked into the latter briefly now and I think the issue with that is that it's rather Vue specific event and it doesn't seem to propagate. I had no success with `change` event either.

So here I think we could stay with the original approach?

```
radioBtn.$on('input', () => {
    this.handleInputChange(index);
  });
```

(but while still making sure there's only one `keyup` event listener which is the main issue)

 I'm sorry this was such a hassle, I didn't expect that `input`/`change` will be troubling.

MisRob:
Thanks a lot @muditchoudhary! Just answered. Seems we're close, great work.

lib/KRadioButtonGroup.vue
MisRob: I think this should still be possible to do with the recursion approach

MisRob: As far as I can say, I think we can do `input` handler assignment during the recursion,  around time when you're saving a radio button to the list, right?

muditchoudhary: > As far as I can say, I think we can do `input` handler assignment during the recursion, around time when you're saving a radio button to the list, right?
Thank you!! 

Yes, we can do that in the recursion method too.

MisRob: Okay, good!

muditchoudhary:
Hi @MisRob I have pushed the commit. Could you please check?

Also, please confirm I need to target the develop branch, right?

Thank you!!

muditchoudhary:
There were some much mege conflicts with `KCard` too. I resolved it

MisRob:
Thanks @muditchoudhary! Yes, `develop`, that's right. Would you mind also updating https://github.com/learningequality/kolibri/pull/12325 to point to this latest work, so that we can QA again in Kolibri? There were lots of changes, so it'd be great to re-test.

MisRob:
> Would you mind also updating https://github.com/learningequality/kolibri/pull/12325 to point to this latest work, so that we can QA again in Kolibri? There were lots of changes, so it'd be great to re-test.


@muditchoudhary I have some time so I will do this now

lib/KRadioButtonGroup.vue
```diff
@@ -0,0 +1,118 @@
...
+    },
+
+    methods: {
+      handleInputChange(idx) {
+        if (this.isFirefox) {
+          this.setChecked(idx);
+          this.focusedRadioIdx = idx;
+        }
+      },
+      queryAndAddRadioBtns(children) {
```
MisRob, line 70: The recursion never gets to a third child of a childrens array. I know that our motivating use-case has only two grid items so it works there, but the point of recursion here is to make it work no matter what. It'd be best to fix it so it's as robust as possible. You can reproduce by

```
<KGrid>
  <KRadioButtonGroup>
    <KGridItem>
      <KRadioButton
        v-model="exampleValue"
        label="Option A"
        value="val-a"
      />
      <KRadioButton
        v-model="exampleValue"
        label="Option B"
        value="val-b"
      />
    </KGridItem>
    <KGridItem>
      <KRadioButton
        v-model="exampleValue"
        label="Option C"
        description="This one is special!"
        buttonValue="val-c"
      />
      <KRadioButton
        v-model="exampleValue"
        label="Truncated label. Adjusting your browser window size to see this in action."
        buttonValue="val-d"
        truncateLabel
      />
    </KGridItem>
    <KGridItem>
      <KRadioButton
        v-model="exampleValue"
        label="Option E"
        description="This one is special!"
        buttonValue="val-e"
      />
      <KRadioButton
        v-model="exampleValue"
        label="Option F"
        buttonValue="val-f"
        truncateLabel
      />
    </KGridItem>
  </KRadioButtonGroup>
</KGrid>
```

and then see it's not possible to navigate with keyboard to the last two checkboxes:

Peek 2024-08-16 12-57

MisRob: I think it has to do with the way it is structured - the only thing that makes the function run on a sibling child is `if (children.length > 1) this.queryAndAddRadioBtns(children[1].$children);` but that's not sufficient and can't cover all children.

For recursions, this is a typical structure:

```
FUNCTION traverseChildren(array)
  FOR EACH child IN array
    PROCESS child
    IF child HAS children
      CALL traverseChildren(child.children)
  END FOR
END FUNCTION
```

I think that using it should resolve it.

If that's helpful, one thing that helps to check recursion implementation is to test on crazy unexpected nested structures :) If recursion is implemented correctly it will always work on all children on all levels.

For debugging, this piece may help:
```
queryAndAddRadioBtns(children) {
   console.log('[queryAndAddRadioBtns]', children.map(child => child.$el));
   ...
```

Then just have a look at a browser console and you will see in detail how your recursion progresses on elements and if it covers them all.

MisRob: Minor additional note - since this is a bit more complex piece, would you please add a comment above the function briefly summarizing what it does and why (example of grid usage)?

lib/KRadioButtonGroup.vue
```diff
@@ -0,0 +1,118 @@
...
+        this.focusRadio(newFocusedRadioIdx, event);
+      },
+      focusNextRadio(event) {
+        const newFocusedRadioIdx =
+          this.focusedRadioIdx === this.lastRadioIdx
+            ? this.firstRadioIdx
+            : this.focusedRadioIdx + 1;
+        this.focusRadio(newFocusedRadioIdx, event);
+      },
+      setChecked(radioIdx) {
```
MisRob: Minor note, reading through what those methods actually do, it seems to me that `setChecked` would better be named `focusRadio` and `focusRadio` would better be named `setChecked` :-)?

MisRob (dismissed):
@muditchoudhary Thanks again, Mudit. This is fantastic progress and I think we're close to merge. Ccan't wait to see it live (as a Firefox user myself :) 

The only important thing to resolve is the recursion bug. Besides I left a couple of minor non-blocking notes that are optional.

muditchoudhary:
@MisRob Thank you!! so much :heart:  for reviewing the code and helping me with the recursive logic. 
I have updated the recursive logic and methods name as you have suggested. I updated the KDS fork version too with latest commit for QA.

Once the QA successfully tests the PR, I will rebase this branch with develop for merging.

Edit:
Here is the playground.vue code which I have used for testing KRadioButtonGroup with new recursive logic
https://gist.github.com/muditchoudhary/66f1663e457d5dd455d5190b7058ccd5

MisRob:
Great @muditchoudhary, thanks a lot too! Joy on my side. I will have a look at the latest changes, hopefully some time this week, and also ask my colleagues for final QA run on Kolibri. And thank you for sharing the playground code, that's helpful.

MisRob:
@muditchoudhary All new code looks well to me, and testing here on KDS passes. We will wait for the final QA confirmation from https://github.com/learningequality/kolibri/pull/12325 and then I will merge.

FYI I updated the changelog. I appreciate you started the documentation page with the example. I pushed some more updates to the documentation. Cross-referenced with `KRadioButton` and moved your example to `KRadioButton` to  ensure it's clear that `KRadioButton` is meant to always used within `KRadionButton` group.

I think that as soon as https://github.com/learningequality/kolibri/pull/12325 passes and there is a new KDS release with this PR merged in it, we can install the release to https://github.com/learningequality/kolibri/pull/12325 and even merge it to Kolibri. There are still many other places in Kolibri where we will want to wrap `KRadioButton`'s - I can open a follow-up issue for that.

Thanks a lot for this solid piece of work - progressing from the research, communication with @radinamatic, proof of concept, to the final implementation. So glad this is completed :)

MisRob:
@muditchoudhary I also added the example with grids to docs, since I don't think people would figure out this is actually supported :)!

muditchoudhary:
Thanks a lot @MisRob for all the help and support. I'm excited too, to see this PR merging in the codebase. 

Also thank you for updating the docs.

MisRob:
QA in https://github.com/learningequality/kolibri/pull/12325 passed so merging. Great work @muditchoudhary.

MisRob (approved):

muditchoudhary:
Thank you :) @MisRob

PR #654, opened 2024-06-09
KTable - Week 3

BabyElias:
## Description
- Added sorting as a composable function `(useSorting.js)`
- Included scope for external sorting function (when data received from backend) using `useLocalSorting` prop, still work in progress though
- Added scoped slots
- Made sorting an optional feature with `sortable` prop
- Fixed naming convention for files and icons for sorting
- Added constants for strings


docs/pages/playground.vue
```diff
@@ -21,7 +21,39 @@
...
+    <h1>KTable Component Example</h1>
+    <h2>Sortable Table</h2>
+    <KTable
+      :headers="headers"
+      :rows="rows"
+      caption="Sortable Table"
+      :useLocalSorting="true"
+      sortable
+    >
+      <template #header="{ header, index }">
```
MisRob, line 33: Great, I appreciate you also thought of headers customization.

docs/pages/playground.vue
```diff
@@ -21,7 +21,39 @@
...
+        <span>{{ header.label }} (Custom)</span>
+      </template>
+      <template #cell="{ content, rowIndex, colIndex }">
```
MisRob, line 36: Cool :)!

lib/KTable/KTableGridItem.vue
```diff
@@ -16,22 +18,27 @@
...
       },
-      cellIndex: {
+      colIndex: {
         type: Number,
         required: true,
       },
     },
+    computed: {
+      textAlign() {
+        return this.dataType === 'numeric' ? 'right' : 'left';
```
MisRob, line 36: Nice thing about having the constants in the composable is that you could then import and use them in this file as well

lib/KTable/KTableGridItem.vue
```diff
@@ -16,22 +18,27 @@
         type: [String, Number],
         required: true,
       },
-      align: {
+      dataType: {
```
MisRob, line 21: For all `KTable` and `KTableGridItem` props where we expect a set of limited values, it'd be great to have validator functions as well. This can be done during the stage of API finalization. Examples https://github.com/learningequality/kolibri-design-system/blob/release-v4/lib/KImg/index.vue#L172-L175

lib/KTable/index.vue
```diff
@@ -0,0 +1,240 @@
...
+      <caption v-if="caption">
+        {{ caption }}
+      </caption>
+      <thead>
+        <tr>
+          <th
+            v-for="(header, index) in headers"
+            :key="index"
+            tabindex="0"
+            :aria-sort="sortable ? getAriaSort(index) : null"
```
MisRob, line 14: Do we need to use `sortable && header.dataType !== DATA_TYPE_OTHERS` here as well or is it not needed? Same for `@click` below.

lib/KTable/useSorting.js
MisRob, line 1: Nice overall! Some high-level feedback for next iterations:

- Could you please place the file to `lib/composables`? That's where all composables are kept. See the directory for examples.
- Would recommend developing a test suite for this composable as you keep working. You can see some examples of tests for composables in `lib/composables`. Please test sorting of all data types and for dates use the standard that we chatted about in https://learningequality.slack.com/archives/C0LE9NLCE/p1717788213871049.

docs/pages/playground.vue
MisRob, line 1: I love seeing working examples for various modes! And I think it's handy to have the option to sort by default in the table without the need to use the composable from outside it, appreciate the idea (even though I'd suggest to keep both options available, see below).

As you keep on sorting logic architecture, could we also add some more examples for the following use-cases we talked about together, particularly

- Sorting done on backend (that's one of the Kolibri use cases we're working towards)
- Sorting done on frontend with `useSorting` but from outside the `KTable` (will ensure flexibility in regard to overriding any sorting logic in the future easily)

These are not to replace the use-cases you already have, but are additions. I think that having examples for them early on will ensure that as you keep building new logic, it supports the table in all its expected modes. To sum it up, in this project we will work with
- Sorting turned off
- Already sorted data received from backend (`useSorting` not utilized by the table internally neither externally)
- Sorting on frontend and internally by table via `useSorting`
- Sorting on frontend but only externally via `useSorting` (that is, the table doesn't use sort internally)

Also, we'll want to have documentation for all these use-cases on the documentation pages for `KTable` and `useSorting`. This could be done as soon as we're finalizing the table API or, if writing how something is used helps with development, feel free to draft it as you develop, it's really up to you.

MisRob (approved):
Wonderful progress! Thanks @BabyElias. I left a few tips and directions, if you'd like to discuss any of these, please post comment even after I merge, or we can chat in one of our meetings.

lib/KTable/useSorting.js
BabyElias: Table
-emits event
-sorting (not all datatypes)
-stores data well

Composable
- Test for sorting of all datatypes

lib/KTable/index.vue
BabyElias: create a separate function to check for sortability

PR #655, opened 2024-06-13
KTable - [Trial] Added changeSort event handler

BabyElias:
## Description

- Added changeSort event handler as discussed for implementation of backend sorting logic. For demo purposes, this table imitates the behaviour of 'Externally sorted on Frontend' tables since there was no backend API that could be used for testing.
- Enhanced KTable testing conditions.


MisRob:
Thanks @BabyElias, this reflect well everything we talked about and am pretty confident that it will make the usage of the table for the Kolibri's facility users seamless ;)

MisRob (approved):

lib/KTable/index.vue
```diff
@@ -82,6 +88,25 @@
         }
       });
 
+      watch(
```
MisRob, line 91: Yes, this is an important piece to ensure reactivity

lib/__tests__/KTable.spec.js
```diff
@@ -29,9 +39,34 @@ describe('KTable.vue', () => {
...
         headers,
         rows,
+        useLocalSorting: true,
       },
     });
     const trElements = wrapper.findAll('tbody tr');
     expect(trElements.length).toBe(rows.length);
   });
+
+  it('should emit changeSort event on header click when useLocalSorting is false', async () => {
```
MisRob, line 49: Nice, thank you

PR #657, opened 2024-06-17
[KTable] Accessibility for Loading States

BabyElias:
- Tried implementing accessibility for loading states, doesn't work still though
- Refined sorting for dates in UTC format using lodash sorting utility

### DIFF
--- a/docs/pages/playground.vue
+++ b/docs/pages/playground.vue
@@ -2,16 +2,18 @@
 
   <div id="app">
     <h1>KTable Component Example</h1>
-    <h2>Sortable Table with Local Sorting</h2>
+
+    <!-- Local Sorting Table Example -->
+    <h2>Local Sorting Table</h2>
     <KTable
       :headers="headers"
-      :rows="localRows"
-      caption="Sortable Table"
+      :rows="rows"
+      caption="Local Sorting Table"
       :useLocalSorting="true"
       sortable
     >
       <template #header="{ header, index }">
-        <span>{{ header.label }} (Local Sort)</span>
+        <span>{{ header.label }} (Local)</span>
       </template>
       <template #cell="{ content, rowIndex, colIndex }">
         <span v-if="colIndex === 1">{{ content }} years old</span>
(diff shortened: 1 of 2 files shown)


docs/pages/playground.vue
```diff
@@ -103,4 +151,30 @@
...
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  margin: -1px;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  border-width: 0;
+  visibility: hidden;
```
MisRob: Hi @BabyElias, followig-up on our chat. So I played around this page for a bit and I think that this line is causing trouble. When I keep it, my Orca screen reader doesn't announce the updates. After removing it, it does. I think it is caused by

> Using a visibility value of hidden on an element will remove it from the [accessibility tree](https://developer.mozilla.org/en-US/docs/Learn/Accessibility/What_is_accessibility#accessibility_apis). This will cause the element and all its descendant elements to no longer be announced by screen reading technology. ([source](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility#accessibility_concerns))

MisRob: Here's [some guidance on hiding the ARIA live regions](https://www.sarasoueidan.com/blog/accessible-notifications-with-aria-live-regions-part-2/#choose-an-appropriate-hiding-technique-if-the-live-region-isn%E2%80%99t-visible):



I think you don't need to be much worried about it here though since it's just for testing, so whatever technique works for the screen reader will do.

lib/composables/useSorting.js
```diff
@@ -14,16 +15,7 @@ export default function useSorting(headers, rows, useLocalSorting) {
...
-      const aValue = a[sortKey.value];
-      const bValue = b[sortKey.value];
-
-      if (sortOrder.value === SORT_ORDER_ASC) {
-        return aValue > bValue ? 1 : aValue < bValue ? -1 : 0;
-      } else {
-        return aValue < bValue ? 1 : aValue > bValue ? -1 : 0;
-      }
-    });
+    return _.orderBy(rows.value, [row => row[sortKey.value]], [sortOrder.value]);
```
MisRob, line 18: Neat!

docs/pages/playground.vue
MisRob, line 1: The loading demo is looking great

MisRob:
@BabyElias I think we can merge after you commit the screen reader fix and confirm it works for you as well

docs/pages/playground.vue
BabyElias: Ahh, just removing that makes the NVDA screen reader on my system too announce the loading states. Thank you so much for your time in debugging this! Made the required changes.

MisRob: 🙂

PR #645, opened 2024-05-22
Node.js upgrade to v18 along with dependency upgrades and linting fixes

bjester:
## Description

Upgrades Node.js to v18, along with Kolibri-Tools to v0.16, Jest to v29 and Nuxt to v2.15. Additionally, this pins dependencies in order to maintain compatibility between Nuxt and kolibri-tools, applies fixes for new linting errors, and cleans up testing output through better error handling.

#### Issue addressed

Addresses https://github.com/learningequality/kolibri-design-system/issues/439

## Changelog

  - **Description:** Upgrades Node.js to v18, along with Kolibri-Tools to v0.16, Jest to v29 and Nuxt to v2.15
  - **Products impact:** none
  - **Addresses:** #439 
  - **Components:** none
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** Netlify configuration needs to be updated to use v18 as well

  - **Description:** Component error handling now use `@error` listener to avoid polluting test output, nor suppressing `console.*` in tests 
  - **Products impact:** any
  - **Addresses:** n/a
  - **Components:** `KImg`, `KTabs`, `KTabsList`
  - **Breaking:** yes
  - **Impacts a11y:** no
  - **Guidance:** The `KImg` component may now emit an `Error` object in its `@error` listener when incorrectly configured, in addition to an `UiEvent|Event` when an image fails to load. Consumers should type check the value.

## Steps to test

1. Run `yarn run test`
2. Verify tests pass
3. Run `yarn run dev`
4. Verify documentation

### At a high level, how did you implement this?

- Upgraded Node.js
- Upgraded dependencies to highest version we could support
- Cleaned up test output throwing warnings and other errors
- Pinned dependency resolutions as needed

### Does this introduce any tech-debt items?

Yes https://github.com/learningequality/kolibri-design-system/issues/646

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [x] Critical and brittle code paths are covered by unit tests
- [x] The change is described in the changelog section above

## Reviewer guidance

- [x] Is the code clean and well-commented?
- [x] Are there tests for this change?
- [ ] Are all UI components LTR and RTL compliant (if applicable)?
- [ ] _Add other things to check for here_

## After review

- [ ] The changelog item has been pasted to the `CHANGELOG.md`

PR #670, opened 2024-06-21
Percy and jest-puppeteer environment setup for visual testing

KshitijThareja:
## Description

This PR introduces the environment setup and configuration for introducing visual testing to the KDS repository.

## Steps to test

1. Sign up for a Percy.io account and create a new percy webapp project.
2. Install all the dependecies by running `yarn install` in the root directory of your local copy of KDS.
3. Copy the project token and set it as an environment variable.
4. Run `yarn test:visual` to execute visual tests.

To implement this, several new packages like `jest-puppeteer`, `@percy/cli`, etc were added to the project's dependecies and several changes were made to the current testing environment to allow for visual tests to be run along with the unit tests.

--

## Testing checklist

## Reviewer guidance

## After review


.eslintrc.js
```diff
@@ -20,4 +20,19 @@ esLintConfig.settings['import/resolver'].nuxt = {
   nuxtSrcDir: 'docs',
 };
 
+// Remove linting errors for the globals defined in the jest-puppeteer package
+esLintConfig.env = {
+  ...esLintConfig.env,
+  'jest': true,
```
AlexVelezLl: I dont think we should add this here, because this lint config is for the whole package, including the development environment where we are not running jest, and I think this could cause unwanted behaviours. Probably we could set these settings in a specific file for tests?

.eslintrc.js
```diff
@@ -20,4 +20,19 @@ esLintConfig.settings['import/resolver'].nuxt = {
...
+};
+
+esLintConfig.globals = {
```
AlexVelezLl, line 24: This is the only place where we can add this? What if we add these global vars in the jest.conf file corresponding to the visual testing, I tihnk it would be more appropriate to put it there

.percy.yml
```diff
@@ -0,0 +1,19 @@
+version: 2
+snapshot:
+  widths:
+    - 375
+    - 1280
+  minHeight: 1024
+  percyCSS: ""
+  enableJavaScript: false
```
AlexVelezLl: I think we could add a comment here explaining why one of these props `enableJavaScript` is false but `cliEnableJavaScript` is true. So anyone else who read this code could understand the reasons of this decision

jest.conf/visual.index.js
```diff
@@ -0,0 +1,32 @@
+const path = require('node:path');
+
+process.env.JEST_PUPPETEER_CONFIG = require.resolve('../jest-puppeteer.config.js');
```
AlexVelezLl: I think we already have this env variable declared in the script in the package.json isnt it? Would we need to define it again?

jest.conf/setup.visual.js
```diff
@@ -0,0 +1,72 @@
+import 'regenerator-runtime/runtime';
+import '@testing-library/jest-dom';
+import * as Aphrodite from 'aphrodite';
+import * as AphroditeNoImportant from 'aphrodite/no-important';
+
+// eslint-disable-next-line import/no-unresolved
```
AlexVelezLl: It seems like most of this file is almost the same as setup.js, so I think that we could just import setup.js and adapt the things that we need on top of that for it to run with visual tests.

jest.conf/testUtils.js
```diff
@@ -10,6 +10,17 @@ export const resizeWindow = (width, height = 768) => {
   });
 };
 
+export function takeScreenshot() {
```
AlexVelezLl: It seems like this method is used to check if we can take a screenshot isnt it? So probably a better naming like "canTakeScreenshot" could work better, because "takeScreenshot" sounds like this method actually takes the screenshot

jest.conf/setup.visual.js
```diff
@@ -0,0 +1,72 @@
...
+import 'mock-match-media/jest-setup';
```
AlexVelezLl: I think we could standarize better the naming/organization of the visual testing config files. We have a `setup.visual.js` and a `visual.index.js`, I think we could prefer one of both, if we want to prefix the `visual.` or put it before the `.js`. I would prefer to have both `visual.setup.js` and `visual.index.js` so they will be sorted together. Other option could be to have them in a subfolder, but this is something we can discuss.

start-server-and-test.js
```diff
@@ -0,0 +1,102 @@
+const { spawn } = require('child_process');
+const fetch = require('node-fetch');
+
+/* eslint-disable no-console */
+
+const serverUrl = 'http://localhost:4000';
+const serverTimeout = 360000; // 6 minutes
```
AlexVelezLl: This means that the max time the visual test could take is 6 minutes right? Why 6 minutes?

start-server-and-test.js
```diff
@@ -0,0 +1,102 @@
+const { spawn } = require('child_process');
+const fetch = require('node-fetch');
+
+/* eslint-disable no-console */
+
+const serverUrl = 'http://localhost:4000';
```
AlexVelezLl: For these types of constants, we use to name them in caps like `SERVER_URL` or `SERVER_TIMEOUT`

start-server-and-test.js
```diff
@@ -0,0 +1,102 @@
...
+    }
+    await new Promise(resolve => setTimeout(resolve, 1000));
+  }
+  throw new Error('Server did not start within the timeout period');
+};
+
+const killPortProcess = async port => {
+  try {
+    const fkill = await import('fkill');
+    await fkill(`:${port}`, { force: true });
```
AlexVelezLl: I dont think we should kill other user processes, usually if the port is already allocated, then we just throw an error and wait for the user to kill them, but not us. And without this, I think we can get rid of the `fkill` dependency

start-server-and-test.js
```diff
@@ -0,0 +1,102 @@
...
+      [
+        'percy',
+        'exec',
+        '-v',
+        '--',
+        'jest',
+        '--config',
+        'jest.conf/visual.index.js',
+        '-i',
+        './lib/KImg/__tests__/KImg.spec.js',
```
AlexVelezLl: Is this temporal right? Just until we finish setting up percy.

start-server-and-test.js
AlexVelezLl: nvm just noticed that this was the sever start timeout right?

start-server-and-test.js
```diff
@@ -0,0 +1,102 @@
+const { spawn } = require('child_process');
```
AlexVelezLl: I think this file could be named as "startVisualTests" or "runVisualTests", and running the visual tests imply starting the dev server. What do you think?

lib/KImg/__tests__/KImg.spec.js
```diff
@@ -1,92 +1,104 @@
...
-    expect(error).toHaveBeenCalled();
-    expect(error.mock.calls[0][0]).toBeInstanceOf(Event);
-    expect(error.mock.calls[0][0]).toEqual(e);
-  });
+      expect(error).toHaveBeenCalled();
+      expect(error.mock.calls[0][0]).toBeInstanceOf(Event);
+      expect(error.mock.calls[0][0]).toEqual(e);
+    });
+  }
+  if (takeScreenshot()) {
```
AlexVelezLl: Hmm I think that we sould prefer this to be more transparent for developers. I think what @bjester was suggesting was to run the unit tests and that we could run an "snapshot" function that will only be excecuted if we were on visual tests mode. Something like
```js
    it(`renders without any errors when a valid 'src' and 'altText' are provided`, () => {
      const error = jest.fn();
      const wrapper = makeWrapper({ // This method should also allow to render the component dynamically on the dev server, so we can take the snapshot later.
        propsData: { src: '/le-logo.svg', altText: 'LE logo' },
        listeners: { error },
      });

      takeSnapshot(); // This will call an util function that will take the percy snapshot if we are running visual tests. If we are running unit tests we will just ignore it.

      expect(wrapper.exists()).toBe(true);
      expect(error).not.toHaveBeenCalled();

      const img = wrapper.find('img');
      expect(img.exists()).toBe(true);
      expect(img.attributes('src')).toBe('/le-logo.svg');
      expect(img.attributes('alt')).toBe('LE logo');
    });
```

I think that was what @bjester was talking about, but you can correct me if I misunderstood :open_hands:

lib/KImg/__tests__/KImg.spec.js
```diff
@@ -1,92 +1,104 @@
...
+    describe('KImg Visual Tests', () => {
+      it('renders correctly (mock test)', async () => {
+        await global.page.goto('http://localhost:4000/', { waitUntil: 'networkidle2' });
```
AlexVelezLl: I think the `global` object could be ommited here

package.json
```diff
@@ -16,25 +16,33 @@
...
     "frame-throttle": "3.0.0",
     "fuzzysearch": "1.0.3",
+    "jest-puppeteer": "^10.0.1",
     "lodash": "4.17.21",
     "popper.js": "1.16.1",
+    "puppeteer": "^22.11.0",
     "purecss": "2.2.0",
     "tippy.js": "4.2.1",
     "vue-intl": "3.1.0",
+    "wait-on": "^7.2.0",
```
AlexVelezLl: I dont see this `wait-on` lib used anywhere, probably it was someething you were testing and then removed? :sweat_smile:

AlexVelezLl (commented):
Woooo this is super exciting! :tada: :tada:  Thank you @KshitijThareja!!! This is looking super good. I just left some minor comments/questions, but this is looking great! :smiley:

jest.conf/setup.visual.js
KshitijThareja: I had thought about that initially. Will make the changes soon 👍🏻

.eslintrc.js
KshitijThareja: I had followed the documentation for this. But I understand your concern. Will see if we can have this somewhere else.

jest.conf/visual.index.js
KshitijThareja: Oh that'd be my mistake. Must have forgotten to remove this after having added it in the package.json script later.

jest.conf/testUtils.js
KshitijThareja: Sure, I'll add the changes in my next commit.

jest.conf/setup.visual.js
KshitijThareja: I guess we can go with having the prefix as 'visual' for such files. That'd indeed be better.

start-server-and-test.js
KshitijThareja: Yeah, that's the server timeout. I just assumed the worst case time as I was tired with the constant timeouts while testing 😅

start-server-and-test.js
KshitijThareja: Sure, will keep that in mind.

start-server-and-test.js
KshitijThareja: But we'll be informing them that there was another process running on that port before killing it. What prompted me to add this was that when I used to cancel a test execution midway, sometimes the port would still be occupied, causing the next test run to fail.
Isn't having this automated beneficial as opposed to the user having to kill the processes manually, considering the user meant to execute visual tests knowing that it will need port 4000?

start-server-and-test.js
KshitijThareja: Yeah. As I mentioned earlier, I am using it only for testing whatever changes I am making now.

start-server-and-test.js
KshitijThareja: Sure, seems much better 👍🏻

package.json
KshitijThareja: Oh I had used this while configuring jest-puppeteer but removed that part later on. Will remove this dependency.

jest.conf/setup.visual.js
KshitijThareja: Would it be okay if I created separate setup files for unit and visual tests and put all the common setup configuration in the original setup.js and import it in the test specific setup files?

AlexVelezLl: What are the differences between the two? @KshitijThareja Is there something in setup.js that we dont need in visual.setup.js?

start-server-and-test.js
AlexVelezLl: Hmm so we would probably need to handle the cancel action? and kill the dev server there. Would it be possible? The thing is that if a user has another process unrelated to kds running in that port, we cant just kill it, since it would be a bit invasive

jest.conf/setup.visual.js
KshitijThareja: There's the `window property` definition and the` TEST_TYPE` env var declaration that's present in setup.js which is changed in the visual.setup.js. Simply importing it would cause errors in the testing environment.

start-server-and-test.js
KshitijThareja: I understand that. Will look into it. Maybe giving the user a nice warning and gracefully exiting the testing environment would be better.

jest.conf/setup.visual.js
AlexVelezLl: You mean the `scrollTo` override?
```JS
Object.defineProperty(window, 'scrollTo', { value: () => {}, writable: true });
```
What could be the problem with this? And the Test type wouldnt be a problem because at the end if we import 'setup.js' we will be excecuting that code earlier, and then we can just override it in the `visual.setup.js`. But if the scrollTo override is a big problem, we will definetely need to separate them in different files

lib/KImg/__tests__/KImg.spec.js
bjester: yes @AlexVelezLl, I think that would be ideal, as it is simpler and minimizes the implementation in each test case required to take a screenshot.

The main reason I brought up the [testing library](https://testing-library.com/docs/) was because of the differences in how the test might need to be set up relative to unit testing. I supposed the testing library could be leveraged to achieve the simpler implementation we ideally seek, because the API has similarities between the different framework-specific packages. 

I can see there would likely be some challenges in moving forward with the ideal approach. I also think it's where some careful software engineering can be employed to put the pieces together, and where some real value will be added in this project to keep maintenance and overhead low.

jest.conf/setup.visual.js
KshitijThareja: The window property is only defined for jsdom environments. Since we are using a node environment defined as a part of the jest-puppeteer package, it will cause the following error:

> ● Test suite failed to run
> 
>     The error below may be caused by using the wrong test environment, see https://jestjs.io/docs/configuration#testenvironment-string.
>     Consider using the "jsdom" test environment.

start-server-and-test.js
bjester, line 1: As an alternative to having this scripted within KDS, you could use a package like `concurrently`. I think there should be a way for it to send SIGTERM to the server process after the test process exits

.eslintrc.js
bjester: This seems like it would be required in order to omit `global` when accessing these, in order for linting to pass.

jest.conf/setup.visual.js
AlexVelezLl: In our regular check-in meetings we saw that we could add a type check of `window` so we can reuse this file :open_hands:.

lib/KImg/__tests__/KImg.spec.js
AlexVelezLl: We decided to keep the current change for now to avoid blocking the PR, since more decisions will be needed. For next steps we will first determine how we will be rendering the components in the dev server, and after we can continue with how we can abstract this logic for developers.

KshitijThareja:
Hi @bjester @AlexVelezLl :)
I've added the component rendering mechanism for visual testing. Sharing some sample test results:

Screenshot from 2024-06-30 17-03-56

For this test, I rendered the `KButton` component from within the tests written in `KButton.spec.js`.

I'd love to hear your thoughts and suggestions on this.

startVisualTests.js
```diff
@@ -0,0 +1,140 @@
...
+        }
+      });
+      resolve();
+    });
+  });
+};
+
+const start = async () => {
+  let server;
+  try {
```
AlexVelezLl, line 156: I think we should validate the prerrequisites, like having the PERCY_TOKEN set, since here. To avoid running the dev server and percy if we already know that it  cant take snapshots

docs/pages/testing-playground.vue
```diff
@@ -0,0 +1,42 @@
...
+    <component :is="component" v-bind="componentProps" />
+  </div>
+
+</template>
+
+
+<script>
+
+  export default {
+    name: 'TestingPlayground',
```
AlexVelezLl: I think this component could be renamed to something like VisualTestingPlayground or something related to visual testing. Also we can add some jsdocs here explaining the purpose of this component.

lib/buttons-and-links/__tests__/KButton.spec.js
```diff
@@ -1,67 +1,109 @@
...
+                type: 'RENDER_COMPONENT',
+                component: component,
+                props: props,
+              },
+              '*'
+            );
+          },
+          { component, props }
+        );
+        await page.waitForSelector('#testing-playground');
```
AlexVelezLl, line 90: Something that could be wrong here is if we want to run several tests, when we take the second snapshot, we are not necessarily wating for it to re-render. Because the #testing-playground selector is always visible, and is not an indicator that the new render have already been made. I think we could use the `await this.$nextTick()` inside the component, but this would need direct method invocation or listeting to a response event. I think an alternative could be to define a method in the `window` object inside the testingPlayground component such as: 
``` javascript
mounted() {
  window.renderComponent = async (data) => {
        this.component = data.component;
        this.componentProps = data.props;
        await this.$nextTick();
  };
} 
```

And play with it to see how it goes. This way we can just call the await window.renderComponent(...); and when the promise resolves we will be ensuring that the re-render has already happened. 

If you have other Ideas how to ensure we are waiting for the re-render (or if the previous idea doesnt work) let us know and we can discuss it :open_hands:.

.gitignore
```diff
@@ -25,3 +25,6 @@ kolibri-design-system.iml
 # env
 .envrc
 .python-version
+
+# local
+myenv
```
AlexVelezLl: Is this an env file to save the `PERCY_TOKEN` env variable right?

lib/KImg/__tests__/KImg.spec.js
```diff
@@ -1,92 +1,103 @@
...
-    expect(error).toHaveBeenCalled();
-    expect(error.mock.calls[0][0]).toBeInstanceOf(Event);
-    expect(error.mock.calls[0][0]).toEqual(e);
-  });
+      expect(error).toHaveBeenCalled();
+      expect(error.mock.calls[0][0]).toBeInstanceOf(Event);
+      expect(error.mock.calls[0][0]).toEqual(e);
+    });
+  } else {
+    describe('KImg Visual Tests', () => {
```
AlexVelezLl: I think we can clean these changes and keep the KButton instead =)

lib/buttons-and-links/__tests__/KButton.spec.js
AlexVelezLl: (I just added a new unit test with the current changes and it worked ok, the snapshot showed the page updated. So it doesnt seem to be a blocker for now, but wonder if in the future we can find some race conditions because of this)

package.json
```diff
@@ -16,22 +16,29 @@
...
     "autosize": "3.0.21",
     "css-element-queries": "1.2.0",
     "date-fns": "1.30.1",
     "frame-throttle": "3.0.0",
     "fuzzysearch": "1.0.3",
+    "jest-puppeteer": "^10.0.1",
     "lodash": "4.17.21",
     "popper.js": "1.16.1",
+    "ps-tree": "^1.2.0",
+    "puppeteer": "^22.11.0",
```
AlexVelezLl: These new dependencies should be on the `devDepencencies` prop, since they are just for development purposes.

AlexVelezLl (commented):
Thank you @KshitijThareja! This looks good, I ran the tests, and I didn't find any problems in Percy's dashboard. I just left some minor concerns, let me know what you think!

package.json
KshitijThareja: Sure. Will change that accordingly :+1:

lib/KImg/__tests__/KImg.spec.js
KshitijThareja: Sure :+1:

.gitignore
KshitijThareja: I'm sorry I forgot to remove it during the cleanup. Will do that in the next set of changes

lib/buttons-and-links/__tests__/KButton.spec.js
KshitijThareja: I understand your concern. There will be changes in the way tests are written currently. We can discuss this once we are done working on the backbone of the visual testing mechanism. Once we have the setup complete, we can look for ways to streamline the tests.

docs/pages/testing-playground.vue
KshitijThareja: Sure :+1:

startVisualTests.js
KshitijThareja: That'd indeed be better. Will add this check in my next changes.

start-server-and-test.js
KshitijThareja: I did find a way to handle such situations from within the script. I was going through some custom server configurations and found that one of them employed the use of ps-tree to kill all the child processes that may have been created while running the dev server. So when you stop the execution of the command midway or just let it complete, the child processes are detected an killed too, thereby providing a proper cleanup after its execution is complete.
And if the port is occupied by another process, we can just give the user a warning and stop server initialization there itself.

lib/KImg/__tests__/KImg.spec.js
```diff
@@ -23,26 +20,20 @@ describe('KImg', () => {
...
-    makeWrapper({
-      propsData: { src: '/le-logo.svg', altText: undefined },
-      listeners: { error },
-    });
-    expect(error).toHaveBeenCalled();
-    expect(error.mock.calls[0][0]).toBeInstanceOf(Error);
-    expect(error.mock.calls[0][0].message).toBe(
-      'Missing required prop - provide altText or indicate isDecorative'
-    );
+    expect(() =>
```
AlexVelezLl: Hi @KshitijThareja! I think we can let this file without modifications. In this case it seems it is the expected behaviour to pass an error listener to catch these errors. See https://github.com/learningequality/kolibri-design-system/pull/645#discussion_r1617393477.

.eslintrc.js
AlexVelezLl: Hi @KshitijThareja! why do we need this property here in the esLintConfig env?

KshitijThareja: I had followed troubleshooting guidelines as specified here: [Jest-Puppeteer](https://github.com/argos-ci/jest-puppeteer?tab=readme-ov-file#troubleshooting:~:text=module.exports%20%3D%20%7B%0A%20%20env%3A%20%7B%0A%20%20%20%20jest%3A%20true%2C%0A%20%20%7D%2C)
This was basically to recognize jest's global variables and not flag their use while running linting tests.
But on taking a closer look now, I can see that this property has already been included in the main esLint config import (`kolibri-tools/.eslintrc`). I'll make the required changes.

lib/KImg/__tests__/KImg.spec.js
KshitijThareja: Sure. I'll make the changes. I was currently working on finding a way to use `concurrently` as discussed, so was a bit busy with that. I'll get this resolved by today at the earliest.

AlexVelezLl (commented):
Thank you @KshitijThareja!!! :tada: This looks good. I have run some tests and it works fine, so the setup is working as expected. Follow up about concurrently will happen in #676.

Failed checks are because of node version update and missing changelog, but we dont need to add it in this PR as is still a feature WIP. So nothing else to not from my side, but will let the final approve and merge to @bjester :open_hands:.

bjester:
One last thing @KshitijThareja: please separate the visual tests from the existing `yarn test` so we can keep the GH test check passing for now. We should add a followup for the GH Actions integration which runs the visual tests and the Percy diff checks.

bjester (approved):

PR #671, opened 2024-06-24
DRAFT: Updates `DD` to `dd` and `YYYY` to `yyyy` to fix error from `date-fns/format`

nucleogenesis:
## Description

The error reported in https://github.com/learningequality/kolibri/issues/12350 shows that `date-fns` is complaining about the use of all-caps date formatting throughout.

Initially, I fixed this in the `DataPage` component in Kolibri, but then the complaint extended to KDS components.

I am not sure if this is the proper fix, but I'm making the PR as I made the changes to be sure they'd fix the issue, and they do.

But, will wait for thoughts from @LianaHarris360


nucleogenesis:
The issue was resolved with a reverting of a PR (see https://github.com/learningequality/kolibri/issues/12350)

PR #672, opened 2024-06-25
Bump kolibri-tools, fix lint errors

nucleogenesis:
## Summary

In support of #671 and #578, which also then support fixing this critical bug in Kolibri: https://github.com/learningequality/kolibri/issues/12350

- Bumps to use the latest version published to NPM of the kolibri-tools package
- Adds a lot of `// eslint-disable-next-line no-console` -- I think our console errors are generally purposeful and useful for development -- but I'm not sure if we should just disable that rule. Open to thoughts on this

#### Notable outstanding linting issue though

I kept getting this error for what seems to be every file:

`TypeError: Class extends value undefined is not a constructor or null`

Some light research suggests it's a circular import issue but I'm not super sure where to dig on that and don't want to spend time on it unless this, overall, is a correct step forward.

One thing I did was run the linter _only_ on Vue files, which was successful. Then I did the same with _only_ JS files, which was successful. But if I do both at the same time, `TypeError: Class extends value undefined is not a constructor or null`

### DIFF
--- a/docs/common/DocsAnchorTarget.vue
+++ b/docs/common/DocsAnchorTarget.vue
@@ -30,10 +30,12 @@
         required: true,
         validator(value) {
           if (!value.startsWith('#')) {
+            // eslint-disable-next-line no-console
             console.error(`'anchor' prop value '${value}' must start with a '#'`);
             return false;
           }
           if (!value.match(/^#[\w\-.:]*$/)) {
+            // eslint-disable-next-line no-console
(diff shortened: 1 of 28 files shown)


bjester:
kolibri-tools has already been updated here: https://github.com/learningequality/kolibri-design-system/pull/645

nucleogenesis:
I'm a dork, #645 is already doing this

PR #682, opened 2024-07-16
Added scroll and custom widths

BabyElias:
- Added custom min-width and width attributes to each column, that can be determined by the user.
- Added scroll behaviour in x and y directions for the table.
- Ensured text wrapping within cells based on min-width attribute
- Added sticky row and column headers.
- Fixed header focus states.
- Added keyboard and mouse hover effects for entire row.
- Updated scrollIntoView for arrow key movements.
- Added sortColumn icon.
- Enhanced keyboard navigation using tab key.
- Theming


lib/KTable/index.vue
```diff
@@ -120,12 +141,20 @@
...
         SORT_ORDER_DESC,
         DATA_TYPE_OTHERS,
+        getHeaderStyle,
       };
     },
     props: {
       headers: {
         type: Array,
         required: true,
+        validator: function(value) {
```
MisRob, line 160: Nice

MisRob:
All new addition make sense to me :)

lib/KIcon/iconDefinitions.js
```diff
@@ -69,6 +69,10 @@ const KolibriIcons = {
     defaultColor: themePalette().grey.v_400,
   },
   sort: { icon: require('./precompiled-icons/le/sort.vue').default },
+  sortColumn: {
+    icon: require('./precompiled-icons/le/sortColumn.vue').default,
+    defaultColor: themePalette().grey.v_400,
```
MisRob: @BabyElias Since this is quite a general icon, I'd recommend we don't override the default color here (dropdown up and down don't have it set either) and just keep it by default.

Instead, you can just set it on `KIcon`'s `color` prop in the table component.

lib/KTable/index.vue
```diff
@@ -220,41 +261,81 @@
         }
         if (nextCell) {
           nextCell.focus();
```
MisRob, line 293: @BabyElias I tested and seems good to me. Would you add a comment summarizing what's the newly added code for scrolling achieving?

BabyElias: Sure!

BabyElias: Done

lib/KIcon/iconDefinitions.js
BabyElias: Done

Issue #684, opened 2024-07-17
[Visual Testing]: Abstract the test API for developers


AlexVelezLl:
## Product
KDS.

## Description

The setup of the visual testing framework in KDS has been carried out successfully. But the current way of doing snapshots involves a lot of boilerplate code, which makes the visual tests complex to write, complex to read, and the maintainability of the code is low. Example of the current behaviour:

```javascript
describe('KButton', () => {
  if (!canTakeScreenshot()) {
    //... <unit tests>
  } else {
          beforeAll(async () => {
        await page.goto('http://localhost:4000/testing-playground', { waitUntil: 'networkidle2' });
      });

      async function renderComponent(component, props) {
        await page.evaluate(
          ({ component, props }) => {
            window.postMessage(
              {
                type: 'RENDER_COMPONENT',
                component: component,
                props: props,
              },
              '*'
            );
          },
          { component, props }
        );
        await page.waitForSelector('#testing-playground');

        const isComponentRendered = await page.evaluate(() => {
          const testing_playground = document.querySelector('#testing-playground');
          return testing_playground && testing_playground.children.length > 0;
        });

        if (!isComponentRendered) {
          // eslint-disable-next-line no-console
          console.error('Component did not render in the testing playground');
        }
      }

      it('renders correctly with default props', async () => {
        await renderComponent('KButton', { text: 'Test Button' });
        await percySnapshot(page, 'KButton - Default');
      });
    });
  }
```

So there are a lot of things that we can abstract:
* Avoid developers having to make a huge wrap of the current unit tests with an `if(canTakeSnapshot())`.
  * Find a cleaner way to select the tests that will be executed when we are running visual tests and unit tests.
* Avoid developers from having to write and know about the `page.goto()` method and any other boilerplate code, and abstract it so that it is done automatically when we want to run a visual test.
* Abstract the writing of the renderComponent function in all test suites.

## The Value Add

This will allow for much easier, faster and more maintainable development of visual tests. And it will contribute a lot to developers easily adopting this pattern.

KshitijThareja:
Thanks @AlexVelezLl for creating the issue. Will have the PR up soon :)

AlexVelezLl:
Closed by #685.

PR #685, opened 2024-07-17
Add abstraction logic for simplifying writing visual tests

KshitijThareja:
## Description
This PR aims to abstract the logic for various functions used for component rendering, taking snapshots for visual tests and separating the two types of tests so that they can run independently.

#### Issue addressed
#684 

## Changelog

- [#PR no]
  - **Description:** Abstracts the test API for developers.
  - **Products impact:** -
  - **Addresses:** #684 
  - **Components:** -
  - **Breaking:** -
  - **Impacts a11y:** -
  - **Guidance:** -

[#PR no]: PR link

This was done by creating new supporting functions in the `testUtils.js` to provide the necessary functionalities which were earlier written directly as a part of tests.

## Testing checklist

## Reviewer guidance

## After review


jest.conf/testUtils.js
```diff
@@ -29,3 +22,56 @@ export const testAfterResize = testFunction => {
...
+export async function takeSnapshot(page, name) {
+  if (process.env.TEST_TYPE == 'visual') {
+    await percySnapshot(page, name);
+  } else {
+    // eslint-disable-next-line no-console
+    console.log(`Skipping Percy snapshot: ${name}`);
+  }
+}
+
+export function describeUnit(name, fn) {
```
AlexVelezLl: As far as possible, we should leave the unit tests intact, modifying them in this way would mean that we would have to:
* migrate all the tests we have right now.
* educate all devs who write tests and get them used to this new way of writing "describes"
* and reviewers should be more attentive so that we do not miss any.

So this makes the maintainability of the code a little more complex, makes bugs harder to find (It wouldn't be unusual to inadvertently end up with a describeVisual inside a describeUnit that never runs and we don't realize it). And above all because "describes" is a fairly standardized method already. Then we should look for another way that only changes the way of writing tests for the new functionality we are introducing, and avoid causing breaking changes with previous versions as much as possible.

I would suggest that having the [Visual] filter achieves that since with it we would not be introducing breaking changes for the unit tests, we would not have to do migrations, and it is also more flexible than the describe blocks, because we could have the flexibility to apply it as well to tests that are within other descriptions that are not necessarily visual. For example we could have

```js
describe('KButton', () => {
  describe('user interactions', () => {
    it('some test testing user interactions');
    it('some test testing user interactions');
    it('[Visual] hover state');
  });
});
```

Or if we have multiple visual testing about user interactions:
```js
describe('KButton', () => {
  describe('user interactions', () => {
    it('some test testing user interactions');
    it('some test testing user interactions');
    it('[Visual] hover state');
  });
});
```

And we have the flexibility of grouping these tests together with other unit tests to have them semantically together, describes inside other unit tests describes, or tests within other unit tests describes. The only restriction would be that we shouldnt have an unit test inside a visual block, but semantically it's a little more obvious to see it.

Let us know what you think!

jest.conf/testUtils.js
```diff
@@ -29,3 +22,56 @@ export const testAfterResize = testFunction => {
...
+  if (process.env.TEST_TYPE != 'visual') {
+    describe(name, fn);
+  }
+}
+
+export function describeVisual(name, fn) {
+  if (process.env.TEST_TYPE == 'visual') {
+    describe(name, () => {
```
AlexVelezLl: I think that we dont need to do this redirection for every describe block right? It sounds like is something that we probably need to do just once before running all tests. So probably it would be better to have this in the setup file where we set up the dev server.

lib/buttons-and-links/__tests__/KButton.spec.js
```diff
@@ -1,109 +1,75 @@
...
-  if (!canTakeScreenshot()) {
-    describe('icon related props', () => {
-      it('should render an icon before the text with the icon string passed to the `icon` prop', () => {
-        const wrapper = shallowMount(KButton, {
-          propsData: {
-            icon: 'add',
-          },
-        });
-        expect(wrapper.find('[data-test="iconBefore"]').exists()).toBe(true);
+  describeUnit('icon related props', () => {
```
AlexVelezLl: Something that should trigger all our alerts should be when we see that we need to wrap all the previous tests into something new, this is a bit similar to when we needed to wrap them into an `if(canTakeSnapshot())`, and this would cause us to have to do a lot of refactor, and also all the git blame of tests files would fall to the person who does that refactor, making it more difficult to trace the previous code changes.

lib/buttons-and-links/__tests__/KButton.spec.js
```diff
@@ -1,109 +1,75 @@
...
-          console.error('Component did not render in the testing playground');
-        }
-      }
-
-      it('renders correctly with default props', async () => {
-        await renderComponent('KButton', { text: 'Test Button' });
-        await percySnapshot(page, 'KButton - Default');
-      });
+  describeVisual('KButton Visual Tests', () => {
+    it('renders correctly with default props', async () => {
```
AlexVelezLl, line 70: Wooo, loved it, it looks super great and easy to use :tada: :star_struck:

lib/buttons-and-links/__tests__/KButton.spec.js
```diff
@@ -1,109 +1,75 @@
...
+      await renderComponent(page, 'KButton', { text: 'Test Button' });
+      await takeSnapshot(page, 'KButton - Default');
```
AlexVelezLl: Just one thing to note is that since page is global we could just use it inside these functions, and with this, the dev shouldnt even know about it existence.

The problem arises if we need to run multiple visual tests in parallel, and we cant have just one page for every render, but this is something that we will need to discuss in the future.

jest.conf/testUtils.js
```diff
@@ -29,3 +22,56 @@ export const testAfterResize = testFunction => {
   };
   animationFrameId = requestAnimationFrame(assertAfterResize);
 };
+
+export async function renderComponent(page, component, props) {
```
AlexVelezLl: Just wonder about the organization of these functions, because if we just read `renderComponent` inside a file called `testUtils`, that doesnt give us any clue that we are using it for visual testing. We could create another utils file for visual testing utilities, or we could rename it. I'll leave it up to you.

jest.conf/testUtils.js
```diff
@@ -1,5 +1,9 @@
 import { setMedia } from 'mock-match-media';
 
+const percySnapshot = require('@percy/puppeteer');
```
AlexVelezLl: Can't we use the import from syntax here?

jest.conf/testUtils.js
```diff
@@ -29,3 +22,56 @@ export const testAfterResize = testFunction => {
...
+        },
+        '*'
+      );
+    },
+    { component, props }
+  );
+  await page.waitForSelector('#testing-playground');
+
+  const isComponentRendered = await page.evaluate(() => {
+    const testing_playground = document.querySelector('#testing-playground');
```
AlexVelezLl: Just a note that in the future we should look for a more reliable way to know if the component was actually rendered.

jest.conf/testUtils.js
```diff
@@ -29,3 +22,56 @@ export const testAfterResize = testFunction => {
...
+    return testing_playground && testing_playground.children.length > 0;
+  });
+
+  if (!isComponentRendered) {
```
AlexVelezLl: I wonder if we could add something like `expect(isComponentRendered).toBe(true)` so this could be the assertion to know if a visual test pass. And if it dont, we could see more easily that an error is happening, because a console.error could go unnoticed in the console.

jest.conf/testUtils.js
```diff
@@ -29,3 +22,56 @@ export const testAfterResize = testFunction => {
...
+    console.error('Component did not render in the testing playground');
+  }
+}
+
+export async function takeSnapshot(page, name) {
+  if (process.env.TEST_TYPE == 'visual') {
+    await percySnapshot(page, name);
+  } else {
+    // eslint-disable-next-line no-console
+    console.log(`Skipping Percy snapshot: ${name}`);
```
AlexVelezLl: Now that we are running this method just in visual tests, I think we can remove this else block.

AlexVelezLl (changes requested):
Thank you @KshitijThareja! We have achieved a good level of abstraction. But Im little bit concerned about the maintainability of the code in the future with this new way of writing describe blocks. Apart from that just left a couple of notes here and there. Thank you! Good work so far! :tada:

jest.conf/testUtils.js
KshitijThareja: If we consider the additional overhead of migrating all the tests, you are right. I had concerns about that too, but since the idea was okayed in the meetings, I chose to go along with it. 
I'll shift this to use the [Visual] tags, but we'd have to see a lot of tests marked as skipped in the console output. One way to handle that would be to use a `custom reporter`. Or we can just use `verbose=false` with the test commands to reduce the output verbosity in that case. 
Let me know what you'd prefer!

jest.conf/testUtils.js
KshitijThareja: Sure. Will make the changes :+1:

lib/buttons-and-links/__tests__/KButton.spec.js
KshitijThareja: That's quite understandable. Will work on shifting to using [Visual] tags in that case.

lib/buttons-and-links/__tests__/KButton.spec.js
KshitijThareja: Yeah. Once we start writing visual tests considering all the cases, the situation should become clear.

jest.conf/testUtils.js
KshitijThareja: Well, in that case it'd be better to have another file for visual testing utility functions.

jest.conf/testUtils.js
AlexVelezLl: Oh, Im sorry, a thousand apologies if we okayed you in that direction, there was probably some misunderstanding on my side :').

Regarding the test outputs. For now I think there would be no problems in seeing the skipped tests, I was testing it and I see that Jest does report all the non-skipped test suites grouped at the end, and if there is an error it always reports them all at the end too, so they still have visibility. However if it becomes a problem I think we could consider the custom reporter option. Would love to know @bjester's opinion on that too.

jest.conf/testUtils.js
KshitijThareja: Sure, will replace this check with the assertion.

jest.conf/testUtils.js
KshitijThareja: One thing we can do is capture the inital content of the testing playground page and compare it with the html content captured after rendering the component and run a check to ensure that they both are different.
Does it seem like a good option?

jest.conf/testUtils.js
KshitijThareja: Sorry I had overlooked this. Will update it to use the import from syntax here :+1:

lib/buttons-and-links/__tests__/KButton.spec.js
KshitijThareja: Thanks a lot :smile: 
I will work on the suggestions provided and update this PR by tomorrow.

jest.conf/testUtils.js
KshitijThareja: Sure :+1:

jest.conf/testUtils.js
KshitijThareja: Oh no problem, I should have understood the requirements clearly. 
For now I'll update the PR with the required changes. We can discuss and work on this later if the need arises.

KshitijThareja:
Hi @AlexVelezLl :wave: 
The PR has been updated with the requested changes :)

jest.conf/testUtils.js
AlexVelezLl: Lets see how it works :)

jest.conf/visual.setup.js
```diff
@@ -4,4 +4,9 @@ import { percySnapshot } from '@percy/puppeteer';
 // Set the test type to visual
 process.env.TEST_TYPE = 'visual';
 
+const TESTING_PLAYGROUND_URL = 'http://localhost:4000/testing-playground';
 global.percySnapshot = percySnapshot;
+
+global.beforeAll(async () => {
+  await page.goto(TESTING_PLAYGROUND_URL, { waitUntil: 'networkidle2' });
```
AlexVelezLl, line 11: Just realized that we already do this page.goto after waiting for devserver to be up https://github.com/KshitijThareja/kolibri-design-system/blob/61c98f97fd2811dde712ae2b41f75b7bd587d92a/jest.conf/visual.index.js#L80. So I think we wouldnt need to do this redirection anymore right? Can we try just removing this beforeAll and check if it works without this?

jest.conf/visual.testUtils.js
```diff
@@ -0,0 +1,43 @@
+import percySnapshot from '@percy/puppeteer';
+
+export async function renderComponent(component, props) {
+  const initialState = await page.evaluate(() => {
```
AlexVelezLl: I think we could improve this name to something like `beforeRenderState` or `beforeRenderSnapshot` so the name suggests we are going to do something with this variable after rendering the component

jest.conf/visual.testUtils.js
```diff
@@ -0,0 +1,43 @@
...
+          props: props,
+        },
+        '*'
+      );
+    },
+    { component, props }
+  );
+  await page.waitForSelector('#testing-playground');
+
+  await page.waitForFunction(
```
AlexVelezLl, line 25: It would be nice if we add here some comments of the motivation for this check, so anybody else who read this code knows the intention of this wait.

package.json
```diff
@@ -16,8 +16,8 @@
     "precompile-svgs": "node utils/precompileSvgs/index.js && yarn run pregenerate",
     "precompile-custom-svgs": "node utils/precompileSvgs/index.js --custom && yarn run pregenerate",
     "_lint-watch-fix": "yarn lint -w -m",
-    "test:percy": "PERCY_LOGLEVEL=error npx percy exec -v -- jest --config jest.conf/visual.index.js -i ./lib/buttons-and-links/__tests__/KButton.spec.js",
-    "test": "jest --config=jest.conf/index.js",
+    "test:percy": "PERCY_LOGLEVEL=info npx percy exec -v -- jest --config jest.conf/visual.index.js -i ./lib/buttons-and-links/__tests__/KButton.spec.js -t '\\[Visual\\]'",
+    "test": "jest --config=jest.conf/index.js -t '^(?!.*\\[Visual\\])'",
```
AlexVelezLl: To avoid complex regex patterns, we can also use the `--testPathIgnorePatterns` option, so we just replicate the regex of the previous script just changing the option flag `-t` with `--testPathIgnorePatterns` https://jestjs.io/docs/cli#--testpathignorepatternsregexarray

KshitijThareja: Sure. Will try it out :)

jest.conf/visual.testUtils.js
KshitijThareja: Sure. I'll add the comments :+1:

jest.conf/visual.testUtils.js
KshitijThareja: I agree with that. Will make the changes soon :+1:

jest.conf/visual.setup.js
KshitijThareja: The page used there is a separate page instance using puppeteer and doesn't refer to the global page defined in jest-puppeteer package.  I remember trying to use jest-puppeteer's setup here, but ran into some errors as the environment provided by jest-puppeteer is not initialized at that point of time.

AlexVelezLl: Ooooh, I get it. Havent noticed that. Thank you!!

package.json
KshitijThareja: Just read that the flag `--testPathIgnorePatterns` is used to ignore specific test files based on their path and not the tests within it. Maybe if we had something like `--testNameIgnorePatterns` it would have been easier.

AlexVelezLl: Ooooh it's true, I didn't realize, it seems that jest doesn't have an ignore name pattern. Well, we could keep the regex then

AlexVelezLl: Just tested the regex and it seems to be ok :tada:

AlexVelezLl (commented):
Thank you @KshitijThareja!! LGTM! Code looks good, the new way of writing visual tests seems quite clean and easy to use and we have not required a large migration from the previous tests, and there do not seem to be any regressions :tada:. On my side it is already ok and ready to merge. Will left the final approve and merge to @bjester in case he has any comments. Thank you!

KshitijThareja:
Hello @AlexVelezLl !!
Support for test blocks to append [Visual] tag automatically has been added :)

jest.conf/index.js
```diff
@@ -10,6 +10,7 @@ const moduleNameMapper = {
 module.exports = {
   rootDir: path.resolve(__dirname, '..'),
   moduleFileExtensions: ['js', 'json', 'vue'],
+  testNamePattern: '^(?!.*\\[Visual\\])',
```
AlexVelezLl, line 13: Wooo, thats great, I researched if we had something like this for jest.config but I found nothing in the docs, but its great that this is working :tada:

AlexVelezLl (approved):
Thank you @KshitijThareja!! This is looking soo good :D. Code changes make sense and just tested it and it is working as expected. Its ready to go!!

Issue #683, opened 2024-07-17
[GSOC] Github workflow and PR checks for visual differences


bjester:
## Product

KDS itself

## Desired behavior

As an engineer reviewing KDS pull requests, I would like to see what visual differences resulted from the PR's changes. A visual testing GH workflow should be created that:
- executes the visual tests specifically
- only runs if components or tests have changed (for instance, it does not run if a PR only changes another GH workflow)
- only runs after unit tests have passed (to prevent using Percy credits unnecessarily)
- adds a failed PR check when there are visual differences
- provides a link to Percy for PR reviewers to either approve or reject the visual differences
- resets or re-runs the failed PR check when a reviewer approves the visual differences (if possible)

The existing unit testing workflow should remain unchanged.

## Current behavior

No such workflow or functionality currently exists

## (Optional) The Value Add

This will help contributors making changes to KDS, and reviewers of those changes, be more confident that there aren't visual regressions caused by the changes.

## (Optional) Possible Tradeoffs

It adds another step in the process of PR reviews, but a step that would otherwise be a manual process. Insufficient visual test coverage could also expose weaknesses in the process.

KshitijThareja:
Thanks @bjester for creating the issue. Will have the PR up soon :)

AlexVelezLl:
Closed by #686

PR #686, opened 2024-07-17
Introduce visual testing to the existing Javascript tests workflow

KshitijThareja:
## Description

This PR expands upon the existing JavaScript tests workflow to add a visual testing job in addition to the unit testing job.

#### Issue addressed
#683 

## Changelog

- [#PR no]
  - **Description:** Add visual testing job to the current Javascript test workflow.
  - **Products impact:** -
  - **Addresses:** #683 
  - **Components:** -
  - **Breaking:** -
  - **Impacts a11y:** -
  - **Guidance:** -

[#PR no]: PR link

## Steps to test

1. Make some changes to either the test files or the Vue components (Only the `KButton` component for now).
2. Make a PR with the above changes.
3. Wait for the Visual test job to finish and see the results in the comment generated.

## Testing checklist

## Reviewer guidance

## After review


.github/workflows/tests.yml
```diff
@@ -39,3 +39,72 @@ jobs:
...
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.OS }}-node-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.OS }}-node-
+      - name: Install dependencies
+        run: |
+          yarn --frozen-lockfile
+          npm rebuild node-sass
+      - name: Start server
```
AlexVelezLl: Could we avoid this step if we use `test:visual` for running the visual tests? With that command we already run the server and in the jest config we already wait for it to be running

.github/workflows/tests.yml
```diff
@@ -39,3 +39,72 @@ jobs:
...
+        run: |
+          yarn dev &
+          sleep 60 &&
+          curl -sSf http://localhost:4000/testing-playground
+        env:
+          SHELL: /bin/bash
+      - name: Run visual tests
+        id: run-visual-tests
```
AlexVelezLl: Why do we need these ids on these steps?

.github/workflows/tests.yml
KshitijThareja: The reason for me to use this was to generate the `jsdocs` and `environment.js` files so that the server can be started. The `test:visual` command is set to use the `yarn dev-only` command for starting the server as it's fast, but it doesn't generate those files, which are required for the server to run properly.

.github/workflows/tests.yml
KshitijThareja: Ahh I was trying different ways I could capture the step outputs. The `run-visual-tests id` is not needed, but the `extract-url id` is being used to refer to the extracted url output so that it can be accessed later. I'll remove the run-visual-tests id from there.

.github/workflows/tests.yml
AlexVelezLl: Oooh, I see. I think we can run `yarn pregenerate` (or the needed generate commands that are pre-requisites) and then `yarn test:visual`, so we dont have the logic of waiting for the server to be running in the workflow. `yarn dev` start some other processes also as the lint watch and api watch that we probably dont need to run in this context.

KshitijThareja: I've made the required changes :)

.github/workflows/tests.yml
```diff
@@ -39,3 +39,67 @@ jobs:
...
+          path: '**/node_modules'
+          key: ${{ runner.OS }}-node-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.OS }}-node-
+      - name: Install dependencies
+        run: |
+          yarn --frozen-lockfile
+          npm rebuild node-sass
+      - name: Extract jsdocs and environment info 
+        run: |
```
AlexVelezLl: I think we dont need these pipes in these two steps right? (extract jsdocs and run visual tests). As we only have a one-line command

KshitijThareja: Oh, Sorry for the oversight. I've updated it accordingly :+1:

KshitijThareja:
@AlexVelezLl The PR has been updated with the required changes

AlexVelezLl (approved):
Thank you @KshitijThareja! Looks good to me! Lets test this with #710

PR #687, opened 2024-07-19
Add live regions logic

### LINKED ISSUE #668: Implement KLiveRegion and useKLiveRegion
## Summary

This is a proposal for a new component and an accompanying composable, `<KLiveRegion>` and `useKLiveRegion`.

`<KLiveRegion>` will render itself as two live regions, one with a `status` role and another with an `alert` role:

```html
<div role="status">
  {{ statusMessage }}
</div>

<div role="alert">
  {{ alertMessage }}
</div>

```

A single `<KLiveRegion>` is supposed to be used in an app and high in the app's structure.

`useKLiveRegion` can be imported to multiple places in an app that can send `status` or `alert` messages to `<KLiveRegion>` via `useKLiveRegion.sendStatusMessage(<string>)` or `useKLiveRegion.sendAlertMessage(<string>)`.

The documentation will contain brief guidance with relevant references guiding developers on when to use or not use live regions, and what types of information should be sent as status messages and what should be alert messages.

## Background

Originally motivated by the ongoing sortable table work. Upon discussing with @radinamatic, it seems that the table will need to utilize live region in some use-cases, particularly to communicate loading/loaded states when re-fetching data when sorting is done on the backend. However, the table itself shouldn’t contain the live region as this would only add another live region to the already many regions in Kolibri. Additionally, a Kolibri audit revealed we use many live regions and we wrap whole components in them, which can lead to a number of problems.

## The Value Add

- Encapsulates commonly re-used logic
- By design enforces best practices that prevent messages from getting lost, conflicting with each other,  or from irrelevant and confusing information being announced. Such problems are often caused by:
    - multiple live regions
    - live region loaded into the DOM too late
    - wrapping rich HTML content by live region
    - wrapping interactive content by live region
- Possible integration with some KDS components that could send messages to the live region by default (to be decided, see the discussion below)

## **Possible Tradeoffs**

- Not aware of any

## Related

- https://github.com/learningequality/kolibri/issues/12326

## Resources

- [https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions)
- [https://www.sarasoueidan.com/blog/accessible-notifications-with-aria-live-regions-part-1/](https://www.sarasoueidan.com/blog/accessible-notifications-with-aria-live-regions-part-1/)
- [https://www.sarasoueidan.com/blog/accessible-notifications-with-aria-live-regions-part-2/](https://www.sarasoueidan.com/blog/accessible-notifications-with-aria-live-regions-part-2/)

MisRob:
## Description

- Adds logic that inserts ARIA live assertive and polite regions to an application's document body during KDS initialization and documents this on the new "Installation" page
- Relatedly adds `useKLiveRegion` composable with public methods for updating the live regions with assertive and polite messages. 

See "Comments" section below on why this strategy, after all.

#### Issue addressed

- Closes https://github.com/learningequality/kolibri-design-system/issues/668
  - Note that the interface is slightly different than I originally proposed. KDS takes care of inserting the live regions to DOM instead of having to have KLiveRegion. This makes things more robust and easier in consuming apps, and prepares ground for root teleport element. 

## Changelog

- [#687]
  - **Description:** Adds logic that inserts ARIA live assertive and polite regions to an application's document body during KDS initialization and documents this on the new "Installation" page. Relatedly adds `useKLiveRegion` composable with public methods for updating the live regions with assertive and polite messages. 
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/668
  - **Components:** `useKLiveRegion`
  - **Breaking:** No
  - **Impacts a11y:** Yes. It will fix several places utilizing live regions that don't work in our applications at all. Furthemore, it follows the recommended practices that will fix major a11y issues with live regions we're having.
  - **Guidance:** Find all polite and live regions (or roles) in an application. Remove them and instead use `useKLiveRegion.sendPoliteMessage` and `useKLiveRegion.sendAssertiveMessage` to update the live regions that KDS inserted to document body during installation.

## Steps to test

- See the new [Installation docs page](https://deploy-preview-687--kolibri-design-system.netlify.app/installation/)
- See the [new useKLiveRegion docs page](https://deploy-preview-687--kolibri-design-system.netlify.app/usekliveregion/) and try [the demo at its bottom part ](https://deploy-preview-687--kolibri-design-system.netlify.app/usekliveregion/#demo)
- Try it out in Kolibri https://github.com/learningequality/kolibri/pull/12475

- In KThemePlugin.js insert live region elements to `document.body` via plain JavaScript when DOM is loaded
- Publicly expose `useKLiveRegion.sendPoliteMessage` and `useKLiveRegion.sendAssertiveMessage` to update the region elements

Opposite - after we use it more widely, it will fix several places utilizing live regions that don't work in our applications at all. Furthermore, it follows the recommended practices that will fix major a11y issues with live regions we're having.

## Testing checklist

## Reviewer guidance

## After review

## Comments

I explored several approaches regarding inserting live region elements and their many combinations, some of the more important ones:

- Mounting `KLiveRegion.vue` as a Vue component
  - Makes setup complex by requiring attaching to the root Vue app element rather than simply to `document.body`
  - Its `{{ message }}` changes not being announced properly when compared to using plain JS to update elements' `textContent`

- Adding `mounted` hook to KThemePlugin's mixin, where we check if `this.$root === this` and if so, we insert the live regions to DOM at this point
  - live regions seemed to be ready fast at first, but testing https://github.com/learningequality/kolibri/pull/12475 showed that when we need to do the announcement from an element that is soon in DOM, it's not sufficient
  - as @AlexVelezLl noted, this also adds many conditions to every `mounted` to every KDS component, not just when mounted for the first time, but every time (called hundreds of times when navigating in Kolibri)

- Adjusting Vue prototype's `$mount` and some other places in a way similar to above
  - same issues as above 

- Exposing `mountLiveRegions` publicly and instructing apps to call it in the root app `mounted`
  - even though this mitigated the issue with many `mounted` conditions being executed, it really didn't help with the live regions not being ready in time when announcing very shortly after Kolibri loaded

Note that @rtibbles also mentioned some other Vue prototype places that we may want to experiment with. However, at the end of the day, since we need to update `DOM` as soon as possible, after seeing above, I can hardly imagine something faster than utilizing `document.readyState` and `DOMContentLoaded` event.

The original reason for not using this strategy was my wish to attach `KLiveRegion` as Vue component as well as a Nuxt problem. However, those can be overcome easily and I think reliability and robustness in consumer apps is the most important criteria here.

### DIFF
--- a/docs/pages/installation.vue
+++ b/docs/pages/installation.vue
@@ -1,0 +1,40 @@
+<template>
+
+  <DocsPageTemplate>
+
+    <DocsPageSection title="1. Install the plugin" anchor="#install-plugin">
+      <p>Kolibri Design System (KDS) is a Vue plugin. Register it in your application alongside scripts that are needed for its full functioning:</p>
+
+      <!-- eslint-disable -->
+      <DocsShowCode language="javascript">
+        import KThemePlugin from 'kolibri-design-system/lib/KThemePlugin';
+        import trackInputModality from 'kolibri-design-system/lib/styles/trackInputModality';
+        import trackMediaType from 'kolibri-design-system/lib/styles/trackMediaType';
+
+        Vue.use(KThemePlugin);
+
+        trackInputModality();
+        trackMediaType();
+      </DocsShowCode>
+      <!-- eslint-enable -->
+
+      This ensures the following:
+
+      <ul>
+        <li>Installs <code>$themeBrand</code>, <code>$themeTokens</code> <code>$themePalette</code>, and <code>$computedClass</code> helpers on all Vue instances.</li>
+        <li>Provides <code>$coreOutline</code>, <code>$inputModality</code>, <code>$mediaType</code>, and <code>$isPrint</code> computed properties as well as <code>$print</code> method to all Vue instances.</li>
+        <li>Globally registers all KDS Vue components.</li>
+        <li>Inserts assertive and polite ARIA live regions to your application's document body (see <DocsInternalLink href="/usekliveregion" text="useKLiveRegion" /> to understand how to utilize them).</li>
+      </ul>
+    </DocsPageSection>
+
+    <DocsPageSection title="2. Initialize theme" anchor="#initialize-theme">
+      <p>
+        Until this section is better documented (TODO link GH issue), refer to <DocsExternalLink href="https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/styles/initializeTheme.js" text="Kolibri's initializeTheme.js" />.
+      </p>
+    </DocsPageSection>
+
+  </DocsPageTemplate>
+
+</template>
+
--- a/docs/pages/installation.vue
+++ b/docs/pages/installation.vue
@@ -21,16 +21,16 @@
       This ensures the following:
 
       <ul>
-        <li>Installs <code>$themeBrand</code>, <code>$themeTokens</code> <code>$themePalette</code>, and <code>$computedClass</code> helpers on all Vue instances.</li>
+        <li>Installs <code>$themeBrand</code>, <code>$themeTokens</code> <code>$themePalette</code>, and <code>$computedClass</code> helpers on all Vue instances (see <DocsInternalLink href="/colors/#usage" text="Colors" />).</li>
         <li>Provides <code>$coreOutline</code>, <code>$inputModality</code>, <code>$mediaType</code>, and <code>$isPrint</code> computed properties as well as <code>$print</code> method to all Vue instances.</li>
         <li>Globally registers all KDS Vue components.</li>
-        <li>Inserts assertive and polite ARIA live regions to your application's document body (see <DocsInternalLink href="/usekliveregion" text="useKLiveRegion" /> to understand how to utilize them).</li>
+        <li>Inserts assertive and polite ARIA live regions to your application's document body (see <DocsInternalLink href="/usekliveregion" text="useKLiveRegion" />).</li>
       </ul>
     </DocsPageSection>
 
     <DocsPageSection title="2. Initialize theme" anchor="#initialize-theme">
       <p>
-        Until this section is better documented (TODO link GH issue), refer to <DocsExternalLink href="https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/styles/initializeTheme.js" text="Kolibri's initializeTheme.js" />.
+        Until this section is better documented, refer to <DocsExternalLink href="https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/styles/initializeTheme.js" text="Kolibri's initializeTheme.js" />.
       </p>
     </DocsPageSection>
 
--- a/docs/pages/usekliveregion.vue
+++ b/docs/pages/usekliveregion.vue
@@ -1,0 +1,123 @@
+<template>
+
+  <DocsPageTemplate apiDocs>
+
+    <DocsPageSection title="Overview" anchor="#overview">
+      <p>A composable that offers <code>sendPoliteMessage</code> and <code>sendAssertiveMessage</code> functions that send polite and assertive messages to their corresponding ARIA live regions.</p>
+    </DocsPageSection>
+
+    <DocsPageSection title="When to use live regions" anchor="#usage">
+      <p>Before sending messages to live regions, always research carefully if you really need it for the task ahead. Live regions can be <DocsExternalLink href="https://www.sarasoueidan.com/blog/accessible-notifications-with-aria-live-regions-part-2/#avoid-live-regions-if-you-can" text="buggy and inconsistent" />. There are often better alternatives, such as utilizing <DocsExternalLink href="https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes" text="WAI-ARIA attributes" />. A good rule of thumb is to use live regions only when there's no other way.</p>
+    </DocsPageSection>
+
+    <DocsPageSection title="Usage" anchor="#usage">
+      <p>Since polite and assertive regions are inserted to an application's document body automatically <DocsInternalLink href="/installation#install-plugin" text="during the KDS installation process" />, the only thing you need to do to deliver messages is to import and call <code>sendPoliteMessage</code> or <code>sendAssertiveMessage</code> from any place in your application.</p>
+
+      <p>These two methods are also used internally from some KDS components to provide a11y out of the box. Always check that you don't send messages to announce updates that are already being announced from KDS to prevent from duplicate announcements.</p>
+
+      <h3>Polite message</h3>
+
+      <p>Sending a polite message updates the text content of <code>aria-live="polite"</code> region. <em>Use it to send messages that can wait to be announced until the user is idle. This message should typically be the most commonly used.</em></p>
+
+      <p>Send polite messages with <code>sendPoliteMessage(message)</code>:</p>
+
+      <!-- eslint-disable -->
+      <!-- prevent prettier from changing indentation -->
+      <DocsShowCode language="javascript">
+        import useKLiveRegion from 'kolibri-design-system/lib/composables/useKLiveRegion';
+
+        export default {
+          setup() {
+            const { sendPoliteMessage } = useKLiveRegion();
+            sendPoliteMessage('Polite message');
+          }
+        };
+      </DocsShowCode>
+      <!-- eslint-enable -->
+
+      <h3>Assertive message</h3>
+
+      <p>Sending an assertive message updates the text content of <code>aria-live="assertive"</code> region. <em>It should be used with caution because it disrupts the user's flow. Use it only to send messages that require immediate attention, such as errors.</em></p>
+
+      <p>Send assertive messages with <code>sendAssertiveMessage(message)</code>:</p>
+
+      <!-- eslint-disable -->
+      <!-- prevent prettier from changing indentation -->
+      <DocsShowCode language="javascript">
+        import useKLiveRegion from 'kolibri-design-system/lib/composables/useKLiveRegion';
+
+        export default {
+          setup() {
+            const { sendAssertiveMessage } = useKLiveRegion();
+            sendPoliteMessage('Assertive message');
+          }
+        };
+      </DocsShowCode>
+      <!-- eslint-enable -->
+    </DocsPageSection>
+
+    <DocsPageSection title="Demo" anchor="#demo">
+      <p>Send messages below and turn on your screen reader or observe the content of <code>&lt;div id="k-live-region"&gt;</code> in the browser console.</p>
+
+      <DocsShow language="html">
+        <KTextbox label="Polite message" :value="politeMessageInput" @input="updatePoliteMessage" />
+        <KButton @click="sendPoliteMessage(politeMessageInput)">
+          Send
+        </KButton>
+      </DocsShow>
+
+      <DocsShow language="html">
+        <KTextbox label="Assertive message" :value="assertiveMessageInput" @input="updateAssertiveMessage" />
+        <KButton @click="sendAssertiveMessage(assertiveMessageInput)">
+          Send
+        </KButton>
+      </DocsShow>
+    </DocsPageSection>
+
+    <DocsPageSection title="Related" anchor="#related">
+      <ul>
+        <li>
+          <DocsInternalLink href="/installation#initialize-dom" text="KDS installation step" /> that attaches live regions to an application's DOM
+        </li>
+      </ul>
+    </DocsPageSection>
+  </DocsPageTemplate>
+
+</template>
+
+
+<script>
+
+  import { ref } from '@vue/composition-api';
+  import useKLiveRegion from '../../lib/composables/useKLiveRegion';
+
+  export default {
+    setup() {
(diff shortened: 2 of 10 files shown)


lib/KThemePlugin.js
```diff
@@ -66,6 +67,12 @@ export default function KThemePlugin(Vue) {
         return this.$mediaType === 'print';
       },
     },
+    mounted() {
+      if (this.$el === this.$root.$el) {
```
AlexVelezLl: I just wonder if there is a problem with the order in which the mounted hooks are executed. Because a parent component finishes mounting when all its children have been mounted, so if, for some reason in the first render, some child component calls sendPoliteMessage in its mounted hook, in theory _mountLiveRegion would not have been called yet because it would be the last one mounted hook to be called on the first render.

btw I'm also not sure if calling sendPoliteMessage on a mounted hook is an appropriate use case, but I was also thinking about the use case of inserting the tooltips container div, and was wondering if the same strategy could be used for it.

MisRob: Thank you Alex, hm yes, I think that may happen, even though I haven't seen it yet amongst many other issues I experienced. I haven't implemented yet, but hoped this TODO item should make everything even more bulletproof, including the situation you are describing if I understand it right:

https://github.com/learningequality/kolibri-design-system/pull/687/files#diff-eacdfbe0cbeff51092f2072e0b0ba63206fed4c9341494075cd77c3667ce6382R72-R74.

How does that look to you? Any other ideas?

MisRob: The safest option would be to always check for the presence of the region in dom before sending the message but that sounded to me like potentially lots of queries and also am not sure if it would cause some delays in delivering the message to the screen reader. That's why the idea about just catching error.

AlexVelezLl: Oh but as we already have these items in the window object, we can just check them. One Idea I had was to check _and mount_ the live-region if needed on demand inside the `sendAssertiveMessage` and `sendPoliteMessage`, something just like `if (!window.politeLiveRegion) { _mountLiveRegion)() }`, and avoid mounting it in the root mounted hook.

But yes, that could potentially cause a little delay on the first call, but it would ensure that no message would be missed. We could meassure the performance impact of mounting it on demand, and see if it actually cause a delay issue.

What do you think?

MisRob: > One Idea I had was to check and mount the live-region if needed on demand inside the sendAssertiveMessage 

Yes, that's a check (and potential re-mount) that will be good  (even though my first thinking was rather using try/catch since it seems to be more robust than relying on window check since it would inform us any time something goes wrong?)

> and avoid mounting it in the root mounted hook

I don't follow exactly how that would help - one of the main objectives is to get live regions to DOM as soon as possible (but after it's more or less ready) and if we completely removed using the mounted hook, the mounting would need to wait until we call `'sendAssertiveMessage'` the first time. From the resource I saw, this is rather discouraged.

But if motivation for that suggestion is to ensure that no message would be missed as you say later, combining both these approaches should do, I believe?

AlexVelezLl: Yes, combining both would do the work. My main thought about removing the check on the mounted hook was more about avoiding these comparison operations, because even though a comparison is super quick, if we do this for every mounted component, it would result in hundreds of comparisons, and Im not sure if it would result in a performance issue, I dont think so, but that was the main thought behind removing it if we already had the check inside the send methods.

MisRob: > My main thought about removing the check on the mounted hook was more about avoiding these comparison operations

I see. I thought it should hopefully not be such an issue since it's common to use `mounted` hooks in many components anyways, but definitely a valid concern since the condition is run many more times now than it would otherwise. In a sense, that's what I was trying to do in the first version that was just an event observing if a page has finished loading (however, there was some trouble recognizing that Vue parts are ready) Let me spend a while on some more experimentation and see if we can improve this aspect.

MisRob: So @AlexVelezLl, to address your concern and some another issues I experienced, I compromised a bit KDS Nuxt dev experience and return to DOMContentLoaded approach. It has shown to be the most robust approach that gets the elements to DOM much earlier than other methods. I will say more in the PR description.

And regarding re-mounting after sending the message unsuccessfully - we'd need to detect that the mounting has finished or add some wait timeout and hope it works. However, I don't think it's needed with this strategy because DOMContentLoaded will be ready before any JS is performed (and out of caution, an error will still get reported)

lib/KThemePlugin.js
```diff
@@ -39,13 +40,34 @@ import KCard from './KCard';
...
  * Also, set up global state, listeners, and styles.
  */
 export default function KThemePlugin(Vue) {
+  // Note that if DOM live regions need to be demostrated
+  // on the KDS website, and therefore attached to the DOM,
+  // just call _mountLiveRegion() in the relevant documentation
+  // page's 'mounted' (see 'docs/pages/usekliveregio.vue' for an example)
+  if (!isNuxtServerSideRendering()) {
+    const onDomReady = () => {
+      _mountLiveRegion();
```
MisRob, line 60: I imagine we would add other features, such as inserting root element for teleporting, here to `onDomReady`

docs/pages/installation.vue
```diff
@@ -0,0 +1,40 @@
...
+        <li>Installs <code>$themeBrand</code>, <code>$themeTokens</code> <code>$themePalette</code>, and <code>$computedClass</code> helpers on all Vue instances.</li>
+        <li>Provides <code>$coreOutline</code>, <code>$inputModality</code>, <code>$mediaType</code>, and <code>$isPrint</code> computed properties as well as <code>$print</code> method to all Vue instances.</li>
+        <li>Globally registers all KDS Vue components.</li>
+        <li>Inserts assertive and polite ARIA live regions to your application's document body (see <DocsInternalLink href="/usekliveregion" text="useKLiveRegion" /> to understand how to utilize them).</li>
+      </ul>
+    </DocsPageSection>
+
+    <DocsPageSection title="2. Initialize theme" anchor="#initialize-theme">
+      <p>
+        Until this section is better documented (TODO link GH issue), refer to <DocsExternalLink href="https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/styles/initializeTheme.js" text="Kolibri's initializeTheme.js" />.
```
MisRob, line 33: TODO @MisRob: Open the mentioned issue and link it here before merging

MisRob:
cc @BabyElias I hope you can use this from your table work after we get it merged and released. You're welcome to review too, all of it, or parts of it! Whatever work time-wise.

lib/KThemePlugin.js
MisRob, line 60: @AlexVelezLl @rtibbles side note, I imagine we would add other features, such as inserting root element for teleporting, here to `onDomReady`

docs/pages/installation.vue
```diff
@@ -0,0 +1,40 @@
...
+
+        trackInputModality();
+        trackMediaType();
+      </DocsShowCode>
+      <!-- eslint-enable -->
+
+      This ensures the following:
+
+      <ul>
+        <li>Installs <code>$themeBrand</code>, <code>$themeTokens</code> <code>$themePalette</code>, and <code>$computedClass</code> helpers on all Vue instances.</li>
```
AlexVelezLl: I see value in adding links to the references of what these objects helpers/properties means, i.e. https://deploy-preview-687--kolibri-design-system.netlify.app/colors/#usage. Although I see that not all of them have documentation as $print :sweat_smile:.

docs/pages/usekliveregion.vue
```diff
@@ -0,0 +1,123 @@
+<template>
+
+  <DocsPageTemplate apiDocs>
+
+    <DocsPageSection title="Overview" anchor="#overview">
+      <p>A composable that offers <code>sendPoliteMessage</code> and <code>sendAssertiveMessage</code> functions that send polite and assertive messages to their corresponding ARIA live regions.</p>
```
AlexVelezLl: Its totally optional, but I think we can add a link to this "Aria live regions" to documentation about what aria live regions are. (It would have helped me :sweat_smile: since I didnt know many of these a11y properties)

docs/pages/usekliveregion.vue
```diff
@@ -0,0 +1,123 @@
...
+        <KButton @click="sendAssertiveMessage(assertiveMessageInput)">
+          Send
+        </KButton>
+      </DocsShow>
+    </DocsPageSection>
+
+    <DocsPageSection title="Related" anchor="#related">
+      <ul>
+        <li>
+          <DocsInternalLink href="/installation#initialize-dom" text="KDS installation step" /> that attaches live regions to an application's DOM
```
AlexVelezLl: I dont see an `installation#initialize-dom` section in the page, maybe should it be `/installation#install-plugin`?

lib/KThemePlugin.js
AlexVelezLl: Yes, that makes sense to me :)

lib/KThemePlugin.js
```diff
@@ -39,13 +40,34 @@ import KCard from './KCard';
...
+      document.removeEventListener('DOMContentLoaded', onDomReady);
+    };
+
+    if (document.readyState === 'loading') {
+      document.addEventListener('DOMContentLoaded', onDomReady);
```
AlexVelezLl, line 65: What would happen if we just mount directly the live region here, without waiting for `DOMContentLoaded`? I think we can just call it directly here. And in any case we will be ensuring that this is called before any other component is mounted or any other composable is called (Because installing the KDS plugin is a pre-requisite for using any KDS component in the app).

lib/composables/useKLiveRegion/__tests__/index.spec.js
```diff
@@ -0,0 +1,84 @@
...
+const { sendPoliteMessage, sendAssertiveMessage } = useKLiveRegion();
+function removeWhitespaceFromHtml(htmlString) {
+  // https://stackoverflow.com/a/33108909
+  return htmlString.replace(/>\s+|\s+</g, function(m) {
+    return m.trim();
+  });
+}
+
+// check that document.body.innerHTML includes html string
+function assertDocumentBodyIncludes(htmlString) {
```
AlexVelezLl: This sounds like a custom matcher :eyes:, we could include this in our `jest.conf/setup.js`:
```javascript
global.expect.extend({
  toBeInDom(received) {
    const pass = removeWhitespaceFromHtml(document.body.innerHTML).includes(
      removeWhitespaceFromHtml(received)
    );
    if (pass) {
      return {
        message: () => `expected ${received} not to be in the document body`,
        pass: true,
      };
    } else {
      return {
        message: () => `expected ${received} to be in the document body`,
        pass: false,
      };
    }
  },
});
```

And then we can reuse this in any test file, and just write something simple like this:
```javascript
expect(`
  <div id="k-live-region" class="visuallyhidden">
    <div aria-live="polite">
    </div>
    <div aria-live="assertive">
    </div>
  </div>
`).toBeInDom();
```
That looks super elegant :smiley:.

AlexVelezLl (commented):
Thanks you @MisRob! This looks good to me! Just had a concern about what would happen if we directly mounted the live region in the KThemePlugin, and a couple of optional things I noticed.

lib/KThemePlugin.js
rtibbles: The only issue would be is if this is executing in the `<head>` tag of the page, and so the body element doesn't exist yet.

In Kolibri, this would work, but in other contexts it might not depending on when execution occurs.

AlexVelezLl: Oooh, thats true, alright then, thank you!

MisRob: Yes, the point is not only to insert it before Vue is loaded, but also after HTML is ready.

Besides, `DOMContentLoaded` listener will be added only if `document.readyState === 'loading'`. If the document is not loading, then live regions will be inserted right away as you suggest @AlexVelezLl.

Perhaps we won't run into it, but it's at least a polite thing to do that doesn't harm anything and increases robustness a bit.

AlexVelezLl: Yees, I agree. Thank you!

lib/composables/useKLiveRegion/__tests__/index.spec.js
MisRob: Neat, thank you. I didn't know it's so easy to achieve this. Will give it a try.

docs/pages/usekliveregion.vue
MisRob: Ah, will fix. Thanks

docs/pages/usekliveregion.vue
MisRob: Sure, why not! Thanks for feedback.

docs/pages/installation.vue
MisRob: That's a good idea. No problem to link places that I know of in this PR, but I won't document more things myself at this point :) Hopefully over time.

MisRob:
All feedback is resolved @AlexVelezLl, thank you

AlexVelezLl (approved):
LGTM! Thank you @MisRob! :)

PR #688, opened 2024-07-25
update below title slot styling

AllanOXDi:
#### Issue addressed

Addresses #*PR# HERE*

## Changelog
[#688]
  - **Description:**  Update`KCard` styling 
  - **Products impact:**  Card updates 
  - **Addresses:** 
  - **Components:** `KCard` 
  - **Breaking:** N0
  - **Impacts a11y:** No
  - **Guidance:**
 [#688] https://github.com/learningequality/kolibri-design-system/pull/688

## Steps to test

1. Step 1
2. Step 2
3. ...

## Testing checklist

## Reviewer guidance

## After review

### DIFF
--- a/lib/KCard/index.vue
+++ b/lib/KCard/index.vue
@@ -34,7 +34,7 @@
           as progressive loading experience.
         -->
         <span
-          v-if="$slots.thumbnailPlaceholder"
+          v-if="$slots.thumbnailPlaceholder && !thumbnailSrc"
           class="thumbnail-placeholder"
         >
           <!-- @slot Places content to the thumbnail placeholder area. -->
--- a/lib/KCard/index.vue
+++ b/lib/KCard/index.vue
@@ -34,7 +34,7 @@
           as progressive loading experience.
         -->
         <span
-          v-if="$slots.thumbnailPlaceholder && !thumbnailSrc"
+          v-if="$slots.thumbnailPlaceholder"
           class="thumbnail-placeholder"
         >
           <!-- @slot Places content to the thumbnail placeholder area. -->
--- a/lib/KCard/index.vue
+++ b/lib/KCard/index.vue
@@ -334,8 +334,8 @@
 
   .below-title {
     order: 4;
-    height: 18px; /* (1) */
-    min-height: 18px; /* (2) */
+    height: 26px; /* (1) */
+    min-height: 26px; /* (2) */
     margin: 0 $spacer $spacer;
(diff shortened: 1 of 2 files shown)


lib/KCard/index.vue
```diff
@@ -34,7 +34,7 @@
           as progressive loading experience.
         -->
         <span
-          v-if="$slots.thumbnailPlaceholder"
+          v-if="$slots.thumbnailPlaceholder && !thumbnailSrc"
```
MisRob, line 37: One of the functions of the placeholder background and element (such as icon) is to show while the image is loading over it. I've tried to explain with the comment above "Show it even when thumbnail source is provided - then the placeholder as progressive loading experience." Adding this condition will break it. I assume there was a situation where it caused some unexpected problems - could you elaborate? Perhaps we can think of a solution while still keeping progressive loading experience feature.

MisRob: Even though the major function of the placeholder background and this slot's content is to serve as placeholder when the thumbnail is not available at all, it also has the additional function to show at least something while the image is loading over it. I've tried to explain with the comment above on line 33.

Adding this condition would break this additional feature. I assume there was a situation where not having the condition caused some unexpected problems - could you elaborate what was the motivation for this change? Perhaps we can think of a solution that would still allow us to keep the progressive loading experience.

AllanOXDi: The motivation around this was to prevent the placeholder from showing when the thumbnailSrc is provided.

AllanOXDi: Actually - on a second thought . we actually don't need both. This will do` v-if="$slots.thumbnailPlaceholder"`

MisRob: > was to prevent the placeholder from showing when the thumbnailSrc is provided

Ah okay, well we'd want to show it even in this case, if the slot is provided, for the reasons I mentioned above.

So if

> This will do v-if="$slots.thumbnailPlaceholder"

is sufficient, let's just keep it that way.

MisRob: > was to prevent the placeholder from showing when the thumbnailSrc is provided

Ah okay, well we'd want to show it even in this case, if the slot is provided, for the reasons I mentioned above -  to achieve something being displayed below the image when it's still loading, for example on slower networks

So if

> This will do `v-if="$slots.thumbnailPlaceholder"`

is sufficient, let's just keep it that way.

MisRob (approved):
Good to go

Issue #701, opened 2024-07-27
KCard: Add an option to configure thumbnail alignment


MisRob:
## Summary

Add an option to configure the thumbnail area alignment via a new prop  `thumbnailAlign`.

## References

 This need is motivated by new additions to the cards family in Kolibri, such as
- https://github.com/learningequality/kolibri/issues/12318
- https://github.com/learningequality/kolibri/issues/12317
- https://github.com/learningequality/kolibri/issues/12319

## Current behavior

Currently, `KCard`  with `layout="horizontal"` and `thumbnailDisplay="large"` renders as

<img src="https://github.com/user-attachments/assets/2ae09dec-5a0e-4646-973d-d09418b4b185" width="400" />

and with `layout="horizontal"` and `thumbnailDisplay="small"` as

<img src="https://github.com/user-attachments/assets/e2088eb9-514d-4427-a92d-dfc99ad2ca0b" width="400" />

There is no `KCard` API yet that allows to configure whether the thumbnail area is displayed on the left side or the right side.

## Expected behavior

There is a new `thumbnailAlign` prop that accepts `'left'` and `'right'` values, and it decides on which side is the thumbnail area displayed, particularly in the layouts mentioned above.

All other layouts either don't have the thumbnail area or it takes the whole width of the card so they should simply ignore this prop. This would ideally be mentioned in the prop's JSDoc string.

## Acceptance criteria

- [x] Implemented as specified above
- [x] No regressions in any other layout
- [x] New prop has a custom validator and is documented
- [ ] Supports RTL

AllanOXDi:
fixed in https://github.com/learningequality/kolibri-design-system/pull/706

Issue #665, opened 2024-06-19
Ensure KCard supports RTL


MisRob:
## Blocked by

- https://github.com/learningequality/kolibri-design-system/pull/625

## Summary

`KCard` needs to support right-to-left languages. This was difficult to preview and test when developing solely in KDS, but as we're transitioning to Kolibri now, we can preview it more easily and ensure that everything works as expected.

One major area is the placement of the thumbnail area. If the thumbnail area appears on the right side in LTR, it should appear on the left side in RTL:

![Screenshot from 2024-06-18 21-12-00](https://github.com/learningequality/kolibri-design-system/assets/13509191/df59bc06-efb2-4a0c-8b65-c82a76b70bf2)

There may be more places to consider, and reviewing the card in all its layout modes in regard to RTL is one of this issue's tasks.

## Guidance
- [RTL language support](https://kolibri-dev.readthedocs.io/en/develop/i18n.html#rtl-language-support) Kolibri guide
- See [Bidirectionality](https://m2.material.io/design/usability/bidirectionality.html) guidelines to get a sense of what should be mirrored
- It is recommended to work on this in parallel to developing Kolibri cards based on `KCard` (such as https://github.com/learningequality/kolibri/issues/12317, https://github.com/learningequality/kolibri/issues/12318, or https://github.com/learningequality/kolibri/issues/12319)

## Out of scope
- `KCard` shouldn't be concerned about RTL for any elements that are provided by products via slots. It's the product's responsibility to ensure RTL for such elements.
  - Example: In the illustration above, `KCard` ensures that the `#aboveTitle` slot is rendered on the right side. But it is not (and cannot be) concerned whether icons provided to that slot from a product are mirrored.

## Acceptance criteria
- [ ] All `KCard` modes are reviewed for RTL
- [ ] It changes its layout appropriately when used in an RTL language
- [ ] It is tested in Kolibri in at least one place

AllanOXDi:
Currently  this is what KCard  supports by default, no need to reinvent the wheel for now . We will need to do a thorough QA on Kolibri with the new cards that will be built especially with the card grid.  @MisRob and I agreed  to close this after testing  with folder cards and resource cards on Kolibri.

PR #706, opened 2024-07-29
Add options to configure thumbnail alignment

AllanOXDi:
## Description
This PR introduces a new `thumbnailAlign` prop to the KCard component, enabling developers to configure the alignment of the thumbnail area within the card. This  new` thumbnailAlign `prop accepts 'left' and 'right' values, and it decides on which side is the thumbnail area displayed, particularly in the horizontal layout.

#### Issue addressed
https://github.com/learningequality/kolibri-design-system/pull/706

### With the left and right align options 
Screenshot 2024-07-31 at 09 10 19
Screenshot 2024-07-31 at 09 08 18

## Changelog

[#706]
  - **Description:** Add new alignment options to `KCard` 
  - **Products impact:**  new alignment 
  - **Addresses:** [#701](https://github.com/learningequality/kolibri-design-system/issues/701)
  - **Components:** `KCard` 
  - **Breaking:** N0
  - **Impacts a11y:** No
  - **Guidance:**
[#706]: https://github.com/learningequality/kolibri-design-system/pull/706

## Testing checklist

## Reviewer guidance

## After review

### DIFF
--- a/lib/KCard/index.vue
+++ b/lib/KCard/index.vue
@@ -83,6 +83,11 @@
     LARGE: 'large',
   };
 
+  const thumbnailAlign = {
+    LEFT: 'left',
+    RIGHT: 'right',
+  };
+
   function cardValidator(allowedValues, propName) {
     return function(value) {
       if (!Object.values(allowedValues).includes(value)) {
@@ -177,8 +182,8 @@
        */
       thumbnailAlign: {
         type: String,
-        default: 'left',
-        validator: value => ['left', 'right'].includes(value),
+        default: thumbnailAlign.LEFT,
+        validator: value => [thumbnailAlign.LEFT, thumbnailAlign.RIGHT].includes(value),
       },
     },
     computed: {
--- a/lib/KCard/index.vue
+++ b/lib/KCard/index.vue
@@ -485,7 +485,7 @@
 
     .thumbnail {
       position: absolute;
-      width: 30%;
+      width: 35%;
       min-width: 80px;
       right: 0;
       height: 100%;
--- a/lib/KCard/index.vue
+++ b/lib/KCard/index.vue
@@ -4,7 +4,7 @@
     :to="to"
     :title="title"
     :headingLevel="headingLevel"
-    :titleLines="titleLines"
+    :titleLines="titleLine"
     :class="['k-card', rootClass, thumbnailAlignClass]"
     :headingStyles="headingStyles"
   >
@@ -375,7 +375,7 @@
   .below-title {
     order: 4;
     min-height: 26px; /* (2) */
-    margin: 0 $spacer  0 $spacer;
+    margin: 0 $spacer 0 $spacer;
     overflow: hidden; /* (1) */
   }
 
@@ -487,8 +487,7 @@
         left: $spacer;
       }
       .above-title,
-      .below-title
-      {
+      .below-title {
         margin-right: $spacer;
         margin-left: 40%;
       }
--- a/lib/KCard/index.vue
+++ b/lib/KCard/index.vue
@@ -215,7 +215,7 @@
         */
       stylesAndClasses() {
         /* In px. Needs to be the same as $spacer variable in styles part */
-        const SPACER = 24;
+        const SPACER = 12;
 
         const headingCommonStyles = {
           order: 3,
@@ -335,7 +335,7 @@
 <style lang="scss" scoped>
 
   /* Needs to be the same as SPACER constant in JavaScript part */
-  $spacer: 24px;
+  $spacer: 12px;
 
   /*
         Just couple of comments that are referenced from several places:
@@ -369,9 +369,8 @@
 
   .below-title {
     order: 4;
-    height: 26px; /* (1) */
     min-height: 26px; /* (2) */
-    margin: 0 $spacer $spacer;
+    margin: 0 $spacer  0 $spacer;
     overflow: hidden; /* (1) */
   }
 
@@ -483,8 +482,8 @@
         left: $spacer;
       }
       .above-title,
-      .below-title,
-      .footer {
+      .below-title
+      {
         margin-right: $spacer;
         margin-left: 40%;
       }
--- a/lib/KCard/index.vue
+++ b/lib/KCard/index.vue
@@ -185,6 +185,11 @@
         default: 'left',
         validator: cardValidator(thumbnailAlignOptions, 'thumbnailAlign'),
       },
+      /**
+       * Specifies the number of lines allowed for the title before truncation occurs.
+       * @type {number}
+       * @default 2
+       */
       titleLines: {
         type: Number,
         required: false,
@@ -240,7 +245,6 @@
           return {
             rootClass: 'vertical-with-large-thumbnail',
             thumbnailAlignClass: undefined,
-            titleLine: this.titleLines,
             thumbnailAspectRatio: undefined,
             headingStyles: {
               ...headingCommonStyles,
@@ -255,7 +259,6 @@
           return {
             rootClass: 'vertical-with-small-thumbnail',
             thumbnailAlignClass: undefined,
-            titleLine: this.titleLines,
             thumbnailAspectRatio: undefined,
             headingStyles: {
               ...headingCommonStyles,
@@ -270,7 +273,6 @@
           return {
             rootClass: undefined,
             thumbnailAlignClass: undefined,
-            titleLine: this.titleLines,
             thumbnailAspectRatio: undefined,
             headingStyles: {
               ...headingCommonStyles,
@@ -285,7 +287,6 @@
           return {
             rootClass: 'horizontal-with-large-thumbnail',
             thumbnailAlignClass: `thumbnail-align-${this.thumbnailAlign}`,
-            titleLine: this.titleLines,
             thumbnailAspectRatio: undefined,
             headingStyles: {
               ...headingCommonStyles,
--- a/lib/KCard/index.vue
+++ b/lib/KCard/index.vue
@@ -202,14 +202,8 @@
       thumbnailStyles() {
         return this.stylesAndClasses.thumbnailStyles;
       },
-
       thumbnailAlignClass() {
-        if (this.layout === 'horizontal' &&
-          (this.thumbnailDisplay === 'large' || this.thumbnailDisplay === 'small' )
-        ) {
-          return `thumbnail-align-${this.thumbnailAlign}`;
-        }
-        return '';
+        return this.stylesAndClasses.thumbnailAlignClass;
       },
       /*
           Returns dynamic classes and few style-like data that CSS was cumbersome/impossible to use for ,or are in need of using theme, grouped by all possible combinations of layouts.
@@ -240,6 +234,7 @@
         if (this.layout === 'vertical' && this.thumbnailDisplay === 'large') {
           return {
             rootClass: 'vertical-with-large-thumbnail',
+            thumbnailAlignClass:undefined,
             titleLines: 3,
             thumbnailAspectRatio: undefined,
             headingStyles: {
@@ -254,6 +249,7 @@
         if (this.layout === 'vertical' && this.thumbnailDisplay === 'small') {
           return {
             rootClass: 'vertical-with-small-thumbnail',
+            thumbnailAlignClass:undefined,
             titleLines: 3,
             thumbnailAspectRatio: undefined,
             headingStyles: {
@@ -268,6 +264,7 @@
         if (this.layout === 'vertical' && this.thumbnailDisplay === 'none') {
           return {
             rootClass: 'vertical-with-no-thumbnail',
+            thumbnailAlignClass:undefined,
             titleLines: 2,
             thumbnailAspectRatio: undefined,
             headingStyles: {
@@ -282,6 +279,7 @@
         if (this.layout === 'horizontal' && this.thumbnailDisplay === 'large') {
           return {
             rootClass: 'horizontal-with-large-thumbnail',
+            thumbnailAlignClass:`thumbnail-align-${this.thumbnailAlign}`,
             titleLines: 3,
             thumbnailAspectRatio: undefined,
             headingStyles: {
@@ -298,6 +296,7 @@
         if (this.layout === 'horizontal' && this.thumbnailDisplay === 'small') {
           return {
             rootClass: 'horizontal-with-small-thumbnail',
+            thumbnailAlignClass:`thumbnail-align-${this.thumbnailAlign}`,
             titleLines: 2,
             thumbnailAspectRatio: '1:1',
             headingStyles: {
@@ -314,6 +313,7 @@
         if (this.layout === 'horizontal' && this.thumbnailDisplay === 'none') {
           return {
             rootClass: 'horizontal-with-no-thumbnail',
+            thumbnailAlignClass:undefined,
             titleLines: 2,
             thumbnailAspectRatio: undefined,
             headingStyles: {
--- a/lib/KCard/index.vue
+++ b/lib/KCard/index.vue
@@ -182,8 +182,8 @@
        */
       thumbnailAlign: {
         type: String,
-        default: thumbnailAlign.LEFT,
-        validator: value => [thumbnailAlign.LEFT, thumbnailAlign.RIGHT].includes(value),
+        default: 'left',
+        validator: cardValidator(thumbnailAlign,'thumbnailAlign'),
       },
     },
     computed: {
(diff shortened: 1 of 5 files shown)


AllanOXDi:
@MisRob and @AlexVelezLl -Noting that Support for  RTL will be address separately  #665

MisRob:
Hi @AllanOXDi, I'd like to suggest one update that will help with the following

- `thumbnailAling` as is now takes effect only in `horizontal` layout with `large` thumbnail, and the whole implementation is very tightly coupled to this use case. Even though this is the first motivating use-case, there's no need to limit it to this particular case, by means of API as well as internal implementation. It is intended to take effect in any case where it makes sense, for example in the horizontal layout with small thumbnail.
- Keeping the style organization simple as best as we can by limiting the number of conditions and keeping them at a single place, in both JavaScript and styles sections.

I think something similar to the following could help to open it up and make a bit clearer as well:

1. Renaming `rootClass` to `rootClasses` and using it on root element as `:class="rootClasses"`
2. Updating `stylesAndClasses` existing conditions to something like:

```
stylesAndClasses() {
      const commonRootClasses = [
         'k-card',
        this.thumbnailAlign === ThumbnailAlign.LEFT ? 'thumbnail-left' : 'thumbnail-right'
      ]
   
      if (this.layout === 'vertical' && this.thumbnailDisplay === 'large') {
        return {
          rootClasses: [...commonRootClasses, 'vertical-with-large-thumbnail'],
          ...
      }
      if (this.layout === 'vertical' && this.thumbnailDisplay === 'small') {
        return { rootClasses: [...commonRootClasses, 'vertical-with-small-thumbnail'],
         ...
        }
      }
     // update `rootClass` in this way for all other existing branches
    },
  },
```
This ensures all thumbnail layouts have information on what alignment we've chosen.

3. In styles, just updating relevant layouts, currently the following two. All styles related to the thumbnail alignment are grouped within them, in these two new classes:

```
.horizontal-with-large-thumbnail
   &.thumbnail-left {
     
   }

  &.thumbnail-right {
     
   }
}

.horizontal-with-small-thumbnail
   &.thumbnail-left {
     
   }

  &.thumbnail-right {
     
   }
}
```


If you can think about further improvements, I welcome them, this is one of the approaches I could think of right now.

lib/KCard/index.vue
```diff
@@ -278,10 +317,10 @@
   $spacer: 12px;
 
   /*
-      Just couple of comments that are referenced from several places:
-      - (1) Intentionally fixed. Cards on the same row of a grid should have the same overall height and their sections too should have the same height so that information is placed consistently. As documented, consumers need to ensure that contents provided via slots fits well or is truncated.
-      - (2) Solves issues with fixed height in a flex item
-    */
+        Just couple of comments that are referenced from several places:
```
AlexVelezLl, line 332: Just wondering if are these extra spaces enforced by the linter? :sweat_smile:

AlexVelezLl (commented):
I think it would be good to have this variation in the documentation too, showing some examples. What do you think? @AllanOXDi @MisRob

lib/KCard/index.vue
```diff
@@ -246,6 +271,25 @@
             },
           };
         }
+        if (
```
MisRob: **[required]** Just from code it seems to me that ` thumbnailAlignClass` is taking care of everything already. Is this whole `if` section possibly a leftover from the first PR version that just needs to be removed?

lib/KCard/index.vue
```diff
@@ -418,4 +487,23 @@
     }
   }
 
+  .horizontal-with-large-thumbnail-and-right-alignment {
```
MisRob: **[required]** Relatedly to https://github.com/learningequality/kolibri-design-system/pull/706/files#r1709907101, it seems to me this whole class needs to be removed. `horizontal-with-large-thumbnail` already contains styles related to left/right alignment in its `thumbnail-align-left` and `thumbnail-align-right`.

MisRob:
@AllanOXDi I think this looks better now, thank you. I appreciate you simplified my original suggestion even further :+1:  Let's just finish cleanup of the previous implementation.

lib/KCard/index.vue
```diff
@@ -167,6 +172,19 @@
...
+       * Only applies to horizontal layouts with 'small' or 'large' thumbnail display.
+       * Ignored in other layouts.
+       * @type {String}
+       * @values 'left', 'right'
+       * @default 'left'
+       */
+      thumbnailAlign: {
+        type: String,
+        default: thumbnailAlign.LEFT,
+        validator: value => [thumbnailAlign.LEFT, thumbnailAlign.RIGHT].includes(value),
```
MisRob: Nothing blocking, just could be nice to use the `cardValidator` you implemented some time ago?

lib/KCard/index.vue
```diff
@@ -167,6 +172,19 @@
...
+       * Controls the alignment of the thumbnail area in horizontal layouts.
+       * Only applies to horizontal layouts with 'small' or 'large' thumbnail display.
+       * Ignored in other layouts.
+       * @type {String}
+       * @values 'left', 'right'
+       * @default 'left'
+       */
+      thumbnailAlign: {
+        type: String,
+        default: thumbnailAlign.LEFT,
```
MisRob, line 185: Needs to be `default: 'left'` because the documentation build won't otherwise show the correct default value.

MisRob: We need to hardcode `default: 'left'` here so the documentation shows the correct value:

docs

MisRob:
> I think it would be good to have this variation in the documentation too, showing some examples. What do you think? @AllanOXDi @MisRob

@AlexVelezLl @AllanOXDi Yes

MisRob:
@AllanOXDi **[required]** I've just noticed that the thumbnail alignment still doesn't take effect for the horizontal layout with the small thumbnail which is expected behavior. Using the approach you've chosen, we can achieve that by

```
stylesAndClasses() {
      if (this.layout === 'horizontal' && this.thumbnailDisplay === 'large') {
        return {
          rootClass: 'horizontal-with-large-thumbnail',
          thumbnailAlignClass: `thumbnail-align-${this.thumbnailAlign}`
          ...
        }
      }
      if (this.layout === 'vertical' && this.thumbnailDisplay === 'small') {
        return {
           rootClass: 'vertical-with-small-thumbnail',
           thumbnailAlignClass: ''
           ...
        }
      }
     if (this.layout === 'horizontal' && this.thumbnailDisplay === 'small') {
        return {
           rootClass: 'vertical-with-small-thumbnail',
           thumbnailAlignClass: `thumbnail-align-${this.thumbnailAlign}`
           ...
        }
      }
    },
  },
```

I think this is a nice merging of both approaches that will most importantly resolve the remaining issue for small thubmnail, and additionally also remove some duplicative conditions by keeping `stylesAndClasses` as a source of truth.

MisRob:
@AllanOXDi **[required]** I've just noticed that the thumbnail alignment still doesn't take effect for the horizontal layout with the small thumbnail which is expected behavior. Using the approach you've chosen

```
BaseCard :class="['k-card', rootClass, thumbnailAlignClass]">
```

we can achieve that by

```
stylesAndClasses() {
      if (this.layout === 'horizontal' && this.thumbnailDisplay === 'large') {
        return {
          rootClass: 'horizontal-with-large-thumbnail',
          thumbnailAlignClass: `thumbnail-align-${this.thumbnailAlign}`
          ...
        }
      }
      if (this.layout === 'vertical' && this.thumbnailDisplay === 'small') {
        return {
           rootClass: 'vertical-with-small-thumbnail',
           thumbnailAlignClass: ''
           ...
        }
      }
     if (this.layout === 'horizontal' && this.thumbnailDisplay === 'small') {
        return {
           rootClass: 'vertical-with-small-thumbnail',
           thumbnailAlignClass: `thumbnail-align-${this.thumbnailAlign}`
           ...
        }
      }
    },
  },
```

I think this is a nice merging of both approaches that will most importantly resolve the remaining issue for small thumbnail.

Additionally, I think that keeping everything in `stylesAndClasses` as a single source of truth will help you (and potentially any future devs) to understand what is applied exactly to what layouts - since we can now remove `thumbnailAlignClass` computed property, and instead have it defined in the existing conditional branches.

MisRob:
@AllanOXDi **[required]** I've just noticed that the thumbnail alignment still doesn't take effect for the horizontal layout with the small thumbnail which is expected behavior. Using the approach you've chosen

```
BaseCard :class="['k-card', rootClass, thumbnailAlignClass]">
```

we can achieve that by

```
stylesAndClasses() {
      if (this.layout === 'horizontal' && this.thumbnailDisplay === 'large') {
        return {
          rootClass: 'horizontal-with-large-thumbnail',
          thumbnailAlignClass: `thumbnail-align-${this.thumbnailAlign}`
          ...
        }
      }
      if (this.layout === 'vertical' && this.thumbnailDisplay === 'small') {
        return {
           rootClass: 'vertical-with-small-thumbnail',
           thumbnailAlignClass: undefined
           ...
        }
      }
     if (this.layout === 'horizontal' && this.thumbnailDisplay === 'small') {
        return {
           rootClass: 'vertical-with-small-thumbnail',
           thumbnailAlignClass: `thumbnail-align-${this.thumbnailAlign}`
           ...
        }
      }
    },
...
...
...
  },
```

I think this is a nice merging of both approaches that will most importantly resolve the remaining issue for small thumbnail. Additionally, keeping everything in `stylesAndClasses` as a single source of truth will help you (and potentially any future devs) to understand what is applied exactly to what layouts - since we can now have it defined in the existing conditional branches rather than having to add the same very conditional branches to `thumbnailAlignClass` computed property which is removed in this proposal.

After this is setup, you can add new styles for `.horizontal-with-small-thumbnail` using the very same structure you used for styling `.horizontal-with-large-thumbnail`, that is

```
.horizontal-with-small-thumbnail {
  &.thumbnail-align-left {

  }

  &.thumbnail-align-right {

  }
}
```

AllanOXDi:
Hi @MisRob ,just noting that  the `KTextTruncator`on different layouts would render different tiltle lines. And the solution   was to re-introduce a titleLine prop that gives it a default lines to handle it.  When the KCard prop(`titleLines`) is not specified it defaults to 2lines.

AllanOXDi:
Also, I have added a few different layouts on the playground for testing

lib/KCard/index.vue
```diff
@@ -205,10 +231,19 @@
...
+              ? `calc(40% + ${SPACER}px)`
+              : `${SPACER}px`,
+        };
+
         if (this.layout === 'vertical' && this.thumbnailDisplay === 'large') {
           return {
             rootClass: 'vertical-with-large-thumbnail',
-            titleLines: 3,
+            thumbnailAlignClass: undefined,
+            titleLine: this.titleLines,
```
MisRob: Feel free to remove `titleLine` from here (and all other layouts) and pass ` titleLines` prop value directly to `BaseCard`s'  `:titleLines="titleLines"`. I think with your latest update that makes `titleLines` into `KCard`'s prop, this will make sense (I don't recall exactly why I removed it, probably just some experimentation I forgot to cleanup - apologies)

lib/KCard/index.vue
```diff
@@ -167,13 +172,31 @@
...
+       * @type {String}
+       * @values 'left', 'right'
+       * @default 'left'
+       */
+      thumbnailAlign: {
+        type: String,
+        default: 'left',
+        validator: cardValidator(thumbnailAlignOptions, 'thumbnailAlign'),
+      },
+      titleLines: {
```
MisRob, line 193: Would be nice to have JSDoc string above this title as well so it shows in the documentation

lib/KCard/index.vue
```diff
@@ -194,7 +220,7 @@
         */
       stylesAndClasses() {
         /* In px. Needs to be the same as $spacer variable in styles part */
-        const SPACER = 24;
+        const SPACER = 12;
```
MisRob: **[required]** Was the change from 24 to 12 intentional? I think we will need to keep it due https://github.com/learningequality/kolibri-design-system/pull/709.

If this was part of the fix related to the lack of space on cards, that could be fixed by increasing the overall height in all layouts where it's needed. The overall heights are defined in the css part, for example:

```
 .horizontal-with-large-thumbnail {
    ...
    height: 240px;
    ...
```

Same applies to `$spacer`.

I think it'd be fine to revert to 24 in this PR and then do the height fix in a follow-up PR if it'd be better, feel free to decide.

lib/KCard/index.vue
```diff
@@ -394,6 +432,32 @@
     .footer {
       width: calc(60% - 2 * #{$spacer}); /* same as heading width defined in script */
     }
+
+    &.thumbnail-align-left {
```
MisRob, line 421: Nice, I appreciate you kept only styles specific to thumbnail alignment in this block, while keeping the more general styles applied to both layouts outside. Makes it easy to understand.

lib/KCard/index.vue
```diff
@@ -394,6 +432,32 @@
...
+      .thumbnail {
+        left: 0;
+      }
+
+      .above-title,
+      .below-title,
+      .footer {
+        margin-right: $spacer;
+        margin-left: calc(40% + #{$spacer});
```
MisRob: **[required]** Related to comment for line 457. This seems to be redundant style overlapping with the flex align technique that is already in place. Basically, these slots are already aligned and take the correct width by means of `width: calc(60% - 2 * #{$spacer});` and `align-items: flex-end;`, both set above. Please double-check but I think we can remove this line.

lib/KCard/index.vue
```diff
@@ -394,6 +432,32 @@
...
+
+    &.thumbnail-align-right {
+      .thumbnail {
+        right: 0;
+      }
+
+      .above-title,
+      .below-title,
+      .footer {
+        margin-right: calc(40% + #{$spacer});
```
MisRob: **[required]** Related to the comment for line 445. I think this line needs to be removed. Instead, let's simply switch `flex-end` / `flex-start` :) So the previous:

```
.horizontal-with-large-thumbnail {
  align-items: flex-end; // this took care of aligning slots to right
}
```

will become

```
.horizontal-with-large-thumbnail {
  &.thumbnail-align-left {
      align-items: flex-end; // align slots to right (aka end) as previously, just make sure it's applied only when thumbnail is supposed to be on the left
   }

    &.thumbnail-align-right {
      align-items: flex-start; // and here when the thumbnail is supposed to be on right, align slots to left (aka start)
    }
}
```

The main reason is that it wouldn't make sense to keep using `flex-end` just to override that behavior via setting the margins in exactly the opposite way.

lib/KCard/index.vue
```diff
@@ -394,6 +432,32 @@
...
+        margin-left: $spacer;
+      }
+    }
   }
 
   .horizontal-with-small-thumbnail {
```
MisRob, line 448: **[required]** Would you please double-check this layout too and make sure to utilize `flex-start/flex-end` as mentioned in more detail for similar issue in `.horizontal-with-large-thumbnail`?

lib/KCard/index.vue
```diff
@@ -416,6 +480,25 @@
     .footer {
       width: calc(100% - 2 * #{$spacer});
     }
+
+    &.thumbnail-align-left {
+      .thumbnail {
+        top: $spacer;
```
MisRob: I think this can be removed because it's already defined on line 469

lib/KCard/index.vue
```diff
@@ -416,6 +480,25 @@
...
+      .above-title,
+      .below-title {
+        margin-right: $spacer;
+        margin-left: 40%;
+      }
+    }
+
+    &.thumbnail-align-right {
+      .thumbnail {
+        top: $spacer;
```
MisRob: I think this can be removed because it's already defined on line 469

lib/KCard/index.vue
```diff
@@ -416,6 +480,25 @@
...
+        right: $spacer;
```
MisRob, line 477: I think it makes sense to have this line at this place because it related to alignment :+1: Let's just cleanup duplicate  `right: $spacer;` from line 470

lib/KCard/index.vue
```diff
@@ -205,10 +231,19 @@
           height: '100%',
         };
 
+        const titleAlignStyle = {
```
MisRob: **[required]** I believe that `titleAlignStyle` can be removed completely after you utilize switching between `flex-end` and `flex-start`. Explained in mode detail in other comments.

MisRob (commented):
Hi @AllanOXDi, good progress! I think we're almost there. Left a few minor cleanup comments. There's only two changes I consider required:

- (1) The spacer value regression - however if we revert here, I think the issue it attempted to solve could be addressed in another PR since it's not related to thumbnails
- (2) Not mixing two techniques to achieve left/right alignment of slots. We can simply utilize flex features for aligning items as that's what is already in place. Overriding it with margins is not needed and on some occasions it creates conflicting or unnecessary styles. Not that using the margins technique would be generally bad - it's definitely valid one! My concern is about not mixing the two approaches. Let me know if you wanted to chat or had questions.

I think after this is done, all should be in place for merge.

MisRob:
> Hi @MisRob ,just noting that the KTextTruncatoron different layouts would render different tiltle lines. And the solution was to re-introduce a titleLine prop that gives it a default lines to handle it. When the KCard prop(titleLines) is not specified it defaults to 2lines.

Yes, thank you! I think this was my oversight, I don't know why I did it honestly.

MisRob:
> Also, I have added a few different layouts on the playground for testing

This helps a lot - thank you

lib/KCard/index.vue
AllanOXDi: I think we need still need this to handle  switching between the title.

lib/KCard/index.vue
AllanOXDi: I have double checked and  we will still need this in this layout out.

MisRob: Would you explain what's the reason (or what gets broken) on your side?

On my current playground page it works. I've checked in more layouts in the browser console, here are just some examples. See the browsers console - margins percentage calculation is not there anymore.

Screenshot from 2024-08-15 13-08-17

In the layouts that have the thumbnail on the right side though I needed to set `flex-start` as mentioned in other comments. Did you do that?

Here's an example for the title, with no margins percentage calculation, and with `flex-start`:

Screenshot from 2024-08-15 13-13-28

It's possible I am missing something in some of the other layouts, perhaps? But will need to understand what you did exactly at first. I can try to reproduce whatever you experienced if you let me now what steps you did. Or you could make sure to update flex alignment correctly and then push it, and show me what happens on the playground page when you remove the margins. The flex-start/end should be updated anyways.

lib/KCard/index.vue
MisRob: Response here https://github.com/learningequality/kolibri-design-system/pull/706#discussion_r1718275985

lib/KCard/index.vue
MisRob: If this turns out to be difficult in writing, we can meet on a call :) In any case, I will need to understand what's going on exactly on your part.

MisRob (approved):
Well done! Good to go :)

I am observing two issues on the playground page, none of them blocking and even related to this work so I will open follow-ups later on.

Issue #689, opened 2024-07-26
[Visual Testing]: Implement visual tests of KButton


AlexVelezLl:
## Product
KDS.

## Desired behavior

After #685, we ended up with a very straightforward way of writting visual tests. Now lets implement exhaustive tests for one component - KButton, and test the whole visual tests process with this component.

We should take the snapshot of the different visual combinations of KButton:
- KButton as primary raised button
- KButton as primary flat button
- KButton as secondary raised button
- KButton as secondary flat button
- KButton as basic link
- KButton as raised button disabled
- KButton as flat button disabled
- KButton as raised button with hover state
- KButton as flat button with hover state
- KButton with icon and afterIcon
- KButton with default slot

## Current behavior
Currently we just have the basic example with default props.

KshitijThareja:
Hi @AlexVelezLl 
Thanks for creating the issue. I'll try to create a PR for this soon :)

AlexVelezLl:
Thank you @KshitijThareja!

AlexVelezLl:
Closed by #710.

PR #710, opened 2024-07-31
Implement visual tests for KButton component

KshitijThareja:
## Description
This PR aims to introduce concrete visual tests for the` KButton` component, supplemented by modifications to the existing `testing playground` file and the `renderComponent` function to allow for usage of slots in the visual tests.

#### Issue addressed
#689 

Addresses #*PR# HERE*

## Changelog

- [#PR no]
  - **Description:** Implement visual tests for KButton component along with some other modifications.
  - **Products impact:** -
  - **Addresses:** #689 
  - **Components:** -
  - **Breaking:** -
  - **Impacts a11y:** -
  - **Guidance:** -

[#PR no]: PR link

## Steps to test

1. Set the PERCY_TOKEN env variable in your environment.
2. Run `yarn test:visual` to execute visual tests
3. Check Percy dashboard to see visual diffs generated.

This was accomplished by referring to the Puppeteer documentation for implementing abstract methods for interacting with the components. The changes to the render function and playground file were made in order to render the components if slots were passed from within the tests.

## Testing checklist

## Reviewer guidance

## After review


docs/pages/testing-playground.vue
```diff
@@ -7,7 +7,17 @@
     Please do not modify the contents of this file.
   -->
   <div id="testing-playground" style="padding: 24px">
-    <component :is="component" v-bind="componentProps" />
+    <component :is="component" v-bind="componentProps">
+      <!-- Render default slot if provided -->
+      <template v-if="slots.default">
+        <!-- eslint-disable-next-line vue/no-v-html -->
+        <span v-html="slots.default"></span>
```
AlexVelezLl: Why should this be in as span? I think a div would be more appropiate in case we need to add more elements inside

docs/pages/testing-playground.vue
```diff
@@ -7,7 +7,17 @@
...
+      </template>
+      <!-- Render named slots passed to the component -->
+      <template v-for="(slot, name) in namedSlots" #[name]>
+        <component :is="slot.content" v-bind="slot.props" :key="name" />
```
AlexVelezLl: I think it would be better if we follow the same pattern in both default slot and named slot, I would vote to render it inside a v-html. Or in any case, if rendering inside a div gives any error, we can be flexible and provide both options. If slots.name is String, then render it inside a v-html, if not, we can render it as you are doing it here. What do you think?

docs/pages/testing-playground.vue
```diff
@@ -32,9 +42,25 @@
          * @type {Object} The props to be passed to the dynamically rendered component.
          */
         componentProps: {},
+        /**
+         * @type {Object} The slots to be passed to the dynamically rendered component.
```
AlexVelezLl, line 46: Here in the jsdoc we can specify the structure and give guidance on the object structure.

jest.conf/visual.testUtils.js
```diff
@@ -1,23 +1,24 @@
 import percySnapshot from '@percy/puppeteer';
 
-export async function renderComponent(component, props) {
+export async function renderComponent(component, props, slots = {}) {
```
AlexVelezLl, line 27: Here we need to add a jsdoc explaining these props, and explaining the structure slots should follow

jest.conf/visual.testUtils.js
```diff
@@ -40,8 +41,30 @@ export async function renderComponent(component, props) {
   global.expect(isComponentRendered).toBe(true);
 }
 
-export async function takeSnapshot(name) {
+export async function takeSnapshot(name, options = {}) {
```
AlexVelezLl, line 78: Here would be good also to have some jsdoc, and let users know what options could we use (I think it should be a lot of options that percySnapshot receives, in any case we could link to their documentation).

lib/buttons-and-links/__tests__/KButton.spec.js
```diff
@@ -67,9 +73,94 @@ describe('KButton', () => {
   });
 
   describe.visual('KButton Visual Tests', () => {
-    it('renders correctly with default props', async () => {
-      await renderComponent('KButton', { text: 'Test Button' });
-      await takeSnapshot('KButton - Default');
+    it('renders correctly with different appearances', async () => {
```
AlexVelezLl: IMO each snapshot should be a test case. Unit test cases should be simple and test just one thing. If we want to group multiple test cases like for example with the disable state, we could use a describe block inside de discribe.visual block and group them.

docs/pages/testing-playground.vue
KshitijThareja: If I take the KButton test with default slot as an example, if you use div, the button will become twice the size, with the upper part showing slot content and the lower part showing the content passed to button itself. That's the reason for using span here.

docs/pages/testing-playground.vue
KshitijThareja: Won't the named slots be other components that are being nested in a particular component? In that case they'll always have a name and specific props attached to them just like the parent component, which is being handled correctly in the current scenario.
I thought that default slots are used to handle the case in which slots.name is expected to be a string. I may be wrong though; I am not fully aware of how slots are used throughout KDS.

jest.conf/visual.testUtils.js
KshitijThareja: Sure, will get it done 👍

jest.conf/visual.testUtils.js
KshitijThareja: Sure 👍

docs/pages/testing-playground.vue
AlexVelezLl: Hmm I see, but there could be some components where a span could also be a problem. What if we add something like a "wrapperEl" prop that defines the element container, and a "content" that defines the v-html prop? Any other idea is welcome :open_hands:.

lib/buttons-and-links/__tests__/KButton.spec.js
KshitijThareja: So, if I am getting it correctly, for the case where we are testing different appearances of KButton for example, we need to write them all in separate `it` blocks and group those it blocks in one `describe` block?

docs/pages/testing-playground.vue
AlexVelezLl: Not necessarily, we can have html elements that are not vue components as part of named slots (e.g. this [#divider in KListWithOverflow](https://github.com/learningequality/studio/blob/4dcc60d35de6ab1910185982a3d0e8da60fd5b7f/contentcuration/contentcuration/frontend/channelEdit/views/CurrentTopicView.vue#L103), It is a correct Vue syntax, there are actually no differences in the restrictions of what we can do in the default slot vs named slots.

lib/buttons-and-links/__tests__/KButton.spec.js
AlexVelezLl: Yes, although Im not sure if we should group them as "different appearances", in general all snapshots are different appearances.

AlexVelezLl: Oh different appearances means different values for the `appearance` prop. I hadnt get it, in that case its up to you if it makes sense to group them, you can group them, if not its fine too.

docs/pages/testing-playground.vue
AlexVelezLl: I think we can combine both and have something like:
```javascript
{
  slots: {
    default: {
      element: "div", // or any vue component like KIcon
      elementProps: {
        class: "some-class"
      },
      innerHTML: "<div> Some nested <a>content</a> </div>"
    }
  }
}
```

KshitijThareja:
Hi @AlexVelezLl. I've updated the PR based on the suggestions 👍

lib/buttons-and-links/__tests__/KButton.spec.js
```diff
@@ -67,9 +67,102 @@ describe('KButton', () => {
...
-      await renderComponent('KButton', { text: 'Test Button' });
-      await takeSnapshot('KButton - Default');
+    describe('renders correctly with different appearances', () => {
+      it('renders correctly as primary raised button', async () => {
+        await renderComponent('KButton', {
+          text: 'Raised Button',
+          primary: true,
+          appearance: 'raised-button',
+        });
+        await takeSnapshot('KButton - Primary Raised Button', { widths: [400], minHeight: 512 });
```
AlexVelezLl: As we will probably have the same sizes for (almost) all the screenshots of a single component, I think we can make this a constant object defined in the context of the describe.visual block and use the same options object thoughout the tests.

AlexVelezLl (approved):
Thanks @KshitijThareja!! Code looks good to me, and snapshots are working as expected! I think we have reach to a very neat and flexible api for slots elements, which is super great. I left just a minor observation, once it is fixed, please feel free to hit that merge button.

We are almost thereee!!! Congratulations :tada: :tada: :tada: :tada:.

lib/buttons-and-links/__tests__/KButton.spec.js
KshitijThareja: I've made the changes as requested ✅

AlexVelezLl: Thank you!!

PR #718, opened 2024-08-09
Fix KDateRange failing tests on prs opened on last day of the month

### LINKED ISSUE #713: KDateRange test failures: KDateRange test failure impacting month-end pull requests
## Product

KDS
## Expected behavior

A test failure has been identified in the KDS framework located at `lib/KDateRange/__tests__/KDateRange.spec.js`, and is likely causing test failures on new pull requests submitted on the final day of the month.

## Actual behavior

KDateRange tests pass as expected.

## Additional information

Screenshot from 2024-07-31 21-10-07

LianaHarris360:
## Description

This pull request fixes the failing `KDateCalendar` component tests that would occur on open pull requests on the last day of the month. This was fixed by setting dates manually in the tests. The `KDateCalendar` has also been updated to display the month of the property `lastAllowedDate`.

#### Issue addressed

Addresses #713 

## Changelog

- [#718]
  - **Description:** This pull request resolves failing `KDateCalendar` component tests that occurred on the last day of the month in open pull requests by setting dates manually in the tests. Additionally, the `KDateCalendar` is updated to show the month of the `lastAllowedDate` property.
  - **Products impact:** none
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/713
  - **Components:** -
  - **Breaking:** no
  - **Impacts a11y:** -
  - **Guidance:** -

[#718]: https://github.com/learningequality/kolibri-design-system/pull/718

## Testing checklist

## Reviewer guidance

## After review

### DIFF
--- a/lib/KDateRange/KDateCalendar.vue
+++ b/lib/KDateRange/KDateCalendar.vue
@@ -157,8 +157,9 @@
         },
         numOfDays: 7,
         isFirstChoice: this.selectedStartDate == null ? true : false,
-        activeMonth: new Date().getMonth() - 1 == -1 ? 11 : new Date().getMonth() - 1,
-        activeYearStart: new Date().getFullYear(),
+        activeMonth:
+          this.lastAllowedDate.getMonth() - 1 == -1 ? 11 : this.lastAllowedDate.getMonth() - 1,
+        activeYearStart: this.lastAllowedDate.getFullYear(),
       };
     },
     computed: {
--- a/lib/KDateRange/__tests__/KDateRange.spec.js
+++ b/lib/KDateRange/__tests__/KDateRange.spec.js
@@ -123,7 +123,8 @@ describe('KDateCalendar component', () => {
   it('the month and year display on the left-side calendar view should display the previous month of the lastAllowedDate', async () => {
     const previousMonthDisplay = getKDateCalendarElements(wrapper).previousMonthDisplay();
     const lastAllowedDate = wrapper.props().lastAllowedDate;
-    const previousMonth = new Date(lastAllowedDate) - 1;
+    const previousMonth = new Date(lastAllowedDate);
+    previousMonth.setMonth(lastAllowedDate.getMonth() - 1);
     const date = Vue.prototype.$formatDate(previousMonth, { month: 'long', year: 'numeric' });
     expect(previousMonthDisplay.text()).toBe(date);
   });
--- a/lib/KDateRange/__tests__/KDateRange.spec.js
+++ b/lib/KDateRange/__tests__/KDateRange.spec.js
@@ -90,9 +90,9 @@ describe('KDateRange component', () => {
 
 const KDATECALENDAR_PROPS = {
   firstAllowedDate: new Date(2022, 0, 1),
-  lastAllowedDate: new Date(),
+  lastAllowedDate: new Date(2022, 10, 1),
   selectedStartDate: new Date(2022, 8, 1),
-  selectedEndDate: new Date(),
+  selectedEndDate: new Date(2022, 10, 1),
   previousMonthText: 'Previous Month',
   nextMonthText: 'Next Month',
 };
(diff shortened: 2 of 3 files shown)


lib/KDateRange/KDateCalendar.vue
```diff
@@ -157,8 +157,9 @@
         },
         numOfDays: 7,
         isFirstChoice: this.selectedStartDate == null ? true : false,
-        activeMonth: new Date().getMonth() - 1 == -1 ? 11 : new Date().getMonth() - 1,
-        activeYearStart: new Date().getFullYear(),
+        activeMonth:
```
AlexVelezLl, line 160: I'm curious to know why 1 is subtracted from the active month. It means that the activeMonth will always be the month before the lastAllowedDate right? But if we are in January, the previous month would cause the year to also have to be reduced, but I see that activeYearStart is taken only from getFullYear.

lib/KDateRange/__tests__/KDateRange.spec.js
```diff
@@ -120,16 +120,18 @@ describe('KDateCalendar component', () => {
     expect(wrapper.exists()).toBe(true);
   });
 
-  it('the month and year display on the left-side calendar view should contain the previous month', async () => {
+  it('the month and year display on the left-side calendar view should display the previous month of the lastAllowedDate', async () => {
     const previousMonthDisplay = getKDateCalendarElements(wrapper).previousMonthDisplay();
-    const previousMonth = new Date().setDate(0);
+    const lastAllowedDate = wrapper.props().lastAllowedDate;
+    const previousMonth = new Date(lastAllowedDate) - 1;
```
AlexVelezLl: I think a safest way of getting the previous month could be
```suggestion
    const previousMonth = new Date(lastAllowedDate);
    previousMonth.setMonth(lastAllowedDate.getMonth() - 1);
```

And it will not depend if the lastAllowedDate is the first day of the month, and it also handles if lastAllowedDate is in january.

lib/KDateRange/KDateCalendar.vue
LianaHarris360: This is because of how the Date() constructor in JavaScript sets the months. It uses an integer value to represent the month, beginning with 0 for January to 11 for December.

lib/KDateRange/__tests__/KDateRange.spec.js
LianaHarris360: Yes, good catch! Since the `lastAllowedDate` is manually set to November in this test case, we won't have to worry about January dates, but it would definitely be the safer approach.

lib/KDateRange/KDateCalendar.vue
AlexVelezLl: Oh no, I meant what if the `lastAllowedDate` was for example `new Date(2024, 0, 1)`, then activeMonth would be 11, but activeYearStart would be 2024, so the active date would be December 2024 instead of December 2023. But I looked at the code and I see that this `activeYearStart` value is updated in the [created hook](https://github.com/LianaHarris360/kolibri-design-system/blob/13b4365315ca186768767a6876afde5c711fb621/lib/KDateRange/KDateCalendar.vue#L205) if this happens, so there is no problem.

Anyways I think we could avoid the ternary here, and the check in the created hook if we do something similar to get the previous month:
```javascript
    data() {
      const previousMonth = new Date(this.lastAllowedDate);
      previousMonth.setMonth(lastAllowedDate.getMonth() - 1);
      return {
        ...,
        activeMonth: previousMonth.getMonth(),
        activeYearStart: previousMonth.getFullYear(),
      };
    },
```
But it is totally optional, although I think it would clarify the intention that activeMonth and activeYearStart refer to the previous month of the lastAllowedDate.

AlexVelezLl (approved):
lgtm! And code seems to be working even if `lastAllowedDate` is the last day of the month.

Issue #330, opened 2022-05-20
Create prototype (non-accessible) SortableTable


marcellamaki:
### Overview 

This is an initial prototyping issue for the creation of a reusable `KSortableTable` component. For context, please see the epic. 

### Expect Component API 

The API should build on the current `CoreTable` and KDP's `SortableHeader`. This would likely require most of the existing props, although they could be renamed or shared (if a similar prop is currently used in both components). 

`header` and `body` slots can be used, as in `CoreTable`, so that the table can be created to suit the data being displayed.

Core Table's props
```
selectable: {
        type: Boolean,
        default: false,
        required: false,
      },
      emptyMessage: {
        type: String,
        default: null,
      },
```
SortableHeader's props 
```
title: {
        type: String,
        required: true,
      },
      column: {
        type: String,
        required: true,
      },
      pos: {
        type: String,
        default: 'right',
      },
      sort: {
        type: Boolean,
        default: true,
      },
```

During this stage, the visual design of the basic `CoreTable` component does not need to change. 

### Acceptance Criteria
- Creation of a prototype table that contains a sortable header which accurately updates table content when clicking sorting for ascending and descending
- At this stage, accessibility and design specifications are not acceptance criteria

nick2432:
can i work on this?

MisRob:
Hi @nick2432, I'm sorry, this is not ready to be developed yet. We recommend issues labeled as "help wanted" or "good first issue".

We currently have contributing opportunities in three repositories. You can see the contributing guidelines including links to issues suitable for contribution for each repository here:

- [Kolibri Contributing Guidelines](https://github.com/learningequality/kolibri/blob/release-v0.16.x/CONTRIBUTING.md)

- [Kolibri Design System Contributing Guidelines](https://github.com/learningequality/kolibri-design-system/blob/main/CONTRIBUTING.md)

- [Studio Contributing Guidelines](https://github.com/learningequality/studio/blob/unstable/CONTRIBUTING.md)

MisRob:
Closed by GSOC `KTable` project

Issue #328, opened 2022-05-06
`KSortableTable` Component


marcellamaki:
## Overview
The `KSortableTable` Component is an update to work begun with `CoreTable` and `SortableHeader` to create a KDS Component that will  

1. Have an easy-to-use, standard sortable table experience
2. Work across Kolibri and KDP
3. Meet best practices for screen-reader accessibility

### Impact and Timeline
The feature is in support of design updates for 0.16, for the "change facility" feature where a sortable table is required. By addressing this problem one time in KDS, we will be able to use this component across our products.

This is a Q2 project that should be completed well in advance of 0.16, although it will be deployed with that release. 

### Working Assumptions
- CoreTable and SortableHeader will be useful references 
- This component will be useful across products, and will be easily reusable

### Out of Scope
- This component does not need to handle filtering or searching
- This component does not need to manage pagination

### Acceptance Criteria 
The **core purpose** of making this a KDS component is for accessibility. Therefore, accessibility best practices will be the major part of the acceptance criteria. The other major goal is that the use of this component _should not_ leave too much to the developer. The core pieces of accessibility criteria should be the default behavior (i.e. not requiring extra configuration) and not easily modifiable. 

#### MVP

- **Full compliance to accessibility criteria is the baseline for our MVP.**
- Documentation for the new component added to KDS site

#### Additional notes
- Specs of the visual design of the table (colors, fonts, spacing, etc.) _may be changed_ as needed in support of accessibility best practices, but any changes should be done in consultation with Jessica and Radina. Changes to the spec that are _not_ made in support of accessibility are a no.
- Implementation of this component in Kolibri is out of scope of this epic -- this is just about the creation of the component 

### Resources 
[Figma](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=194%3A2920)

External resources on Accessible Sortable Tables: 
[USWDS: Sortable Tables](https://designsystem.digital.gov/components/table/#sortable-table-rows)
[Inclusive Components: Sortable Tables](https://inclusive-components.design/data-tables/)
[Multi-column sortable tables](https://adrianroselli.com/2021/06/multi-column-sortable-table-experiment.html)

MisRob:
May be related https://github.com/learningequality/kolibri-design-system/issues/220

MisRob:
Closed by GSOC `KTable` project

PR #727, opened 2024-08-13
Initial Implementation Of KTable Component

BabyElias:
## Description
This PR introduces the `KTable` component to KDS. The component is a flexible and customizable table component designed to handle a variety of data presentation needs within our application. It supports advanced features such as local sorting, sticky headers, keyboard navigation, and dynamic column resizing, making it suitable for both simple and complex data tables. 

#### Issue addressed

[328](https://github.com/learningequality/kolibri-design-system/issues/328)
[330](https://github.com/learningequality/kolibri-design-system/issues/330)

## Note to Reviewers :
- The table supports both `tab` key and `arrow` key navigations to move between the cells.
- The width of columns can be dynamically set by using the `width` and `min-width` props.
- When viewport is shrinked, based on min-width/default width properties, scroll is enabled in the x & y-directions.
- During scrolling in vertical direction, the column headers remain fixed and during scrolling in horizontal direction, the row headers remain fixed.
- If content of a cell exceeds its size, text-wrapping takes place.
- On mouse hover, the entire row over which mouse is hovered gets highlighted.
- When navigating using keyboard (tab/arrow keys), the cell, the entire row containing the cell as well as the column header containing the cell is highlighted to maintain better readability and context.
- The table supports another mode (not yet documented) - `Sortable table with external sorting method`. This external sorting can include cases where sorted data is fetched from backend, or use of custom sort functions with the table. An example of the same has been included in [Playground](https://deploy-preview-727--kolibri-design-system.netlify.app/playground/) page as Backend Sorted Table. In this example, clicking on any header will reverse the contents of the column - just an attempt to show the difference between external sorting and uselocalsorting. 

## Changelog

  - **Description:** Initial implementation of `KTable` component
  - **Products impact:** New Component
  - **Addresses:** [#328](https://github.com/learningequality/kolibri-design-system/issues/328)
  - **Components:** KTable
  - **Breaking:** No
  - **Impacts a11y:**  Yes
  - **Guidance:** 

## Steps to test
Sample tables have been added on the Playground Page for testing.

- https://deploy-preview-727--kolibri-design-system.netlify.app/playground/
- https://deploy-preview-727--kolibri-design-system.netlify.app/ktable/

## Testing checklist

## Reviewer guidance

## After review


MisRob:
@BabyElias I linked the docs build preview in the pull request description because if I recall you don't have access to 'Checks' on the PR

docs/pages/ktable.vue
```diff
@@ -0,0 +1,167 @@
+<template>
+
+  <DocsPageTemplate apiDocs>
+    <DocsPageSection title="Overview" anchor="#overview">
+      <p>
+        The <code>KTable</code> component is a flexible and customizable table component designed to handle a variety of data presentation needs within our application. It supports advanced features such as local sorting, sticky headers, keyboard navigation, and dynamic column resizing, making it suitable for both simple and complex data tables.
```
MisRob:
```suggestion
        The <code>KTable</code> component is an accessible and customizable table component designed to handle a variety of data presentation needs. The component is suitable for both simple and complex data tables. It offers:
        <ul>
          <li>Built-in sorting as well as integration with externally sorted data</li>
          <li>Keyboard navigation</li>
          <li>Dynamic column resizing</li>
          <li>Sticky headers</li>
        </ul>
```

Just an idea that may help to make the language a bit more clear and concise.

docs/pages/ktable.vue
```diff
@@ -0,0 +1,167 @@
...
+
+  <DocsPageTemplate apiDocs>
+    <DocsPageSection title="Overview" anchor="#overview">
+      <p>
+        The <code>KTable</code> component is a flexible and customizable table component designed to handle a variety of data presentation needs within our application. It supports advanced features such as local sorting, sticky headers, keyboard navigation, and dynamic column resizing, making it suitable for both simple and complex data tables. 
+      </p>
+    </DocsPageSection>
+    <DocsPageSection title="Usage" anchor="#usage">
+      <!--Non-Sortable Table-->
+      <h3>Table without Sorting functionality</h3>
```
MisRob:
```suggestion
      <h3>Table without sorting functionality</h3>
```

**[required] Would you decapitalize this title and all other titles on this page?** The convention on the KDS website is not to capitalize titles. It will help with consistency and orientation.

Total nitpick, shorter section titles, such as _"Without sorting"_, _"With built-in sorting"_, and (future) _"With externally sorted data"_ or similar, would do. On this page, it's clear we talk about the table.

You can see more details in the [Writing guidance](https://design-system.learningequality.org/writing/).

docs/pages/ktable.vue
```diff
@@ -0,0 +1,167 @@
...
+        ]
+      };
+    },
+  </DocsShowCode>
+
+      <DocsShow block>
+        <KTable
+          :headers="headers"
+          :rows="rows"
+          caption="Non Sortable Table"
```
MisRob:
```suggestion
          caption="Non-sortable table"
```

Explained in other comments.

docs/pages/ktable.vue
```diff
@@ -0,0 +1,167 @@
...
+      <h3>Table with Frontend Sorting</h3>
+      <p>
+        The <code>KTable</code> can be used with local sorting functionality, allowing you to sort data on the client side without the need for server requests. This example demonstrates a table with local sorting enabled.
+      </p>
+      <!-- eslint-disable -->
+      <DocsShowCode language="html">
+        <KTable
+          :headers="headers"
+          :rows="rows"
+          caption="Local Sorting Table"
```
MisRob:
```suggestion
          caption="Table with built-in sorting"
```

Explained in other comments.

lib/KTable/KTableGridItem.vue
```diff
@@ -0,0 +1,91 @@
+<template>
+
+  <td
+    :class="$computedClass(coreOutlineFocus)"
+    :style="{ textAlign: textAlign, minWidth: minWidth, width: width }"
+    tabindex="0"
+    role="gridcell"
```
MisRob, line 8: **[required] Let's place `data-focus="true"` below this line and remove `[role="gridcell"]` addition from the trackInputModality.js**. Details in the related comment.

lib/KTable/index.vue
```diff
@@ -0,0 +1,480 @@
...
+              [$computedClass(coreOutlineFocus)]: true,
+              sortable: sortable && header.dataType !== DATA_TYPE_OTHERS,
+              'sticky-header': true,
+              'sticky-column': index === 0,
+            }"
+            :style="[getHeaderStyle(header),
+                     sortKey === index ? { color: $themeBrand.primary.v_1000 } : { color: $themePalette.grey.v_800 },
+                     { backgroundColor: $themePalette.white } ,
+                     focusedColIndex === index ? { backgroundColor: $themePalette.grey.v_50 } : {}]"
+            role="columnheader"
```
MisRob: **[required] Let's place `data-focus="true"` below this line and remove `[role="columnheader"]` addition from the trackInputModality.js**. Details in the related comment.

lib/styles/trackInputModality.js
```diff
@@ -27,6 +27,8 @@ function setUpEventHandlers(disableFocusRingByDefault) {
     'button',
     // Custom KDS attribute. When attached to an element, it will receive a focus ring. Typically used for a11y optimizations, such as for KCard's <li>.
     '[data-focus=true]',
+    '[role=gridcell]',
+    '[role=columnheader]',
```
MisRob: **[required] Let's revert these two lines and instead apply `data-focus=true` on the corresponding elements**. Elements in this list trigger the keyboard modality and I think generally that's not expected from all elements with `role=gridcell` and `role=columnheader`. Perhaps it would not break anything, but this script has a long history of behaving troublesome and there's already `data-focus=true` ready for custom use-cases so we'll go safe :) I recall you mentioned some trouble so I tried it and seemed to work well on my end - let me know if that's the case for you too.

lib/composables/useSorting/index.js
MisRob, line 1: Nice!

The only improvement I could think of is that it would be good to have at least a few docstring comments for data and functions exported by this composable. Even though it's not a complex logic and the test suite helps a lot in understanding its purpose, it's not apparent on a first glance (to me at least :) So I think it will make orientation easier for some of us. Just a brief summary would do. JSDoc format is commonly used.

lib/composables/useSorting/__tests__/index.spec.js
MisRob, line 1: Nice! Neat and clear, meaningful coverage.

lib/composables/useSorting/index.js
```diff
@@ -0,0 +1,47 @@
...
+import _ from 'lodash';
+
+export const SORT_ORDER_ASC = 'asc';
+export const SORT_ORDER_DESC = 'desc';
+export const DATA_TYPE_STRING = 'string';
+export const DATA_TYPE_NUMERIC = 'numeric';
+export const DATA_TYPE_DATE = 'date';
+export const DATA_TYPE_OTHERS = 'others';
+
+export default function useSorting(headers, rows, useLocalSorting) {
```
MisRob: **[required] Would you move this file to `KTable` directory?** I'm sorry for recommending `composables` location, I assumed it would be public or private shared composable but at this point it is private and tightly coupled to the `KTable` component. That's fine, just better to have them together to make the connection clear.

lib/composables/useSorting/__tests__/index.spec.js
```diff
@@ -0,0 +1,97 @@
...
+    const { sortedRows } = useSorting(headers, rows, useLocalSorting);
+    expect(sortedRows.value).toEqual(rows.value);
+  });
+
+  it('should sort rows by string column in ascending order', () => {
+    const { handleSort, sortedRows } = useSorting(headers, rows, useLocalSorting);
+
+    handleSort(0); // Sort by 'Name'
+
+    expect(sortedRows.value).toEqual(_.orderBy(rows.value, [row => row[0]], [SORT_ORDER_ASC]));
```
MisRob: One testing pattern is to check against the actual result rather than computing it. Here it would be

```
expect(sortedRows.value).toEqual([
   ['Alice', 28, new Date(1992, 8, 10)],
   ['Jane', 25, new Date(1995, 10, 20)],
   ['John', 30, new Date(1990, 5, 15)],
])
```
The main reason is robustness of tests. For example here, if you used `_.orderBy` incorrectly in both the composable and here in the test, you would get a 'false positive' - passing test for a buggy logic.

I guess opinions on this will vary. I tend to prefer this approach. Added benefits are simplicity and immediately understanding what the function does and what data structure it returns, without slightest mental overhead. This is an opinionated tip and as with other non-required comments, up to you :)

lib/composables/useSorting/index.js
```diff
@@ -0,0 +1,47 @@
...
+  const sortKey = ref(null);
+  const sortOrder = ref(null);
+
+  const sortedRows = computed(() => {
+    if (!useLocalSorting.value || sortKey.value === null) return rows.value;
```
MisRob:
```suggestion
    if (!useLocalSorting.value || sortKey.value === null || sortOrder.value ===null) return rows.value;
```

**[required] I think we need yet another check here?** Seems that without it, `_.orderBy` below may potentially run with an invalid `null` order value. And it doesn't make sense to sort until we know in which order anyways.

I'm not entirely sure if we will actually run into any problems but it's quite common that `computed` runs before all values it depends on are ready and then it throws an unexpected error just to run correctly in another millisecond when the values are there, causing developers to investigate mysterious console errors that actually don't break anything for users :)

lib/composables/useSorting/index.js
```diff
@@ -0,0 +1,47 @@
+import { ref, computed } from '@vue/composition-api';
+import _ from 'lodash';
+
+export const SORT_ORDER_ASC = 'asc';
+export const SORT_ORDER_DESC = 'desc';
+export const DATA_TYPE_STRING = 'string';
+export const DATA_TYPE_NUMERIC = 'numeric';
+export const DATA_TYPE_DATE = 'date';
+export const DATA_TYPE_OTHERS = 'others';
```
MisRob: Since these values will need to be used by consumers, would you consider using `'string'`, `'number'`, `'date'`, and `'undefined'` so the naming is closer to JavaScript types? It's not the same thing, but I think it would be a bit more intuitive for JavaScript developers.

lib/KTable/KTableGridItem.vue
```diff
@@ -0,0 +1,91 @@
...
+      coreOutlineFocus() {
+        return {
+          ':focus': {
+            ...this.$coreOutline,
+            outlineOffset: '-2px',
+          },
+        };
+      },
+      textAlign() {
+        return this.dataType === DATA_TYPE_NUMERIC ? 'right' : 'left';
```
MisRob:
```suggestion
       const alignLtr = this.dataType === DATA_TYPE_NUMERIC ? 'right' : 'left';
       if (this.isRtl && alignLtr === 'right') { return 'left' }
       if  (this.isRtl && alignLtr === 'left') {return 'right'}
       return alignLtr;
```

or something like that..

**[required] Ensures that in right-to-left (RTL) languages, text alignment will be the opposite.**

RTLSCSS framework takes care automatically for anything in `<styles>` but for these dynamic JavaScript styles, we need to use `isRtl` utility. Details in [Kolibri developer documentation > RTL language support](https://kolibri-dev.readthedocs.io/en/develop/i18n.html#rtl-language-support).

lib/KTable/KTableGridItem.vue
```diff
@@ -0,0 +1,91 @@
...
+    },
+    methods: {
+      onKeydown(event) {
+        // Check if the focused element is a button or actionable item
+        const focusedElement = event.target;
+        if (
+          event.key === 'Enter' &&
+          (focusedElement.tagName === 'BUTTON' ||
+            focusedElement.tagName === 'A' ||
+            focusedElement.hasAttribute('role'))
```
MisRob: What was the intention behind having `focusedElement.hasAttribute('role')` here?

lib/KTable/KTableGridItem.vue
```diff
@@ -0,0 +1,91 @@
...
+            outlineOffset: '-2px',
+          },
+        };
+      },
+      textAlign() {
+        return this.dataType === DATA_TYPE_NUMERIC ? 'right' : 'left';
+      },
+    },
+    methods: {
+      onKeydown(event) {
```
MisRob, line 60:
```suggestion
      // Ensures that clickable elements within a table cell, such as buttons and links, can be clicked with ENTER key. 
      onKeydown(event) {
```

`KTable` is a component with the most a11y nuances in our codebase that not many people are familiar with. So I think a few function summary comments like here will be important in long-run, so we know what we're potentially breaking when updating code, and for easier orientation. 

I don't think you need to comment `KTable` file exhaustively, the code is pretty self-explanatory. Perhaps just very tricky/unexpected lines (that you've already commented on), and additionally these functions summaries should do.

lib/__tests__/KTable.spec.js
MisRob, line 1: Very well done! Good choices of what to test and what not to test, robust tests design, and good structure and clarity. Thanks for the comments too. By the way, I was caught by surprise to see that keyboard navigation can be tested in such a straightforward manner. Learned something new. Hard to believe these were some of your first test suites (if I recall :)

lib/KTable/index.vue
MisRob, line 1: Tons-of-good-work :)

lib/KTable/index.vue
```diff
@@ -0,0 +1,480 @@
...
+      </caption>
+      <thead>
+        <tr>
+          <th
+            v-for="(header, index) in headers"
+            :key="index"
+            tabindex="0"
+            :aria-sort="sortable && header.dataType !== DATA_TYPE_OTHERS ? getAriaSort(index) : null"
+            :class="{
+              [$computedClass(coreOutlineFocus)]: true,
```
MisRob: I think this could be simplified a bit?

```suggestion
              $computedClass(coreOutlineFocus): true,
```

lib/KTable/index.vue
```diff
@@ -0,0 +1,480 @@
...
+              'sticky-header': true,
+              'sticky-column': index === 0,
+            }"
+            :style="[getHeaderStyle(header),
+                     sortKey === index ? { color: $themeBrand.primary.v_1000 } : { color: $themePalette.grey.v_800 },
+                     { backgroundColor: $themePalette.white } ,
+                     focusedColIndex === index ? { backgroundColor: $themePalette.grey.v_50 } : {}]"
+            role="columnheader"
+            :aria-colindex="index + 1"
+            @click="sortable && header.dataType !== DATA_TYPE_OTHERS ? handleSort(index) : null"
```
MisRob: I think moving these conditions to `handleSort` on line 128 would help because some conditions are here and another ones are in `handleSort`, making it a bit more difficult to build a mental picture of what happens under what circumstances.

lib/KTable/index.vue
```diff
@@ -0,0 +1,480 @@
...
+            @keydown="handleKeydown($event, -1, index)"
+          >
+            <!--@slot Scoped slot for customizing the content of each header cell. Receives the following slot props - `header`: The current header object & `index`: The index of the current header.-->
```
MisRob:
```suggestion
            <!--@slot Scoped slot for customizing the content of each header cell. Provides a header object `header` and its column index `colIndex`.-->
```

Related to the suggested slot changes. Plus a few tweaks for clarity.

lib/KTable/index.vue
```diff
@@ -0,0 +1,480 @@
...
+
+            <slot name="header" :header="header" :index="index">
```
MisRob:
```suggestion
            <slot name="header" :header="header" :colIndex="index">
```

Few tweaks for clarity and consistency with the `cell` slot. Note that this would also need to be updated in the documentation examples.

lib/KTable/index.vue
```diff
@@ -0,0 +1,480 @@
...
+            }"
+            :style="[colIndex === 0 ? { backgroundColor: $themePalette.white } : {},(hoveredRowIndex === rowIndex || focusedRowIndex === rowIndex) && colIndex === 0 ? { backgroundColor: $themePalette.grey.v_50 } : {}
+            ]"
+            role="gridcell"
+            :aria-colindex="colIndex + 1"
+            @keydown="handleKeydown($event, rowIndex, colIndex)"
+          >
+            <template #default="slotProps">
+
+              <!-- @slot Scoped slot for customizing the content of each data cell. Receives the following slot props - `content`: The content of the current cell,`rowIndex`: The index of the current row & `colIndex`: The index of the current column.-->
```
MisRob:
```suggestion
              <!--@slot Scoped slot for customizing the content of each data cell. Provides the content of a data cell `content`, its row index `rowIndex`, its column index `colIndex`, and the corresponding whole row object `row`.-->
```

Few tweaks for clarity and consistency with the `header` slot. Also added `row`.

lib/KTable/index.vue
```diff
@@ -0,0 +1,480 @@
...
+            @click="sortable && header.dataType !== DATA_TYPE_OTHERS ? handleSort(index) : null"
+            @keydown="handleKeydown($event, -1, index)"
+          >
+            <!--@slot Scoped slot for customizing the content of each header cell. Receives the following slot props - `header`: The current header object & `index`: The index of the current header.-->
+
+            <slot name="header" :header="header" :index="index">
+              {{ header.label }}
+            </slot>
+            <span v-if="sortable && header.dataType !== DATA_TYPE_OTHERS" class="sort-icon">
+              <span v-if="sortKey === index && sortOrder === SORT_ORDER_ASC"><KIcon icon="dropup" :color="sortKey === index ? $themeBrand.primary.v_1100 : $themePalette.grey.v_800 " /></span>
```
MisRob: **[required] Some conditions repeat many times, and the template part is full so I'd recommend some smaller abstractions.** One example here, we could extract `sortKey === index` to a new `computed`, perhaps `isColumnSortActive(colIndex)` or similar. This would also help with understanding. There are several other places in the template part that would similarly benefit.

lib/KTable/index.vue
```diff
@@ -0,0 +1,480 @@
...
+        SORT_ORDER_ASC,
+        SORT_ORDER_DESC,
+        DATA_TYPE_OTHERS,
+        getHeaderStyle,
+      };
+    },
+    /* eslint-disable kolibri/vue-no-unused-properties */
+    props: {
+      /**
+       * An array of objects representing the headers of the table. Each header should have a label and a dataType.
```
MisRob:
```suggestion
       * An array of objects `{ label, dataType, minWidth, width }`representing the headers of the table. The `dataType` can be one of `'string'`, `'numeric'`, `'date'`, or `'others'`. `label` and `dataType` are required. `minWidth` and `width` are optional.
```

Few tweaks for clarity and to make the generated documentation format nice.

lib/__tests__/KTable.spec.js
```diff
@@ -0,0 +1,151 @@
...
+      propsData: {
+        headers,
+        items,
+        useLocalSorting: false,
+      },
+    });
+
+    const thElements = wrapper.findAll('th');
+    await thElements.at(0).trigger('click');
+    expect(wrapper.emitted().changeSort).toBeTruthy();
```
MisRob, line 80: We'd also typically check that it emitted with the correct payload. Because it's part of the public API.

lib/KTable/index.vue
```diff
@@ -0,0 +1,480 @@
...
+        return {
+          ':focus': {
+            ...this.$coreOutline,
+            outlineOffset: '-2px',
+          },
+        };
+      },
+    },
+    methods: {
+      handleKeydown(event, rowIndex, colIndex) {
```
MisRob, line 296: Related to the comment about summary docstrings for functions, to provide one more example I think here we will benefit from a high-level info. Something like:

```suggestion
/**
* Takes care of
*  - keyboard navigation focus trap
*  - the arrow keys navigation flow
*  - the tab keys navigation flow
*  - access to focusable elements within a cell via tab and shift tab keys
* - triggering sort on the enter key
*  - header highlight
*/
      handleKeydown(event, rowIndex, colIndex) {
```

lib/KTable/index.vue
```diff
@@ -0,0 +1,480 @@
...
+        const key = event.key;
+        const totalRows = this.rows.length;
+        const totalCols = this.headers.length;
+
+        let nextRowIndex = rowIndex;
+        let nextColIndex = colIndex;
+        // Identify all focusable elements inside the current cell
+        const currentCell = this.getCell(rowIndex, colIndex);
```
MisRob: I finally get why we were seeing the previous element from the focus ring on the call - this all is happening in `keydown`. Makes sense :)!

lib/KTable/index.vue
```diff
@@ -0,0 +1,480 @@
...
+        const focusableElements = currentCell
+          ? Array.from(
+              currentCell.querySelectorAll(
+                'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])'
+              )
+            )
+          : [];
+        const focusedElementIndex = focusableElements.indexOf(document.activeElement);
```
MisRob: Well done. I appreciate you think far beyond the use-case you currently work with and design the approach robustly. If we ever happen to have several clickable elements in a cell, like a list of links for example, we're ready to go :) And even if that doesn't happen, it's a very helpful practice to not make assumptions/dependencies to particular contexts outside the table's control :+1:

lib/KTable/index.vue
```diff
@@ -0,0 +1,480 @@
...
+            break;
+          case 'Enter':
+            if (rowIndex === -1 && this.sortable) {
+              this.handleSort(colIndex);
+            }
+            break;
+          case 'Tab':
+            if (focusableElements.length > 0) {
+              if (!event.shiftKey) {
+                if (focusedElementIndex < focusableElements.length - 1) {
```
MisRob, line 375:
```suggestion
                // if navigating between more focusable elements within the cell
                if (focusedElementIndex < focusableElements.length - 1) {
```

Just a comment suggestion - it became clear, but took me some time to get it because at first I didn't realize there could be more focusable elements.

lib/KTable/index.vue
```diff
@@ -0,0 +1,480 @@
...
+                  } else if (rowIndex < totalRows - 1) {
+                    nextColIndex = 0;
+                    nextRowIndex = rowIndex + 1;
+                  } else {
+                    nextColIndex = 0;
+                    nextRowIndex = -1;
+                  }
+                }
+              } else {
+                if (focusedElementIndex < focusableElements.length - 1) {
```
MisRob, line 391:
```suggestion
                 // if navigating between more focusable elements within the cell
                if (focusedElementIndex < focusableElements.length - 1) {
```

lib/KTable/index.vue
```diff
@@ -0,0 +1,480 @@
...
+                } else if (rowIndex > 0) {
+                  nextColIndex = totalCols - 1;
+                  nextRowIndex = rowIndex - 1;
+                } else {
+                  nextColIndex = totalCols - 1;
+                  nextRowIndex = totalRows - 1;
+                }
+              }
+            }
+            this.focusCell(nextRowIndex, nextColIndex);
```
MisRob: **[required] When I navigate with TAB key, lines 350-355 are executed and then lines 361-367 are executed too, causing duplicate calls of the same functions, which I don't think is intended, nor desired. Will need to be fixed.**

Relatedly, I think what contributes here is the structure of the code that has grown gradually and now there's many branches that make things like this hard to spot. Would you like to think if there are some ways of re-structuring the handler in a simpler way? We can chat.

lib/KTable/index.vue
```diff
@@ -0,0 +1,480 @@
...
+      getCell(rowIndex, colIndex) {
+        if (rowIndex === -1) {
+          return this.$el.querySelector(`thead th:nth-child(${colIndex + 1})`);
+        } else {
+          return this.$el.querySelector(
+            `tbody tr:nth-child(${rowIndex + 1}) td:nth-child(${colIndex + 1})`
+          );
+        }
+      },
+      scrollCellIntoView(cell) {
```
MisRob, line 454: @AlexVelezLl You're welcome to review anything/everything/nothing, but I'd like to particularly ask you for support with this part. Could I get your eyes on performance, please? Would be much appreciated :) 

I gave the function look and on a basic level it felt pretty safe due to being called only on a key press. On the other hand, we frequently read kinds of information from DOM that, if I recall, are causing re-calculations/re-layouts. Do you think we need to optimize?

Relatedly there's a few other `querySelector`s and perhaps some another ways of accessing DOM scattered here and there in the PR, if you'd like to preview them too.

MisRob (changes requested):
@BabyElias Very good-quality work overall. Well done! Thank you. I'm leaving a few questions and some feedback in code comments. Feedback that is not marked as required isn't blocking, rather tips, optional nice-to-have, or just for consideration or discussion.

MisRob:
@BabyElias I think this would be also a good point to think about the table's public API. Even though some minor APIs may be subject to change, you've got to a great place of having the most important pieces functioning robustly, so it'd be great to reflect that in the public interface and documentation, especially as we're moving towards using the table in production 🥳 (long story short, you're doing such a good progress, so now there's some extra work 😁) So on a high-level, I'd like to offer a proposal that should help with clarity and reflect how the table behaves. Let's discuss together, feedback is welcome.

- **(1) Provide the simplest behavior by default. Gradually progress towards more advanced or less frequent use-cases configured by means of optional APIs**
  - **Why:** A good component-design practice that makes using and documenting components easy and clear.
- **(2) Update the language around the table sorting modes to reflect the separation of concerns** 
  - There's no need to use _"backend"_, _"frontend"_, _"server"_ and similar to _define_ the table's sorting modes.  _By this I do not mean that these terms are prohibited_ :-) Later, when you document the integration with the externally sorted data, it'll be perfect to use the concrete example based on the server-side sorting use-case we work with as long as it's communicated as an example rather than the defining behavior of the table.
  - **Why:** The table doesn't and is not supposed to know about all these things. Its only concern is whether it should sort data internally or not. One example - sorted data doesn't need to necessarily come from the server, it could be the result of the client-side external custom sort function. For the table, it makes no difference. This is intended separation of concerns.

**Put together:**

|  | Proposed API |
| -------- |  --------------- |
| **Non-sortable** | `<KTable>` |
| **Sortable with the default sorting enabled** (simpler and the most frequent sorting use-case) | `<KTable sortable>` |
| **Sortable with the built-in sorting disabled** (rather advanced and less frequent use-case) | `<KTable sortable disableBuiltInSorting>`  or `<KTable sortable disableDefaultSorting>` or similar|

As part of this, language used on the documentation page would be updated accordingly. Instead of 'backend', 'frontend' etc., we'd use 'externally sorted', 'built-in', 'default' or similar.

MisRob:
@radinamatic you can start testing on the playground before the Kolibri PR is finished. It has more use-cases, so it will be useful in any case.

MisRob:
@radinamatic @LianaHarris360 @marcellamaki 

It's fine to review existing places, just note there's no full RTL support yet - will be done by the end of GSOC or there will be a follow-up issue.

lib/KTable/index.vue
```diff
@@ -342,16 +342,16 @@
...
                 } else {
-                  nextColIndex = totalCols - 1;
-                  nextRowIndex = totalRows - 1;
+                  // Allow default behavior when reaching the first cell
+                  return;
                 }
               }
             }
-            this.focusCell(nextRowIndex, nextColIndex);
+            this.focusCell(nextRowIndex, nextColIndex, true);
```
MisRob: What was the intention behind the third `true` parameter here? I don't see it being used in the function.

MisRob:
@radinamatic @LianaHarris360 @marcellamaki @AlexVelezLl Ready for your feedback! You're all welcome  to review as much as you'd like or only chime in for some parts. There is also a Kolibri PR coming. I think I reviewed this PR carefully and would be nice to have more space for other reviewers so if someone would like to be the main assignee on the other PR, I would be grateful. I could still chime in briefly if needed. It would also help me time-wise since I will be re-reviewing this PR at some point.

@BabyElias Thanks for all the guidance and playground preparation, very helpful

Can't wait to see this live :) Thanks everybody.

MisRob:
@radinamatic @LianaHarris360 @marcellamaki @AlexVelezLl 

Some context for reviews and QA:
- It's fine to improve existing places in code, and resolve blocking issues if we spot some in Kolibri, just note there's no nuanced RTL support yet
- There's generally lots of features and modes we'll want to document, however at this stage, I asked @BabyElias to focus on documenting only high-level, with one example for each major table mode (third one, for externally sorted data, is about to come in another PR).

There will be some work on this by the end of GSOC if there is time left or there will be follow-up issues. There's already lots of great things done or in progress :)

lib/KTable/KTableGridItem.vue
BabyElias: In case there are elements like progress bar (I saw it in some Kolibri tables that had some information related to assessment questions), then the progress bar would hopefully have a role and be focusable within the cell. I am not sure if not focussing the progress bar/any other such representative elements within the cell would allow screen readers to read their content or not.

lib/__tests__/KTable.spec.js
BabyElias: Thank youu :)

lib/KTable/KTableGridItem.vue
MisRob: Ah I see. As always you think broad and ahead :)! That's a good consideration. I think for such cases we'd need to call `focus` rather than `click` though? I don't quite see how `click` would help with non-clickable elements.

MisRob: With this I am not suggesting we do some branching here to call `focus`. I am not entirely sure how it should work for progress bars. It's just that `click` seems to only make sense for clickable elements to me. Perhaps better remove line 71 for now and clarify later when the need arises? I think as soon as we start introducing the table in more places, we will always want to do a11y QA run on the feature anyways so it should surface. Generally KDS can't do everything on its own, particularly for content provided via slots - so this is quite a common approach.

MisRob: Okay not even `focus` needed per @radinamatic's recommendation

docs/pages/ktable.vue
```diff
@@ -0,0 +1,173 @@
...
+          :rows="rows"
+          caption="Non Sortable Table"
+          :sortable="false"
+        >
+
+          <template #header="{ header, index }">
+            <span>{ header.label } (Backend)</span>
+          </template>
+          <template #cell="{ content, rowIndex, colIndex }">
+            <span v-if="colIndex === 2">{ content } (City)</span>
```
AlexVelezLl, line 34: I dont know if you have already discussed about this. But I think as a future enhancement we could add dynamic slots for content cells depending on the headers. For example, we could add an "id" prop to the headers, and then dynamically pass a slot with that id, so we dont need to check the colIndex. It could be something like:
```vue
        <KTable
          :headers="headers"
          :rows="rows"
          caption="Non-sortable table"
          :sortable="false"
        >
          <template #header="{ header, index }">
            <span>{{ header.label }} (Backend)</span>
          </template>
          <template #cell="{ content }">
            <span> {{ content }} </span>
          </template>
          <template #city="{ content }">
            <span>{{ content }} (City)</span>
          </template>
        </KTable>
```
(Note that for the rest of the headers we could use the #cell slot as fallback, so we dont need to define a slot for each header, and if we want just to customize a single column with this we could avoid to pass a cell slot)

So it is a little more maintainable, and if we change the position of the headers we wouldnt have to change the template or have to deal with counting which index corresponds to each attribute.

Its just an Idea, but we could discuss if it seems relevant for the KTable usage and if it would be useful for other products to have this flexibility.

lib/styles/trackInputModality.js
BabyElias: Made the required changes

lib/composables/useSorting/__tests__/index.spec.js
BabyElias: This approach actually makes more sense in terms of testing the sorting functionality. Made the required changes. Thanks for this note!

lib/KTable/index.vue
BabyElias: I was trying a different approach to implement the tab key behaviour earlier that involved use of this 'true' prop. Seems I forgot to clean it up :

lib/composables/useSorting/index.js
BabyElias: It's always better to have double checks, than a potentially breaking code. Made the change.

docs/pages/ktable.vue
```diff
@@ -0,0 +1,173 @@
+<template>
+
+  <DocsPageTemplate apiDocs>
+    <DocsPageSection title="Overview" anchor="#overview">
+      <p>
+        The <code>KTable</code> component is an accessible and customizable table component designed to handle a variety of data presentation needs. The component is suitable for both simple and complex data tables. It offers:
+        <ul>
+          <li>Built-in sorting as well as integration with externally sorted data</li>
```
AlexVelezLl: Since we already support "externally sorted data" it would be good to share some examples in the documentation to know how to use it.

docs/pages/ktable.vue
```diff
@@ -0,0 +1,173 @@
...
+            <span v-if="colIndex === 2">{{ content }} (City)</span>
+            <span v-else>{{ content }}</span>
+          </template>
+        </KTable>
+      </DocsShow>
+       <!-- eslint-enable -->
+      <!-- Frontend Sorting Example-->
+      <h3>Table with Frontend Sorting</h3>
+      <p>
+        The <code>KTable</code> can be used with local sorting functionality, allowing you to sort data on the client side without the need for server requests. This example demonstrates a table with local sorting enabled.
```
AlexVelezLl: Is this sortable appllied to every column? I mean, every column will show the sort Icon right, Can we exclude any column?. Also what does "useLocalSorting" refers to? What happens if it is false? Would be good to have these explained in the documentation :). Or if it is explained in the props section would be good to make the reference to that section.

docs/pages/ktable.vue
```diff
@@ -0,0 +1,173 @@
...
+          :rows="rows"
+          caption="Table with built-in sorting"
+          :useLocalSorting="true"
+          sortable
+        >
+
+          <template #header="{ header, index }">
+            <span>{ header.label } (Backend)</span>
+          </template>
+          <template #cell="{ content, rowIndex, colIndex }">
```
AlexVelezLl: Would be great to explain that these slots are optional just for more specific use cases, and if we dont pass any, we will still have the headers and labels rendered in the table. So this shows the flexibility of our KTable and see how something as basic as:
``` Vue
<KTable
    :headers="headers"
    :rows="rows"
    caption="Non-sortable table"
 />
```

would render the table too :D.

That being said, I think that if in this section we are going to highlight how to enable the sorting view, it would be better to have just the minimal necessary setup code so people can focus more on the "sortable" prop, and dont need to look in more detail to see if any of the slots or header has something else to enabling the sorting feature. If for this example I see something like this:
```Vue
        <KTable
          sortable
          useLocalSorting
          :headers="headers"
          :rows="rows"
          caption="Table with built-in sorting"   
        />
```
I will inmediately notice that the only props needed for the table to be sortable are these two props `sortable` and `useLocalSorting' which reduces the cognitive load of the reader.

docs/pages/ktable.vue
AlexVelezLl: Reading the code I just noticed that if the dataType of the header is "others" we dont get the column sortable. I think that is worth mentioning here too.

lib/KTable/index.vue
```diff
@@ -0,0 +1,484 @@
...
+              <span v-else><KIcon icon="sortColumn" :color="$themePalette.grey.v_800" /></span>
+            </span>
+          </th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr
+          v-for="(row, rowIndex) in finalRows"
+          :key="rowIndex"
+          :style="hoveredRowIndex === rowIndex || focusedRowIndex === rowIndex ? { backgroundColor: $themePalette.grey.v_50 } : {}"
```
AlexVelezLl: Would be better if we have this in a computed prop so its easier for people to understand the logic.

lib/KTable/index.vue
```diff
@@ -0,0 +1,484 @@
...
+            :content="col"
+            :dataType="headers[colIndex].dataType"
+            :minWidth="headers[colIndex].minWidth"
+            :width="headers[colIndex].width"
+            :rowIndex="rowIndex"
+            :colIndex="colIndex"
+            :class="{
+              'sticky-column': colIndex === 0,
+            }"
+            :style="[colIndex === 0 ? { backgroundColor: $themePalette.white } : {},(hoveredRowIndex === rowIndex || focusedRowIndex === rowIndex) && colIndex === 0 ? { backgroundColor: $themePalette.grey.v_50 } : {}
```
AlexVelezLl: Idem. Ternaries in the template of the component are a little bit harder to read. As this is a little more complex logic we could even add some comments in the computed prop to explain the logic. This could probably be simplified with `if ... else if ... else` as I see both items in the array are setting a backgroundColor.

lib/KTable/index.vue
```diff
@@ -0,0 +1,484 @@
...
+       * The caption of the table
+       */
+      caption: {
+        type: String,
+        required: true,
+      },
+      /**
+       * Enables or disables local sorting of the table data.
+       */
+      useLocalSorting: {
```
AlexVelezLl: While reading the code I noticed this useLocalSorting refers to if it is sorted in the backend (or externally) or in the frontend. I didnt get it by just reading the props section, I think we could specify more what does local sorting refers to, since as a external user that hasnt been involved in the project I got a little bit confused.

lib/KTable/index.vue
```diff
@@ -0,0 +1,484 @@
...
+        const currentCell = this.getCell(rowIndex, colIndex);
+        const focusableElements = currentCell
+          ? Array.from(
+              currentCell.querySelectorAll(
+                'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])'
+              )
+            )
+          : [];
+        const focusedElementIndex = focusableElements.indexOf(document.activeElement);
+        switch (key) {
```
AlexVelezLl, line 304: [Optional] I think this could be broken down in smaller handler methods for each key case, and explain a little bit the logic in each of them. For example just reading the code I didnt realize that if you where at the beginning of the row and press left the focus changes to the last item in the previous row. We could have perhaps a handleArrowUp() that returns [nextRowIndex, nextColumnIndex] and just do something like
```javascript
case 'ArrowUp':
  [nextRowIndex, nextColIndex] = this.handleArrowUp();
case 'ArrowDown':
  [nextRowIndex, nextColIndex] = this.handleArrowDown();
```
and so on.

And its even more important with the tab, as it has an even more complex logic.

lib/KTable/index.vue
AlexVelezLl: Sure, I will!

lib/KTable/index.vue
```diff
@@ -0,0 +1,484 @@
...
+          }
+          if (cellRect.left < wrapperRect.left + stickyColumnWidth) {
+            tableWrapper.scrollLeft -= wrapperRect.left + stickyColumnWidth - cellRect.left;
+          }
+        }
+      },
+      focusCell(rowIndex, colIndex) {
+        let nextCell;
+        if (rowIndex === -1) {
+          nextCell = this.$el.querySelector(`thead th:nth-child(${colIndex + 1})`);
```
AlexVelezLl: I think its better to have vue refs to the elements to make this operation more lighter to the browser, as this method is going to be called on every keypress.

AlexVelezLl (commented):
This looks super great!! The interface is super flexible and easy to use, enabling sorting is super easy, the keyboard navigation works like a charm. Thank you very much for this great work @BabyElias!!

I think the documentation can be improved a little to better demonstrate the flexibility and how easy to use is this component, and the code can be refactored in a couple of places, but overall it looks great :smiley:.

(Apologies if I pointed to something that was already been discused. I didn't get too deep into Misha's review).

docs/pages/ktable.vue
MisRob: Feel fee to brainstorm together, just chiming in to mark as a possible _follow-up_.
 
(this could indeed be useful @AlexVelezLl, however with @BabyElias we agreed to only focus on high-level public API adjustments, and even those are _not required_ - the table was originally meant to be just a prototype :)

MisRob: Feel fee to brainstorm together, just chiming in to mark as a possible _follow-up_.
 
(this could indeed be useful @AlexVelezLl, however with @BabyElias we agreed to only focus on high-level public API adjustments, and even those are _not required_ - the table was originally meant to be just a prototype :)

MisRob: Feel fee to brainstorm together, just chiming in to mark as a _follow-up_.
 
(this could indeed be useful @AlexVelezLl, however with @BabyElias we agreed to only focus on high-level public API adjustments for logic that already exists, rather than adding new features, and even those are not required - the table was originally meant to be just a prototype :)

docs/pages/ktable.vue
MisRob: Thank you, this will be done later, together with the second Kolibri use-case.

lib/KTable/index.vue
MisRob: Note that this prop may be subject to change and hopefully it helps to address this as well https://github.com/learningequality/kolibri-design-system/pull/727#issuecomment-2295168580

docs/pages/ktable.vue
MisRob: That's not to dismiss the comment, it's very good! Just doing a bit of triaging here.

AlexVelezLl: Yes, yes! This could be a future improvement. I can do a follow up proposal issue if you agree with this

docs/pages/ktable.vue
AlexVelezLl: Alright, thank you!

docs/pages/ktable.vue
MisRob: Oh lovely, thanks so much for your offer! I am fine chiming in later on if you'd like to hear on something particular, but for now will let @BabyElias to enjoy the discussion and make some conclusions together with you if she'd like to and there's time :) (but no pressure @BabyElias, no problem to categorize this as post-GSOC)

BabyElias: That sounds like a good addition to KTable, thanks for suggesting this @AlexVelezLl !
I didn't have an idea about dynamic slots before this, read about them now for the first time. I agree that they will be more useful than the existing slot arrangement. However, I am not sure if I can cover them during the GSoC period or not. As suggested by @MisRob, I guess we can add it to triage/future enhancements and then maybe I can look into it myself post GSoC as well :)

AlexVelezLl: Perfect! I'm gonna create the issue just to immortalize the idea, and we can discuss it later on, no pressure to look at it for now :hugs:.

lib/KTable/index.vue
BabyElias: Gives me this error when doing so.
[{
	"resource": "/home/anoushka/kolibri-design-system/lib/KTable/index.vue",
	"owner": "_generated_diagnostic_collection_name_#0",
	"code": "8010",
	"severity": 8,
	"message": "Type annotations can only be used in TypeScript files.",
	"source": "ts",
	"startLineNumber": 16,
	"startColumn": 49,
	"endLineNumber": 16,
	"endColumn": 53
}]

docs/pages/ktable.vue
BabyElias: done

docs/pages/ktable.vue
BabyElias: done

lib/KTable/index.vue
AlexVelezLl: There is a performance issue here, but not because of layout or repainting, but because of scripting. Having more than 2 seconds scripting in a test of 6 seconds just tabbing at a normal speed like an user could do is a lot.

image

This is proably because we have several querySelectors to get the current cell, to get the focusable elements inside the cell, to get the next cell, to get the sticky header, to get the sticky column, and to get the header to highlight. Besides with the fact that querySelector and querySelectorAll has a less performance in several cases compared to the traditionals getElementBy* https://dev.to/wlytle/performance-tradeoffs-of-queryselector-and-queryselectorall-1074. 

I think replacing all of these calls with [vue template refs](https://www.w3schools.com/vue/vue_refs.php) will help a lot with this

lib/KTable/index.vue
```diff
@@ -0,0 +1,484 @@
...
+        // Identify all focusable elements inside the current cell
+        const currentCell = this.getCell(rowIndex, colIndex);
+        const focusableElements = currentCell
+          ? Array.from(
+              currentCell.querySelectorAll(
+                'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])'
+              )
+            )
+          : [];
+        const focusedElementIndex = focusableElements.indexOf(document.activeElement);
```
AlexVelezLl: We use `currentCell`, `focusableElemens` and `focusedElementIndex` just when we navigate with tabs, so its better to move this into the tabs handler block to avoid these calculations for the rest of keypresses :)

lib/KTable/index.vue
MisRob: Thanks for looking into this @AlexVelezLl, this is a great catch.

lib/KTable/index.vue
MisRob: My bad, I didn't notice this is attribute of a wrapping object

lib/KTable/index.vue
BabyElias: done

lib/KTable/index.vue
BabyElias: Added the logic as a computed property, using if-else statements.

lib/KTable/index.vue
BabyElias: I have fixed the duplicate call issue, still thinking about how we can re-structure the handler in a simpler way. There are a lot of interdependent functions being executed along with the handler, so it might be a task to figure out an appropriate refactor. Still working on it.

lib/KTable/index.vue
BabyElias: Seems like a repeated condition, because handlesort already has checks to handle 'Others' datatype. Made the change.
```
      const handleSort = index => {
        if (headers.value[index].dataType === DATA_TYPE_OTHERS) {
          return;
        }
```

lib/KTable/KTableGridItem.vue
```diff
@@ -0,0 +1,97 @@
...
+      coreOutlineFocus() {
+        return {
+          ':focus': {
+            ...this.$coreOutline,
+            outlineOffset: '-2px',
+          },
+        };
+      },
+      textAlign() {
+        const alignLtr = this.dataType === DATA_TYPE_NUMERIC ? 'right' : 'left';
```
LianaHarris360: Is there a particular reason numeric data types are right aligned compared to other data types? I'm curious about this because there seems to be some misalignment compared to the centered headers when creating classes in the Kolibri pr: 
classListAlignment

lib/KTable/index.vue
```diff
@@ -0,0 +1,505 @@
...
+  };
+
+</script>
+
+
+<style scoped>
+.k-table-wrapper {
+  overflow: auto;
+  position: relative;
+  height:300px;
```
LianaHarris360: Setting a minimum height or adjusting the height to auto could help to remove this side scroll bar and enhance the table's responsiveness in Kolibri.
ktablewithscrollbar

lib/KTable/KTableGridItem.vue
BabyElias: This was inspired from the MS Excel default alignment for numbers. In MS Excel, text and other datatypes are aligned to the left while numbers are aligned to the right. This ideally helps with :
- Enhanced readability since numbers are generally read from right-to-left. So, this makes numbers of different lengths to be visualised/compared easily.  
- Better visibility as can be distinguished from other datatypes.
As an example, in the below image, data presented in the right-aligned format seems more readable than the data aligned in left format.
image

lib/KTable/index.vue
BabyElias: Hey @AlexVelezLl,
I made the changes in replacing queryselectors with getElementbyTagName and Vue refs. Can you please check now if the performance is any better?

lib/KTable/index.vue
BabyElias: done

lib/KTable/index.vue
BabyElias: done

lib/KTable/index.vue
BabyElias: Just a heads up on this one, I tried refactoring the code to the suggested format. All navigations work perfect except the tab key navigation. The code is same as the one that I have been using so far and am unable to figure out what might have gone wrong. Attaching a snippet of that code here (haven't pushed it yet). Are you or @MisRob able to spot anything that might be an issue?
```
handleKeydown(event, rowIndex, colIndex) {
        const key = event.key;
        const totalRows = this.rows.length;
        const totalCols = this.headers.length;

        let nextRowIndex = rowIndex;
        let nextColIndex = colIndex;

        switch (key) {
          case 'ArrowUp':
            [nextRowIndex, nextColIndex] = this.handleArrowUp(rowIndex, colIndex, totalRows);
            break;
          case 'ArrowDown':
            [nextRowIndex, nextColIndex] = this.handleArrowDown(rowIndex, colIndex, totalRows);
            break;
          case 'ArrowLeft':
            [nextRowIndex, nextColIndex] = this.handleArrowLeft(
              rowIndex,
              colIndex,
              totalRows,
              totalCols
            );
            break;
          case 'ArrowRight':
            [nextRowIndex, nextColIndex] = this.handleArrowRight(
              rowIndex,
              colIndex,
              totalRows,
              totalCols
            );
            break;
          case 'Enter':
            if (rowIndex === -1 && this.sortable) {
              this.handleSort(colIndex);
            }
            break;
          case 'Tab':
            [nextRowIndex, nextColIndex] = this.handleTab(
              event,
              rowIndex,
              colIndex,
              totalRows,
              totalCols
            );
            break;
          default:
            return;
        }

        this.focusCell(nextRowIndex, nextColIndex);
        this.focusedRowIndex = nextRowIndex === -1 ? null : nextRowIndex;
        this.focusedColIndex = nextColIndex;

        this.highlightHeader(nextColIndex);

        event.preventDefault();
      },
      
      handleTab(event, rowIndex, colIndex, totalRows, totalCols) {
        let nextRowIndex = rowIndex;
        let nextColIndex = colIndex;
        const currentCell = this.getCell(rowIndex, colIndex);
        const focusableElements = [];

        if (currentCell) {
          const buttons = currentCell.getElementsByTagName('button');
          const links = currentCell.getElementsByTagName('a');
          const inputs = currentCell.getElementsByTagName('input');
          const selects = currentCell.getElementsByTagName('select');
          const textareas = currentCell.getElementsByTagName('textarea');

          focusableElements.push(...buttons, ...links, ...inputs, ...selects, ...textareas);
        }
        console.log('Current cell:', currentCell);
        console.log('Focusable elements:', focusableElements);
        const focusedElementIndex = focusableElements.indexOf(document.activeElement);
        if (focusableElements.length > 0) {
          if (!event.shiftKey) {
            if (focusedElementIndex < focusableElements.length - 1) {
              focusableElements[focusedElementIndex + 1].focus();
              event.preventDefault();
              return [rowIndex, colIndex];
            } else {
              if (colIndex < totalCols - 1) {
                nextColIndex = colIndex + 1;
              } else if (rowIndex < totalRows - 1) {
                nextColIndex = 0;
                nextRowIndex = rowIndex + 1;
              } else {
                return [rowIndex, colIndex];
              }
            }
          } else {
            if (focusedElementIndex < focusableElements.length-1) {
              focusableElements[focusedElementIndex + 1].focus();
              event.preventDefault();
              return [rowIndex, colIndex];
            } else {
              if (colIndex > 0) {
                nextColIndex = colIndex - 1;
              } else if (rowIndex > 0) {
                nextColIndex = totalCols - 1;
                nextRowIndex = rowIndex - 1;
              } else {
                return [rowIndex, colIndex]; // Allow default tab behavior when reaching the last cell
              }
            }
          }
        } else {
          if (!event.shiftKey) {
            if (colIndex < totalCols - 1) {
              nextColIndex = colIndex + 1;
            } else if (rowIndex < totalRows - 1) {
              nextColIndex = 0;
              nextRowIndex = rowIndex + 1;
            } else {
              return [rowIndex, colIndex];
            }
          } else {
            if (colIndex > 0) {
              nextColIndex = colIndex - 1;
            } else if (rowIndex > 0) {
              nextColIndex = totalCols - 1;
              nextRowIndex = rowIndex - 1;
            } else {
              return [rowIndex, colIndex];
            }
          }
        }

        return [nextRowIndex, nextColIndex];
      },
```

lib/KTable/index.vue
BabyElias: I have added a prop to the table- `height` - that can be used to set the height of the table. It is set to `auto` by default for now.

docs/pages/ktable.vue
```diff
@@ -0,0 +1,156 @@
+<template>
+
+  <DocsPageTemplate apiDocs>
+    <DocsPageSection title="Overview" anchor="#overview">
+      <p>
+        The <code>KTable</code> component is an accessible and customizable table component designed to handle a variety of data presentation needs. The component is suitable for both simple and complex data tables. It offers:
+        <ul>
+          <li>Built-in sorting as well as integration with externally sorted data</li>
```
marcellamaki: I know Alex previously commented here, but I am actually not totally sure what this mean. I think if the distinction here is sorting without an API call, or sorting with an API call, it would be clearer to say that. Built-in sorting more or less makes sense to me, but the "externally sorted data" is not super clear to me.

MisRob: Hi @marcellamaki, thank you.

I suggested this phrasing. If it's not very clear, we should improve it, however I wouldn't recommend using the language of API calls here.

> if the distinction here is sorting without an API call, or sorting with an API call

The thing is that this is not the distinction. Rather it simply is - _does the table do sorting or does it receive sorted data from outside_? Here we want to say that it is designed flexibly enough to work well in both modes. When receiving sorted data from outside, it's not necessarily from waiting on an API call, it could actually be from frontend as well (won't go into detail much, but can elaborate if needed). Relatedly, KDS components as such don't really need to (and imho better don't) have information about all the details of all contexts they may be used in - that's intentional separation of concerns.

That's to provide context, not to dismiss. I think your comment shows we need to improve it. Taking the use-cases I mentioned into account, can you think of better way to communicate it? Perhaps we could just say it very simply "Offers built-in sorting by default" and "Integrates with already sorted data"/"Can receive already sorted data"? Or any other ideas?

lib/KTable/index.vue
MisRob: @BabyElias Do we need a new prop? I would assume that `<KTable :style="{ height: ... }" />` should work if we remove the hardcoded value from inside that @LianaHarris360 pointed out? I think you could even set it to `'auto'` on line 469 to provide default behavior. It should still be properly overriden by `style`.

lib/KTable/KTableGridItem.vue
MisRob: Thanks for feedback.

It seems to me that what @LianaHarris360 pointed out is larger issue - we don't have borders within the table which may make some information harder to categorize (e.g. does 'user' belong to 'Class name' or 'Coaches' column?)

I can also see the value in what @BabyElias is saying - so not exactly sure if solution would be to show some borders or enforce different alignments or even make them configurable.

@BabyElias when we get to open some follow-up issues (later on), we can also include this one in it - may need some input from more folks.

lib/KTable/index.vue
MisRob: I will have a look soon :)

lib/KTable/index.vue
BabyElias: Didn't check that one, ideally it should work. Looking into it.

lib/KTable/KTableGridItem.vue
BabyElias: Okay!

lib/KTable/index.vue
MisRob: Nothing immediately obvious, I think I'd need to see this in a working branch

BabyElias: https://github.com/BabyElias/kolibri-design-system/tree/code-refactor-trial
The trial branch. @MisRob

MisRob: @BabyElias One of the issues with TAB I observe is that it gets stuck on a cell that has focusable element. This is caused by a difference in how the 'after refactor' version works, even though code looks quite the same.

I think the most clear way to point you I could think of would be to recommend you do

```
console.log('will focus cell', nextRowIndex, nextColIndex);
this.focusCell(nextRowIndex, nextColIndex);
```

and see what happens in both versions.

Before, `focusCell` wasn't called right before focus goes to the button in the cell.
After, `focusCell` is called every time, causing some issues later on that can be observed at various places in code.

Looking at the code, I wonder if `return` statements could be a culprit. My guess would be that this is just one problematic place I spotted.

Here, I'd recommend to not transfer the code 1:1 during this refactor, but focus line by line what needs to happen from the user point of view, and how that translates to the new structure.

MisRob: Again as for other things, this is optional, and if you'd prefer we open a follow-up, that's alright. I think it may even be preferable since before we had a stable version, this needs more re-thinking than expected, and there's already lots of things in this PR :)

BabyElias: Okay! I'll give it another try today. In case that does not work up, we'll add it to the follow up issues :)

lib/KTable/index.vue
BabyElias: It works! Reverted this change and set default height of table as 'auto'

docs/pages/ktable.vue
marcellamaki: > "Offers built-in sorting by default" and "Integrates with already sorted data"

this sounds great to me! I know it isn't really that different, but for some reason, I find that easier to understand. :)

MisRob: Well apparently it is - seems the simpler the better :) Thanks

MisRob: @BabyElias could we make this tweak here please? Later on as we continue building the documentation, either during GSOC or after, this will set up the language nicely.

BabyElias: Yes, sure!

lib/KTable/index.vue
BabyElias: Tried today, still to no avail. It's something up with the return statements for sure. But what is that, I have still not been able to figure out. Tried a few different combinations with those statements but no results as such. Think it might be better to add it to post GSoC task-list :(

MisRob: Yes, no problem :) Better to take some time to think through it

CHANGELOG.md
MisRob, line 1: Leaving a note here that before merging it'd be good to revert CHANGELOG.md changes since the new GH action will now update the CHANGELOG.md after merging this PR by pasting the changelog from the PR description. Otherwise we'd have duplicate entries.

BabyElias: done

lib/KTable/index.vue
```diff
@@ -0,0 +1,524 @@
...
+      handleRowMouseLeave() {
+        this.hoveredRowIndex = null;
+      },
+      setHighlightHeader(header, highlight) {
+        header.style.backgroundColor = highlight
+          ? this.$themePalette.grey.v_50
+          : this.$themePalette.white;
+      },
+      highlightHeader(colIndex) {
+        const headers = this.$el.querySelectorAll('thead th');
```
AlexVelezLl: (optional) Here we can use the template refs as we already set them for header cells.

lib/KTable/index.vue
AlexVelezLl: Hey, just unresolved to say the reason why the tab doesnt work in those cases as I imagined it could happen when I suggested the refactor :sweat_smile:. (@BabyElias You dont need to read all this comment right now, its still optional and it can be solved in a follow-up issue without problems and you can then revisit it).

Right now the return statements cuts the execution of the `handleKeydown` method, for example here:

```javascript
if (focusedElementIndex < focusableElements.length - 1) {
  focusableElements[focusedElementIndex + 1].focus();
  event.preventDefault();
  return;
}
```
This means that if `focusedElementIndex < focusableElements.length - 1` none of the code below will be executed because we are already focusing the element that needs to be focused and nothing more needs to be done (we dont need the `nextColIndex`, `nextRowIndex`, nor the focusCell method that is called below, etc). 

But in the refactored implementation that return statement is not going to cut the execution of the `handleKeydown` but the execution of the `handleTab` method, and there is the error. When refactoring and breaking down methods that has a return statement in the middle of the method, we need to ensure that we keep that behaviour after the refactor and make sure the method keeps stopping the execution in the same place it was doing it before the refactor. Right now with the current refactor in https://github.com/BabyElias/kolibri-design-system/tree/code-refactor-trial the code is:

```javascript
if (focusedElementIndex < focusableElements.length - 1) {
  focusableElements[focusedElementIndex + 1].focus();
  event.preventDefault();
  return [rowIndex, colIndex];
}
```

So, you are returning the current `rowIndex, colIndex` as a result of `handleTab`, returning this means that the given cell is the one that need to be focused when you press the tab, but that is an error, as the `handleKeydown` is suposed to stop its execution and avoid calling the `focusCell` method if that condition `focusedElementIndex < focusableElements.length - 1` is met, because we are already focusing the thing that need to be focused inside the handle tab method (the line `focusableElements[focusedElementIndex + 1].focus();`) and nothing more needs to be done after the `handleTab` method returns.

But, as we are returning `return [rowIndex, colIndex];` from `handleTab` that means that this is the cell that needs to be focused as a result of `handleKeyDown`, but its not the case as `handleKeyDown` has aready done its job after focusing the focusableElement, and we need to cut its execution there.

Soo, we can see it as `handleTab` process and returns the cell that needs to be focused when we press tab, but in this specific case we need to return something that tells the caller method that there is no cell to focus, as handleTab already took care of that. So an option is to just keep the empty return statements there inside the `handleTab` mthod that makes the return to be `undefined`(or return `null`) instead of the current `return [rowIndex, colIndex];`. So returning these falsy values will mean `there is no cell that need to be focused` and then we can cut the execution of the `handleKeydown` method there.

So, a fixed example would be:
```javascript
// inside `handleTab`
if (focusedElementIndex < focusableElements.length - 1) {
  focusableElements[focusedElementIndex + 1].focus();
  event.preventDefault();
  return null;
}
```

```javascript
// inside handleKeyDown
case 'Tab':
  const nextCell = this.handleTab(
    event,
    rowIndex,
    colIndex,
    totalRows,
    totalCols
  );
  if (!nextCell) {
    // cut the executing if there is no cell to be focused
    return;
  }
  [nextRowIndex, nextColIndex] = nextCell;
  break;
```

And repeat this for every `return;` statement that stopped the execution of the `handleKeyDown` method before the refactor :). Hope this helps :hugs:. And again its not required that you do this right away :sweat_smile:.

lib/KTable/index.vue
AlexVelezLl: Hey! Just tested it, and this is working so much better! Thank you @BabyElias!!

AlexVelezLl (commented):
Thank you @BabyElias!! This is looking great! Nothing more to say on my side. Thanks!

lib/KTable/index.vue
MisRob: Thank you @AlexVelezLl, much appreciated, we both got stuck here. @BabyElias Alex noticed what's going on, so I assured him it's still valuable to post here since it will help us to unblock it eventually even though in the scope of this PR we may not pursue refactor anymore. With that we can reference his guidance from a new issue as we log them tomorrow together.

MisRob:
Everybody thanks for feedback and wonderful work with resolving it all @BabyElias. Talk tomorrow - we will log follow-ups and then open for final QA in Kolibri. If it shows no issues, will be ready for merge :partying_face:

lib/KTable/index.vue
MisRob: This has been logged to a doc from which we will open follow-ups. Marking as resolved in the scope of this PR.

BabyElias:
Fixed the failing test in KTable, the existing failing test comes from KDateRange.

lib/KTable/index.vue
BabyElias: done

lib/KTable/index.vue
BabyElias: done

lib/KTable/index.vue
BabyElias: Added to Follow-Up issues.

CHANGELOG.md
MisRob, line 1: @BabyElias would you remove any changes to the CHANGELOG.md here?

docs/pages/playground.vue
MisRob, line 1: Let's also cleanup the playground page before merge. You could do it in a separate commit so we can find the samples easily if needed.

lib/KTable/useSorting/__tests__/index.spec.js
```diff
@@ -0,0 +1,112 @@
...
+  SORT_ORDER_ASC,
+  SORT_ORDER_DESC,
+  DATA_TYPE_STRING,
+  DATA_TYPE_NUMERIC,
+  DATA_TYPE_DATE,
+  DATA_TYPE_OTHERS,
+} from '../';
+
+describe('useSorting', () => {
+  let headers, rows, useLocalSorting;
```
MisRob, line 12: I think here we need to update `useLocalSorting` to the updated API

lib/__tests__/KTable.spec.js
```diff
@@ -0,0 +1,155 @@
...
+    const rows = [
+      ['Alice', 25, 'New York'],
+      ['Bob', 30, 'Los Angeles'],
+      ['Charlie', 35, 'San Francisco'],
+    ];
+    const wrapper = mount(KTable, {
+      propsData: {
+        headers,
+        rows,
+        useLocalSorting: true,
```
MisRob: `useLocalSorting` is the obsolete API, so this place will need to be updated

MisRob (approved):
@BabyElias I read through the latest implementation and haven't noticed anything blocking, so I think we're ready :)! Well done.

Just leaving a few cleanup notes before merge.

Also I can see the linter check is failing, would you run `yarn lint-fix` and commit?

lib/KTable/KTableGridItem.vue
jtamiace: @LianaHarris360 and @MisRob thank you for pointing this out. I noticed this as well and wondered if it was intentional. I do think that table headers should be aligned with the alignment of the data, and never be center-aligned since it makes it a more difficult to tell which header is associated with which data.

MisRob: Thank you @jtamiace. That will help, yes. And should we always align to the left (right in RTL) or would you prefer to keep the numeric type on the right?

jtamiace: Let's keep numeric type on the right

MisRob: :+1: Thanks, all clear :) @BabyElias I know we originally discussed doing this in the follow-up, but now wondering if it'd make sense to have it here since it seems to be a simple change with a pretty high impact on the user experience and we have several days before the new release? It also depends on other work you have in progress. So no pressure, rather if it doesn't cause much trouble.

BabyElias: I think doing this change in the current PR itself would be better. No trouble at all :) Working on it!

MisRob: Lovely, thank you

lib/composables/useSorting/index.js
BabyElias: Done

lib/KTable/index.vue
BabyElias: done

lib/__tests__/KTable.spec.js
BabyElias: done

CHANGELOG.md
BabyElias: copy pasted the entire changelog.md file as in Develop branch currently, still it shows some changes in changelog :|

docs/pages/playground.vue
BabyElias: done

BabyElias:
Ran yarn lint-fix, still shows failing lint errors.

lib/KTable/KTableGridItem.vue
```diff
@@ -0,0 +1,97 @@
...
+    computed: {
+      coreOutlineFocus() {
+        return {
+          ':focus': {
+            ...this.$coreOutline,
+            outlineOffset: '-2px',
+          },
+        };
+      },
+      textAlign() {
```
MisRob: As it's supposed to work same for both headers and data, now that we calculate this in `KTable`, could we just pass it down from `KTable` via prop to `KTableGridItem`?

MisRob: Then `textAlign` computed here can be removed

docs/pages/ktable.vue
```diff
@@ -0,0 +1,153 @@
...
+      </p>
+      <!-- eslint-disable -->
+      <DocsShowCode language="html">
+        <KTable
+          :headers="headers"
+          :rows="rows"
+          caption="Non Sortable Table"
+        >
+
+          <template #header="{ header, index }">
```
MisRob: I think here we'd need to update `index` to `colIndex`? Same for other examples.

docs/pages/ktable.vue
```diff
@@ -0,0 +1,153 @@
...
+            <span>{ header.label } (Backend)</span>
+          </template>
+          <template #cell="{ content, rowIndex, colIndex }">
```
MisRob, line 33: I'm a bit confused here whether we want to use `colIndex` or `colindex`. I don't think it matters that much which one, let's just make sure we're referencing the correct thing and naming is same across all files as  JS is case-sensitive.

MisRob:
> Ran yarn lint-fix, still shows failing lint errors.

@BabyElias can you try to run it multiple times locally until it shows no error? I don't know why but I needed to run it twice. It will make some changes to `docs/pages/ktable.vue` that you will then need to commit. In case it doesn't work I can push it.

lib/KTable/index.vue
```diff
@@ -0,0 +1,564 @@
...
+                       { backgroundColor: $themePalette.white } ,
+                       isColumnFocused(index) ? { backgroundColor: $themePalette.grey.v_50 } : {},
+                       { textAlign: getTextAlign(header.dataType) }]"
+              role="columnheader"
+              data-focus="true"
+              :aria-colindex="index + 1"
+              @click="sortable ? handleSort(index) : null"
+              @keydown="handleKeydown($event, -1, index)"
+            >
+              <slot name="header" :header="header" :colindex="index">
```
MisRob: I think here you intended `colIndex`?

Would be nice to double-check on other places as sometimes it's `colindex` and sometimes `colIndex` (now talking only about the context of scoped slot, for other variables it's fine). As JS is case-sensitive, let's make sure we use one name for the same thing - in examples, documentation, and internal implementation

MisRob:
Caught few typos in the latest updates, but other than that, looks good! Thank you :)

CHANGELOG.md
BabyElias: done using `git checkout develop -- CHANGELOG.md`

lib/KTable/KTableGridItem.vue
BabyElias: done

docs/pages/ktable.vue
BabyElias: done

lib/KTable/index.vue
BabyElias: Checked all places, done required changes.

MisRob (approved):
Thank you, @BabyElias, let's release this :) Well done

PR #744, opened 2024-08-21
KCheckbox: Remove local state management, rely on given props

nucleogenesis:
## Description

KCheckbox had local data which are initialized to the given props for `indeterminate` and `checked` and then updated in a watcher. This resulted in that clicking a checkbox would always toggle it's `isCurrentlyChecked` value and setting `isCurrentlyIndeterminate` to false.

The changes here instead have KCheckbox simply use the given prop values and does nothing to manage any internal state around whether it shows `indeterminate` and/or `checked`.

This results in users of KCheckbox needing to be mindful of how they manage the values they pass into KCheckbox's props as the component will now reflect their values at all time.

#### Issue addressed

This permits Studio's clipboard to have the ability to tell KCheckbox what it should be looking like.

## Changelog

  - **Description:** Removes internal state management for checked & indeterminate in KCheckbox.
  - **Products impact:** Updated API
  - **Addresses:** https://github.com/learningequality/studio/issues/4636
  - **Components:** KCheckbox
  - **Breaking:** No
  - **Impacts a11y:** No
  - **Guidance:** If you use KCheckbox, it is your responsibility to handle the `change` event and update whether or not the given `checked` and `indeterminate` props reflect the reality that you expect.

## Steps to test

Click all of the checkboxes and the states of indeterminacy vs checked-ness -- particularly in nested situations.

The most common in Kolibri is content selection. In Studio, this issue was found while debugging a problem with the Clipboard.

KOLIBRI:

- Resource selection during import
- During Quiz creation
- During Lesson creation

STUDIO:

- Clipboard
- Channel editor

## Testing checklist

## Reviewer guidance

## After review

### DIFF
--- a/docs/pages/kcheckbox.vue
+++ b/docs/pages/kcheckbox.vue
@@ -7,11 +7,13 @@
       <DocsShow>
         <KCheckbox
           label="Some label"
-          :checked="true"
+          :checked="checked1"
+          @change="checked1 = !checked1"
         />
         <KCheckbox
           label="Another label"
-          :checked="false"
+          :checked="checked2"
+          @change="checked2 = !checked2"
         />
       </DocsShow>
     </DocsPageSection>
@@ -39,7 +41,9 @@
       <DocsShow>
         <KCheckbox
           label="Topic is selected"
-          :indeterminate="true"
+          :indeterminate="indeterminate3"
+          :checked="checked3"
+          @change="handle3"
         />
       </DocsShow>
       <p>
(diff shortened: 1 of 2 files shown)


AlexVelezLl:
Code changes make sense to me. But I wonder if removing the local state management is a breaking change.

I would suggest to make a different PR to remove the internal state targeting to develop and explore there its implications. And leave this PR with the removal of the `this.isCurrentlyIndeterminate = false;` line when we are toggling the check, which I suspect would be the needed change to let the consumer have total control in the combination of the four possible combinations of checked/unchecked/indeterminate/non-indeterminate states (as we have a watcher for the `checked` and `indeterminate` states.

For example, this change breaks the docs page as it relies on the KCheckbox internal state:

https://github.com/user-attachments/assets/33b05a26-be8e-45cf-9481-9b541399675f

And there could be some other places where we could be relying on the KCheckbox internal state too (also because the `checked` prop isnt even marked as required).

CHANGELOG.md
MisRob, line 1: Leaving a note here that before merging it'd be good to revert CHANGELOG.md changes since the new GH action will now update the CHANGELOG.md after merging this PR by pasting the changelog from the PR description. Otherwise we'd have duplicate entries.

AlexVelezLl (approved):
lgtm!

AlexVelezLl:
Thank you @nucleogenesis!

Issue #699, opened 2024-07-26
KCard: Complete vertical/horizontal layouts with no thumbnail 


MisRob:
## Summary

This is one of the follow-ups to https://github.com/learningequality/kolibri-design-system/pull/625

Check that `KCard` supports vertical/horizontal layouts with no thumbnail well and add some examples to the documentation page.

This is to ensure that it can support cards like

![Screenshot from 2024-07-26 21-30-16](https://github.com/user-attachments/assets/407bc6c0-033b-48b6-a480-159b93ab79a9)

## Guidance

- Try if cards like above can be easily achieved by the current `KCard` or if we need to add some internal code or new interface APIs to be able to do so
- I think there's no need to distinguish between `'vertical'` and `'horizontal'` layout in this case when `'thumbnailDisplay'` is `'none'`. So the expected behavior would be that any time `thumbnailDisplay='none'`,  `KCard`'s output would be such as illustrated on the designs.

## References

[Figma](https://www.figma.com/design/1tQQ0Sw1yDjzO66hyjLVPM/Kolibri-UX-Fixes?node-id=1097-30500)

MisRob:
With @AllanOXDi we agreed to use the same approach for hiding the slot areas when empty as in https://github.com/learningequality/kolibri-design-system/issues/702. That is, by adding `preserverFooter` and `preserveBelowTitle` props.

MisRob:
Note that here we will preserve 24px spaces introduced here https://github.com/learningequality/kolibri-design-system/issues/704 (slightly different from the screenshot)

AllanOXDi:
fix in  #699

PR #752, opened 2024-08-25
fixes card display without a thumbnail

AllanOXDi:
#### Issue addressed
https://github.com/learningequality/kolibri-design-system/issues/699

## Changelog

  - **Description:**  Update`KCard` to complete vertical/horizontal layouts with no thumbnail
  - **Products impact:**  Card updates 
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/pull/752
  - **Components:** `KCard` 
  - **Breaking:** N0
  - **Impacts a11y:** No
  - **Guidance:**

## Testing checklist

## Reviewer guidance

## After review


lib/KCard/index.vue
```diff
@@ -418,7 +422,7 @@
   /************* Layout-specific styles *************/
 
   .vertical-with-large-thumbnail {
-    height: 480px; /* (1) */
+    height: auto; /* (1) */
```
AlexVelezLl, line 429: Seems like this change is breaking the height of BelowTitle in vertical cards with large thumbnails
image

lib/KCard/index.vue
```diff
@@ -407,7 +411,7 @@
   }
 
   .thumbnail-placeholder {
-    position: absolute;
+    position: relative;
```
AlexVelezLl: Why do we need this as relative now? :eyes: Not saying that its wrong, but would like to know the reason

lib/KCard/index.vue
```diff
@@ -13,12 +13,13 @@
       <slot name="title"></slot>
     </template>
     <template #default>
-      <div class="thumbnail">
+      <div :class="isThumbnailPresent ? 'thumbnail' : null">
```
AlexVelezLl: Do we need this div if isThumbnailPresent is false? Given that in that situation, thsi div would be just simply a div without any styles I think we can safely add a v-if="isThumbnailPresent".

docs/pages/playground.vue
```diff
@@ -8,22 +8,678 @@
...
+
+    <ul class="horizontal-cards-smaller">
+      <KCard
+        class="card"
+        :to="{ name: '/' }"
+        :headingLevel="2"
+        layout="horizontal"
+        thumbnailDisplay="none"
+        thumbnailAlign="right"
+        title="This is a sample resource title"
```
AlexVelezLl: In this example I see there is more margin in the top than in the bottom, which causes the title to be slightly misaligned.

docs/pages/playground.vue
```diff
@@ -8,22 +8,678 @@
...
+              :layout12="{ span: 4 }"
+              :layout8="{ span: 2 }"
+            >
+              <div style="float:right">
+                <KIcon icon="infoOutline" class="icon-size" />
+              </div>
+            </KGridItem>
+          </KGrid>
+        </template>
+        <template #belowTitle>
```
AlexVelezLl: It seems that we are missing some padding at the bottom if we dont have footer.

image

AlexVelezLl (commented):
Thank you @AllanOXDi! I left a couple of questions as I saw some padding issues in some examples.

MisRob (changes requested):
@AllanOXDi We can't condition the presence of the thumbnail area by the presence of the thumbnail source and other factors you used to determine whether the area should be displayed. We would loose functionality that shows thumbnail placeholders when a thumbnail source is not available.

The only determining factor is `thumbnailDisplay` prop. Basically the idea is quite simple - if the thumbnail area is configured to be present via this prop, is shows no matter whether the thumbnail source is provided. If the thumbnail source is not provided in this case, the thumbnail area shows as a placeholder.

So let's just:

```
<div
  v-if="thumbnailDisplay !== ThumbnailDisplays.NONE"
  class="thumbnail"
>
```

All other conditions need to be reverted.

This is important because we quite frequently display placeholders in Kolibri. One way to observe this would be on playground - try to see what happens when you don't provide `thumbnailSrc` but still want to display the thumbnail area via `thumbnailDisplay`. Also I'd recommend you try to use `#thumbnailPlaceholder` slot.

Again, visual tests will help us greatly one day :) Until then, with every change we will need to manually preview the card in all related modes.

lib/KCard/index.vue
MisRob: This whole condition will need to be changed https://github.com/learningequality/kolibri-design-system/pull/752#pullrequestreview-2265059042

lib/KCard/index.vue
```diff
@@ -13,12 +13,13 @@
...
         <!-- 
           Render KImg even if thumbnailSrc is not provided since in that case
           KImg takes care of showing the gray placeholder area.
         -->
         <KImg
+          v-if="thumbnailSrc"
```
MisRob: Relatedly to my other comments, let's remove this line (see the comment on line 17-20)

lib/KCard/index.vue
```diff
@@ -34,7 +35,7 @@
           as progressive loading experience.
         -->
         <span
-          v-if="$slots.thumbnailPlaceholder"
+          v-if="$slots.thumbnailPlaceholder && !thumbnailSrc"
```
MisRob, line 38: Here to we'd want to remove ` && !thumbnailSrc`.  Similar reason as other changes, see the comment on lines 34-35.

MisRob: Here too we'd want to remove ` && !thumbnailSrc`.  Similar reason as other changes, see the comment on lines 34-35.

lib/KCard/index.vue
```diff
@@ -350,6 +351,9 @@
         }
         return {};
       },
+      isThumbnailPresent() {
```
MisRob: Will need to be removed in favor of checking `thumbnailDisplay` configuration. https://github.com/learningequality/kolibri-design-system/pull/752#pullrequestreview-2265059042

lib/KCard/index.vue
MisRob: Good catch, however it's fine in the scope of this PR. There's some limitations around heights that I am taking care of in https://github.com/learningequality/kolibri-design-system/pull/748 and this will to be resolved.

lib/KCard/index.vue
MisRob: It needs to stay `absolute`, otherwise placeholder element won't be positoned as intended via other styles right below

docs/pages/playground.vue
MisRob: I took care of this in https://github.com/learningequality/kolibri-design-system/pull/748

docs/pages/playground.vue
MisRob: I took care of this in https://github.com/learningequality/kolibri-design-system/pull/748

CHANGELOG.md
MisRob, line 1: Before merging, would you remove CHANGELOG.md changes from this PR completely and filled in the template in the pull request description instead? The new action will then paste it to CHANGELOG.md from there.

lib/KCard/index.vue
```diff
@@ -13,7 +13,10 @@
       <slot name="title"></slot>
     </template>
     <template #default>
-      <div class="thumbnail">
+      <div
+        v-if="thumbnailDisplay !== 'none'"
```
MisRob: I think here we could `v-if="thumbnailDisplay !== ThumbnailDisplays.NONE"` because we have the constants available in this component.

If you

```
data() {
  return { ThumbnailDisplays }
}
```

then the constants will become available in the template. Putting constants to `data` or `computed` is quite a common pattern in our frontend.

MisRob (approved):
Thank you for the updates @AllanOXDi, all good!

Just two cleanup notes that would be good to push before merging and then feel free to merge.

MisRob (changes requested):
I will put changes request here temporarily so we don't merge accidentally before changelog.md cleanup finished.

MisRob (approved):
Thank you

Issue #693, opened 2024-07-26
Refactor KBreadcrumbs to utilize KListWithOverflow


MisRob:
<img height="20px" src="https://i.imgur.com/c7hUeb5.jpeg">

❌ **This issue is not open for contribution. Visit <a href="https://learningequality.org/contributing-to-our-open-code-base/" target="_blank">Contributing guidelines</a>** to learn about the contributing process and how to find suitable issues.

<img height="20px" src="https://i.imgur.com/c7hUeb5.jpeg">

In times when [`KListWithOverflow` component](https://design-system.learningequality.org/klistwithoverflow/) didn't yet exist, we used the "duplicate markup" technique in the [`KBreadcrumbs` component](https://design-system.learningequality.org/kbreadcrumbs/) to achieve their overflow behavior when breadcrumb items were collapsed to the dropdown menu on the left:

![breadcrumbs](https://github.com/user-attachments/assets/c4f58830-7d2d-4fd3-8054-2f7445972cf6)

The duplicate markup technique is based on keeping a hidden copy of visible breadcrumbs (`breadcrumbs-offscreen`) in the markup so that elements are available for recalculations:

[https://github.com/learningequality/kolibri-design-system/blob/9d2a147de7e81c945ca0eab7ffb2c4719fd2bb06/lib/KBreadcrumbs.vue#L84-L118](https://github.com/learningequality/kolibri-design-system/blob/9d2a147de7e81c945ca0eab7ffb2c4719fd2bb06/lib/KBreadcrumbs.vue#L84-L118)

Now that we have `KListWithOverflow` that is designed to take care of overflowing items to a menu, it would be good to remove this other implementation from `KBreadcrumbs` and refactor them to utilize `KListWithOverflow` instead.

## The Value Add

Easier maintenance, development efficiency.

## Guidance

Note that `KListWithOverflow` will likely need some updates to be able to display the button on the left side, and possibly some other adjustments so that these two components can collaborate smoothly.

## Acceptance criteria

- [ ]  There are no regressions in places where `KBreadcrumbs` and `KListWithOverflow` are used:
    - Live examples on their documentation pages
    - In Kolibri and Studio

sruthin21:
Hey @MisRob I want to work on this Issue Please Assign this to me

AlexVelezLl:
[...]

PR #790, opened 2024-09-28
Refactor KBreadcrumbs to utilize KListWithOverflow

sruthin21:
## Description

Task1 
 1. Added OverFlowDirection Prop for the kListwithoverflow component
 Task2
 1. Updated KBreadcrums to use KListWithOverflow.

#### Issue addressed

Addresses #693.

## Changelog

## Steps to test

1. Step 1
2. Step 2
3. ...

## Testing checklist

## Reviewer guidance

### DIFF
--- a/lib/KBreadcrumbs.vue
+++ b/lib/KBreadcrumbs.vue
@@ -1,378 +1,110 @@
 <template>
-
-  <div
-    v-show="showSingleItem || crumbs.length > 1"
-    :class="{ 'breadcrumbs-collapsed': collapsedCrumbs.length }"
-  >
-    <nav class="breadcrumbs" v-bind="$attrs" :aria-label="$attrs.ariaLabel">
-      <div
-        v-show="collapsedCrumbs.length"
-        class="breadcrumbs-dropdown-wrapper"
+  <KListWithOverflow
+  overflowDirection="start"
+  :items="breadcrumbs"
+>
+  <template #item="{ item }">
+    <li>
+      <KRouterLink
+        v-if="item.link"
+        :text="item.text"
+        :to="item.link"
       >
-        <KIconButton
-          size="small"
-          :icon="showDropdown ? 'chevronUp' : 'chevronDown'"
-          appearance="raised-button"
-          @click="showDropdown = !showDropdown"
+        <template #text="{ text }">
+          <span >{{ text }}</span>
+        </template>
+      </KRouterLink>
+      <span v-else>{{ item.text }}</span>
+    </li>
+  </template>
+  <template #more="{ overflowItems }">
+    <KIconButton
+      size="small"
+      icon="chevronUp"
+      appearance="raised-button"
+    >
+      <template #menu>
+        <KDropdownMenu
+          :options="overflowItems"
         />
-        <div
-          v-if="showDropdown"
-          class="breadcrumbs-dropdown"
-          :style="{ background: $themeTokens.surface }"
-        >
-          <ol class="breadcrumbs-dropdown-items">
-            <li
-              v-for="(crumb, index) in collapsedCrumbs"
-              :key="index"
-              class="breadcrumbs-dropdown-item"
-            >
-              <KRouterLink
-                v-if="crumb.link"
-                class="krouter-item"
-                :text="crumb.text"
-                :to="crumb.link"
-              >
-                <template #text="{ text }">
-                  <span class="breadcrumbs-crumb-text" dir="auto">{{ text }}</span>
-                </template>
-              </KRouterLink>
-              <span v-else dir="auto">{{ crumb.text }}</span>
-            </li>
-          </ol>
-        </div>
-      </div>
-
-      <ol class="breadcrumbs-visible-items">
-        <template v-for="(crumb, index) in crumbs">
-          <li
-            v-if="index !== crumbs.length - 1"
-            v-show="!crumb.collapsed"
-            :key="index"
-            class="breadcrumbs-visible-item breadcrumbs-visible-item-notlast"
-          >
-            <KRouterLink
-              v-if="crumb.link"
-              :text="crumb.text"
-              :to="crumb.link"
-              dir="auto"
-            >
-              <template #text="{ text }">
-                <span class="breadcrumbs-crumb-text">{{ text }}</span>
-              </template>
-            </KRouterLink>
-            <span v-else>{{ crumb.text }}</span>
-          </li>
-
-          <li
-            v-else
-            :key="index"
-            class="breadcrumb-visible-item-last breadcrumbs-visible-item"
-          >
-            <span
-              class="breadcrumbs-crumb-text"
-              :style="{ maxWidth: lastBreadcrumbMaxWidth }"
-              dir="auto"
-            >
-              {{ crumb.text }}
-            </span>
-          </li>
-        </template>
-      </ol>
-    </nav>
-
-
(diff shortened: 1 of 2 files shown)


lib/KBreadcrumbs.vue
```diff
@@ -1,378 +1,110 @@
...
-      showSingleItem: {
-        type: Boolean,
-        default: false,
-      },
+import KListWithOverflow from './KListWithOverflow.vue';
+import KDropdownMenu from './KDropdownMenu.vue';
+export default {
+  name: 'KBreadcrumbs',
+  props: {
+    breadcrumbs: {
```
AlexVelezLl, line 45: Here we should not change the public API as this would be a breaking change, we should keep the same API as before, or at least that if we make any change to the API, that this is not a breaking change. So we should keep the `items` and `showSingleItem` props, and they should keep working exactly the same as before the refactor.

lib/KBreadcrumbs.vue
```diff
@@ -1,378 +1,110 @@
...
-      },
-      /**
-       * By default, the breadcrums will be hidden when the length of items is 1.
-       * When set to `true`, a breadcrumb will be shown even when there is only one.
-       */
-      showSingleItem: {
-        type: Boolean,
-        default: false,
-      },
+import KListWithOverflow from './KListWithOverflow.vue';
```
AlexVelezLl, line 40: Here we dont need to import there components, as they are already registered in the Vue context, we can use it without it.

lib/KListWithOverflow.vue
```diff
@@ -195,11 +207,17 @@
...
 
-        maxWidth = Math.ceil(maxWidth);
-        this.overflowItems = overflowItemsIdx.map(idx => this.items[idx]);
+        this.overflowItems = overflowItemsIdx.map((idx) => this.items[idx]);
         this.isMoreButtonVisible = overflowItemsIdx.length > 0;
-        list.style.maxWidth = `${maxWidth}px`;
+        list.style.maxWidth = `${Math.ceil(maxWidth)}px`;
         list.style.maxHeight = `${maxHeight}px`;
+
+        this.setVisibleItems(directionIndexes, overflowItemsIdx);
```
AlexVelezLl, line 215: We are not doing anything with this `visibleItems` array. I would suggest to remove them. Also I dont note any added value of removing the `maxWidth = Math.ceil(maxWidth);` sentence. I think we can keep these lines as before:
```javascript
        maxWidth = Math.ceil(maxWidth);
        this.overflowItems = overflowItemsIdx.map(idx => this.items[idx]);
        this.isMoreButtonVisible = overflowItemsIdx.length > 0;
        list.style.maxWidth = `${maxWidth}px`;
        list.style.maxHeight = `${maxHeight}px`;
      },
```

lib/KListWithOverflow.vue
```diff
@@ -152,29 +161,32 @@
           itemsSizes.push(itemSize);
         }
 
+        if (this.overflowDirection === 'start') {
+              itemsSizes.reverse();
```
AlexVelezLl, line 165: We dont need to reverse the itemSizes because we are already reverting the `directionIndexes`, and when we iterate `directionIndexes`, we are using the same index to access the list.children which is not reverted and the itemSizes which is reverted. So we are not matching the same information for the children and its size.

lib/KListWithOverflow.vue
```diff
@@ -152,29 +161,32 @@
...
-          // If the item dont fit in the available space or if we have already
-          // overflowed items, we hide it. This means that once one item overflows,
-          // all the following items will be hidden.
+        directionIndexes.forEach((i) => {
+          const itemWidth = itemsSizes[i].width;
           if (itemWidth >= availableWidth || overflowItemsIdx.length > 0) {
             overflowItemsIdx.push(i);
-            item.style.visibility = 'hidden';
-            item.style.position = 'absolute';
+            list.children[i].style.visibility = 'hidden';
```
AlexVelezLl, line 177: I think its better to revert this change and have the `item` object to avoid repeating the `list.children[i]`

lib/KListWithOverflow.vue
```diff
@@ -152,29 +161,32 @@
...
+        const indexSequence = [...Array(list.children.length).keys()];
+        const directionIndexes =
+          this.overflowDirection === 'start' ? indexSequence.reverse() : indexSequence;
+
         const overflowItemsIdx = [];
-        for (let i = 0; i < list.children.length; i++) {
-          const item = list.children[i];
-          const itemWidth = itemsSizes[i].width;
 
-          // If the item dont fit in the available space or if we have already
```
AlexVelezLl, line 160: Lets keep these comments, as they are useful for new team members that doesnt know the logic here

lib/KListWithOverflow.vue
```diff
@@ -195,11 +207,17 @@
           maxWidth -= removedDividerWidth;
```
AlexVelezLl, line 207: We need to also check the logic of the fix dividers Visibility, because for example here, when we have dividers, we can have this situation where some overflowed items are visible:

https://github.com/user-attachments/assets/ac572531-7806-4462-a841-45a08f04bc37

lib/KListWithOverflow.vue
```diff
@@ -276,5 +294,8 @@
   .more-button-wrapper {
     visibility: hidden;
   }
+  .more-button-wrapper.start-button {
```
AlexVelezLl, line 297: We cant have this solution for the position of the button, because Keyboard navigation would not be the same. We need to keep the current behaviour in KBreadcrumbs that if the more button appears at the beginning, the keyboard navigation should focus it first, before the rest items.

https://github.com/user-attachments/assets/dbf2abb9-4a9e-471b-be0f-eb2081e959f3

lib/KBreadcrumbs.vue
```diff
@@ -1,378 +1,110 @@
...
+  margin-top: 8px;
+  margin-bottom: 8px;
+  font-size: 16px;
+  font-weight: bold;
+  line-height: 32px;
+  white-space: nowrap;
+}
+
+.link {
+  color: blue;
```
AlexVelezLl, line 82: Here we shouldnt add color styles like these. We cant set burned colors, instead we need to use our brand colors, you can find the guides [here](https://design-system.learningequality.org/colors/).

Although, I think most of these styles arent really needed. Do they?

AlexVelezLl (changes requested):
Hey @sruthin21! Thanks for your hard work. This is looking great! However my major concern is that we should make this change a non-breaking change, and avoid modifying the public API of the KBreadcrumbs and keep the external functionality the same as before.

One method to ensure that we are not breaking anything is to constantly check the KBreadcrumbs page http://localhost:4000/kbreadcrumbs and see that it looks and behave exactly the same as our current version of kbreadcrumbs which you can see in the production docs site https://design-system.learningequality.org/kbreadcrumbs/.

As you can see now we dont show the breadcrumbs in the docs page:

| Before | After |
| --- | --- |
| image | image |

AlexVelezLl:
Also, a side note! When we open a PR we should keep and fill the form template that appears when we create the PR, not remove it, I have pasted it again to the PR description :hugs:.

AlexVelezLl:
Hey @sruthin21! Are you still interested in continuing working in this issue?

MisRob:
Hi @sruthin21, by deleting your fork, unfortunately we won't be able to finish this PR. I will unassign you from https://github.com/learningequality/kolibri-design-system/issues/693, but feel free to message us and re-open if you'd like.

Issue #775, opened 2024-09-10
Update KDS Theme Token values


marcellamaki:
When upgrading our themes for the rebrand, the new numerical token values for colors did not correspond exactly to the material specifications. This is fine in main Kolibri, but presents a challenge for upgrading custom themed Kolibri.

While we will need to make these changes in some places in Kolibri as well, the first thing will be to make the changes in KDS.

The updated mapping: 
![Screenshot 2024-09-06 at 12 14 29 PM](https://github.com/user-attachments/assets/6554cc04-50f1-4517-97b1-f5df682b65e7)

AlexVelezLl:
Closed by #782

PR #796, opened 2024-10-04
KCardGrid part 3: Add advanced grid configuration and loading skeletons

MisRob:
## Description

Adds new features to `KCardGrid`:
- Option to override base layouts partially or completely via the new prop `layoutOverride`
- Loading skeletons and a way to configure them via the new prop `skeletonsConfig`
  - New development-only `debug` prop to help with skeletons configuration

These APIs were previously discussed and agreed on during the beta stage.

See commit messages for details.

## Changelog

  - **Description:** Adds an option to override `KCardGrid` base layouts partially or completely via the new prop `layoutOverride`
  - **Addresses:** Allows advanced grids customization
  - **Components:** `KCardGrid`
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -

  - **Description:** Adds loading skeletons to `KCardGrid` and a way to configure them via the new prop `skeletonsConfig`
  - **Addresses:** Ensures smooth loading experience
  - **Components:** `KCardGrid`
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -

## Steps to test

- Read new documentation sections for the two new features and preview live examples
  -  https://deploy-preview-796--kolibri-design-system.netlify.app/kcardgrid/#layout-customization
  - https://deploy-preview-796--kolibri-design-system.netlify.app/kcardgrid/#loading-state

## Testing checklist

- [ ] Critical and brittle code paths are covered by unit tests - will do as follow-up

## Reviewer guidance

## Comments

I got stuck on some issues with mocking composables in KDS that prevented me from completing unit tests. This is being discussed and I plan to follow-up later, together with some other maintenance tasks I am going to do.


MisRob:
For code reviewers, I don't think you need to review the large diff on documentation pages - rather just preview the generated documentation. One of the maintenance tasks I am planning to do in a follow-up PR is to break doc pages and examples to sub-components internally so it's easier to updated in the future.

MisRob:
Linking feedback from designers - [Slack thread](https://learningequality.slack.com/archives/CTG8XDTCL/p1728307006686509?thread_ts=1728305999.363109&cid=CTG8XDTCL)

lib/cards/KCardGrid.vue
```diff
@@ -16,28 +55,40 @@
...
+        skeletonThumbnailDisplay,
+        skeletonThumbnailAlign,
+      } = useGridLoading(props);
 
       const gridStyle = ref({});
       const gridItemStyle = ref({});
 
       watch(
-        currentLevelConfig,
+        currentBreakpointConfig,
```
rtibbles, line 77: Just out of curiosity - what is the reason we're doing a watch and assign here, rather than a computed prop? If there's a specific reason, might be good to have a code comment accordingly, like we do in other places where we use a similar pattern.

lib/cards/KCardGrid.vue
```diff
@@ -77,6 +136,47 @@
...
+      loading: {
+        type: Boolean,
+        default: false,
+      },
+      // eslint-enable-next-line kolibri/vue-no-unused-properties
+      /**
+       * Configures loading skeletons
+       */
+      // eslint-disable-next-line kolibri/vue-no-unused-properties
+      skeletonsConfig: {
```
rtibbles, line 155: Hopefully upgrading to newer versions of eslint eventually will prevent us having to do this so much?

lib/cards/KCardGrid.vue
```diff
@@ -77,6 +136,47 @@
           return [LAYOUT_1_1_1, LAYOUT_1_2_2, LAYOUT_1_2_3].includes(value);
         },
       },
+      // eslint-enable-next-line kolibri/vue-no-unused-properties
+      /**
+       * Overrides the base grid `layout` for chosen breakpoints levels
+       */
+      // eslint-disable-next-line kolibri/vue-no-unused-properties
+      layoutOverride: {
```
rtibbles, line 135: I am not finding the rationale for doing this as an array rather than an object keyed by breakpoint, as it was in the [commit linked](https://github.com/learningequality/kolibri-design-system/commit/61e95ea084127811a58d880f7ee1a4e0f131c113) in the beta proposal.

My only real concern is that it felt hard to reason about what the resulting layout would be, because of the ability to mix in multiple layouts. Having the object specced form is more verbose, but I think easier to understand exactly what is happening.

Have I missed something else in the motivation here?

rtibbles (commented):
Main question is just about the form of the layout overrides.

lib/cards/KCardGrid.vue
MisRob: Thank you, ah yes - I admit I forgot the reason for this particular place. I think it was related to general problem I had - I couldn't make the `computed` approaches work in the accompanying composables. I was able to track it to `useKResponsiveWindow`, but am still not clear what exactly was going on. Was planning to debug and discuss it later together with the unit testing issue, would that be okay?

lib/cards/KCardGrid.vue
MisRob: That'd be nice

lib/cards/KCardGrid.vue
MisRob: Yes, that's a good note, thank you. I should have mentioned it - it didn't come up until I dived deeper with loading skeletons work.

The main reason was consistency of `layoutOverride` with the `skeletonsConfig`. I think it'd make sense to use the same format so that it's not confusing.

I considered using the more verbose object version for `skeletonsConfig` at first for its explicitness, but in comparison to the `layoutOverride` that won't be used that much, `skeletonsConfig` is required for every grid and it should have defined every breakpoint. When I was implementing examples on the documentation page, the verbosity of the original format started coming up quite frequently in ways that felt overwhelming. Compare:

**Version 1**

```
skeletonsConfig: [
    {
      breakpoints: [0, 1, 2, 3, 4, 5, 6, 7],
      count: 3,
      height: '400px',
      orientation: 'vertical',
      thumbnailDisplay: 'large',
      thumbnailAlign: 'left'
    },
    {
      breakpoints: [4, 5, 6, 7],
      height: '220px',
      orientation: 'horizontal'
    }
  ],
};
```

**Version 2**

```
skeletonsConfig: {
  0: {
     count: 3,
     height: '400px',
     orientation: 'vertical',
     thumbnailDisplay: 'large',
  },
  1: {
     count: 3,
     height: '400px',
     orientation: 'vertical',
     thumbnailDisplay: 'large',
  },
  2: {
     count: 3,
     height: '400px',
     orientation: 'vertical',
     thumbnailDisplay: 'large',
  },
  3: {
     count: 3,
     height: '400px',
     orientation: 'vertical',
     thumbnailDisplay: 'large',
  },
  4: {
     count: 3,
     height: '220px',
     orientation: 'horizontal'
     thumbnailDisplay: 'large',
     thumbnailAlign: 'left'
  },
  5: {
     count: 3,
     height: '220px',
     orientation: 'horizontal'
     thumbnailDisplay: 'large',
     thumbnailAlign: 'left'
  6: {
     count: 3,
     height: '220px',
     orientation: 'horizontal'
     thumbnailDisplay: 'large',
     thumbnailAlign: 'left'
  },
  7: {
     count: 3,
     height: '220px',
     orientation: 'horizontal'
     thumbnailDisplay: 'large',
     thumbnailAlign: 'left'
  },
}
```

Of course, this could be mitigated by saving some configs to variables on product side.

> My only real concern is that it felt hard to reason about what the resulting layout would be, because of the ability to mix in multiple layouts.

Yes, that's right, theoretically you can define a breakpoint multiple-times. I actually even recommended this in **Show more** section of https://deploy-preview-796--kolibri-design-system.netlify.app/kcardgrid/#loading-state, thinking that it will be intuitive. I realize that may be just an assumption based on being used to CSS where the bottom-up approach is quite common (e.g. mobile-first approach, where you defined base styles on all breakpoints and then go up and override where needed). Based on your comment, I think it's not as clear as intended, so we'd better discuss.

In terms of next steps, do you have a strong preference for either or perhaps alternative API ideas that would align all hese perspectives better? It'd be also fine to raise this for discussion with more developers (likely when I'm back from vacation though).

lib/cards/KCardGrid.vue
MisRob: Opened issue https://github.com/learningequality/kolibri-design-system/issues/801

lib/cards/KCardGrid.vue
MisRob: For now, I at least attempted to improve documentation by making more clear the bottom-up strategy

MisRob: For now, I at least attempted to improve documentation by making more clear the bottom-up strategy is intended

MisRob:
**Latest updates**

- Resolves feedback from designers regarding the loading state timing, loading skeletons count, and border radius fix ([Slack thread](https://learningequality.slack.com/archives/CTG8XDTCL/p1728305999363109)).
- Updates [the "Loading state" documentation section](https://deploy-preview-796--kolibri-design-system.netlify.app/kcardgrid/#loading-state) accordingly and adds ways to preview the loading state timing optimizations
- Simplifies examples and documentation of the `skeletonsConfig`
- Opened follow-ups
  - https://github.com/learningequality/kolibri-design-system/issues/801
  - https://github.com/learningequality/kolibri-design-system/issues/802

I think this is all I could complete here at this point. We still need to decide on https://github.com/learningequality/kolibri-design-system/pull/796#discussion_r1795864784 - I will leave to @rtibbles's consideration whether it's blocking or if it can be addressed as a follow-up if we figure out some changes are needed.

rtibbles:
I don't think my comments here are blocking - I think by using the grid and skeletons in practice, we may discover that there are some sensible defaults we end up using, and the more advanced props format doesn't matter as much as it is rarely used - or we come up with a brainwave of a better way.

lib/cards/SkeletonCard.vue
```diff
@@ -0,0 +1,134 @@
...
+    class="skeleton-card"
+    :style="{ height: height }"
+    :orientation="orientation"
+    :thumbnailDisplay="thumbnailDisplay"
+    :thumbnailAlign="thumbnailAlign"
+  >
+    <template #title>
+      <span
+        class="title-placeholder"
+        :style="{ backgroundColor: $themePalette.grey.v_50 }"
```
AlexVelezLl: I just merged #775 into develop, so these two uses of theme palette will need to be updated to `v_100` :).

MisRob: Thank you, @AlexVelezLl, done.

MisRob:
All is ready from my side @rtibbles

rtibbles (approved):
I say we merge this, get it into an RC and take the API for a few practice laps!

Issue #793, opened 2024-10-03
KIconButton: Set `tooltip` value as fallback if `ariaLabel` is missing


AlexVelezLl:
🌱 Are you new to the codebase? Welcome! Please see [the contributing guidelines](https://github.com/learningequality/kolibri-design-system/blob/develop/CONTRIBUTING.md).

## Product
Kolibri, Studio, etc.

## Description

In `KIconButton` we have these two props related but not exactly the same: `ariaLabel` and `tooltip`.  For accessibility purposes we dont want these props to be the same, but if the user forgot to send an `ariaLabel` prop but they sent a `tooltip` prop, the ariaLabel should take the value of the `tooltip` prop.

This means, the aria label of the icon button will be set by the `ariaLabel` prop, and have the `tooltip` value as fallback if `ariaLabel` is missing.

## The Value Add
This will improve accessibility in places where we have tooltips but no aria labels. And in the future we can just send the `tooltip` prop if we want the `ariaLabel` to have the same value.

shivam-daksh:
@AlexVelezLl , Could you assign this to me?

MisRob:
Hi @shivam-daksh, thanks again. I will assign you this one alongside https://github.com/learningequality/kolibri-design-system/issues/795 so that you can switch tasks. Regarding other issues you volunteered for, let's leave it for later when these two are resolved.

MisRob:
Closed by https://github.com/learningequality/kolibri-design-system/pull/798

PR #798, opened 2024-10-08
Improve accessibility by using tooltip as fallback for ariaLabel in KIconButton

shivam-daksh:
## Description

This PR addresses an accessibility issue in the `KIconButton` component. Previously, the `ariaLabel` and `tooltip` props were treated independently, and it was possible for a developer to omit the `ariaLabel` prop while setting a `tooltip`. This could lead to accessibility issues, as `ariaLabel` is essential for screen readers, but the `tooltip` is not.

#### Issue addressed

#793 

Addresses 
It is resolved by the adding a new custom computed property `computedAriaLabel` which will dynamically sets the `ariaLabel` and `tooltip` props.

## Changelog

  - **Description:**  Adds custom computed property `computedAriaLabel` that dynamically sets the `ariaLabel` for `KIconButton` based on the `tooltip` prop when `ariaLabel` is not provided.
  - **Products impact:** none 
  - **Addresses:** #793 
  - **Components:** `KIconButton`
  - **Breaking:** no
  - **Impacts a11y:** Improves accessibility in places where we have tooltips but no aria- label attribute set for `KIconbutton`.
  - **Guidance:** -

## Testing checklist

## Reviewer guidance

### DIFF
--- a/lib/buttons-and-links/KIconButton.vue
+++ b/lib/buttons-and-links/KIconButton.vue
@@ -10,6 +10,7 @@
     text=""
     v-on="$listeners"
   >
+    <!-- if no "position" is passed as a prop, defaults to bottom, as previously -->
     <UiTooltip
       v-if="tooltip"
       :zIndex="24"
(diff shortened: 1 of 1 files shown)


LianaHarris360 (approved):
The code changes are clean and correctly sets the `aria-label` attribute using the `tooltip` property as needed. I updated the ChangeLog section so that the Github action could set it as well. Thank you @shivam-daksh

PR #818, opened 2024-11-07
Issues with keyboard navigation on KDropdownMenu

### LINKED ISSUE #588: Issues with keyboard navigation on KDropdownMenu

Sahil-Sinha-11:
…p down position

#### Issue addressed

Addresses #588

### After Video
[](https://github.com/user-attachments/assets/549e99bb-8a8c-414f-9dd7-414fde337327)

## Changelog

  - **Description:** keyboard navigation on KDropdownMenu 
  - **Addresses:**Issues with keyboard navigation on KDropdownMenu #588.
  - **Components:** kDropdownMenu.vue
  - **Breaking:** -
  - **Impacts a11y:** - 
  - **Guidance:** -

## Steps to test

1. Step 1
2. Step 2
3. ...

## Testing checklist

## Reviewer guidance

### DIFF
--- a/lib/KDropdownMenu.vue
+++ b/lib/KDropdownMenu.vue
@@ -232,11 +232,6 @@
         this.$emit('close');
         this.$refs.popover.close();
       },
-      focusOnButton() {
-        if (this.$refs.button) {
-          this.$refs.button.$el.focus();
-        }
-      },
     },
   };
 
--- a/lib/KDropdownMenu.vue
+++ b/lib/KDropdownMenu.vue
@@ -179,9 +179,7 @@
         // let prevSibling = focusedElement.previousElementSibling;
 
         //function to check if the sibling is divider
-        const isDivider = element => {
-          return element && element.classList.contains('is-divider');
-        };
+        const isFocusable = element => element && element.tabIndex >= 0;
 
         // manage rotating through the options using arrow keys
         // UP arrow: .keyCode is depricated and should used only as a fallback
@@ -190,7 +188,7 @@
 
           // Checking if previous sibling is divider and if yes then skip it
           let prevSibling = focusedElement.previousElementSibling;
-          while (prevSibling && isDivider(prevSibling)) {
+          while (prevSibling && !isFocusable(prevSibling)) {
             prevSibling = prevSibling.previousElementSibling;
           }
           prevSibling
@@ -202,7 +200,7 @@
 
           //Chekcing if next sibling is divider and skipping it
           let sibling = focusedElement.nextElementSibling;
-          while (sibling && isDivider(sibling)) {
+          while (sibling && !isFocusable(sibling)) {
             sibling = sibling.nextElementSibling;
           }
           sibling ? this.$nextTick(() => sibling.focus()) : this.$nextTick(() => this.setFocus());
--- a/lib/KDropdownMenu.vue
+++ b/lib/KDropdownMenu.vue
@@ -32,7 +32,6 @@
 <script>
 
   import { computed } from '@vue/composition-api';
-  // import { get } from 'lodash';
   import UiMenu from './keen/UiMenu';
   import UiPopover from './keen/UiPopover';
   import useKContextMenu from './composables/_useKContextMenu';
@@ -147,9 +146,9 @@
     },
     methods: {
       handleOpen() {
+        this.lastFocusElement = document.activeElement;
         this.$nextTick(() => this.$nextTick(() => this.setFocus()));
         window.addEventListener('keydown', this.handleOpenMenuNavigation, true);
-        this.lastFocusElement = document.activeElement;
       },
       setFocus() {
         this.$refs.menu.$el.querySelector('li').focus();
@@ -163,11 +162,9 @@
             focusedElement.classList.contains('ui-popover-focus-redirector') ||
             focusedElement.classList.contains('ui-menu-option'))
         ) {
-          this.focusOnButton();
-        }
-
-        if (this.lastFocusElement) {
-          this.lastFocusElement.focus();
+          if (this.lastFocusElement) {
+            this.lastFocusElement.focus();
+          }
         }
 
         window.removeEventListener('keyup', this.handleKeyUp, true);
@@ -201,11 +198,6 @@
         const popoverIsOpen = popover.clientWidth > 0 && popover.clientHeight > 0;
         // set current element and its siblings
         let focusedElement = document.activeElement;
-        // let sibling = focusedElement.nextElementSibling;
-        // let prevSibling = focusedElement.previousElementSibling;
-
-        //function to check if the sibling is divider
-        // const isFocusable = element => element && element.tabIndex >= 0;
 
         // manage rotating through the options using arrow keys
         // UP arrow: .keyCode is depricated and should used only as a fallback
@@ -213,7 +205,7 @@
           event.preventDefault();
 
           // Checking if previous sibling is divider and if yes then skip it
-          let prevSibling = this.getPreviousFocusableSibling(focusedElement);
+          const prevSibling = this.getPreviousFocusableSibling(focusedElement);
           prevSibling
             ? this.$nextTick(() => prevSibling.focus())
             : this.$nextTick(() => lastChild.focus());
@@ -221,7 +213,7 @@
         } else if ((event.key == 'ArrowDown' || event.keyCode == 40) && popoverIsOpen) {
           event.preventDefault();
           //Chekcing if next sibling is divider and skipping it
-          let sibling = this.getNextFocusableSibling(focusedElement);
+          const sibling = this.getNextFocusableSibling(focusedElement);
           sibling ? this.$nextTick(() => sibling.focus()) : this.$nextTick(() => this.setFocus());
           // if a TAB key, not an arrow key, close the popover and advance to next item in the tab index
         } else if ((event.key == 'Tab' || event.keyCode == 9) && popoverIsOpen) {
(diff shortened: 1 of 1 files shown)


lib/KDropdownMenu.vue
```diff
@@ -175,19 +175,36 @@
...
         // set current element and its siblings
         let focusedElement = document.activeElement;
-        let sibling = focusedElement.nextElementSibling;
-        let prevSibling = focusedElement.previousElementSibling;
+        // let sibling = focusedElement.nextElementSibling;
+        // let prevSibling = focusedElement.previousElementSibling;
+
+        //function to check if the sibling is divider
+        const isDivider = element => {
+          return element && element.classList.contains('is-divider');
```
AlexVelezLl: This is a good way to determine if the element is a divider, but is a little specific. If anything changes and in the future we get rid of this specific class. This would break this condition.

A more general solution could be to ask if the element has a tabIndex >= 0, so this means it is focusable and with this, we can skip the dividers.

lib/KDropdownMenu.vue
```diff
@@ -175,19 +175,36 @@
...
           prevSibling
             ? this.$nextTick(() => prevSibling.focus())
             : this.$nextTick(() => lastChild.focus());
           // DOWN arrow
         } else if ((event.key == 'ArrowDown' || event.keyCode == 40) && popoverIsOpen) {
           event.preventDefault();
+
+          //Chekcing if next sibling is divider and skipping it
+          let sibling = focusedElement.nextElementSibling;
+          while (sibling && isDivider(sibling)) {
```
AlexVelezLl: We could write two separate methods here, to breakdown a little bit this method. so we can just do something like

```javascript
const sibling = this.getNextFocusableSibling(focusedElement);
const prevSibling = this.getPrevFocusableSibling(focusedElement);
```

lib/KDropdownMenu.vue
Sahil-Sinha-11: Oh right, that seems more logical to skip any other element [like divider] in a more generalized way. Thanks @AlexVelezLl .

lib/KDropdownMenu.vue
Sahil-Sinha-11: Hey @AlexVelezLl , I've defined prevSibling Already here
Screenshot 2024-11-11 at 12 21 48 AM
and the variable sibling Is for next focused element.

Sahil-Sinha-11:
Hey @Sahil-Sinha-11,  I made the suggested changes. 🤗 
PS: sorry for the late reply, actually my university exams are next week so quite loaded with tests. 😅

Sahil-Sinha-11:
Hey @AlexVelezLl , the focus works like this, should I return it to the last focused button?

https://github.com/user-attachments/assets/4c1a3ca4-6bd5-444a-929d-a421ef216488

lib/KDropdownMenu.vue
AlexVelezLl: Yes, my suggestion was more to avoid writing more logic to query the next/previouss element in the same `handleOpenMenuNavigation` method. So lets define two new methods in this component `getNextFocusableSibling` and `getPrevFocusableSibling` where we should do the focusable check and the while sentence.

e.g.
```javascript
methods: {
  ...
  getNextFocusableSibling(element) {
    let sibling = focusedElement.nextElementSibling;
    while (sibling && !isFocusable(sibling)) {
      sibling = sibling.nextElementSibling;
    }
    return sibling;
  }
}
```
and lets refactor the definition of these variables `prevSibling` and `sibling` to use the result of these methods.

AlexVelezLl:
Hey @Sahil-Sinha-11. Yes!! It should be working like that, when we close the dropdown, the last focused element is the element that should be focused. Thank you!! Loking forward to see this change pushed in the PR :100:.

> PS: sorry for the late reply, actually my university exams are next week so quite loaded with tests. 😅

No worriees! :sweat_smile: There is no rush, please take your time :hugs:.

Sahil-Sinha-11:
Hey @AlexVelezLl, I made the changes, please review it, thanks.

lib/KDropdownMenu.vue
```diff
@@ -32,6 +32,7 @@
 <script>
 
   import { computed } from '@vue/composition-api';
+  // import { get } from 'lodash';
```
AlexVelezLl: We can get rid of this line

lib/KDropdownMenu.vue
```diff
@@ -147,6 +149,7 @@
       handleOpen() {
         this.$nextTick(() => this.$nextTick(() => this.setFocus()));
         window.addEventListener('keydown', this.handleOpenMenuNavigation, true);
+        this.lastFocusElement = document.activeElement;
```
AlexVelezLl: Although there is no problem with having this line as the last line of the `handleOpen` method, we can move it to be the first line of the method, so the intention of storing the last focused element before we focus anything else is clearer :).

lib/KDropdownMenu.vue
```diff
@@ -162,8 +165,31 @@
         ) {
           this.focusOnButton();
         }
+
+        if (this.lastFocusElement) {
```
AlexVelezLl: Above this, you can see there is a call to the `this.focusOnButton()` method, which apparently was an old behaviour we had in KDropdownMenu, but its no longer useful as now KDropdownMenu doesnt render any button anymore.

But if in the past, we were making this call in those conditions, it was for some reason (perhaps because of an edge case im not aware of), but to play safe, lets just replace that `focusOnButton()` call with this lastFocusElement call, for this focus to occur on those conditions too, so we would have something like:
```javascript
        if (
          popover.contains(focusedElement) &&
          (focusedElement.classList.contains('ui-popover') ||
            focusedElement.classList.contains('ui-popover-focus-redirector') ||
            focusedElement.classList.contains('ui-menu-option'))
        ) {
          if (this.lastFocusElement) {
            this.lastFocusElement.focus();
          }
        }
```
And then we can remove the `focusOnButton` method :).

lib/KDropdownMenu.vue
```diff
@@ -175,19 +201,27 @@
         const popoverIsOpen = popover.clientWidth > 0 && popover.clientHeight > 0;
         // set current element and its siblings
         let focusedElement = document.activeElement;
-        let sibling = focusedElement.nextElementSibling;
-        let prevSibling = focusedElement.previousElementSibling;
+        // let sibling = focusedElement.nextElementSibling;
```
AlexVelezLl: Great catch! We dont need to make these calls always. We can remove these lines too.

lib/KDropdownMenu.vue
```diff
@@ -175,19 +201,27 @@
...
+        //function to check if the sibling is divider
+        // const isFocusable = element => element && element.tabIndex >= 0;
 
         // manage rotating through the options using arrow keys
         // UP arrow: .keyCode is depricated and should used only as a fallback
         if ((event.key == 'ArrowUp' || event.keyCode == 38) && popoverIsOpen) {
           event.preventDefault();
+
+          // Checking if previous sibling is divider and if yes then skip it
+          let prevSibling = this.getPreviousFocusableSibling(focusedElement);
```
AlexVelezLl: lets define these `prevSibling` and `sibling` as const, as we wont need to reassign them anymore.

AlexVelezLl (changes requested):
Woo :tada:. This is working soo smothly @Sahil-Sinha-11. Thank you!! I have left a couple of comments just for some code clean up, thanks!

Sahil-Sinha-11:
Hey @AlexVelezLl, I made the suggested changes, please review them, thanks.

AlexVelezLl:
Thank youu @Sahil-Sinha-11!! Code changes look good to me! :) . I have just opened a [PR in Kolibri](https://github.com/learningequality/kolibri/pull/12838) with your changes for QA to test this in Kolibri and confirming that this is the expected behaviour, after that we can merge your PR. Thanks @Sahil-Sinha-11! Good job :hugs:

AlexVelezLl (approved):
Thanks @Sahil-Sinha-11!

MisRob:
Thanks @Sahil-Sinha-11 and @AlexVelezLl, good chunk of work here :)!

PR #804, opened 2024-10-20
Refactor handleKeydown Method for Improved Readability and Maintainability

shivam-daksh:
## Description

This PR addresses the issue described in [Issue #795](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.esm.html). The `handleKeydown` method in the `KTable` component has been refactored to improve readability and maintainability while preserving the existing functionality and accessibility of the keyboard navigation.
#### Issue addressed

#795

Addresses #*PR# HERE*

### Summary

In [#727](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.esm.html), we introduced a new component—KTable—which provides a flexible, accessible table layout with enhanced keyboard navigation capabilities. As part of this, the handleKeydown method is responsible for managing keyboard navigation between table cells, handling focus, and managing the arrow, tab, and other keys.

The current implementation of the handleKeydown method has grown more complex over time, making it harder to maintain and reason about. The goal of this issue is to refactor the method into a more readable and maintainable format while preserving the functionality and accessibility of the keyboard navigation.

### Changes Made

1. Refactored handleKeydown Method:
-- Broke down the logic into smaller, modular methods for handling different key events.
-- Ensured that the Tab key mimics the functionality of the ArrowRight key.
-- Ensured that the Shift+Tab key mimics the functionality of the ArrowLeft key and moves to the last cell when the first cell is focused.
```
methods: {
  handleKeydown(event, rowIndex, colIndex) {
    const key = event.key;
    const totalRows = this.rows.length;
    const totalCols = this.headers.length;

    let nextRowIndex = rowIndex;
    let nextColIndex = colIndex;

    switch (key) {
      case 'ArrowUp':
        ({ nextRowIndex, nextColIndex } = this.handleArrowUp(rowIndex, colIndex, totalRows));
        break;
      case 'ArrowDown':
        ({ nextRowIndex, nextColIndex } = this.handleArrowDown(rowIndex, colIndex, totalRows));
        break;
      case 'ArrowLeft':
        ({ nextRowIndex, nextColIndex } = this.handleArrowLeft(rowIndex, colIndex, totalRows, totalCols));
        break;
      case 'ArrowRight':
        ({ nextRowIndex, nextColIndex } = this.handleArrowRight(rowIndex, colIndex, totalRows, totalCols));
        break;
[...]

Issue #802, opened 2024-10-15
Improve unit test coverage of KCardGrid and KCard


MisRob:
## Summary

In https://github.com/learningequality/kolibri-design-system/pull/796, I got stuck on some issues with mocking composables in KDS that prevented me from completing unit tests, particularly for new `KCardGrid` composables. The goal of this issue is to debug in more depth and complete the unit tests for all the important use-cases implemented in _lib/cards_.

MisRob:
Closing in favor of https://github.com/learningequality/kolibri-design-system/issues/1001

Issue #801, opened 2024-10-15
Optimize KCardGrid's internal usage of watchers


MisRob:
## Summary

In https://github.com/learningequality/kolibri-design-system/pull/796, I experienced problems with reactivity when using `computed` in some places.  I was able to track it to `useKResponsiveWindow`, but am still not clear what exactly was going on. I temporarily resolved it by using `watch` even in places that would benefit from `computed`. The goal of this issue is to audit all watchers in _lib/cards_ files and refactor to `computed` where possible or debug and comment on why `watch` has to be used exactly.

## The Value Add

- Reduces tech debt
- `computed` model is more intuitive in some places
- `computed` is generally more performant than  `watch`

MisRob:
Closing in favor of https://github.com/learningequality/kolibri-design-system/issues/1000

Issue #772, opened 2024-09-04
[Proposal]: Add default sort to KTable


BabyElias:
## Product
KDS

## Description :
We plan to include a feature in KTable that will allow the table to be sorted by default based on any column when initially loaded. 
For eg :
![Frame 572](https://github.com/user-attachments/assets/89209f15-8668-4b09-9459-21ccb54ee364)
## Desired behavior 
In order to implement this, we propose two ways of configuring the API. Not sure, which one would actually be more suited from a user perspective :
- Defining default sorter state with KTable itself, something like :
```
<KTable
  :headers="headers"
  :rows="rows"
  caption="Table with built-in sorting"
  sortable
  :default="{column: 'Full Name', state:'asc'}"
/>
```
- Defining default sorter state when defining `headers` , something like :
```
<KTable
  :headers="headers"
  :rows="rows"
  caption="Table with built-in sorting"
  sortable
/>
data() {
  return {
    headers: [
      { label: 'Full Name', dataType: 'string', default : SORT_ASC },
      { label: 'Age', dataType: 'numeric' },
      { label: 'City', dataType: 'string' },
    ],
}
```

MisRob:
Thanks @BabyElias!

(1) When I looked at `default` with newcomer's eyes, it wasn't very clear to me what it's supposed to do. So I think we could improve the name. Perhaps `defaultSortOrder`, `sortOrder`, `defaultSort`?

(2) Imagining using `:default="{column: 'Full Name', state:'asc'}"` to me didn't feel very intuitive compared to configuring right on `headers`. Probably because we already have various configurations on `headers`, such as width, so the latter option fits that mental picture and feels cohesive.

Putting it together, maybe something like

```
data() {
  return {
    headers: [
      { label: 'Full Name', dataType: 'string', defaultSort : SORT_ASC },
      { label: 'Age', dataType: 'numeric' },
      { label: 'City', dataType: 'string' },
    ],
}
```

???

This is not meant to be conclusion. Just feedback based on my first thoughts - we can build on top of it or perhaps it will help to spark some alternative ideas. Glad to discuss. Let's see if we can also hear from others.

BabyElias:
The second approach using headers has also been my personal preference. It's more intuitive and easier to configure that way.

AlexVelezLl:
[...]

PR #824, opened 2024-11-13
Add default sort to `KTable`

### LINKED ISSUE #794: Add default sort to KTable
🌱 **Are you new to the codebase? Welcome! Please see [the contributing guidelines](https://github.com/learningequality/kolibri-design-system/blob/develop/CONTRIBUTING.md).**

## Summary

Add a new feature to `KTable` that will allow the table to be sorted by default when initially loaded based on a chosen column. To achieve this:

- (1) Each entry of `headers` will be expected to have a unique column identifier `columnId`
- (2) There will be a new prop `defaultSort` to receives a configuration object `{ columnId, direction }`. `columnId` references a `columnId` defined in `headers`. This specifies a column by which the table should be sorted when initially loaded. `direction` can be `'asc'` for ascending or `'desc'` for descending sort direction. `defaultSort` prop is not required.

This feature is optional. The table should not be sorted by any column when initially loaded unless `defaultSort` is provided.

### Usage example

```
<KTable
  sortable
  :headers="headers"
  :rows="rows"
  :defaultSort="{columnId: 'nameCol', direction: 'asc'}"
  ...
/>
```

```
data() {
  return {
    headers: [
      { columnId: 'nameCol', label: 'Name', dataType: 'string' },
      { columnId: 'ageCol', label: 'Age', dataType: 'number' },
      { columnId: 'citCol', label: 'City', dataType: 'string' },
    ],
    rows: [
      ['John Doe', 28, 'New York'],
      ['Alice Johnson', 30, 'Houston'],
      ['Emily Davis', 27, 'Philadelphia'],
    ]
  }
}
```

makes the table entries to be sorted by name in ascending direction when initially loaded.

### Related tasks

**Validation**
- `KTable` throws a warning when `columnId` is not present in any entry of `headers`. This should be the case no matter of whether `defaultSort` is used or not.
- `KTable` throws a warning when `defaultSort`'s `direction` is a wrong value. Only allowed values are `'asc'` and `'desc'`.
- `KTable` throws a warning when `defaultSort`'s `columnId` wasn't found in `headers`.

**Documentation**
- All existing `KTable` documentation examples are updated with `columnId` in `headers`
- There's a new sub-section in `KTable`'s documentation "Usage" section that explains this new feature

## References

- Initial proposal discussed and decisions made in https://github.com/learningequality/kolibri-design-system/issues/772

## Use-cases

- Some Kolibri tables with names will be sorted by full name in ascending order when page loaded ([Figma designs](https://www.figma.com/design/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=194-2920&node-type=canvas&t=AlyLhE0o8MBkXAWE-0))

## Guidance

- Use [the playground page](https://github.com/learningequality/kolibri-design-system/blob/develop/dev_docs/03_how_to_update_library.md#1-preview-updates-in-the-kolibri-design-system-environment) for development
- For documentation updates
  - Follow [“Writing” guidelines](https://design-system.learningequality.org/writing/). Use language that is consistent with other documentation pages
  - Refer to ["How to update the documentation website"](https://github.com/learningequality/kolibri-design-system/blob/develop/dev_docs/04_how_to_update_docs.md#how-to-update-the-documentation-website) and code of the existing documentation pages

## Acceptance criteria

- [ ] The feature is implemented as specified above
- [ ] The feature is covered by unit tests
- [ ] There are no regressions in existing `KTable` usages
- [ ] `KTable` has new validations as specified above
- [ ] `KTable` documentation page is updated as specified above

EshaanAgg:
## Description
Adds default sorting functionality to the `KTable` component

#### Issue addressed
Fixes #794 

## Changelog

  - **Description:** Adds default sorting functionality feature
  - **Addresses:** [#794](https://github.com/learningequality/kolibri-design-system/issues/794) 
  - **Components:** KTable
  - **Breaking:**  no
  - **Impacts a11y:** no
  - **Guidance:** -
 
  - **Description:**  Adds requirement for `columnId` attribute in all `header` objects
  - **Addresses:** [#794](https://github.com/learningequality/kolibri-design-system/issues/794) 
  - **Components:** KTable
  - **Breaking:**  yes
  - **Impacts a11y:** no
  - **Guidance:** Add a unique column identifier `columnId` to all `header` objects 
  
  - **Description:** Renames `disableDefaultSorting` prop to `disableBuiltinSorting`
  - **Addresses:** [#794](https://github.com/learningequality/kolibri-design-system/issues/794) 
  - **Components:** KTable
  - **Breaking:**  yes
  - **Impacts a11y:** no
  - **Guidance:** Rename all occurrence of `disableDefaultSorting`

## Steps to test
Test out the component in the playground!

- I have removed the `disableDefaultSorting` prop, as now we can use the `deafultSort` attribute to convey whether we want to sort the rows by default. I also removed the logic regarding emitting the `change sort` event as the same felt to be moot now. 
- I have reworked the sorting logic not just to take a boolean of `useLocalSort` (which sorted the rows by their complete values) but updated the API to allow developers to choose the `columnId` they want to sort on.  
- Better test coverage of the sorting composable with default sorting 
- Added exhaustive prop validation for the new `deafultSort` prop
- Added more tests to the `KTable` component
- I added an example related to `defaultSort` on the documentation pages as well. 

Please let me know if you see any possible regressions due to my implementation!

No

## Testing checklist

## Reviewer guidance

### DIFF
--- a/docs/pages/ktable.vue
+++ b/docs/pages/ktable.vue
@@ -39,15 +39,15 @@
               { label: 'Age', dataType: 'number', columnId: 'age' },
               { label: 'City', dataType: 'string', columnId: 'city' },
             ],
-          rows: [
-              ['John Doe', 28, 'New York'],
-              ['Jane Smith', 34, 'Los Angeles'],
-              ['Samuel Green', 22, 'Chicago'],
-              ['Alice Johnson', 30, 'Houston'],
-              ['Michael Brown', 45, 'Phoenix'],
-              ['Emily Davis', 27, 'Philadelphia'],
-            ]
-          };
+            rows: [
+                ['John Doe', 28, 'New York'],
+                ['Jane Smith', 34, 'Los Angeles'],
+                ['Samuel Green', 22, 'Chicago'],
+                ['Alice Johnson', 30, 'Houston'],
+                ['Michael Brown', 45, 'Phoenix'],
+                ['Emily Davis', 27, 'Philadelphia'],
+              ]
+            };
         },
       </DocsShowCode>
       <!-- eslint-enable -->
@@ -64,10 +64,9 @@
       <h3>Table with sorting</h3>
       <p>
         The <code>KTable</code> offers built-in sorting functionality. There are 4 permissible data types - <code>string</code>,<code>number</code>,<code>date</code> and <code>undefined</code>. Columns declared with <code>undefined</code> data type are not sortable. This example demonstrates a table with sorting enabled via the <code>sortable</code> prop.
-        
       </p>
       <p>
-       Clicking the same header multiple times toggles the sort direction cyclically in the order of ascending, descending, and unsorted.
+        Clicking the same header multiple times toggles the sort direction cyclically in the order of ascending, descending, and unsorted.
       </p>
 
       <DocsShowCode language="html">
@@ -233,7 +232,7 @@
       </DocsShow>
       <!-- eslint-enable -->
 
-      <!--Table with Default Sort-->
+      <!--Table with default sort-->
       <h3>Table with default sort</h3>
       <p>
         This is an example to show how to use the <code>defaultSort</code> prop to sort the table based on a particular column upon the initial load. The <code>defaultSort</code> attribute can be used irrespective of the <code>sortable</code> attribute.
@@ -264,7 +263,6 @@
           caption="Unsortable Table with Rows Sorted by 'Age' Column"
           :defaultSort="{ columnId: 'age', direction: 'asc' }"
         />
-
       </DocsShowCode>
 
       <!-- eslint-disable -->
@@ -308,7 +306,7 @@
         />
       </DocsShow>
 
-      <!-- Disable Builtin Sorting -->
+      <!-- Disable built-in sorting -->
       <h3>Disable built-in sorting</h3>
       <p>
         For <code>sortable</code> tables, you can use the <code>disableBuiltinSorting</code> prop to disable built-in sort function. This is useful when the table receives already sorted data, for example when sorting is done by backend or a custom sorting function outside the table. In this case, when one of the header sort buttons is clicked, the table won't sort the column itself, but only emit the <code>changeSort</code> event to notify the parent component to handle the sorting logic. The event contains column index of the header and the sort order in its payload.
--- a/docs/pages/ktable.vue
+++ b/docs/pages/ktable.vue
@@ -13,14 +13,14 @@
         </ul>
       </p>
     </DocsPageSection>
-  
+
     <DocsPageSection title="Usage" anchor="#usage">
       <!--Non-Sortable Table-->
       <h3>Table without sorting functionality</h3>
       <p>
         This is an example to show how <code>KTable</code> can be used without any sorting functionality, as a simple table. 
       </p>
-  
+
       <DocsShowCode language="html">
         <KTable
           :headers="headers"
@@ -154,7 +154,7 @@
           };
         },
       </DocsShowCode>
-  
+
       <DocsShow block>
         <KTable
           :headers="slotHeaders"
@@ -471,9 +471,9 @@
         const newRows = rows.reverse();
         return {
           rows: newRows,
-          sortOrder: currentSortOrder === "asc" ? "desc" : "asc",
+          sortOrder: currentSortOrder === 'asc' ? 'desc' : 'asc',
           sortKey: columnIndex,
-        }
+        };
       },
     },
   };
--- a/docs/pages/ktable.vue
+++ b/docs/pages/ktable.vue
@@ -310,21 +310,21 @@
       <!-- Disable Builtin Sorting -->
       <h3>Disable Builtin Sorting & Custom Sorting Logic</h3>
       <p>
-        You can make use of the <code>disableBuiltinSorting</code> attribute to disable all sorting functionality in the table. This is useful when you want to display the data in a particular order or want to define a custom sorting mechanism. 
+        You can make use of the <code>disableBuiltinSorting</code> attribute to disable all sorting functionality by the table component. This is useful when you want to display the data in a particular order or you want to define a custom sorting function. 
       </p>
 
       <p>
-        You should not use this attribute if the <code>sortable</code> attribute is set to <code>false</code>, as in that case the table headers for sorting will not be displayed at all. If the same is set to <code>true</code>, then table emits a <code>changeSort</code> event with column index of the header clicked and the sort order to notify the parent component to handle the sorting logic. 
+        You should not use this attribute if <code>sortable</code> is set to <code>false</code>. If <code>sortable</code> is set to <code>true</code>, then the table component will emit a <code>changeSort</code> event with column index of the header clicked and the sort order to notify the parent component to handle the sorting logic. 
       </p>
 
       <p>
-        You can also define a custom sorting logic using the <code>customSort</code> attribute. The <code>customSort</code> attribute takes a function that accepts the rows, column index and the current sort order as arguments. The function should return an object with the following properties:
+        You can also define custom sorting logic in such a scenario using the <code>customSort</code> attribute. The <code>customSort</code> attribute takes a function that accepts the current rows, column index of the click, and the current sort order as arguments. The function should return an object with the following properties:
         <ul>
           <li><code>rows</code>: The sorted rows</li>
           <li><code>sortOrder</code>: The sort order of the column</li>
           <li><code>sortKey</code>: The column index based on which the sorting was done</li>
         </ul>
-        These values are then used by the component to update the state of the table data and the headers. You can set the <code>sortOrder</code> to one of <code>['asc', 'desc', null]</code> to convey the sort order to the table. The <code>sortKey</code> is the column index based on which the sorting was done, and can be set to <code>null</code> as well to convey that no sorting was done.
+        These values are used by the <code>KTable</code> component to update the table rows and headers. You can set the <code>sortOrder</code> to one of <code>['asc', 'desc', null]</code> to convey the sort order. The <code>sortKey</code> is the column index based on which the sorting was done, and can be set to <code>null</code> as well to convey the default no-sorting state.
       </p>
 
       <DocsShowCode language="html">
--- a/docs/pages/ktable.vue
+++ b/docs/pages/ktable.vue
@@ -308,23 +308,13 @@
       </DocsShow>
 
       <!-- Disable Builtin Sorting -->
-      <h3>Disable Builtin Sorting & Custom Sorting Logic</h3>
+      <h3>Disable Builtin Sorting</h3>
       <p>
         You can make use of the <code>disableBuiltinSorting</code> attribute to disable all sorting functionality by the table component. This is useful when you want to display the data in a particular order or you want to define a custom sorting function. 
       </p>
 
       <p>
         You should not use this attribute if <code>sortable</code> is set to <code>false</code>. If <code>sortable</code> is set to <code>true</code>, then the table component will emit a <code>changeSort</code> event with column index of the header clicked and the sort order to notify the parent component to handle the sorting logic. 
-      </p>
-
-      <p>
-        You can also define custom sorting logic in such a scenario using the <code>customSort</code> attribute. The <code>customSort</code> attribute takes a function that accepts the current rows, column index of the click, and the current sort order as arguments. The function should return an object with the following properties:
-        <ul>
-          <li><code>rows</code>: The sorted rows</li>
-          <li><code>sortOrder</code>: The sort order of the column</li>
-          <li><code>sortKey</code>: The column index based on which the sorting was done</li>
-        </ul>
-        These values are used by the <code>KTable</code> component to update the table rows and headers. You can set the <code>sortOrder</code> to one of <code>['asc', 'desc', null]</code> to convey the sort order. The <code>sortKey</code> is the column index based on which the sorting was done, and can be set to <code>null</code> as well to convey the default no-sorting state.
       </p>
 
       <DocsShowCode language="html">
@@ -334,7 +324,6 @@
           caption="Disable Builtin Sorting Example"
           sortable
           disableDefaultSorting
-          :customSort="customSort"
           @changeSort="changeSortHandler"
         />
       </DocsShowCode>
@@ -362,16 +351,6 @@
           changeSortHandler(index, sortOrder) {
             console.log(`changeSort event emitted with index: ${index} and sortOrder: ${sortOrder}`);
           },
-          customSort(rows, columnIndex, currentSortOrder) {
-            // Demo Implementation: Reverse the order of the rows irrespective of the current sort order
-            console.log(`Custom sorting logic applied for column index: ${columnIndex}`);
-            const newRows = rows.reverse();
-            return {
-              rows: newRows,
-              sortOrder: currentSortOrder === "asc" ? "desc" : "asc",
-              sortKey: columnIndex,
-            }
-          },
         },
       </DocsShowCode>
       <!-- eslint-enable -->
@@ -383,7 +362,6 @@
           caption="Disable Builtin Sorting Example"
           sortable
           disableBuiltinSorting
-          :customSort="customSort"
           @changeSort="changeSortHandler"
         />
       </DocsShow>
@@ -464,16 +442,6 @@
     methods: {
       changeSortHandler(index, sortOrder) {
         console.log(`changeSort event emitted with index: ${index} and sortOrder: ${sortOrder}`);
-      },
-      customSort(rows, columnIndex, currentSortOrder) {
-        // Demo Implementation: Reverse the order of the rows irrespective of the current sort order
-        console.log(`Custom sorting logic applied for column index: ${columnIndex}`);
-        const newRows = rows.reverse();
-        return {
-          rows: newRows,
-          sortOrder: currentSortOrder === 'asc' ? 'desc' : 'asc',
-          sortKey: columnIndex,
-        };
       },
     },
   };
--- a/docs/pages/ktable.vue
+++ b/docs/pages/ktable.vue
@@ -180,47 +180,65 @@
 
       </DocsShowCode>
 
-  <DocsShowCode language="javascript">
-    data() {
-      return {
-        headersWithCustomWidths: [
-          { label: 'Name', dataType: 'string', minWidth: '20px', width: '2%', columnId: 'name' },
-          { label: 'Age', dataType: 'number', minWidth: '100px', width: '33%', columnId: 'age' },
-          { label: 'City', dataType: 'string', minWidth: '200px', width: '25%', columnId: 'city' },
-          {
-            label: 'Joined',
-            dataType: 'date',
-            minWidth: '150px',
-            width: '20%',
-            columnId: 'joined',
-          },
-          {
-            label: 'Misc',
-            dataType: 'undefined',
-            minWidth: '100px',
-            width: '20%',
-            columnId: 'misc',
-          },
-        ],
-        customRows: [
-          ['John Doe', 28, 'New York', '2022-01-15T00:00:00Z', 'N/A'],
-          ['Jane Smith', 34, 'Los Angeles', '2021-12-22T00:00:00Z', 'N/A'],
-          ['Samuel Green', 22, 'Chicago', '2023-03-10T00:00:00Z', 'N/A'],
-          ['Alice Johnson', 30, 'Houston', '2020-07-18T00:00:00Z', 'N/A'],
-        ],
-      };
-    },
+      <!--Table with Default Sort-->
+      <h3>Table with Default Sort</h3>
+      <p>
+        This is an example to show how <code>KTable</code> can be used with the <code>defaultSort</code> attribute to sort the table based on a particular column. This is useful if you are getting unsorted data from an API and want to display it in a sorted manner. The <code>defaultSort</code> attribute can be used irrespective of the <code>sortable</code> attribute.
+      </p>
+
+      <DocsShowCode language="html">
+        <KTable
+          :headers="headers"
+          :rows="rows"
+          caption="Sortable Table with Rows Sorted by 'Age' Column"
+          sortable
+          :defaultSort="{ columnId: 'age', direction: 'asc' }"
+        />
+
+        <KTable 
+          :headers="headers"
+          :rows="rows"
+          caption="Unsortable Table with Rows Sorted by 'Age' Column"
+          :defaultSort="{ columnId: 'age', direction: 'asc' }"
+        />
+
+      </DocsShowCode>
+
+      <DocsShowCode language="javascript">
+        data() {
+          return {
+            headers: [
+              { label: 'Name', dataType: 'string', columnId: 'name' },
+              { label: 'Age', dataType: 'number', columnId: 'age' },
+              { label: 'City', dataType: 'string', columnId: 'city' },
+            ],
+            rows: [
+              ['John Doe', 28, 'New York'],
+              ['Jane Smith', 34, 'Los Angeles'],
+              ['Samuel Green', 22, 'Chicago'],
+              ['Alice Johnson', 30, 'Houston'],
+              ['Michael Brown', 45, 'Phoenix'],
+              ['Emily Davis', 27, 'Philadelphia'],
+            ]
+          };
+        },
       </DocsShowCode>
 
       <DocsShow block>
         <KTable
-          :headers="headersWithCustomWidths"
-          :rows="customRows"
-          caption="Table showing columns with custom widths"
+          :headers="headers"
+          :rows="rows"
+          caption="Sortable Table with Rows Sorted by 'Age' Column"
           sortable
+          :defaultSort="{ columnId: 'age', direction: 'asc' }"
+        />
+        <KTable
+          :headers="headers"
+          :rows="rows"
+          caption="Unsortable Table with Rows Sorted by 'Age' Column"
+          :defaultSort="{ columnId: 'age', direction: 'asc' }"
         />
       </DocsShow>
-      <!-- eslint-enable -->
 
     </DocsPageSection>
 
--- a/docs/pages/ktable.vue
+++ b/docs/pages/ktable.vue
@@ -235,7 +235,7 @@
       <!--Table with Default Sort-->
       <h3>Table with Default Sort</h3>
       <p>
-        This is an example to show how <code>KTable</code> can be used with the <code>defaultSort</code> attribute to sort the table based on a particular column. This is useful if you are getting unsorted data from the backend and want to display it in a sorted manner on load. The <code>defaultSort</code> attribute can be used irrespective of the <code>sortable</code> attribute (as <code>sortable</code> is used to configure whether the client has sorting capabilities or not). 
+        This is an example to show how <code>KTable</code> can be used with the <code>defaultSort</code> attribute to sort the table based on a particular column. This is useful if the table needs to be sorted based on any column from the beginning itself (by default). The <code>defaultSort</code> attribute can be used irrespective of the <code>sortable</code> attribute (as <code>sortable</code> is used to configure whether the client has sorting capabilities or not). 
       </p>
 
       <p>
@@ -243,7 +243,7 @@
       </p>
 
       <p>
-        To make use of <code>defaultSort</code>, please ensure that the <code>disableBuiltinSorting</code> attribute is not set to <code>true</code> as it will disable all sorting functionality.
+        To make use of <code>defaultSort</code>, please ensure that the <code>disableBuiltinSorting</code> attribute is not set to <code>true</code>.
       </p>
 
       <DocsShowCode language="html">
@@ -314,7 +314,17 @@
       </p>
 
       <p>
-        You should not use this attribute if the <code>sortable</code> attribute is set to <code>false</code>, as in that case the table headers for sorting will not be displayed at all. If the same is set to <code>true</code>, then table emits a <code>changeSort</code> event (whose implementation can be defined by the user) which needs to return a <code>sortOrder</code> value for the UI headers to toggle.
+        You should not use this attribute if the <code>sortable</code> attribute is set to <code>false</code>, as in that case the table headers for sorting will not be displayed at all. If the same is set to <code>true</code>, then table emits a <code>changeSort</code> event with column index of the header clicked and the sort order to notify the parent component to handle the sorting logic. 
+      </p>
+
+      <p>
+        You can also define a custom sorting logic using the <code>customSort</code> attribute. The <code>customSort</code> attribute takes a function that accepts the rows, column index and the current sort order as arguments. The function should return an object with the following properties:
+        <ul>
+          <li><code>rows</code>: The sorted rows</li>
+          <li><code>sortOrder</code>: The sort order of the column</li>
+          <li><code>sortKey</code>: The column index based on which the sorting was done</li>
+        </ul>
+        These values are then used by the component to update the state of the table data and the headers. You can set the <code>sortOrder</code> to one of <code>['asc', 'desc', null]</code> to convey the sort order to the table. The <code>sortKey</code> is the column index based on which the sorting was done, and can be set to <code>null</code> as well to convey that no sorting was done.
       </p>
 
       <DocsShowCode language="html">
@@ -324,6 +334,7 @@
           caption="Disable Builtin Sorting Example"
           sortable
           disableDefaultSorting
+          :customSort="customSort"
           @changeSort="changeSortHandler"
         />
       </DocsShowCode>
@@ -349,12 +360,19 @@
         },
         methods: {
           changeSortHandler(index, sortOrder) {
-            // Demo Implementation: Reverse the order of the rows and return the opposite sort order
-            console.log('Index:', index, 'Sort Order:', sortOrder);
-            this.changeSortRows = [...this.changeSortRows.reverse()];
-            return sortOrder === 'asc' ? 'desc' : 'asc';
-          }
-        }
+            console.log(`changeSort event emitted with index: ${index} and sortOrder: ${sortOrder}`);
+          },
+          customSort(rows, columnIndex, currentSortOrder) {
+            // Demo Implementation: Reverse the order of the rows irrespective of the current sort order
+            console.log(`Custom sorting logic applied for column index: ${columnIndex}`);
+            const newRows = rows.reverse();
+            return {
+              rows: newRows,
+              sortOrder: currentSortOrder === "asc" ? "desc" : "asc",
+              sortKey: columnIndex,
+            }
+          },
+        },
       </DocsShowCode>
       <!-- eslint-enable -->
 
@@ -365,6 +383,7 @@
           caption="Disable Builtin Sorting Example"
           sortable
           disableBuiltinSorting
+          :customSort="customSort"
           @changeSort="changeSortHandler"
         />
       </DocsShow>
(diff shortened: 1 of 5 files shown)


EshaanAgg:
Update: I have added the tests for the `defaultSort` prop to the `KTable` suite, but there seems to be some error in the reactive handling that I can't seem to figure out. The outline of the approach that I am trying to follow is like:
- Use the `defaultSort` and `headers` props to get the column index based on which I should be performing the sorting.
- Supply the same to the `useSorting` hook.

As you can see the `useSorting` tests pass, and thus, we can be sure that the hook is handling the data as expected. There probably is some error in the way I am handling the reactive references to the props while calculating the `useDefaultSorting` reference in the setup function. I would greatly appreciate any help with the same!

MisRob:
Hi @EshaanAgg, thanks so much for your work on this :) I will dive into more detailed review as soon as we clarify a few things and debug the reactivity issues. So far, I read the PR description and the updated documentation. Couple of first higher-level thoughts:

> I have removed the disableDefaultSorting prop, as now we can use the deafultSort attribute to convey whether we want to sort the rows by default. I also removed the logic regarding emitting the change sort event as the same felt to be moot now.

The purpose of `disableDefaultSorting` is to turn off internal sorting completely (this is for the case when the table receives sorted data from backend). The function of the `defaultSort` is to configure by which column should the table be sorted on an initial load. We may use it or not. There's some overlap, but I don't think that's the same, right? Seems names I chose are confusing here.

> Docs "The defaultSort attribute can be used irrespective of the sortable attribute."

First thanks for all the docs updates. And I appreciated this particular example. This behavior seems useful to me, thanks!

> there seems to be some error in the reactive handling that I can't seem to figure out

We'll be glad to help. I think I could understand a bit some parts, but I am not sure how to reproduce the issue. Would you post step by step what should I observe and where - e.g. by console logging or any other means.

MisRob:
Flagging this PR for @BabyElias - as your time allows, if you'd like to, you're very much welcome to join the full or partial review :)

MisRob:
Hi @EshaanAgg, thanks so much for your work on this :) I will dive into more detailed review as soon as we clarify a few things and debug the reactivity issues. So far, I read the PR description and the updated documentation. Couple of first higher-level thoughts:

> I have removed the disableDefaultSorting prop, as now we can use the deafultSort attribute to convey whether we want to sort the rows by default. I also removed the logic regarding emitting the change sort event as the same felt to be moot now.

The purpose of `disableDefaultSorting` is to turn off internal sorting completely (this is for the case when the table receives sorted data from backend). The function of the `defaultSort` is to configure by which column should the table be sorted on an initial load. We may use it or not. There's some overlap, but I don't think that's the same, right? Seems names I chose are confusing here.

> Docs "The defaultSort attribute can be used irrespective of the sortable attribute."

First thanks for all the docs updates. And I appreciated this particular example. This behavior seems useful to me, thanks!

> there seems to be some error in the reactive handling that I can't seem to figure out

We'll be glad to help. I think I could understand a bit some parts. Just to be sure, you're referring to the test failure, right?

EshaanAgg:
Thanks a lot, @MisRob, for the response. 

> The purpose of `disableDefaultSorting` is to turn off internal sorting completely (this is for the case when the table receives sorted data from backend). The function of the `defaultSort` is to configure by which column should the table be sorted on an initial load. We may use it or not. There's some overlap, but I don't think that's the same, right? Seems names I chose are confusing here.

I see, but I would like to argue that the presence of such a prop like `useLocalSorting` is somewhat of an antipattern. I say that because what if the user sets `sortable` to `true` and `useLocalSorting` to `false`? In such a case, according to the current implementation of the `useSorting` hook, we would just return the original rows without actually performing any sorting. At the same time, the UI headers would toggle (from `asc` -> `desc` -> `default`) and update their state, which is somewhat misleading (as no actual sorting of the data is being performed while the UI is conveying that the same does indeed occur). 

The introduction of the `defaultSort` helps to clear up this ambiguity, which clearly states the precedence: always listen to the latest update by the user first, and then if no sorting is currently being applied, apply the default sorting by the column provided. This, in my opinion, is a much cleaner API. Having `defaultSort` to be empty effectively disables all the internal sorting in the table.

I am happy to be corrected if my understanding of these functionalities is at fault somewhere! Thanks for pointing the same out!

> We'll be glad to help. I think I could understand a bit some parts. Just to be sure, you're referring to the test failure, right?

Yes. I initially wrote a basic test suite for both the principal components (`useSorting` and `KTable`) with the intended functionality associated with `defaultSort` and then tried to implement the changes to pass them. All the tests for `useSorting` have succeeded, while those for `KTable` have not. I can share a playground script and a brief description of what to look for that helps debug!

BabyElias:
Hey @EshaanAgg, Thank you for working on this issue :)

> I see, but I would like to argue that the presence of such a prop like useLocalSorting is somewhat of an antipattern. I say that because what if the user sets sortable to true and useLocalSorting to false? In such a case, according to the current implementation of the useSorting hook, we would just return the original rows without actually performing any sorting. At the same time, the UI headers would toggle (from asc -> desc -> default) and update their state, which is somewhat misleading (as no actual sorting of the data is being performed while the UI is conveying that the same does indeed occur).

Talking about the presence of an anti-pattern, in a case where `sortable is set to true` and `disableDefaultSorting is set to true` (i.e uselocalSorting set to false)- the implementation for this sorting experience has to be provided by the user and is independent of the `useSorting` hook. As you can see in the code [here](https://github.com/learningequality/kolibri-design-system/blob/31eb6365d8f7b5e351dba8376d860d6e04f7f4be/lib/KTable/index.vue#L141-L154), if disableDefaultSorting is false - the table emits a `changesort` event(whose implementation can be defined by the user) and this event needs to return a `sortOrder` value for the UI headers to toggle. If the sortOrder value is not specified by the user - no toggling effect would occur. You can see an example of such a use-case in the playground page code [here](https://github.com/learningequality/kolibri-design-system/blob/eb923d275021f1d0ba3b97c9cd59a2f6165dbc0c/docs/pages/playground.vue#L26-L49). Here, the changeSort event just simply reverses the columns upon being clicked and since no sortOrder value has been returned by the function - the headers are not toggled.

EshaanAgg:
Thanks @BabyElias! As you stated here, I had missed entirely the functionality of the `changesort` event. Thanks again for clarifying this and resolving my query. 

Do you have any pointers on designing the API for this component going forward? Some queries I had for the same include:
- Can we rename `disableDefaultSorting` to `disableInternalSorting` or anything else that communicates this indented behavior more clearly?
- We should probably document this behavior and the usage of this prop with examples in the documentation section!
- What should the expected behavior be if we set `disableDefaultSorting` to `false` and even provide a `defaultSort` attribute?  Should the sorting occur or not in such a case?

BabyElias:
@EshaanAgg, No worries - the lack of documentation for this mode can so lead to the confusion. About the queries, 
- I think we can rename the prop to `disableBuiltInSorting` for more clarity, we already had discussed this as an option in a PR earlier (for reference, [here](https://github.com/learningequality/kolibri-design-system/pull/727#issuecomment-2295168580))
- Yes, we do need to document this KTable mode - we are currently working on a real-application for this use-case in Kolibri. Once it is done, we will be able to identify the shortcomings of this mode, enhance the API and finally document it in a much better way. 
- From the [conversation](https://github.com/learningequality/kolibri-design-system/issues/772) about how we felt the need to introduce a defaultSort prop to the table - I believe this feature would only be needed for tables that have `disableDefaultSorting` set to false. For tables whose data is fetched from backend, configuring the external API to use the defaultSort option provided by KTable would rather be not needed as we can already request for data from backend in such a way that a column is sorted by default. In my opinion, KTable should just interact with changeSort in such a case and not have any interaction with the backend directly.

Your opinions on this @MisRob ?

MisRob:
Hi @BabyElias and @EshaanAgg, thanks both!

> I think we can rename the prop to disableBuiltInSorting for more clarity

I think that'd help. We can definitely do smaller changes to improve clarity now.

> Yes, we do need to document this KTable mode - we are currently working on a real-application for this use-case in Kolibri. Once it is done, we will be able to identify the shortcomings of this mode, enhance the API and finally document it in a much better way.

Yes, issue here https://github.com/learningequality/kolibri/issues/12705 for reference.

> From the https://github.com/learningequality/kolibri-design-system/issues/772 about how we felt the need to introduce a defaultSort prop to the table - I believe this feature would only be needed for tables that have disableDefaultSorting set to false. For tables whose data is fetched from backend, configuring the external API to use the defaultSort option provided by KTable would rather be not needed as we can already request for data from backend in such a way that a column is sorted by default.

I think that would make sense.

MisRob:
@EshaanAgg

> Yes. I initially wrote a basic test suite for both the principal components (useSorting and KTable) with the intended functionality associated with defaultSort and then tried to implement the changes to pass them. All the tests for useSorting have succeeded, while those for KTable have not. I can share a playground script and a brief description of what to look for that helps debug!

I will have a look at this some time this week. I think I saw strange reactivity issues in another tests when using composables, so I wonder if it's related. If I don't have any success, we can try continuing debugging together or invite someone else from the team.

MisRob:
@BabyElias @EshaanAgg For context, I hope for focused detailed work on table's public API specification at some point. Even though we touched on it a bit, it really wasn't the focus of the GSoC project. So hopefully we will fine-tune all these aspects. If you had some more feedback, now or in the future, even for breaking updates to the public interface, I just opened [a tracking issue](https://github.com/learningequality/kolibri-design-system/issues/829) that should serve us well for discussions.

EshaanAgg:
Thanks a lot for the reviews @MisRob and @BabyElias! Since we had a lot of discussion points and use cases to clear, I felt that it would be best first to update the documentation and then change the code. It would be a great help if you could review the public API along with the examples in the [deploy preview](https://deploy-preview-824--kolibri-design-system.netlify.app/ktable/) and let me know if my understanding of the API, as well as the current implementation, produces the expected results or not.

I also had a query. If you would look at the `Disable Builtin Sorting & Custom Sorting Logic` section in the documentation, I was able to figure out how to implement the custom sorting via the `changeSort` event (thanks a lot to @BabyElias for sharing the playground script!), but I can't seem to figure out how to toggle the headers based on the return value of the function. Is there any existing functionality for the same, or the same remains to be implemented?

If all seems well, I can work on adding the tests (I thought it would be great to add all the documentation examples as integration tests to the component so that there are no regressions in the component, as I had introduced earlier).

BabyElias:
@EshaanAgg
From the code that I see for disableBuiltInSort example on the playground page, I guess the issue is with not returning sortKey as part of the function or with reactivity. In the code that [useSorting](https://github.com/learningequality/kolibri-design-system/blob/develop/lib/KTable/useSorting/index.js) composable uses (responsible for sorting data on frontend when built in sorting is enabled), sortOrder and sortKey both are returned and as we can see in the part code responsible for icon toggle [here](https://github.com/learningequality/kolibri-design-system/blob/c04026a44fd4646a7a4f5763560b1d8ee1f6e2ac/lib/KTable/index.vue#L41-L44) - if sortOrder is not defined, then I don't think icon will be toggled as `isColumnSortActive(index)` will be false.
If even after this the code doesn't work, maybe we can try using ref with sortOrder and sortKey just to ensure it's not an issue with reactivity. Thanks!

BabyElias:
Also, for the documentation -
Thank you so much for all the efforts in putting things there. The examples used are just perfect!
However, I am not sure about the description for _Table with Default Sort_ section :

> This is useful if you are getting unsorted data from the backend and want to display it in a sorted manner on load. The defaultSort attribute can be used irrespective of the sortable attribute (as sortable is used to configure whether the client has sorting capabilities or not).
- Maybe it would be better to replace this with : _This is useful if the table needs to be sorted based on any column from the beginning itself (by default)_. We generally avoid using terms like client/backend in the documentation to reduce complexity and avoid confusion :)
> To make use of defaultSort, please ensure that the disableBuiltinSorting attribute is not set to true as it will disable all sorting functionality.
- Disable all sorting functionality as in referring to the client-side sorting functionality here? I think just simply stating that defaultSort only works when disableBuiltInSorting set to false rather than specifying what actually happens then would also work.

MisRob:
Regarding @BabyElias:

> We generally avoid using terms like client/backend in the documentation to reduce complexity and avoid confusion :)

This was my request for the documentation - even though backend is one of the use-cases, the table could also receive data sorted from outside in any other way (e.g. external sort function that's still run on frontend). So basically the language would ideally be as simple and straightforward as possible and not making much assumption about how the component is used (the only exception I could think of would be a specific example, where we could say: _For example, if receiving backend data..._) I hope that's not way too nitpicky, it's based on my reading and some team feedback of the very first version of docs where we experimented a bit with this.

I will too read and reply to all the comments later in the week, but you two are on it so perhaps I won't have much work left by that time :))!

MisRob:
@EshaanAgg

> Since we had a lot of discussion points and use cases to clear, I felt that it would be best first to update the documentation and then change the code. 

Definitely, thanks for following-up :)!

EshaanAgg:
> Maybe it would be better to replace this with : _This is useful if the table needs to be sorted based on any column from the beginning itself (by default)_. We generally avoid using terms like client/backend in the documentation to reduce complexity and avoid confusion :)
> 
> I think just simply stating that defaultSort only works when disableBuiltInSorting set to false rather than specifying what actually happens then would also work.

Thanks for these suggestions! Completely agree that these changes make the documentation much easier to follow through and understand. Have implemented them in the latest push.


> @EshaanAgg From the code that I see for disableBuiltInSort example on the playground page, I guess the issue is with not returning sortKey as part of the function or with reactivity. In the code that [useSorting](https://github.com/learningequality/kolibri-design-system/blob/develop/lib/KTable/useSorting/index.js) composable uses (responsible for sorting data on frontend when built in sorting is enabled), sortOrder and sortKey both are returned and as we can see in the part code responsible for icon toggle [here](https://github.com/learningequality/kolibri-design-system/blob/c04026a44fd4646a7a4f5763560b1d8ee1f6e2ac/lib/KTable/index.vue#L41-L44) - if sortOrder is not defined, then I don't think icon will be toggled as `isColumnSortActive(index)` will be false. If even after this the code doesn't work, maybe we can try using ref with sortOrder and sortKey just to ensure it's not an issue with reactivity. Thanks!

Yes! It turns out that our approach was a bit wrong. Let me explain. When we emit the `changeSort` event, we essentially notify the parent that he should sort the rows based on the click information, but we did not provide him with any means to do so! That's why in the playground script that you shared, you can change the value of the `rows` prop itself, which causes the complete component to rerender and thus is not the most optimal for the performance reason. The result of the `emit` event handler is never propagated to the child component, and thus it has no way of knowing the new updated `sortKey` and `sortOrder`.

The way that I think we should handle this is by introducing two separate handlers. 
- We retain the `changeSort` event handler, so that the parent is notified when the headers are clicked and he is expected to provide a sorting functionality. 
- We introduce a new prop called `customSort`, which the user can optionally define if he wants to implement a custom sorting logic. This function receives the current rows, the column index of the header, and the current sort key as arguments and is expected to return all three back, which can used by the `KTable` to update the state! I have updated the documentation to include an example.

What do you think about this proposed API change?

MisRob:
> We introduce a new prop called customSort, which the user can optionally define if he wants to implement a custom sorting logic.

@EshaanAgg Would it be fine to eventually move the discussion to a new proposal issue about the custom sort or https://github.com/learningequality/kolibri-design-system/issues/829? I'd be also fine doing it. I don't think the table has full support for it yet and there's more things to be decided about the public interface of custom sort functions. It seems it's not strictly necessary to make the `defaultSort` work well for the current use-case, right?

I think it's great to think about these optimizations, and I'm eager to continue iterating with both of you - just chiming in to help us organize this work so the scope can stay as narrow as possible in this PR for ease of review and discussion. Generally for new public APIs, more time is needed for the whole dev team for feedback, so proposal issues are good place.

EshaanAgg:
Yes, @MisRob! Thanks for the input. I would add a comment on #829 specifying the idea that I had, as well as some examples of the same, and would limit the scope of this PR to just `deafultSort`.

MisRob:
Lovely, thank you @EshaanAgg.

MisRob:
Side note, even though I'm not planning to label https://github.com/learningequality/kolibri-design-system/issues/829 for the wider community for contribution, if this task would be interesting for one or both of you at any point in the future, that'd be a different story. I didn't really get an idea to mention it because it's larger time investment, and I feel we're already receiving so so much in `KTable` follow-up work :) However, there's no time pressure at all, and from this PR it seems there is lots of interest and fruitful api discussions already so at least wanted to mention it's open.

MisRob:
Hi @EshaanAgg and @BabyElias, I think the next step now would be me confirming the documentation and then helping to debug the tests, is that right?

BabyElias:
Yes, the tests are all fixed now I guess- so the documentation review part remains.

MisRob:
Oh I didn't notice that, that's good news. Alright, I will read it in detail soon and confirm.

docs/pages/ktable.vue
```diff
@@ -58,9 +61,12 @@
       </DocsShow>
 
       <!-- Frontend Sorting Example-->
-      <h3>Table with Default Sorting</h3>
+      <h3>Table with Sorting</h3>
```
MisRob: Not introduced by this PR, just an opportunity for little improvement to help [the writing style](https://design-system.learningequality.org/writing/#punctuation) consistency

```suggestion
      <h3>Table with sorting</h3>
```

docs/pages/ktable.vue
```diff
@@ -29,25 +30,27 @@
...
+      <!-- eslint-disable-->
+      <DocsShowCode language="javascript">
+        data() {
+          return {
+            headers: [
+              { label: 'Name', dataType: 'string', columnId: 'name' },
+              { label: 'Age', dataType: 'number', columnId: 'age' },
+              { label: 'City', dataType: 'string', columnId: 'city' },
+            ],
+          rows: [
```
MisRob: Just a small visual improvement, could we indent `rows` one more level?

review-1

docs/pages/ktable.vue
```diff
@@ -58,9 +61,12 @@
       </DocsShow>
 
       <!-- Frontend Sorting Example-->
-      <h3>Table with Default Sorting</h3>
+      <h3>Table with Sorting</h3>
       <p>
-        The <code>KTable</code> can be used with default sorting functionality, allowing you to sort data on the client side without the need for server requests. There are 4 permissible data types - <code>string</code>,<code>number</code>,<code>date</code> and <code>undefined</code>. Columns declared with <code>undefined</code> data type are not sortable. This example demonstrates a table with default sorting enabled.
+        The <code>KTable</code> can be used with sorting functionality, allowing you to sort data on the client side without the need for server requests. There are 4 permissible data types - <code>string</code>,<code>number</code>,<code>date</code> and <code>undefined</code>. Columns declared with <code>undefined</code> data type are not sortable. This example demonstrates a table with sorting enabled. To allow client side sorting, set the <code>sortable</code> attribute to <code>true</code>.
```
MisRob: Again this was pre-existing, so optional. Just spotted a further potential simplification in the context  of what we talked about.

```suggestion
        The <code>KTable</code> offers built-in sorting functionality. There are 4 permissible data types - <code>string</code>,<code>number</code>,<code>date</code> and <code>undefined</code>. Columns declared with <code>undefined</code> data type are not sortable. This example demonstrates a table with sorting enabled via the <code>sortable</code> prop.
        
```

docs/pages/ktable.vue
```diff
@@ -58,9 +61,12 @@
...
+      </p>
+      <p>
+        You can also notice that clicking the same header multiple times toggles the sort direction cyclically in the order of <code>asc</code>, <code>desc</code> and <code>none</code>. The <code>none</code> state is the default state when the table is loaded by default.
```
MisRob: Maybe we could just leave out the second sentence here in favor of explaining it below in the already existing dedicated section for the default state since there is more options?

Plus a few optional tweaks for clarity.

Something like:

```suggestion
       Clicking the same header multiple times toggles the sort direction cyclically in the order of ascending, descending, and unsorted.
```

docs/pages/ktable.vue
```diff
@@ -210,8 +232,141 @@
       </DocsShow>
       <!-- eslint-enable -->
 
-    </DocsPageSection>
+      <!--Table with Default Sort-->
+      <h3>Table with Default Sort</h3>
+      <p>
+        This is an example to show how <code>KTable</code> can be used with the <code>defaultSort</code> attribute to sort the table based on a particular column. This is useful if the table needs to be sorted based on any column from the beginning itself (by default). The <code>defaultSort</code> attribute can be used irrespective of the <code>sortable</code> attribute (as <code>sortable</code> is used to configure whether the client has sorting capabilities or not).
```
MisRob: Few tweaks that may help with clarity and conciseness

```suggestion
        This is an example to show how to use the <code>defaultSort</code> prop to sort the table based on a particular column upon the initial load. The <code>defaultSort</code> attribute can be used irrespective of the <code>sortable</code> attribute.
```

docs/pages/ktable.vue
```diff
@@ -210,8 +232,141 @@
       </DocsShow>
       <!-- eslint-enable -->
 
-    </DocsPageSection>
+      <!--Table with Default Sort-->
+      <h3>Table with Default Sort</h3>
```
MisRob: For the case consistency, more in another comment

```suggestion
      <h3>Table with default sort</h3>
```

docs/pages/ktable.vue
```diff
@@ -210,8 +232,141 @@
       </DocsShow>
       <!-- eslint-enable -->
 
-    </DocsPageSection>
+      <!--Table with Default Sort-->
```
MisRob: Can be cleaned up

docs/pages/ktable.vue
MisRob: > can be used irrespective of the <code>sortable</code> attribute

Relatedly, I wonder if bumping this whole section up on the documentation page, right between non-sortable and sortable examples, would help understanding that this applies in both modes.

Reason being that the documentation flow overall goes from non-sortable => sortable, increasingly revealing more complex scenarios.

docs/pages/ktable.vue
```diff
@@ -210,8 +232,141 @@
...
+      <p>
+        This is an example to show how <code>KTable</code> can be used with the <code>defaultSort</code> attribute to sort the table based on a particular column. This is useful if the table needs to be sorted based on any column from the beginning itself (by default). The <code>defaultSort</code> attribute can be used irrespective of the <code>sortable</code> attribute (as <code>sortable</code> is used to configure whether the client has sorting capabilities or not). 
+      </p>
+
+      <p>
+        The <code>defaultSort</code> attribute takes an object with two properties - <code>columnId</code> and <code>direction</code>. The <code>columnId</code> is the unique identifier of the column based on which the table should be sorted. The <code>direction</code> can be either <code>asc</code> or <code>desc</code>.
+      </p>
+
+      <p>
+        To make use of <code>defaultSort</code>, please ensure that the <code>disableBuiltinSorting</code> attribute is not set to <code>true</code>.
```
MisRob, line 246: Thank you for mentioning this :)

docs/pages/ktable.vue
```diff
@@ -210,8 +232,141 @@
...
+      </p>
+
+      <DocsShowCode language="html">
+        <h4>Sortable Table with Rows Sorted by 'Age' Column</h4>
```
MisRob: For the case consistency, more in another comment

```suggestion
        <h4>Sortable table with rows sorted by 'Age' column</h4>
```

docs/pages/ktable.vue
```diff
@@ -210,8 +232,141 @@
...
+        <KTable
+          :headers="headers"
+          :rows="rows"
+          caption="Sortable Table with Rows Sorted by 'Age' Column"
+          sortable
+          :defaultSort="{ columnId: 'age', direction: 'asc' }"
+        />
+
+        <h4>Unsortable Table with Rows Sorted by 'Age' Column</h4>
```
MisRob:
```suggestion
        <h4>Unsortable table with rows sorted by 'Age' column</h4>
```

docs/pages/ktable.vue
```diff
@@ -210,8 +232,141 @@
...
+
+      <p>
+        The <code>defaultSort</code> attribute takes an object with two properties - <code>columnId</code> and <code>direction</code>. The <code>columnId</code> is the unique identifier of the column based on which the table should be sorted. The <code>direction</code> can be either <code>asc</code> or <code>desc</code>.
+      </p>
+
+      <p>
+        To make use of <code>defaultSort</code>, please ensure that the <code>disableBuiltinSorting</code> attribute is not set to <code>true</code>.
+      </p>
+
+      <DocsShowCode language="html">
```
MisRob, line 249: Would you please split this into two `DocsShowCode` and two `DocsShow` and grouped them together logically? For consistency and it will help us with code toggling improvements that may come soon to KDS.

docs/pages/ktable.vue
```diff
@@ -210,8 +232,141 @@
...
+        <KTable
+          :headers="headers"
+          :rows="rows"
+          caption="Sortable Table with Rows Sorted by 'Age' Column"
+          sortable
+          :defaultSort="{ columnId: 'age', direction: 'asc' }"
+        />
+
+        <h4>Unsortable Table with Rows Sorted by 'Age' Column</h4>
+        <KTable
```
MisRob, line 260: Thanks a lot for this use-case, I think neither of us thought of that this could be applied to non-sortable table as well, so wonderful to see such an improvement, and wonderfully documented.

docs/pages/ktable.vue
```diff
@@ -210,8 +232,141 @@
...
+          :headers="headers"
+          :rows="rows"
+          caption="Unsortable Table with Rows Sorted by 'Age' Column"
+          :defaultSort="{ columnId: 'age', direction: 'asc' }"
+        />
+      </DocsShow>
+
+      <!-- Disable Builtin Sorting -->
+      <h3>Disable Builtin Sorting</h3>
```
MisRob:
```suggestion
      <h3>Disable built-in sorting</h3>
```

docs/pages/ktable.vue
```diff
@@ -210,8 +232,141 @@
...
+        <h4>Unsortable Table with Rows Sorted by 'Age' Column</h4>
+        <KTable
+          :headers="headers"
+          :rows="rows"
+          caption="Unsortable Table with Rows Sorted by 'Age' Column"
+          :defaultSort="{ columnId: 'age', direction: 'asc' }"
+        />
+      </DocsShow>
+
+      <!-- Disable Builtin Sorting -->
```
MisRob: Can be cleaned up

docs/pages/ktable.vue
```diff
@@ -210,8 +232,141 @@
...
+      <h3>Disable Builtin Sorting</h3>
+      <p>
+        You can make use of the <code>disableBuiltinSorting</code> attribute to disable all sorting functionality by the table component. This is useful when you want to display the data in a particular order or you want to define a custom sorting function.
```
MisRob: Very nice, thanks for taking care of documenting this. Few tweaks to improve with clarity:

```suggestion
        For <code>sortable</code> tables, you can use the <code>disableBuiltinSorting</code> prop to disable built-in sort function. This is useful when the table receives already sorted data, for example when sorting is done by backend or a custom sorting function outside the table. In this case, when one of the header sort buttons is clicked, the table won't sort the column itself, but only emit the <code>changeSort</code> event to notify the parent component to handle the sorting logic. The event contains column index of the header and the sort order in its payload.
```

docs/pages/ktable.vue
```diff
@@ -210,8 +232,141 @@
...
+      </DocsShow>
+
+      <!-- Disable Builtin Sorting -->
+      <h3>Disable Builtin Sorting</h3>
+      <p>
+        You can make use of the <code>disableBuiltinSorting</code> attribute to disable all sorting functionality by the table component. This is useful when you want to display the data in a particular order or you want to define a custom sorting function. 
+      </p>
 
+      <p>
+        You should not use this attribute if <code>sortable</code> is set to <code>false</code>. If <code>sortable</code> is set to <code>true</code>, then the table component will emit a <code>changeSort</code> event with column index of the header clicked and the sort order to notify the parent component to handle the sorting logic.
```
MisRob, line 316: I couldn't make a code suggesting on both blocks so note that this text was reviewed and tweaked in the comment above.

MisRob (commented):
@EshaanAgg I saw the docs and these updates are such an enrichment. As always, your work goes beyond all expectations :) `KTable` is shaping up so very nicely overall.

I used this opportunity to leave documentation suggestions. None of them are blocking by any means, and most of them you can commit right here in the UI.

Most importantly, regarding how `KTable` is supposed to work, I confirm this is exactly what I would expect at this stage. Thanks to both of you @BabyElias and @EshaanAgg clarifying this together.

@EshaanAgg let me know if there's something else I should do in this PR as a next step.

EshaanAgg:
Thanks a lot, @MisRob, for the kind words and the detailed review! I will update the PR with the suggested changes concerning the style guide and request a review then! Since the tests are not passing and I was able to iron out the reactivity issues, I don't feel any special next steps are needed as a part of this PR.

docs/pages/ktable.vue
EshaanAgg: Yes totally! Would help in ensuring consistency!

docs/pages/ktable.vue
EshaanAgg: My two cents on the same would be that it's better to have it after the introduction of the `sortable` attribute so that when someone is reading the documentation for the first time, he is first informed of both the "base" variants of the table (sortable and un-sortable), and then is informed about the additional functionalities and APIs that the component has. 

Having it placed between the non-sortable and sortable examples might feel like a jump, as the user isn't aware of the functionality of the `sortable` attribute until after he reads the `sortable` section. 

However, please let me know if you disagree on the same!

docs/pages/ktable.vue
EshaanAgg: Can you please elaborate on the same as to how would you like the nesting/grouping to be?

I had grouped them together in this manner intentionally; for I wanted to show two show a direct comparison between two tables which share the exact same data and the rest of the attributes, but only different in just one `sortable` attribute. 

I can split the example into two blocks, one for each example table (and each table having having it's own HTML `DocsShowCode` and javascript `DocsShowCode` blocks), but then I feel we wouldn't be able to as effectively convey the indented comparison (components with the exact same props and data).

docs/pages/ktable.vue
MisRob: I think you've got a good point, @EshaanAgg. Let's leave as is.

docs/pages/ktable.vue
MisRob: No worries, if that was intentional, we can keep it. I assumed it was random choice.

MisRob:
Lovely @EshaanAgg. Thank you. I will start more detailed code review next week. Seeing the documentation and tests I wouldn't expect any larger changes requested :)

MisRob:
Side note, when you were dealing with reactivity and tests, by any chance have you observed something that would resemble these issues I got stuck on?

- https://github.com/learningequality/kolibri-design-system/issues/802
- https://github.com/learningequality/kolibri-design-system/issues/801

lib/KTable/index.vue
```diff
@@ -175,16 +170,22 @@
...
         required: true,
         validator: function(value) {
-          return value.every(
-            header =>
-              ['label', 'dataType'].every(key => key in header) &&
-              ['string', 'number', 'date', 'undefined'].includes(header.dataType)
+          const uniqueColumnIds = new Set(value.map(h => h.columnId));
+
+          return (
+            uniqueColumnIds.size == value.length &&
```
MisRob, line 182: Very nice.

lib/KTable/index.vue
```diff
@@ -231,6 +224,33 @@
...
+
+          return (
+            ['columnId', 'direction'].every(key => key in value) &&
+            ['asc', 'desc'].includes(value.direction) &&
+            ['string', 'number'].includes(typeof value.columnId)
+          );
+        },
+      },
+      /**
+       * Disables all the sorting functionality provided by the component. This is useful when you want to define your own sorting logic. Refer to the examples above for more details.
```
MisRob: I think we could keep this aligned with the phrasing chosen for the docs example.
```suggestion
       * Disables built-in sort function. This is useful when you want to define your own sorting logic. Refer to the examples above for more details.
```

lib/KTable/index.vue
```diff
@@ -283,6 +303,23 @@
         return colIndex => this.sortable && this.headers[colIndex].dataType !== DATA_TYPE_OTHERS;
       },
     },
+    watch: {
```
MisRob, line 306: First, thanks for covering this!

I wonder if it'd be just fine to have it in `mounted` or similar and be run once. Since `columnId`, as far as we know, won't be dynamically obtained but defined by devs right in the code. That said, I am not sure if the performance  cost of the watcher is so significant. Any thoughts?

lib/KTable/index.vue
```diff
@@ -202,14 +203,6 @@
         type: String,
         required: true,
       },
-
-      /**
-       * Disables the default sorting when sortable is true. Facilitates integration with externally sorted data.
-       */
-      disableDefaultSorting: {
```
MisRob, line 209: Just a note for me we will need to mention `disableBuiltinSorting` -> `disableBuiltinSorting` renaming as a breaking change in the changelog.

lib/KTable/useSorting/__tests__/index.spec.js
MisRob, line 1: Very good coverage overall, thanks

lib/__tests__/KTable.spec.js
```diff
@@ -1,12 +1,26 @@
 import { mount } from '@vue/test-utils';
 import KTable from '../KTable';
 
+const assertTableContent = (wrapper, rows) => {
```
MisRob, line 4: That's helpful, thank you

MisRob (commented):
Thank you, @EshaanAgg, I have finished full code review. Nothing blocking.

lib/KTable/index.vue
EshaanAgg: I see! I guess it's a matter of taste. I felt that lifecycle hooks would not be a good place to put the validation logic, as the hooks are called only once, and thus, if the prop values change in the future, we will not get the same validations again. But I see cases in which people use code to generate column IDs as rare. 

As for the performance cost, I don't think that there would be any overhead practically (as if the props aren't changing, the watcher would never run again, and if they, so it will be run every time they change, which should be the expected behaviour!)

EshaanAgg:
> Thank you, @EshaanAgg, I have finished full code review. Nothing blocking.

:)


> Side note, when you were dealing with reactivity and tests, by any chance have you observed something that would resemble these issues I got stuck on?
> 
> * [Improve unit test coverage of KCardGrid and KCard #802](https://github.com/learningequality/kolibri-design-system/issues/802)
> * [Optimize KCardGrid's internal usage of watchers #801](https://github.com/learningequality/kolibri-design-system/issues/801)

I'm afraid not. The errors that I was facing were mostly due to the old versions of enzyme that we are using for testing, and me getting used to writing UI tests without using the Testing Library!

lib/KTable/index.vue
EshaanAgg: Another (not a strictly) breaking change would be the now mandatory requirement of `columnId` attribute in all the headers, if that may be good to mention in the changelog.

MisRob: Oh yes, definitely. Thank you

lib/KTable/index.vue
MisRob: Thanks @EshaanAgg. I think we can leave this way then.

MisRob:
> I'm afraid not. The errors that I was facing were mostly due to the old versions of enzyme that we are using for testing, and me getting used to writing UI tests without using the Testing Library!

@EshaanAgg nevermind, thank you :)

MisRob:
@EshaanAgg @BabyElias I am signing off soon for holidays till the end of this week, like many of us. I can take care of the changelog updates and then I think we could merge next week. Thanks for such a good chunk of work, again!

MisRob:
Apart from the `columnId` addition, this shouldn't impact any of the currently used logic in Kolibri from the user perspective, so I feel confident merging without the Kolibri QA. Also we will soon do QA on https://github.com/learningequality/kolibri-design-system/pull/804.

MisRob:
Merging, yey :)

MisRob (approved):

PR #825, opened 2024-11-13
More flexible KCard click event handling

MisRob:
## Description

`KCard` is updated for more flexible click event handling. Previously, its title text was always `router-link`, `to` prop was required, and the card would navigate to the target on click. The new behavior is:

- `click` event is always emitted on card click
- `to` prop is not required
- when `to` is provided, the title text is `router-link`, and the card navigates to the target
- when `to` is not provided, the title text is `span`, and the card doesn't navigate anywhere

Contains related a11y changes that increase screen reader reliability overall:

- Make title prop required and allow title customization via a scoped slot
- Ensure reliable screen readers announcements no matter whether card is link or no, and no matter whether the title is customized via the title slot.

#### Issue addressed

A new use-case in Studio that @akolson works on in https://github.com/learningequality/studio/pull/4803 where a side panel needs to be toggled on card click without navigating to a new URL.

Screenshot from 2024-11-13 16-33-50

## Changelog

  - **Description:** Make the `title` prop required
  - **Addresses:** A new use-case in Studio in https://github.com/learningequality/studio/pull/4803 where a side panel needs to be toggled on card click without changing URL.
  - **Components:** KCard
  - **Breaking:** yes
  - **Impacts a11y:** no
  - **Guidance:** Even if you use the `title` slot, pass the title text via the `title` prop.

  - **Description:** Change the `title` slot into a scoped slot
  - **Addresses:** A new use-case in Studio in https://github.com/learningequality/studio/pull/4803 where a side panel needs to be toggled on card click without changing URL.
  - **Components:** KCard
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:**  Consider using the slot's `textTitle` attribute to achieve more intuitive code when customizing the title.

  - **Description:** Emit `click` event when card is clicked.
  - **Addresses:** A new use-case in Studio in https://github.com/learningequality/studio/pull/4803 where a side panel needs to be toggled on card click without changing URL.
  - **Components:** KCard
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -

  - **Description:** Make `to` prop optional and when not provided, do not render the title text as `router-link` but rather as `span`.
  - **Addresses:** A new use-case in Studio in https://github.com/learningequality/studio/pull/4803 where a side panel needs to be toggled on card click without changing URL.
  - **Components:** KCard
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -

  - **Description:** Ensure reliable screen readers announcements no matter whether card is link or no, and no matter whether the title is customized via the title slot
  - **Products impact:** bugfix
  - **Addresses:** A new use-case in Studio in https://github.com/learningequality/studio/pull/4803 where a side panel needs to be toggled on card click without changing URL.
  - **Components:** KCard
  - **Breaking:** no
  - **Impacts a11y:** yes
  - **Guidance:** -

## Steps to test

- You could run the devserver, play around with `to` and `@click` on the [DocsKCard](https://github.com/learningequality/kolibri-design-system/blob/develop/docs/pages-components/DocsKCard.vue#L3-L11) and preview live example on `KCard` docs page.
- Is [the updated documentation section](https://deploy-preview-825--kolibri-design-system.netlify.app/kcard/#click-navigation) clear?
- Can you see any functional, markup, or a11y regressions?

## Testing checklist

## Reviewer guidance

### DIFF
--- a/lib/cards/KCard.vue
+++ b/lib/cards/KCard.vue
@@ -484,11 +484,12 @@
       onThumbnailError() {
         this.thumbnailError = true;
       },
-      clickHandler() {
+      clickHandler(event) {
         /**
-         * Emitted when a card is clicked
+         * Emitted when a card is clicked or pressed with enter.
+         * Contains the DOM event in the payload.
          */
-        this.$emit('click');
+        this.$emit('click', event);
         if (this.to) {
           this.$router.push(this.to);
         }
@@ -505,13 +506,13 @@
          */
         this.$emit('hover', e);
       },
-      onEnter() {
-        this.clickHandler();
+      onEnter(event) {
+        this.clickHandler(event);
       },
       onMouseDown() {
         this.mouseDownTime = new Date().getTime();
       },
-      onClick() {
+      onClick(event) {
         if (this.isSkeleton) {
           return;
         }
@@ -524,7 +525,7 @@
         // which is not typically interpreted as a click event so do not run
         // the click handler.
         if (mouseUpTime - this.mouseDownTime < 200) {
-          this.clickHandler();
+          this.clickHandler(event);
         } else {
           return;
         }
--- a/lib/cards/KCard.vue
+++ b/lib/cards/KCard.vue
@@ -17,9 +17,19 @@
       :class="['card-area', ...cardAreaClasses ]"
       :style="{ backgroundColor: $themeTokens.surface }"
     >
+      <!--
+        Utilizing `visuallyhidden`, `aria-labelledby`,
+        and `aria-hidden` to ensure:
+        - More reliable output for some screen readers
+          in both link and non-link cards
+        - Prevents undesired screen reader announcements
+          when title is customized via the `title` slot
+      -->
+      <span :id="`card-title-${_uid}`" class="visuallyhidden">
+        {{ title }}
+      </span>
       <component
         :is="headingElement"
-        v-if="title || $slots.title"
         class="heading"
         :style="{ color: $themeTokens.text }"
       >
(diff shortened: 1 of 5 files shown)


AllanOXDi (approved):
Looks good to me !Thanks Misha, maybe @akolson and @AlexVelezLl has something to say

lib/cards/KCard.vue
```diff
@@ -54,6 +55,40 @@
...
+          `aria-label` is needed in this mode otherwise some
+          screenreaders don't announce the text in KTextTruncator,
+          for mysterious reasons (likely some internal screenreader logic)
+          -->
+        <span
+          v-else
+          tabindex="0"
+          data-focus="true"
+          class="title"
+          :aria-label="title"
```
AlexVelezLl: Here `title` isnt allways necessarily defined, if we use `$slots.title` we wont get the announcement of any title, because this aria-label will be empty, and the slot will be surrounded by an `aria-hidden`. I think we should do an special management here and get the text of the slot to add this as aria-label when the title is being sent as a slot rather than as a prop.

lib/cards/KCard.vue
```diff
@@ -436,23 +472,26 @@
...
         this.thumbnailError = true;
       },
-      navigate() {
-        if (this.isSkeleton) {
-          return;
+      clickHandler() {
+        /**
+         * Emitted when a card is clicked
+         */
+        this.$emit('click');
```
AlexVelezLl: Its usually useful to emit the event object here too. I have found myself many times updating some KDS components because I need the event object, so I think it doent hurt if we emit it too since the beginning.

AlexVelezLl (changes requested):
Hi @MisRob! I just noted a couple of things I would like to discuss about :)

lib/cards/KCard.vue
MisRob: Thanks @AlexVelezLl, that's a good catch - I completely forgot there's two title pathways.

> get the text of the slot to add this as aria-label when the title is being sent as a slot rather than as a prop

I thought and played around it and it seems to be tricky. (1) I don't see a straightforward way to obtain it. E.g. for

```
<template #title>
  <span>title</span>
</template>
```

`this.$slots.title`  is an array of `VNode`s and I can't see any attribute on `VNode` that would give the text, neither I've been able to access HTML element. Perhaps you can notice something I missed?

(2) However in any case, any HTML content can be passed via the slot, and I think we can't be 100% assuming it would be suitable `aria-label`.

So I wonder if we should just add `ariaLabel` required prop? For the sake of simplicity and explicitness, we may consider requiring it even when the `title` is passed via a prop.

MisRob: Even though maybe requiring it when `title` prop is there is way too strict and would be an unnecessary hassle. Technically, it's not a problem to make customized check for only when the title slot is used. Will wait to hear from you, perhaps you will have opinion.

lib/cards/KCard.vue
MisRob: Ah definitely. Done.

lib/cards/KCard.vue
AlexVelezLl: In KOptionalText we get the slots text in this way: https://github.com/learningequality/kolibri-design-system/blob/39ad629331d74cf90622e86b3a6012b384718fd8/lib/KOptionalText.vue#L54-L74, I think we can combine these approaches, and have an ariaLabel prop, and have this `getSlotsText` as a fallback if we have a slots.title and dont have an aria-label. So for example if we have complex title slots we can use the ariaLabel prop to specifically set the aria-label as needed. But for the most basic use cases like

```
<template #title>
  <span>title</span>
</template>
```

We can live just with the `getSlotsText`.

MisRob: @AlexVelezLl resolved in a way we agreed on Slack

MisRob:
Thanks @AllanOXDi @AlexVelezLl, all feedback is resolved, changelog and PR description updated with new information. I tested screenreaders in all modes. As soon as we have some working features in Kolibri, we will loop in Radina for final confirmation (outside this PR though I'd suggest because there's way more a11y we'll be testing in the final products)

MisRob:
In the new docs example, please ignore the thumbnail overflow issue - I shoud have this fixed in another branch

AlexVelezLl (approved):
Thank you @MisRob! Looks good to me!

Issue #206, opened 2021-03-25
Documentation focus ring is nearly invisible in Firefox


indirectlylit:
## Product

Design system docs

## Expected behavior

Focus ring should be easily visible in all browsers

## Actual behavior

Visible in Chrome but not Firefox on macOS:

| Chrome | Firefox |
|--|--|
| ![2021-03-25 09 10 09](https://user-images.githubusercontent.com/2367265/112505832-649c6700-8d4a-11eb-96e2-3c9ea9218c07.gif) | ![2021-03-25 09 10 25](https://user-images.githubusercontent.com/2367265/112505845-66fec100-8d4a-11eb-8625-696cb6f62ccb.gif) |

nikkuAg:
Is this issue still open? I would like to work on this

MisRob:
Hi @nikkuAg, since we haven't seen activity here for a long time, I will assign you, If you're still interested, you're welcome to messages us and we can assign again. Be well

RONAK-AI647:
Hi @indirectlylit @AlexVelezLl I want an opportunity to work on it , will you please assign it to me ,thank you.

AlexVelezLl:
Hey @RONAK-AI647! For sure! I am assigning this to you. Let us know if you have any question :)

RONAK-AI647:
Hey team members @AlexVelezLl @rtibbles @indirectlylit ,there are some queries in my head ,would be pleased if I get suggestions.

**1)I don't see any focus ring in my system when hover around the documentation.**

https://github.com/user-attachments/assets/63ced118-9a09-437e-bc90-bf037ccb88ba


**2) If the first one is my system error ,then increasing the width of the focus ring is sufficient???
3) Are you sure this is not individual's browser issue? In my case both the browsers gives the same result.**

AllanOXDi:
Hi @RONAK-AI647 

       > 1)I don't see any focus ring in my system when hover around the documentation.
Could you try navigating using a tab key?

RONAK-AI647:
@indirectlylit @MisRob kindly take a look at my PR .

MisRob:
Thanks @RONAK-AI647!

MisRob:
Closed by https://github.com/learningequality/kolibri-design-system/pull/827

PR #827, opened 2024-11-14
Ensure visibility of focus-ring in Firefox

RONAK-AI647:
## Description
Making sure the documentation focus-ring is visible in Firefox(earlier it was only visible in chrome).
#### Issue addressed #206 

Addresses #*PR# HERE*#827

### Before/after screenshots

https://github.com/user-attachments/assets/ceda40b1-2325-440c-9f1d-a151f0695e09

## Changelog

  - **Description:** Ensure visibility of focus-ring on the KDS website in Firefox
  - **Products impact:** -
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/206
  - **Components:** -
  - **Breaking:** -
  - **Impacts a11y:** -
  - **Guidance:** -

## Steps to test

1. Step 1
2. Step 2
3. ...

## Testing checklist

## Reviewer guidance

## Comments
I think now its visible in both browsers clearly,

### DIFF
--- a/docs/assets/main.scss
+++ b/docs/assets/main.scss
@@ -50,8 +50,10 @@ a {
   }
 }
 
-*:focus {
+*:focus-visible {
   outline-color: #757575;
+  outline:5px solid #757575; 
+  outline-offset: 2px;  
 }
 
 *::selection {
--- a/docs/assets/main.scss
+++ b/docs/assets/main.scss
@@ -51,8 +51,8 @@ a {
 }
 
 *:focus-visible {
-  outline:5px solid #757575; 
-  outline-offset: 2px;  
+  outline: 5px solid #757575;
+  outline-offset: 2px;
 }
 
 *::selection {
--- a/docs/assets/main.scss
+++ b/docs/assets/main.scss
@@ -51,7 +51,6 @@ a {
 }
 
 *:focus-visible {
-  outline-color: #757575;
   outline:5px solid #757575; 
   outline-offset: 2px;  
 }


RONAK-AI647:
@MisRob ,please help me here, I don't know whats wrong with my linter, it always fails here , in my local server it shows "done in secs."

docs/assets/main.scss
```diff
@@ -50,8 +50,10 @@ a {
   }
 }
 
-*:focus {
+*:focus-visible {
   outline-color: #757575;
+  outline:5px solid #757575;
```
MisRob: Would you remove line 54? Because we already have `outline: 5px solid #757575;` above and `outline`  is a shorthand for `outline-width` `outline-style` `outline-color`, having `outline-color` defined separately is duplicate.

MisRob (commented):
Thanks @RONAK-AI647! I confirm this works well in both Firefox and Chrome, and the change from `focus` to `focus-visible` have further benefits. Just one cleanup note.

Regarding the linting trouble, I am fine to lint and push before we merge, however, would you mind doing a recording of what you are experiencing so I can understand better? Perhaps we need to fix something.

On my end, it seems to work, this is what I see:

Screenshot from 2024-11-15 20-08-17

docs/assets/main.scss
RONAK-AI647: ok

RONAK-AI647:
When i run the yarn lint-fix command in my terminal ,this is the output


https://github.com/user-attachments/assets/3ba47137-7495-4e6d-9fef-ff33c8eb468a

Here I thought I might have the browser update issue , but that warning shows in your terminal as well  @MisRob .
The linter test yet does not pass here. what should i do?

MisRob:
> When i run the yarn lint-fix command in my terminal ,this is the output

That's interesting, @RONAK-AI647. I don't know what's the cause. Would you like to open an issue so we can start tracking it? I noticed some other contributors facing similar troubles. I'd be grateful if you could include recording or description of the following

(1) Do some change that a linter should fix (e.g. remove `;`)
(2) Then run `yarn lint-fix`
(3) Then show `git status`

It'd be great if you could be as specific as possible and include what change should be fixed and wasn't. Also information about your OS and `yarn` version.

MisRob:
Regarding this PR, I will do the fix and then merge. Thanks a lot :)!

MisRob (approved):

PR #857, opened 2024-12-08
[DRAFT] Explore refactoring KBreadcrumbs to use KListWithOverflow

### LINKED ISSUE #693: Refactor KBreadcrumbs to utilize KListWithOverflow

shruti862:
## Description
Updated lib/KListWithOverflow.vue and lib/KBreadcrumbs.vue files

#### Issue addressed
Addresses issue [#693 ](https://github.com/learningequality/kolibri-design-system/issues/693)

Addresses #*PR# HERE*

## Changelog

## Steps to test

1. Step 1
2. Step 2
3. ...

## Testing checklist

## Reviewer guidance


MisRob:
Converting to draft https://github.com/learningequality/kolibri-design-system/issues/693#issuecomment-2526877826

MisRob:
@shruti862 Also referencing @AlexVelezLl's comments that touches on some areas you mentioned, perhaps it helps https://github.com/learningequality/kolibri-design-system/issues/693#issuecomment-2354513487

MisRob:
@AlexVelezLl Hi Alex, @shruti862 continues this work and already followed some of your guidance. @shruti862 experienced some blockers and I asked to elaborate in this PR on some specific places. After that, would you please be available for giving a direction?

MisRob:
@AlexVelezLl Hi Alex, @shruti862 continues this work and already followed some of your guidance. @shruti862 experienced some blockers and I asked to elaborate in this PR on some specific places. Would you please be available for giving a direction?

https://github.com/learningequality/kolibri-design-system/pull/857/commits/cd730f1821ceef5becb428b01d392dfdf1ae546c

shruti862:
Screenshot from 2024-12-09 14-08-55
1.) Unable to get separators '>' between the items of list
2.) In the dropdown menu ,only the overflowed text is visible, link is not being applied.

AlexVelezLl:
Thanks @shruti862! I will take a look later today :)

lib/KListWithOverflow.vue
```diff
@@ -274,7 +284,15 @@
   }
 
   .more-button-wrapper {
-    visibility: hidden;
+    visibility: visible;
+    order: 0; /* Ensure it remains in its original position for keyboard navigation */
```
AlexVelezLl: We cant have this solution for the position of the button, because Keyboard navigation would not be the same. We need to keep the current behaviour in KBreadcrumbs that if the more button appears at the beginning, the keyboard navigation should focus it first, before the rest items.

lib/KListWithOverflow.vue
```diff
@@ -274,7 +284,15 @@
   }
 
   .more-button-wrapper {
-    visibility: hidden;
+    visibility: visible;
```
AlexVelezLl: Hmm interested to know why is this change needed? The more button starts as hidden because we measure its width first, before doing the next computations, so to avoid it being visible for that first render we initialize its visibility as hidden.

lib/KListWithOverflow.vue
```diff
@@ -129,6 +137,8 @@
        */
       setOverflowItems() {
         const { list, listWrapper, moreButtonWrapper } = this.$refs;
+
+        // Exit early if necessary refs are not available
```
AlexVelezLl: Lets remove these comments, as the code itself is descriptive

lib/KListWithOverflow.vue
```diff
@@ -146,20 +156,19 @@
...
+        const indexSequence = [...Array(list.children.length).keys()];
+        const directionIndexes =
+          this.overflowDirection === 'start' ? indexSequence.reverse() : indexSequence;
         const overflowItemsIdx = [];
-        for (let i = 0; i < list.children.length; i++) {
-          const item = list.children[i];
+        directionIndexes.forEach(i => {
           const itemWidth = itemsSizes[i].width;
-
-          // If the item dont fit in the available space or if we have already
```
AlexVelezLl, line 162: Lets get back these comments that describes a behaviour a little bit more complex

lib/KListWithOverflow.vue
```diff
@@ -274,7 +284,15 @@
...
 
   .more-button-wrapper {
-    visibility: hidden;
+    visibility: visible;
+    order: 0; /* Ensure it remains in its original position for keyboard navigation */
+  }
+
+  /* When the 'start-button' class is added, position visually at the start */
+  .more-button-wrapper.start-button {
+    order: -1; /* Moves visually to the start */
```
AlexVelezLl: This is also causing that when we dont have a more button, we see the items aligned to the right, which is an error 
image

lib/KListWithOverflow.vue
```diff
@@ -254,17 +264,17 @@
...
+    justify-content: space-between; 
     width: 100%;
   }
+
   .list {
     overflow: visible;
     display: flex;
     flex-wrap: wrap;
-    position: relative;
+
```
AlexVelezLl: Why is this change needed?  This list has a position relative to be relative to the hidden elements that are positioned as absolute.

lib/KBreadcrumbs.vue
```diff
@@ -1,381 +1,99 @@
...
-              :style="{ maxWidth: lastBreadcrumbMaxWidth }"  
-            >
-              {{ crumb.text }}
-            </span>
-          </li>
+          <template #text="{ text }">
+            <span class="breadcrumbs-crumb-text">{{ text }}</span>
+          </template>
+        </KRouterLink>
+        <span v-else>{{ item.text }}</span>
```
AlexVelezLl: Just a heads up that we need to keep the same styles as before as much as possible.

For example, before the text weight used to look like this: 
image

Now its weight look like

image

lib/KBreadcrumbs.vue
```diff
@@ -1,381 +1,99 @@
...
+
+    <!-- Render the "more" overflow dropdown and its separator -->
+    <template #more="{ overflowItems }">
+      <KIconButton
+        size="small"
+        icon="chevronDown"
+        appearance="raised-button"
+      >
+        <template #menu>
+          <!-- issue :: dropdown menu shows only text of the overflowItems but not the link , I don't know whether i am doing the right way to reference link or not -->
```
AlexVelezLl: For this you can add these lines in KDropdownMenu: 

https://github.com/shruti862/kolibri-design-system/blob/cd730f1821ceef5becb428b01d392dfdf1ae546c/lib/KDropdownMenu.vue#L20-L25

```vue
      <UiMenu 
        ref="menu" 
        :options="options" 
        :hasIcons="hasIcons" 
        @select="handleSelection" 
      >
        <template #option="{ option }">
          <slot name="option" :option="option"/>
        </template>
      </UiMenu>
```
And then you will be able to customize these dropdown options using this option slot in KDropdownMenu. There we can handle the link behaviour

lib/KBreadcrumbs.vue
```diff
@@ -1,381 +1,99 @@
...
-            }
-          });
+      preparedCrumbs() {
+        const crumbs = [...this.items];
+        if (!this.showSingleItem && crumbs.length <= 1) {
+          return [];
         }
+        // Add separators between items
+        return crumbs.flatMap((item, index) =>
+          index > 0 ? [{ type: 'separator' }, item] : [item]
```
AlexVelezLl: The reason the dividers are not being visible is because we are passing this as `type: "separator"` while the correct name is `type: "divider"`

lib/KBreadcrumbs.vue
```diff
@@ -1,381 +1,99 @@
...
-    margin-top: 8px;
-    margin-bottom: 8px;
-    font-size: 16px;
-    font-weight: bold;
-    line-height: 32px;
-    white-space: nowrap;
+<style scoped>
+  .breadcrumbs-divider {
+    margin: 0 10px;
+    color: black; /* Adjust color for the ">" separator */
```
AlexVelezLl: Lets standarize the name, and lets call it always "divider" instead of "separator"

lib/KBreadcrumbs.vue
```diff
@@ -1,381 +1,99 @@
...
+      <KIconButton
+        size="small"
+        icon="chevronDown"
+        appearance="raised-button"
+      >
+        <template #menu>
+          <!-- issue :: dropdown menu shows only text of the overflowItems but not the link , I don't know whether i am doing the right way to reference link or not -->
+          <KDropdownMenu
+            :options="overflowItems
+              .filter(item => item.type !== 'separator') // Filter out separators
```
AlexVelezLl: idem "divider" instead of separator

lib/KListWithOverflow.vue
```diff
@@ -254,17 +264,17 @@
 
 
 <style scoped>
-
   .list-wrapper {
     display: flex;
-    justify-content: space-between;
+    justify-content: space-between;
```
AlexVelezLl: Lets add a `alignItems: center` instruction here, so the default behaviour is that the items are centered

lib/KBreadcrumbs.vue
AlexVelezLl: Also, lets not use burned colors. Instead we can use our [color tokens](https://design-system.learningequality.org/colors/#tokens), we cant reference these in css, so we will need to set this color in the vue template

AlexVelezLl (changes requested):
Thank you @shruti862!! This is a great start! I have left some instructions to solve the issues you were experiencing, along with some other things I noticed. Thank you!!

lib/KBreadcrumbs.vue
shruti862: @AlexVelezLl   in new commits I tried to keep the same styles.

shruti862:
Screenshot from 2024-12-12 19-01-41
@AlexVelezLl 
I have fixed the keyboard navigation problem and alignment of list items when there is no more button in KListWithOverflow .
Now I am only facing problem in KBreadcrumb , my last visible item of list is getting outside of box even if there is enough space 
left in list-wrapper box. Can you please help me resolving this issue?
Screenshot from 2024-12-12 19-03-45
One more thing I noticed that list items having links differ from whats in the https://design-system.learningequality.org/kbreadcrumbs/ page ,

shruti862:
Screenshot from 2024-12-12 19-12-52
when I mounted the items array , link is present on list tems which is not in the  https://design-system.learningequality.org/kbreadcrumbs/ page

lib/KListWithOverflow.vue
```diff
@@ -15,17 +28,19 @@
           v-if="isDivider(item)"
           name="divider"
           :divider="item"
-        >
-        </slot>
-        <!-- @slot Slot responsible of rendering each item in the visible list -->
+        ></slot>
+        <!-- Item Slot -->
```
AlexVelezLl: Lets gonna revert this change, this comment is useful for KDS docs:
```suggestion
        <!-- @slot Slot responsible of rendering each item in the visible list -->
```

lib/KListWithOverflow.vue
```diff
@@ -277,7 +299,14 @@
   }
 
   .more-button-wrapper {
-    visibility: hidden;
+    visibility: visible;
+  }
+
+  /* When the 'start-button' class is added, position visually at the start */
+  .more-button-wrapper.start-button {
```
AlexVelezLl: I dont think this change is needed anymore, is it?

lib/KListWithOverflow.vue
```diff
@@ -153,15 +177,13 @@
           const itemSize = this.getSize(item);
```
AlexVelezLl, line 196: The issue you are facing with the list width not being calculated correctly is because this `getSize` method currently doesnt support items with margins. So to fix this, lets rewrite this `getSize` method to something like this:
```javascript
      getSize(element) {
        if (!element) {
          return { width: 0, height: 0 };
        }
        let { width, height } = element.getBoundingClientRect();

        const style = element.currentStyle || window.getComputedStyle(element);
        const marginX = parseFloat(style.marginLeft) + parseFloat(style.marginRight);
        const marginY = parseFloat(style.marginTop) + parseFloat(style.marginBottom);

        width += marginX;
        height += marginY;

        return { width, height };
      },
```

After this you wont see the list item in the next line when there is still space for it..

lib/KBreadcrumbs.vue
```diff
@@ -1,317 +1,109 @@
...
+      <!-- Render individual breadcrumb items -->
+      <template #item="{ item }">
+        <li>
+          <KRouterLink
+            v-if="item.link"
+            :text="item.text"
+            :to="item.link"
+          >
+
+            <template #text="{ text }">
```
AlexVelezLl: Be aware that `KRouterLink` doesnt have a #text slot. So we just need to use the default slot here instead

lib/KListWithOverflow.vue
```diff
@@ -153,15 +177,13 @@
...
           itemsSizes.push(itemSize);
         }
-
+        const indexSequence = [...Array(list.children.length).keys()];
```
AlexVelezLl, line 199: There are a couple of things that needs to be changed with this reversed behaviour:

1. The `fixDividersVisibility` method. The intention of this method is to avoid that a divider is placed as the first or last item in the list. And here we calculate this `lastVisibleIdx` https://github.com/shruti862/kolibri-design-system/blob/01aededc7ac60f49278ebf9ff15db89d8aec0581/lib/KListWithOverflow.vue#L248 which will need to change when we are in reversed mode because the lastVisibleIdx is not the same in both cases.
2. We have a couple of little bugs [here](https://github.com/shruti862/kolibri-design-system/blob/01aededc7ac60f49278ebf9ff15db89d8aec0581/lib/KListWithOverflow.vue#L251) and [here](https://github.com/shruti862/kolibri-design-system/blob/01aededc7ac60f49278ebf9ff15db89d8aec0581/lib/KListWithOverflow.vue#L212), because we change the visibility of the item, but we need to change also the position to be `position: absolute` if the item is hidden, and `position: unset` when the item is visible. So, althoguh its not a regression of your PR, fixing this will prevent some bugs that become more apparent when we work with the overflowDirection as "start".

lib/KBreadcrumbs.vue
```diff
@@ -323,77 +115,39 @@
...
-
-    position: absolute;
-    z-index: 8;
-    max-width: 100%;
-    padding: 16px;
-    font-weight: bold;
+  .breadcrumbs-divider {
+    margin-right: 8px;
+    margin-left: 8px;
+    color: #000000;
```
AlexVelezLl: Lets avoid burned color, and lets use the KDS color tokens instead https://design-system.learningequality.org/colors/#tokens-text

lib/KBreadcrumbs.vue
```diff
@@ -323,77 +115,39 @@
...
-    list-style: none;
-
-    .breadcrumbs-collapsed & {
-      // account for dropdown button width
-      width: calc(100% - 55px);
-    }
+    padding: 16px;
+    font-size: 16px;
+    font-weight: bold;
+    color: #4368f5;
```
AlexVelezLl: idem, lets avoid burned colors

lib/KBreadcrumbs.vue
```diff
@@ -323,77 +115,39 @@
...
-      content: '›';
-    }
-  }
-
-  .breadcrumbs-offscreen {
-    position: absolute;
-    left: -1000em;
+    padding: 16px;
+    font-size: 16px;
+    color: #000000;
```
AlexVelezLl: Idem

AlexVelezLl (changes requested):
Hi @shruti862! Thanks! This is looking so much better! I have made some notes that will help to solve the issues that you are experiencing :).

Apologies for the late response :sweat_smile:.

shruti862:
@AlexVelezLl Can you please review my current changes specially the fixDividersVisibility method ::Since the lastVisibleIdx in case of overflowdirection:'start' will always be last element of list.children because overflowIdx array is calculated after reversing the list.children for start direction so I didn't applied any checks on it.

shruti862:
@AlexVelezLl  But now i am facing issue on truncating the last visible item of the list. Could you please guide me how to approach this?? 
Right now the output is looking something like this::
Screenshot from 2025-01-05 23-42-37

lib/KBreadcrumbs.vue
```diff
@@ -1,317 +1,157 @@
...
 
-  <div
-    v-show="showSingleItem || crumbs.length > 1"
-    :class="{ 'breadcrumbs-collapsed': collapsedCrumbs.length }"
-  >
-    <nav
-      class="breadcrumbs"
-      v-bind="$attrs"
-      :aria-label="$attrs.ariaLabel"
+  <div class="breadcrumbs">
```
AlexVelezLl: instead of a div, we will need to have here a `<nav` instead, just as it was before.

lib/KBreadcrumbs.vue
```diff
@@ -1,317 +1,157 @@
...
+          icon="chevronDown"
           appearance="raised-button"
-          @click="showDropdown = !showDropdown"
-        />
-        <div
-          v-if="showDropdown"
-          class="breadcrumbs-dropdown"
-          :style="{ background: $themeTokens.surface }"
         >
-          <ol class="breadcrumbs-dropdown-items">
```
AlexVelezLl, line 27: We also need another change here. In the past, the list container was an `ol`` element, so we need to modify `KListWithOverflow` to receive a prop, lets say `listElement` or so, so we can indicate to the KListWithOverflow that the list should be an ol, so we can have the semantic structure ol > li

lib/KListWithOverflow.vue
```diff
@@ -217,16 +248,24 @@
         }
 
         const { list } = this.$refs;
-        const [firstOverflowedIdx] = overflowItemsIdx;
-        if (this.isDivider(this.items[firstOverflowedIdx])) {
-          overflowItemsIdx.shift();
-        }
+        if (this.overflowDirection === 'start') {
```
AlexVelezLl, line 281: To fix the divider visibility of the overflowed items, we dont need any change actually :sweat_smile:, so we can just have:
```javascript
        const [firstOverflowedIdx] = overflowItemsIdx;
        if (this.isDivider(this.items[firstOverflowedIdx])) {
          overflowItemsIdx.shift();
        }
```

lib/KListWithOverflow.vue
```diff
@@ -217,16 +248,24 @@
...
+          if (this.isDivider(this.items[firstOverflowedIdx])) {
+            overflowItemsIdx.shift();
+          }
 
-        const lastVisibleIdx = firstOverflowedIdx - 1;
-        if (this.isDivider(this.items[lastVisibleIdx])) {
-          const dividerNode = list.children[lastVisibleIdx];
-          dividerNode.style.visibility = 'hidden';
-          return itemsSizes[lastVisibleIdx].width;
+          const lastVisibleIdx = firstOverflowedIdx - 1;
```
AlexVelezLl, line 290: Here is where we need to do the change, because when we have an overflowDirection == start, then we will need to check the visibility of the first item instead. So we can do this instead:

```javascript
if (this.overflowDirection === 'start') {
  const firstVisibleIdx = firstOverflowedIdx + 1;
  if (this.isDivider(this.items[firstVisibleIdx])) {
    const dividerNode = list.children[firstVisibleIdx];
    dividerNode.style.visibility = 'hidden';
    dividerNode.style.position = 'absolute';
    return itemsSizes[firstVisibleIdx].width;
  }
} else {
  const lastVisibleIdx = firstOverflowedIdx - 1;
  if (this.isDivider(this.items[lastVisibleIdx])) {
    const dividerNode = list.children[lastVisibleIdx];
    dividerNode.style.visibility = 'hidden';
    return itemsSizes[lastVisibleIdx].width;
  }
}
```

In general this method should look like:

```javascript
      fixDividersVisibility(overflowItemsIdx, itemsSizes) {
        if (overflowItemsIdx.length === 0) {
          return;
        }

        const { list } = this.$refs;

        const [firstOverflowedIdx] = overflowItemsIdx;
        if (this.isDivider(this.items[firstOverflowedIdx])) {
          overflowItemsIdx.shift();
        }

        if (this.overflowDirection === 'start') {
          const firstVisibleIdx = firstOverflowedIdx + 1;
          if (this.isDivider(this.items[firstVisibleIdx])) {
            const dividerNode = list.children[firstVisibleIdx];
            dividerNode.style.visibility = 'hidden';
            dividerNode.style.position = 'absolute';
            return itemsSizes[firstVisibleIdx].width;
          }
        } else {
          const lastVisibleIdx = firstOverflowedIdx - 1;
          if (this.isDivider(this.items[lastVisibleIdx])) {
            const dividerNode = list.children[lastVisibleIdx];
            dividerNode.style.visibility = 'hidden';
            return itemsSizes[lastVisibleIdx].width;
          }
        }
      },
```

lib/KBreadcrumbs.vue
```diff
@@ -1,317 +1,157 @@
...
-            >
-              {{ crumb.text }}
-            </span>
-          </li>
-        </template>
-      </ol>
-    </div>
+            </KDropdownMenu>
+          </template>
+        </KIconButton>
```
AlexVelezLl: To avoid messing with the dividers visibility after the KIconButton, we can just add the divider besides the KIconButton from the more slot, since for this case, always that we show the more slot, will need to show the divider. So we can have just something like

```vue
      <template #more="{ overflowItems }">
        <div style="display: flex; align-items: center;">
          <KIconButton
            ...
          >
            ...
          </KIconButton>
          <span
            :style="{ color: $themeTokens.text }"
            class="breadcrumbs-divider"
          >›
          </span>
        </div>
      </template>
```

lib/KBreadcrumbs.vue
```diff
@@ -1,317 +1,157 @@
...
-            >
-              <template #text="{ text }">
-                <span class="breadcrumbs-crumb-text">{{ text }}</span>
-              </template>
-            </KRouterLink>
-            <span v-else>{{ crumb.text }}</span>
-          </li>
-
-          <li
-            v-else
```
AlexVelezLl, line 77: Be aware that the last breadcrumb should not be a link, but just a text.

Before:

image

After:

image

Here, `Level 2` shouldnt be a link

lib/KBreadcrumbs.vue
```diff
@@ -1,317 +1,157 @@
...
+      overflowDirection="start"
+      :items="preparedCrumbs"
     >
-      <div
-        v-show="collapsedCrumbs.length"
-        class="breadcrumbs-dropdown-wrapper"
-      >
+      <!-- Render individual breadcrumb items -->
+      <template #item="{ item,index }">
+        <li>
```
AlexVelezLl: We will also need to set the maxWidth that we set to the spans here to the li, for it to know what its its own max width, instead of getting it from the listOverflow container.

lib/KBreadcrumbs.vue
```diff
@@ -1,317 +1,157 @@
...
+          <KRouterLink
+            v-if="item.link"
+            :text="item.text"
+            :to="item.link"
+            dir="auto"
+            :title="item.text"
+          >
+            <template #default="{ text }">
```
AlexVelezLl: <KRouterLink> doesnt have a slot with props, and it doesnt pass the text to the slot, and this is making that we dont render the span with the `breadcrumbs-crumb-text` class. We need to have the slot defined just as `<template #default>` and inside render the text using item.text

AlexVelezLl (changes requested):
Hey @shruti862! Apologies for the late response, this is a big change and needed to find a considerable amount of time to review it well. I have left some suggestions here to fix the problems you are facing, and also some considerations we need to take into account. Let me know if you have any question :hugs:.

shruti862:
Hey @AlexVelezLl , I have done some changes to the code as you suggested , last visible item of list is no longer rendered as a link and its gets truncated when it does not fit in available space , after the changes it is looking like: 
Screenshot from 2025-02-23 20-37-34
But I encountered one more issue that on resizing the container last element gets into the dropdown menu instead of getting truncated :(
Screenshot from 2025-02-24 18-47-23
Can you guide me how to resolve this ??

AlexVelezLl:
Thanks @shruti862! I think for this you can have two options: 
1. Play with the `maxWidth` management of the last breadcrumb item.
2. Add a "showAtLeastOne" prop to KListWithOverflow that prevents from removing the last visible item.

In any case we need for find a better solution for (1) because when we have `showSingleItem = true` in KBreadcrumbs, we should 
not truncate the text. See for example if we had a long string in the first item:

Before:

image

After:

image

shruti862:
Hey @AlexVelezLl ,I have implemented some new changes to KBreadcrumbs. Now, when <code>showSingleItem</code> = true, the text is no longer truncated.
Screenshot from 2025-02-26 00-04-23
Regarding the last visible item not moving into the dropdown menu, I tried implementing your suggestions, but they didn't seem to have any effect. I'm unable to pinpoint what I might be doing wrong.

Could you please review my code and provide feedback on any mistakes? Your insights would be greatly appreciated.

Thanks!

MisRob:
Hi @shruti862, thanks for all this work. I know this is a tricky issue.

> I have implemented some new changes to KBreadcrumbs. Now, when showSingleItem = true, the text is no longer truncated.

I am just chiming in to provide a bit of general contributing guidance on when encountering issues like this. Would you be able to debug it deeper to get a sense on why is that happening? I think after you have a good sense, that would be the most suitable point to discuss next steps.

A little example: _"when showSingleItem = true this doesn't work" --> "this is happening because this particular method does this and this" --> "I could approach it in this way, but that would have negative impact on this another area ... another alternative would be ... any thoughts?"_

Collaborative aspect is important, and it's great when developers can connect and work through challenges together, it's just that we'd need a bit more information. I think this will help to make contributing in any open-source project successful and review process smoother.

shruti862:
Hey @MisRob , I was just letting @AlexVelezLl to know  that previous sub-issue of last visible item when <code>showSingleItem</code>=true being truncated is solved now.
Regarding the other sub-issue I am facing ,I appreciate the guidance  you provided on approaching issues more systematically. I'll take a deeper look into why this is happening and try to pinpoint the root cause.
Thanks !

MisRob:
Ah I totally misunderstood this particular note @shruti862, apologies. Thanks for mentioning it. Yes it was just meant to be a [wrongly chosen] example that nevertheless applies to other issues throughout this work - I think we're on the same page. Thank you and keep up the great work here :)

lib/KListWithOverflow.vue
```diff
@@ -202,7 +237,80 @@
...
+            availableWidth -= itemWidth;
+            const itemHeight = itemsSizes[i].height;
+            if (itemHeight > maxHeight) {
+              maxHeight = itemHeight;
+            }
+          }
+        });
+
+        if (this.showAtLeastOne && overflowItemsIdx.length === itemsSizes.length) {
+          const lastItemIdx = directionIndexes[directionIndexes.length - 1];
```
shruti862: Hey @AlexVelezLl ,for the prop <code>showAtLeastOne</code> , I thought to remove last overflowItem from overflowItems array and also set its visibility to 'visible'.
But I noticed that its visibility is still get set to 'hidden' when available space is less than last item space

lib/KBreadcrumbs.vue
```diff
@@ -1,317 +1,181 @@
...
-  >
-    <nav
-      class="breadcrumbs"
-      v-bind="$attrs"
-      :aria-label="$attrs.ariaLabel"
+  <nav class="breadcrumbs">
+    <KListWithOverflow
+      overflowDirection="start"
+      :items="preparedCrumbs"
+      :showAtLeastOne="true"
```
shruti862, line 7: passed the <code>showAtLeastOne</code> prop as true

shruti862:
As the container size decreases and can no longer accommodate the last visible item, all items, including the "More" button, become hidden. However, the visibility of the "More" button remains set to 'visible'.
[Screencast from 26-02-25 06:52:41 PM IST.webm](https://github.com/user-attachments/assets/99780534-f60b-429e-b93e-1e7f9777ff90)

MisRob:
> As the container size decreases and can no longer accommodate the last visible item, all items, including the "More" button, become hidden. However, the visibility of the "More" button remains set to 'visible'.

Hi @shruti862, thanks for the recording - I would expect the More button never disappear if it has any items (which is this case).

lib/KBreadcrumbs.vue
AlexVelezLl: @shruti862 This havent been yet solved, please lets update this KRouterLink to this:
```Vue
          <KRouterLink
            v-if="item.link && index !== preparedCrumbs.length - 1"
            :to="item.link"
          >
            <span
              class="breadcrumbs-crumb-text"
              :style="{
                maxWidth: index === preparedCrumbs.length - 1 ? lastBreadcrumbMaxWidth : 'none',
              }"
              dir="auto"
            >
              {{ item.text }}
            </span>
          </KRouterLink>
```

lib/KListWithOverflow.vue
```diff
@@ -188,6 +222,7 @@
             const idx = overflowItemsIdx.pop();
             const item = list.children[idx];
             item.style.visibility = 'visible';
+            item.style.position = 'unset';
             maxWidth += itemsSizes[idx].width;
```
AlexVelezLl, line 236: We also need to add this here: maxHeight = Math.max(maxHeight, itemsSizes[idx].height); to update the max height accordingly

lib/KListWithOverflow.vue
AlexVelezLl: The correct condition for this should be something like:
```javascript
        if (this.showAtLeastOne && overflowItemsIdx.length === itemsSizes.length) {
          const firstHiddenIdx = overflowItemsIdx.shift();
          const firstHiddenItem = list.children[firstHiddenIdx];
          firstHiddenItem.style.visibility = 'visible';
          firstHiddenItem.style.position = 'unset';
          maxWidth = availableWidth;
          maxHeight = Math.max(maxHeight, itemsSizes[firstHiddenIdx].height);
        }
```

This should work :)

Although we still need to figure out a better way to handle the text truncation since with this we now need to show the text truncation for widths shorter than 300px too.

lib/KListWithOverflow.vue
AlexVelezLl: (Also, lets uncomment this part and work on top of the current implementation ;)

AlexVelezLl (commented):
Hi @shruti862! I have left some comments, I hope this helps :) We are very close to get it!

shruti862:
> Hi @shruti862! I have left some comments, I hope this helps :) We are very close to get it!

Hey @AlexVelezLl, I appreciate your guidance and support in resolving this issue. I'll go through the changes you suggested and provide an update soon.

lib/KBreadcrumbs.vue
shruti862: Hey @AlexVelezLl , updating KRouterLink as you suggested is removing underline from link text.
So,instead of using default slot ,I have used text slot in new commit.

shruti862:
Hey @AlexVelezLl, I've implemented all the suggested changes. However, I noticed one issue — the last item in the list doesn't show the '...' on text overflow.
image

AlexVelezLl:
Hi @shruti862, It seems there are still some unresolved comments. Please take a look to all my comments in this view https://github.com/learningequality/kolibri-design-system/pull/857/files and make sure you have considered/responded/implemented all of them :). Thanks!

shruti862:
> Hi @shruti862, It seems there are still some unresolved comments. Please take a look to all my comments in this view https://github.com/learningequality/kolibri-design-system/pull/857/files and make sure you have considered/responded/implemented all of them :). Thanks!

Hey @AlexVelezLl , Currently my end semester exams are going on. I will get back to this issue within 2 days :)

AlexVelezLl:
Hey @shruti862! Thanks for letting us know! Please take your time, no rush! Good luck with your exams!

lib/KBreadcrumbs.vue
shruti862: Done :)

lib/KListWithOverflow.vue
shruti862: considered

lib/KListWithOverflow.vue
shruti862: done:)

lib/KListWithOverflow.vue
shruti862: done

lib/KListWithOverflow.vue
shruti862: done

lib/KBreadcrumbs.vue
shruti862: Hey @AlexVelezLl , Since I have done so many recent changes in KBreadcrumb I am confused where to incorporate this change, could you please guide me regrading this in my current implementation

shruti862:
Hey @AlexVelezLl , I have tried to answer all the unresolved comments, please do have a look

MisRob:
Hi @shruti862, in the course of this work, we've noticed some unexpected limitations of `KListWithOverflow` and there's need for some decisions before we can complete this refactor. As you're now working on your GSoC project, I thought it'd be best to let you focus on that, and re-assign this to @AlexVelezLl who will explore what would be possible next steps. Thanks a lot to both!

PR #858, opened 2024-12-09
make ui menu accept maxWidth to avoid text truncation

lokesh-sagi125:
## Description
override the max-width set by this component [UiMenu](https://github.com/learningequality/kolibri-design-system/blob/777ad2f2cbdef9d180550aed431c9550400c9d28/lib/keen/UiMenu.vue#L5). And for this, we will need to update our [KDropdownMenu](https://github.com/learningequality/kolibri-design-system/blob/777ad2f2cbdef9d180550aed431c9550400c9d28/lib/KDropdownMenu.vue#L1) component to accept a new prop maxWidth that will then be passed to the UiMenu component.

#### Issue addressed
https://github.com/learningequality/kolibri/issues/5557

Addresses  https://github.com/learningequality/kolibri/issues/5557

## Changelog

  - **Description:** remove text truncation for the download button
  - **Products impact:** Choose from - none
  - **Addresses:** https://github.com/learningequality/kolibri/issues/5557
  - **Components:** KDropdownMenu,UiMenu
  - **Breaking:** no
  - **Impacts a11y:** yes
  - **Guidance:** -

PR #870, opened 2024-12-11
make ui menu accept maxWidth to avoid text truncation 

lokesh-sagi125:
## Description
override the max-width set by this component [UiMenu](https://github.com/learningequality/kolibri-design-system/blob/777ad2f2cbdef9d180550aed431c9550400c9d28/lib/keen/UiMenu.vue#L5). And for this, we will need to update our [KDropdownMenu](https://github.com/learningequality/kolibri-design-system/blob/777ad2f2cbdef9d180550aed431c9550400c9d28/lib/KDropdownMenu.vue#L1) component to accept a new prop maxWidth that will then be passed to the UiMenu component.

#### Issue addressed
https://github.com/learningequality/kolibri/issues/5557

Addresses  https://github.com/learningequality/kolibri/issues/5557

## Changelog

  - **Description:** remove text truncation for the download button
  - **Products impact:** Choose from - none
  - **Addresses:** https://github.com/learningequality/kolibri/issues/5557
  - **Components:** KDropdownMenu,UiMenu
  - **Breaking:** no
  - **Impacts a11y:** yes
  - **Guidance:** -

### DIFF
--- a/lib/KDropdownMenu.vue
+++ b/lib/KDropdownMenu.vue
@@ -21,6 +21,7 @@
         ref="menu"
         :options="options"
         :hasIcons="hasIcons"
+        :maxWidth="maxWidth"
         @select="handleSelection"
       />
     </UiPopover>
@@ -82,6 +83,11 @@
       hasIcons: {
         type: Boolean,
         default: false,
+      },
+      /** Maximum width for the dropdown menu*/
+      maxWidth: {
+        type: [Number, String],
+        default: null,
       },
       /**
        * The position of the dropdown relative to the button
--- a/lib/keen/UiMenu.vue
+++ b/lib/keen/UiMenu.vue
@@ -8,6 +8,7 @@
     lazy
 
     :class="classes"
+    :style="getWidth"
   >
     <UiMenuOption
       v-for="(option, index) in options"
@@ -59,6 +60,10 @@
         type: Boolean,
         default: false,
       },
+      maxWidth: {
+        type: [Number,String],
+        default: null,
+      },
       iconProps: Object,
       hasSecondaryText: {
         type: Boolean,
@@ -100,6 +105,15 @@
           'has-icons': this.hasIcons,
           'has-secondary-text': this.hasSecondaryText,
         };
+      },
+      getWidth() {
+        if (this.maxWidth){
+          return {
+            maxWidth: this.maxWidth && 
+              (typeof this.maxWidth === 'number' ? `${this.maxWidth}px` : this.maxWidth)
+          };
+        }
+        return {};
       },
       activeOutline() {
         return this.isActive ? this.$coreOutline : {};


MisRob:
Hi @lokesh-sagi125, would you upload some screenshots of how the output would look like for https://github.com/learningequality/kolibri/issues/5557? Also, what changes you did in Kolibri after having this KDS update available to achieve that - do you have a Kolibri working branch you could reference?

At first sight, I don't understand the connection of this work to the problem with the file size being pushed out - this would help me to connect the dots.

MisRob:
Oh I just noticed @AlexVelezLl's https://github.com/learningequality/kolibri-design-system/pull/858#issuecomment-2536238511. Alright, it seems fine (generally I'd recommend you always upload screenshots to pull requests so reviewers don't need to do the detective work :)

One question I had is if the title part gets truncated? It's correct that the title takes less space so the file size is always visible, but if the whole of the title is not visible, we should still show the ellipsis ...

AlexVelezLl:
Oh @MisRob, In this case we should show the whole string without it being truncated because this string is not a user string, but a system string. In this context the string is showing the different files from a resource that can be downloaded (e.g. "Document", "thumbnail", "HTML"). So, if we truncate this string, the message could be a little confusing, and the user might not be able to understand what is they downloading, so thats why Blaine suggested that the better option here is to show the full message and avoid truncating a system string:

> If the text isn't user generated, I don't think there should be text truncation on that at all, because that calls into question the usefulness of the text, if it's being truncated in some languages.

MisRob:
I see @AlexVelezLl, so can I understand right that the string will take more lines in the case you describe? I think that'd be fine, the main thing is that the whole text is visible.

AlexVelezLl:
Hmm no @MisRob, the long string will be all placed in just one line like this, but will not get truncated:

image

MisRob:
Thanks @AlexVelezLl, I looked into it more and I think it should be good for the case we're trying to resolve.

lokesh-sagi125:
hey @MisRob, sorry for the inactivity I have made the changes in the code by testing the components individually, but I can't seem to load the resources into Kolibri like the ones in the production model, @AlexVelezLl has directed with the steps needed, I can't seem to get it to work, could you clarify this one doubt, am I supposed build Kolibri and run `kolibri start`  or run `devserver-with-kids` because when I run dev server, there are no resources being loaded, I keep getting the no resources found error when I try to explore

MisRob:
Hi @lokesh-sagi125, I don't exactly understand what's going on. Have you followed [this guide](https://kolibri-dev.readthedocs.io/en/develop/howtos/development_with_kds.html)? And were you able to run Kolibri [in the usual way](https://kolibri-dev.readthedocs.io/en/develop/getting_started.html), without local Kolibri Design System? If no, please open a new [topic in the Kolibri Discussions](https://github.com/learningequality/kolibri/discussions/categories/q-a) where I'd ask you to describe step by step what you did and upload as full terminal log as possible.

MisRob:
@AlexVelezLl Apart from testing in Kolibri, do code changes on KDS side look good to you?

lokesh-sagi125:
Yes, @MisRob, I can run Kolibri with and without the local Kolibri design system. However, in both cases, I encounter the same issue: I cannot import any channels or resources I could use for testing. I keep getting the "no resources available"
Screenshot 2025-01-22 at 9 36 02 AM
 this issue is the same for anything under the `learn` drop-down menu.(also for the navbar)

AlexVelezLl:
Hey @lokesh-sagi125! You can look at this for reference of how you can import resources to your Kolibri instance! https://kolibri.readthedocs.io/en/latest/manage/resources.html :open_hands:

AlexVelezLl:
> Apart from testing in Kolibri, do code changes on KDS side look good to you?

Yes @MisRob! This is the expected behaviour! With this, we can open a PR in Kolibri setting the maxWidth to `none` to close https://github.com/learningequality/kolibri/issues/5557

MisRob:
Alright, I think it'd be good to merge if you're fine with it @AlexVelezLl. Feel free to proceed. Thank you

MisRob (approved):

AlexVelezLl:
Thanks @MisRob! Merging :open_hands:, @lokesh-sagi125 would you be interested in opening a PR setting the maxWith="none" to the KDropdownMenu in Kolibri?

lokesh-sagi125:
yes @AlexVelezLl  , i am currently working on an issue from studio , will raise a PR once i open a PR for that :)

PR #877, opened 2024-12-20
Add KSelect dropdown to the Overlay layer

### LINKED ISSUE #324: Kmodal should be able to show dropdowns select without making scrollbars appear

### LINKED ISSUE #690: Investigate KSelect’s dropdown being displayed above the input in some cases even when there is enough space below
🌱 **Are you new to the codebase? Welcome! Please see [the contributing guidelines](https://github.com/learningequality/kolibri-design-system/blob/develop/CONTRIBUTING.md).**

## Summary

When there is enough space, `KSelect`'s dropdown should be displayed below the input. However, there are several places in [Kolibri](https://github.com/learningequality/kolibri) where the dropdown is displayed above the input, even though there's enough space below. One example from *Kolibri 0.17.0a0.dev0+git.123.g5cd1f6aa*:

_Coach → Plan → Lessons_

This issue is present in other selects in _Plan_ and _Reports_.

Interestingly, in a very similar place in _Facility → Users_, it works as expected:

The main culprit seems to be the `bottom` positioning set to `62px` on `.ui-select-dropdown`. After commenting it out, the dropdown shows correctly in the _Coach_:

The `bottom` value is calculated in the `calculateSpaceBelow` method:

[https://github.com/learningequality/kolibri-design-system/blob/0cbdf57d24690cb6c23618ba700118a831cd9ffc/lib/KSelect/index.vue#L878-L893](https://github.com/learningequality/kolibri-design-system/blob/0cbdf57d24690cb6c23618ba700118a831cd9ffc/lib/KSelect/index.vue#L878-L893)

The goal of this issue is to find out why exactly the calculation doesn't work well for places in the Coach (or check whether it’s overwritten from somewhere else, even though that’s rather unlikely). After we know more, we can determine the next steps and see if the calculation can be optimized.

## References

Originally reported in [https://github.com/learningequality/kolibri/pull/12079](https://github.com/learningequality/kolibri/pull/12079) (not related to the refactor being tested there - the bug was present before).

AlexVelezLl:
## Description
Finally, a robust solution is proposed to the many known issues with the KSelect dropdown by teleporting KSelect dropdown to the Overlay layer using Popper.

#### Issue addressed

Addresses #324 and #690.

### Before/after screenshots

| Before | After |
| --- | --- |
| image | image | 
| image | image |
| image | image |
| If a modal had a long form, it bugged because of the `overflowY: unset` set because the Modal check for KSelects | Now, we can have scrollables modals even if it has a KSelect inside |
| image | Screenshot from 2024-12-26 16-01-41 |

Screencast showing the behaviour of KSelect in a long modal:

https://github.com/user-attachments/assets/aea8ecd2-9d4c-437c-8762-be8a083f3f73

## Changelog

  - **Description:** Teleport KSelect dropdown to the Overlay layer using Popper.
  - **Addresses:**https://github.com/learningequality/kolibri-design-system/issues/324, https://github.com/learningequality/kolibri-design-system/issues/690.
  - **Components:** KSelect.
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -.

  - **Description:** Removes internal KModal calculations to modify its content height if it had a KSelect inside.
  - **Addresses:**https://github.com/learningequality/kolibri-design-system/issues/324.
  - **Components:** KModal.
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -.

## Steps to test

1. Step 1
2. Step 2
3. ...

I have decided to close the dropdown on scroll, just like the KDropdownMenu, to avoid situations like this:

image

Otherwise we would need to be querying scrollable ancestors elements, and taking into account the height of any absolute/fixed elements like the topbar to automaticlly close the dropdown.

## Testing checklist

## Reviewer guidance

### DIFF
--- a/lib/_Popper.vue
+++ b/lib/_Popper.vue
@@ -9,6 +9,7 @@
      - Allow for appending to a chosen element rather than to body,
        typically to the overlay container element #k-overlay.
        'appendToBody' prop changedto 'appendToEl' and related changes.
+     - Add 'closeOnScroll' prop to close the popper on scroll.
   -->
 
   <span>
@@ -220,12 +221,16 @@
         this.showPopper = false;
       },
 
+      // Debouncing as we don't need to check this on every scroll event on a
+      // "scroll session", just the first one, and although very unlikely, the last one
       doCloseOnScroll: debounce(
         function (event) {
           const popperEl = this.$refs.popper;
           if (
             !popperEl ||
+            // ignore scrolls inside the popper
             this.elementContains(popperEl, event.target) ||
+            // ignore scrolls unrelated to the popper reference, just from the ancestors
             !this.elementContains(event.target, this.referenceElm)
           ) {
             return;
--- a/lib/KModal.vue
+++ b/lib/KModal.vue
@@ -21,11 +21,7 @@
             :tabindex="0"
             role="dialog"
             aria-labelledby="modal-title"
-            :style="[
-              modalSizeStyles,
-              { background: $themeTokens.surface },
-              containsKSelect ? { overflowY: 'unset' } : { overflowY: 'auto' },
-            ]"
+            :style="[modalSizeStyles, { background: $themeTokens.surface }, { overflowY: 'auto' }]"
           >
             <!-- Modal Title -->
             <h1
@@ -65,7 +61,6 @@
                 ]"
                 :class="{
                   'scroll-shadow': scrollShadow,
-                  'contains-kselect': containsKSelect,
                 }"
               >
                 <!-- @slot Main content of modal -->
@@ -216,7 +211,6 @@
       return {
         lastFocus: null,
         maxContentHeight: '1000',
-        containsKSelect: false,
         scrollShadow: false,
         delayedEnough: false,
       };
@@ -278,9 +272,6 @@
       window.addEventListener('focus', this.focusElementTest, true);
       window.setTimeout(() => (this.delayedEnough = true), 500);
 
-      // if modal contains KSelect, special classes & styles will be applied
-      const kSelectCheck = document.querySelector('div.modal div.ui-select');
-      this.containsKSelect = !!kSelectCheck;
       this.updateContentSectionStyle();
     },
     updated() {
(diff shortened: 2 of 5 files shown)


lib/_Popper.vue
```diff
@@ -201,14 +207,42 @@
...
+            this.elementContains(popperEl, event.target) ||
+            // ignore scrolls unrelated to the popper reference, just from the ancestors
+            !this.elementContains(event.target, this.referenceElm)
+          ) {
+            return;
+          }
+
+          this.doClose();
+        },
+        100,
```
LianaHarris360, line 241: This is just out of curiosity, but how was the debounce time 100ms determined?

LianaHarris360 (changes requested):
Thank you for finding a solution to this problem! The code looks good and works correctly in most cases, except on the `MyDownloads.vue` page. The two dropdown menus aren't showing up, and there's an error message in the console related to the `k-overlay` element.

MydownloadsKSelects
Since there are only issues with these two, I'm not entirely sure if changes need to be made in Kolibri or KDS to fix it. I just wanted to mention it in case the problem can be solved in KDS, and if not, it might be considered a breaking change.

Just a quick note: since `KModal` no longer checks for `KSelect`, I believe we can remove the comment in `lib/KSelect/index.vue` that says, "BELOW: KModal targets KeenUiSelect by using div.ui-select selector" on line 11.

lib/_Popper.vue
AlexVelezLl: A 100ms delay is a common choice as it falls within the threshold of the shortest perceivable application response delay, as noted [here](https://stackoverflow.com/questions/536300/what-is-the-shortest-perceivable-application-response-delay). Although it will always depend on the context. If we were debouncing a user keyboard typing action, to lets say, trigger an api request after the user has finished typing we would rather use a >300 ms debounce delay depending on the context https://stackoverflow.com/questions/42361485/how-long-should-you-debounce-text-input.

AlexVelezLl:
Thanks @LianaHarris360! Yes, that was a problem because the KSelect's were being mounted before the `DOMContentLoaded` event was triggered and by that time, the KOverlay layer werent being mounted yet. This will be fixed when https://github.com/learningequality/kolibri-design-system/pull/879 is merged. Will ping Richard again since we had already agreed that this was the best solution :). I will then rebase this PR to have this fixed.

> Just a quick note: since KModal no longer checks for KSelect, I believe we can remove the comment in lib/KSelect/index.vue that says, "BELOW: KModal targets KeenUiSelect by using div.ui-select selector" on line 11.

Thanks! Just removed it :)

AlexVelezLl:
I have just rebased the branch with the fix @LianaHarris360! Could you take another look? :)

LianaHarris360 (approved):
The changes are working as they should and the code logic makes sense, thanks for your work on this! @AlexVelezLl

PR #783, opened 2024-09-16
Remove KResponsiveElementMixin

AlexVelezLl:
## Description

Make useKResponsiveElement public and remove KResponsiveElementMixin.

#### Issue addressed
Addresses #711.

## Changelog

  - **Description:** Removes KResponsiveElementMixin.
  - **Products impact:** removed API.
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/711.
  - **Components:** KResponsiveElementMixin.
  - **Breaking:** yes.
  - **Impacts a11y:** no.
  - **Guidance:** Replace any use of `KResponsiveElementMixin` with the new composable `useKResponsiveElement`.

  - **Description:** Adds new useKResponsiveElement composable.
  - **Products impact:** new API.
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/711.
  - **Components:** -.
  - **Breaking:** no.
  - **Impacts a11y:** no.
  - **Guidance:** -.

  - **Description:** Remove use of KResponsiveElementMixin internally
  - **Products impact:** - none
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/711.
  - **Components:** KBreadcrumbs, KFixedGrid.
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -.
 

## Steps to test

1. Look for any reference to KResponsiveElementMixin in the lib or in the docs.
2. Do KBreadcrumbs or KFixedGrid have any regressions?

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical and brittle code paths are covered by unit tests
- [x] The change is described in the changelog section above

## Reviewer guidance

- [ ] Is the code clean and well-commented?
- [ ] Are there tests for this change?
- [ ] Are all UI components LTR and RTL compliant (if applicable)?
- [ ] _Add other things to check for here_

PR #882, opened 2025-01-09
Add KDS Workflows to keep KDS Roadmap statuses updated

AlexVelezLl:
## Description
Adds two workflows to keep KDS Roadmap statuses updated:
1. Sync KDS Roadmap Project Statuses: Sets up a cronjob to keep KDS Roadmap project statuses synced with the Iteration backlog project items status. 
2. Update KDS Roadmap project statuses on Release: When we have a new kds release, this workflow will check all released PRs and will set their statuses to `Released` and will update the `released in` field to the release version.

## Changelog

  - **Description:** Adds Adds two workflows to keep KDS Roadmap statuses updated
  - **Products impact:** none.
  - **Addresses:** -.
  - **Components:** -.
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -.

## Steps to test

I have tested these actions in the https://github.com/learningequality/test-actions repo. We can dispatch:
* https://github.com/learningequality/test-actions/actions/workflows/sync_kds_roadmap_statuses.yml: For this, we will need to go to the KDS Roadmap project and simulate a situation where we have an updated status in the Iteration Backlog, but not in the KDS Roadmap project.
* https://github.com/learningequality/test-actions/actions/workflows/update_project_items_on_release.yml: For this, we will need to simulate a release adding to the context input an object like this:
  ```{ "payload": { "action": "published", "release": { "body": "Some release notes https://github.com/learningequality/kolibri-design-system/pull/783", "name": "v0.0.1"}, "repository": { "name": "kolibri-design-system", "owner": { "login": "learningequality" } } }, "eventName": "release"}```

### DIFF
--- a/.github/githubUtils.js
+++ b/.github/githubUtils.js
@@ -1,8 +1,5 @@
 const GithubAPI = require('./GithubAPI');
 
-const ITERATION_BACKLOG_PROJECT_NUMBER = 15;
-const KDS_ROADMAP_PROJECT_NUMBER = 29;
-
 /**
  * Will synchronize the KDS Roadmap project statuses synced
  * with the Iteration backlog project items statuses. So if an issue belongs to
@@ -11,9 +8,7 @@ const KDS_ROADMAP_PROJECT_NUMBER = 29;
  * 
  * All issues in KDS Roadmap that already had a status of "RELEASED" will be ignored.
  */
-const synchronizeProjectsStatuses = async (context, github) => {
-  const sourceNumber = ITERATION_BACKLOG_PROJECT_NUMBER;
-  const targetNumber = KDS_ROADMAP_PROJECT_NUMBER;
+const synchronizeProjectsStatuses = async ({context, github, sourceNumber, targetNumber}) => {
   const getTargetStatus = (sourceStatus) => {
     const statusMap = {
       "IN REVIEW": "IN REVIEW",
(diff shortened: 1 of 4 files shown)


.github/githubUtils.js
```diff
@@ -0,0 +1,206 @@
+const GithubAPI = require('./GithubAPI');
+
+const ITERATION_BACKLOG_PROJECT_NUMBER = 15;
```
rtibbles: I think these might be better encoded as org and repo secrets.

rtibbles (approved):
I always wish there was a nicer way to construct GraphQL queries than creating a string template - but that seems to be the preferred way.

As you've tested this elsewhere, I think there's no much point holding this up, so we should just merge, let it run, and make sure it's doing what it ought!

rtibbles (changes requested):
Oh, sorry - actually let's move the ids into secrets that we read in the workflow and then pass into the JS functions.

AlexVelezLl:
Thanks @rtibbles! I've pushed the requeseted changes :)

rtibbles (approved):
Secrets added at the org level and exposed to the KDS repo.

MisRob:
@AlexVelezLl @rtibbles 

@EshaanAgg noted https://github.com/learningequality/kolibri-design-system/issues/906 - would it be possible to limit the action run to LE repo only?

PR #816, opened 2024-11-07
#760-Updating Check Changelog and Update Changelog workflows to make it more robust.

RONAK-AI647:
**The check changelog section and update changelog section is updated to check every field is validated.**

## Description
Their is an attempt to make the check changelog workflow more robust so it checks that every field of changelog item exist and they are different from their default values.
#### Issue addressed
1)check for every field of changelog item.
2)they are different from default values
3)they are properly rendered in the PR description.

Addresses #*PR# 

### Before/after screenshots
**A new file is inserted ,see below.**
image

## Changelog

  - **Description:** Here is an attempt to make the changelog section more robust .Earlier only the "description" field was mandatory ,but this PR ensures that all the fields of the changelog item exist and anything unnecessary is ignored. 
  - **Products impact:** new API
  - **Addresses:**learningequality/kolibri-design-system#760
  - **Components:** N/A
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** This will check if each field is present and ensures flexibility of PRs.

## Steps to test

1. Open check changelog /workflows/github/kolibri-design-system

## Testing checklist

- [ ] Contributor has fully tested the PR manually
- [x] If there are any front-end changes, before/after screenshots are included
- [x] Critical and brittle code paths are covered by unit tests
- [x] The change is described in the changelog section above

## Reviewer guidance

- [x] Is the code clean and well-commented?
- [ ] Are there tests for this change?
- [ ] Are all UI components LTR and RTL compliant (if applicable)?
- [ ] _Add other things to check for here_

## Comments
It was mentioned in "possible tradeoffs" if we could mention all the logic code in a script file to avoid too much of code in gh workflow, an attempt is made to do so.

PR #885, opened 2025-01-11
Updating changelogs.yml/ workflows to make it more robust.

### LINKED ISSUE #760: Make changelog checks more robust

## Product
KDS.

## Desired behavior

Would be helpful to update our [Check Changelog](https://github.com/learningequality/kolibri-design-system/blob/d3d40d2b2ec63e51f509b0eb5221a9fa2770edb3/.github/workflows/changelog.yml) and [Update Changelog](https://github.com/learningequality/kolibri-design-system/blob/d3d40d2b2ec63e51f509b0eb5221a9fa2770edb3/.github/workflows/update_changelog.yml) workflows to make them more robust.

```[tasklist]
### Tasks
- [ ] Update the `Check PR for Changelog` workflow to check that every field of a changelog item exist, that its not empty and that they are different from the "default" values written in the [PR Template](https://github.com/learningequality/kolibri-design-system/blob/d3d40d2b2ec63e51f509b0eb5221a9fa2770edb3/.github/pull_request_template.md#L26).
- [ ] Update the `Check PR for Changelog` workflow to make sure that the `<!-- [DO NOT REMOVE-USED BY GH ACTION] CHANGELOG START -->` and `<!-- [DO NOT REMOVE-USED BY GH ACTION] CHANGELOG END -->` flags are present in the PR description.
- [ ] Find a way to separate multiple changelog items and have them render nicely in the PR description (tbd).
- [ ] Update the `Update Changelog` workflow to ignore any line that does not correspond to any of the changelog item fields.
```
## Possible Tradeoffs

We will probably need to create script files to avoid having too much logic code written inside the gh workflow file.

RONAK-AI647:
## Description
Their is an attempt to make the check changelog workflow more robust so it checks that every field of changelog item exist and they are different from their default values. Please refer PR #816 
#### Issue addressed 
Closes #760

Addresses #*PR# 885

## Changelog

  - **Description:** Updating changelog workflows to check for every field.
  - **Addresses:** #760 
  - **Components:** no
  - **Breaking:**  no
  - **Impacts a11y:** no
  - **Guidance:**  Here is an attempt to make the changelog section more robust .Earlier only the "description" field was mandatory ,but this PR ensures that all the fields of the changelog item exist and anything unnecessary is ignored.

## Steps to test

)check for every field of changelog item.
2)they are different from default values
3)they are properly rendered in the PR description.

## Testing checklist

## Reviewer guidance

## Comments
Please refer PR #816 , for previous comments and reviews by @rtibbles and other reviewers.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -9,23 +9,18 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-      - name: Check PR Description
-        id: check_description
+      - name: Checkout code
+        uses: actions/checkout@v2
+      
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.x
+
+      - name: Install dependencies
         run: |
-          description=$(jq -r ".pull_request.body" "$GITHUB_EVENT_PATH" | awk '/## Changelog/{flag=1; next} /##/{flag=0} flag' | grep -oP '(?<=- \*\*Description:\*\* ).*' | sed 's/^[[:space:]]*//;s/[[:space:]]*$//')
-          echo "$description"
-          default_description="Summary of change(s)"
-          if [ -n "$description" ] && [ "$description" != "$default_description" ]; then
-            echo "containsChangelog=true" >> $GITHUB_OUTPUT
-          else
-              echo "containsChangelog=false" >> $GITHUB_OUTPUT
-          fi
-      - name: Results
+          pip install github
+
+      - name: Check PR fields
         run: |
-          if [[ "${{ steps.check_description.outputs.containsChangelog }}" == "true" || "${{ github.event.pull_request.user.login }}" == "dependabot[bot]" ]]; then
-            echo "PR contains Changelog section/ PR made by Dependabot"
-            exit 0
-          else
-            echo "Changelog section is missing or does not contain the required details"
-            exit 1
-          fi
+          python check_pr_description.py
--- a/.github/workflows/check_pr_description.py
+++ b/.github/workflows/check_pr_description.py
@@ -1,0 +1,3 @@
+import os
+import json
+import re
--- a/.github/workflows/check_pr_description.py
+++ b/.github/workflows/check_pr_description.py
@@ -61,3 +61,17 @@ def check_pr_description(event_path):
     
     return results
 
+
+
+def main():
+    event_path = os.getenv('GITHUB_EVENT_PATH')
+    if not event_path:
+        print("Error: GITHUB_EVENT_PATH environment variable is not set.")
+        exit(1)
+
+    results = check_pr_description(event_path)
+    
+    # Set output variables for GitHub Actions
+    for field, valid in results.items():
+        print(f"contains{field.replace(' ', '')}={str(valid).lower()}")
+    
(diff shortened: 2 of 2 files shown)


.github/workflows/changelog.yml
```diff
@@ -9,23 +9,18 @@ jobs:
...
-          description=$(jq -r ".pull_request.body" "$GITHUB_EVENT_PATH" | awk '/## Changelog/{flag=1; next} /##/{flag=0} flag' | grep -oP '(?<=- \*\*Description:\*\* ).*' | sed 's/^[[:space:]]*//;s/[[:space:]]*$//')
-          echo "$description"
-          default_description="Summary of change(s)"
-          if [ -n "$description" ] && [ "$description" != "$default_description" ]; then
-            echo "containsChangelog=true" >> $GITHUB_OUTPUT
-          else
-              echo "containsChangelog=false" >> $GITHUB_OUTPUT
-          fi
-      - name: Results
+          pip install github
```
rtibbles, line 22: I don't see this actually being used anywhere? This dependency installing seems to be breaking on github actions, so I think you can probably remove this?

.github/workflows/check_pr_description.py
```diff
@@ -0,0 +1,89 @@
...
+    """Check all fields in the PR description and validate their values."""
+    description = get_pr_description(event_path)
+    
+    # Check if changelog markers are present
+    if not validate_changelog(description):
+        return False
+    
+    # Check each field against the default value
+    results = {}
+    for field_name in DEFAULT_VALUES.keys():
```
rtibbles, line 57: No need to change, but you can also iterate over the keys of a dictionary with just:
```
for field_name in DEFAULT_VALUES:
```

.github/workflows/check_pr_description.py
```diff
@@ -0,0 +1,89 @@
...
+
+def check_pr_description(event_path):
+    """Check all fields in the PR description and validate their values."""
+    description = get_pr_description(event_path)
+    
+    # Check if changelog markers are present
+    if not validate_changelog(description):
+        return False
+    
+    # Check each field against the default value
```
rtibbles, line 55: It might be useful to use some of the same logic we are using to parse and extract the changelog for updates here: https://github.com/learningequality/kolibri-design-system/blob/develop/.github/workflows/update_changelog.yml#L27

That way the validation can ensure that it is also readable by the process that will be extracting it.

rtibbles (changes requested):
I think we can get rid of the dependency and it should run.

One thought about potentially reusing parsing code from the extraction action so that we can be doubly sure the validation is validating the same way.

EshaanAgg:
Hi! I just a minor question/feature request which I came across while working on a different PR: 

The `changelog.yml` currently states that we should be creating multiple changelog items for each change:
https://github.com/learningequality/kolibri-design-system/blob/e4449d25512c065573b2ea231ca8925f98a660cd/.github/pull_request_template.md?plain=1#L18-L20

I just want to confirm if the current version of this script supports this functionality? If not, what might be a good way to support it: adding the changes and architecting the template of pull requests for the same in this PR itself or opening a new issue?

RONAK-AI647:
Thanks again @EshaanAgg @rtibbles , get me few days , will be back.

MisRob:
Hi @RONAK-AI647, do you still plan to complete resolving the feedback?

PR #888, opened 2025-01-14
Fix KDateRange confusing selection states and end date reset after selecting start date via keyboard

### LINKED ISSUE #817: KDateRange: Confusing selection states; end date resets after selecting start date via keyboard
## Product

Kolibri Design System 

## Expected behavior

When using the keyboard to select a start date, users should be able to:
- Select the Start Date via the keyboard in the textbox.
- Select the End Date on the calendar using the mouse.

The End Date should only clear if the new start date is after the current End Date. Otherwise, the End Date should remain unchanged.

## Actual behavior

When using the cursor keys to update the start date in `KDateRange`, the End Date is set to a clear state. Additionally, confusing selection states appear when setting the initial date with the keyboard. A subsequent mouse click causes the Start Date to be reset, even if the user then navigates to the End Date input box.

1. Open the KDateRange component and select a Start Date in the textbox using the keyboard.
2. Select a End Date on the calendar view using the mouse.
3. Notice how the Start Date gets updated instead of the End Date.
4. Select a Start and End date on the calendar, then select a new Start Date with a date that is before the End Date.
5. Notice that after the Start Date is updated, the End Date is cleared out.

## Additional information

https://github.com/user-attachments/assets/92f45143-24c5-46f0-a8ad-463093b4a94a

LianaHarris360:
## Description

This pull request addresses the KDateRange bug in which, after initially setting a Start Date in the textbox using the keyboard, selecting another date on the calendar results in the Start Date being reset.

This pull request also clears the End Date after selecting a Start Date, but only if the Start Date is after the End Date. 
**This feature is not replicated when using mouse clicks to select dates on the calendar view, as it would make it difficult to remove or restart the date range values.**

#### Issue addressed

Addresses #817 

### Before/after screenshots

After:
Setting a start date in the text box, then clicking to select the end date on the calendar:

https://github.com/user-attachments/assets/af4b0d99-76c3-46a0-a85f-9a86b391d864

Selecting a start date that is before and after the end date:

https://github.com/user-attachments/assets/1d32a122-a78d-469c-8e7d-b931c22f676c

## Changelog

  - **Description:** Updates KDateRange logic so that end date is cleared only if start date is after end date; users can now set start date via keyboard, then select end date using the calendar view.
  - **Products impact:** none
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/817
  - **Components:** -
  - **Breaking:** no
  - **Impacts a11y:** -
  - **Guidance:** -

## Steps to test

1. Open the KDateRange component and select a Start Date in the textbox using the keyboard.
2. Select a End Date on the calendar view using the mouse. The End Date should be updated as expected.
3. Select a Start and End date on the calendar, then input a new Start Date in the textbox with a date that is before the End Date.
4. The Start Date should be updated, while the End Date remains unchanged.
5. Input a new Start Date in the textbox with a date that is after the End Date.
6. The Start Date should be updated, and the End Date should be reset.

The logic within `setStartDate()` was updated so that if the start date is after the end date, the end date is `null`, otherwise, it is unchanged.

Within `KDateCalendar.vue`, a watcher has been added for the props `selectedStartDate` and `selectedEndDate`, which will update the local `dateRange` object values, so that the calendar dates are consistent with the updates made to the dates in the KDateRange textboxes. `isFirstChoice` has been converted into a computed property and its value is updated based on changes to the dateRange object values.
 

## Testing checklist

## Reviewer guidance

### DIFF
--- a/lib/KDateRange/KDateCalendar.vue
+++ b/lib/KDateRange/KDateCalendar.vue
@@ -243,14 +243,14 @@
        * default value of selected start date
        */
       selectedStartDate: {
-        type: [Date, null],
+        type: Date,
         default: null,
       },
       /**
        * default value of selected end date
        */
       selectedEndDate: {
-        type: [Date, null],
+        type: Date,
         default: null,
       },
       /**
@@ -276,7 +276,6 @@
           end: this.selectedStartDate && this.selectedEndDate ? this.selectedEndDate : null,
         },
         numOfDays: 7,
-        isFirstChoice: this.selectedStartDate == null ? true : false,
         activeMonth:
           this.lastAllowedDate.getMonth() - 1 == -1 ? 11 : this.lastAllowedDate.getMonth() - 1,
         activeYearStart: this.lastAllowedDate.getFullYear(),
@@ -319,6 +318,21 @@
         }
         return this.activeYearStart;
       },
+      isFirstChoice() {
+        // True if there is no start date OR if there is an end date
+        // False if there is a start date but no end date
+        return !this.dateRange.start || !!this.dateRange.end;
+      },
+    },
+    watch: {
+      // Update local dateRange.start whenever selectedStartDate is set
+      selectedStartDate(newVal) {
+        this.dateRange.start = newVal;
+      },
+      // Update local dateRange.end whenever selectedEndDate is set
+      selectedEndDate(newVal) {
+        this.dateRange.end = newVal;
+      },
     },
     created() {
       if (this.activeMonth === 11) this.activeYearStart = this.activeYearStart - 1;
@@ -345,7 +359,6 @@
       getNewDateRange(result, activeMonth, activeYear) {
         const resultDate = new Date(activeYear, activeMonth, result);
         if (!this.isFirstChoice && resultDate < this.dateRange.start) {
-          this.isFirstChoice = false;
           return { start: resultDate };
         }
         const newData = {};
@@ -355,8 +368,6 @@
         } else {
           newData.end = null;
         }
-        // toggle first choice
-        this.isFirstChoice = !this.isFirstChoice;
         newData[key] = resultDate;
         return newData;
       },
--- a/lib/KDateRange/index.vue
+++ b/lib/KDateRange/index.vue
@@ -299,8 +299,19 @@
       },
       /** Updates start date with input from textbox */
       setStartDate(newVal) {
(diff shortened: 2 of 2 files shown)


nucleogenesis (commented):
I have tested locally and the code changes look good to me. Things work as expected and described.

However, I think that the UX is a bit wonky in this particular case:

You cannot edit your end-date selection: Select 10/1/24 for start date, then click 10/31/24 for the end date. Oops, you meant to cilck 11/1/24! When you click 11/1/24, you have now selected 11/1/24 as the start date and basically lose your selection.

Or, similarly, say you select 10/1/24 - 11/1/24 but you really meant to only select all of October! If you click within the selection to change the end date to 10/31/24 it's instead set to your start date.

---

I'm not sure if that's the expected behavior but it feels odd that every time I select a date after having two dates selected that I'm starting my selection over again rather than being able to tweak it a bit.

To be clear, though, the keyboard usage works on both of the text box inputs perfectly in my testing, so this seems to address the issue it's targeting altogether.

LianaHarris360:
I believe it is the expected behavior because if two dates are already set and then a new date (within the selection) is chosen, how do we decide if that new date is meant to be the start or end date? However, being able to directly edit the start and end date via the text box without resetting the date range selection is somewhat of a middle-ground solution.

nucleogenesis:
@LianaHarris360 sorry for the delay in reply here!

I'm interested in @radinamatic and/or @pcenov 's thoughts here.

Particularly I'm stuck on not being able to extend/edit my selection I guess - it feels like I should be able to click 1 day past my end date and extend it but, in any case, things look and feel solid to me.

radinamatic:
Apologies for late reply @LianaHarris360 !

We have reviewed the PR and the proposed solution does fix the original issue without any regressions regarding the keyboard navigation 👏🏽.

Couple of comments for future iterations and improvements:

* @pcenov mentioned that we might consider thinking about adding a default pre-filled date range (would need decisions regarding what would be the most common beneficial range)
* while I concur with comment by @nucleogenesis that the ability to extend the range would be handy, implementing it to be keyboard-navigable would require additional _handlers_ for dragging, and that is not an easy undertaking with a complex  component like this.

radinamatic (approved):
LGTM, thank you @LianaHarris360 ! 💯 :shipit:

PR #504, opened 2023-12-08
/deep/ to ::v-deep

rtibbles:
## Description
* Removes two unused /deep/ selectors in KSelect that were referencing non-existent classes
* Replaces the two remaining /deep/ selectors used in KDS with `::v-deep` https://vue-loader.vuejs.org/guide/scoped-css.html#deep-selectors
* This behaves identically, but has the advantage of not breaking future versions of Sass (i.e. Dart Sass instead of Node Sass, which is deprecated)

#### Issue addressed
This is required as part of an extended yak shave in order to ultimately upgrade to newer NodeJS versions, where building Node Sass becomes increasingly more challenging.

## Changelog

  - **Description:** Remove use of /deep/ in favour of ::v-deep
  - **Products impact:** none
  - **Addresses:** -
  - **Components:** -
  - **Breaking:** no
  - **Impacts a11y:** -
  - **Guidance:** -

## Steps to test

1. Ensure that KSelect still functions as intended when disabled.
2. Ensure that the date input styling in the docs page for the date picker is still correct
3. Ensure that the Icon is still blue on the icons page showing what not to do

## Testing checklist

- [ ] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical and brittle code paths are covered by unit tests
- [ ] The change is described in the changelog section above

## Reviewer guidance

- [ ] Is the code clean and well-commented?
- [ ] Are there tests for this change?
- [ ] Are all UI components LTR and RTL compliant (if applicable)?
- [ ] _Add other things to check for here_

## After review

- [ ] The changelog item has been pasted to the `CHANGELOG.md`

PR #897, opened 2025-01-15
Migrate test from enzyme to vue testing library

### LINKED ISSUE #510: Migrate to Vue Testing Library
## Blocked by

https://github.com/learningequality/kolibri-design-system/issues/439

## Context
[Vue Testing Library](https://testing-library.com/docs/vue-testing-library/intro) is a lightweight wrapper that builds on top of the DOM Testing Library by adding APIs for working with Vue components. It is built on top of [`@vue/test-utils`](https://github.com/vuejs/vue-test-utils), which is currently used in the KDS.

As per the documentation, Vue Testing Library does three things:
1. Re-exports query utilities and helpers from the DOM Testing Library.
2. Hides `@vue/test-utils` methods that conflict with [Testing Library Guiding Principle](https://testing-library.com/docs/guiding-principles).
3. Tweaks some methods from both sources.

## Product
KDS

## The Value Add
I personally advocate for VTL because of two major reasons:
1. Most of the products (OSS or otherwise) are migrating toward the Testing Library (or its framework wrappers like React Testing Library etc.), and thus the community support is much more. The libraries are actively maintained, and using them leads to similar-looking tests irrespective of the framework of your choice. We can also then install `@testing-library/jest-dom` to use [the custom Jest matchers](https://github.com/testing-library/jest-dom#readme) for the DOM, enabling us to write better and more expressive tests. 
2. The Testing Library is based on the principle that we should only test for user interactions, and never for the internal state of components, and thus hides some API exports that allow us to do so. This, in my opinion, again leads to better tests, that check for the functionality of the components and are not heavily invested in the internal implementation of the state and hooks of the component, as it should be for a component library. 

## Possible Tradeoffs
Refactoring the existing tests is additional work, but it would also provide a good foundation for the future if we want to invest heavily in testing. 

## PS
I would be happy to migrate an existing test to VTL to provide a proof of concept for the approach.

Pandaa007:
## Description

- Added Vue testing library to the codebase
- Migrated one test from enzyme to Vue testing library as proof of concept.
- Since the library (node-sass) was deprecated and was causing installation errors in fresh environments as mine, I upgraded it to it's successor sass. Now the installation process completes without any error and we have gotten rid of old dependencies (https://stackoverflow.com/questions/74362396/i-can-not-install-node-sass)

#### Issue addressed

Fixes #510 

## Changelog

  - **Description:** MIgrated test from enzyme to Vue testing library
  - **Products impact:**  none
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/510
  - **Components:** none
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -

## Testing checklist

## Reviewer guidance


MisRob:
cc @EshaanAgg if you're around and would like to join review, you're very welcome

MisRob:
Thank you @Pandaa007!

MisRob:
@rtibbles @bjester wanted to flag this note from the PR description - sounds good to you two? Could it have some implications for Kolibri and Studio?

> Since the library (node-sass) was deprecated and was causing installation errors in fresh environments as mine, I upgraded it to it's successor sass

rtibbles:
> Could it have some implications for Kolibri and Studio?

Immediate implications, no - because this is only a dev dependency, so doesn't affect what gets installed in Kolibri and Studio.

This [PR](https://github.com/learningequality/kolibri-design-system/pull/504) (and the NodeJS upgrade) was the only prerequisite to doing this dependency upgrade so this should be OK within KDS itself.

The main thing is that node-sass was incredibly outdated, had quite a few bugs, and also missing some SASS language features. I think we might have noticed, but it is possible that switching to `sass` may cause changes in styling because of SASS language features actually being supported - or if we start leaning on SASS language features that are not supported in `node-sass` this might have implications for Kolibri and Studio where styling does not behave identically to things in KDS, so if we do retain this dependency upgrade, we would want to expedite the upgrade in Kolibri and Studio also to minimize the chance of mismatches in behaviour. I think this will require finalizing the NodeJS 18 upgrade in Studio, which should hopefully happen in the next month or so, so this may be feasible.

But any merge of this should open correlated issues for sass upgrade on those two repos.

lib/buttons-and-links/__tests__/KButton.spec.js
```diff
@@ -1,67 +1,63 @@
...
-          text: 'test',
-        },
-      });
-      wrapper.trigger('click');
-      expect(wrapper.emitted().click).toBeTruthy();
+    it('should emit a click event when clicked', async () => {
+      const { emitted } = renderComponent({ text: 'test' });
+
+      const button = screen.getByText('test');
+      await fireEvent.click(button);
```
EshaanAgg: As per our testing style guide, we would like to prefer the use of the [`@testing-library/user-event` package over the `fireEvent`](https://kolibri-dev.readthedocs.io/en/develop/frontend_architecture/unit_testing.html#prefer-the-userevent-package-over-fireevent-to-simulate-user-interactions) to simulate interactions wherever possible to maintain consistency. The API is mostly the same between the two packages, so the migration should not be a hassle!

EshaanAgg (commented):
Thanks @Pandaa007! The PR looks great. 

I have added a small nit related to the event simulation package, so it would be really helpful if you can address that! 

Also, might be a bit unrelated to this PR, but it might be nice to have the [`eslint-plugin-jest-dom`](https://www.npmjs.com/package/eslint-plugin-jest-dom) also configured in the repository now, as the same ensures that our tests are using the correct custom matcher's wherever possible! Would greatly appreciate if you can install the same and configure it as well! 

As always, if you feel stuck, please do not hesitate to reach out for help!

lib/buttons-and-links/__tests__/KButton.spec.js
Pandaa007: I've added the plugins in the latest commit. Thanks @EshaanAgg

EshaanAgg (approved):
Thanks for making the changes @Pandaa007! The PR looks good to me now! You can now wait for approval from other members of the team.

lib/buttons-and-links/KButton.vue
```diff
@@ -40,7 +40,7 @@
       v-if="iconAfter"
       :icon="iconAfter"
       :color="iconColor"
-      data-test="iconAfter"
+      data-testid="iconAfter"
```
nucleogenesis, line 48: _Non-blocking question_

Is using `data-testid` a convention of the updated Vue testing lib? I know we've used `data-*` attrs (typically `data-test`) to identify elements in tests but I don't recall a particular attr being used like seems to be the case below w/ the `screen.getByTestId`.

lib/buttons-and-links/__tests__/KButton.spec.js
```diff
@@ -1,67 +1,64 @@
...
-        propsData: {
-          icon: 'add',
-        },
-      });
-      expect(wrapper.find('[data-test="iconBefore"]').exists()).toBe(true);
+    it('should render an icon before the text with the icon string passed to the icon prop', () => {
+      renderComponent({ icon: 'add' });
+
+      const iconBefore = screen.getByTestId('iconBefore');
+      expect(iconBefore).toBeInTheDocument();
```
nucleogenesis, line 24: Love how much better this reads overall <3

nucleogenesis (approved):
On code review, this LGTM - I'm a big fan of how the updated tests are more readable overall and it's a great start of a migration.

lib/buttons-and-links/__tests__/KButton.spec.js
```diff
@@ -1,67 +1,64 @@
...
-        propsData: {
-          text: 'test',
-        },
-      });
-      wrapper.trigger('click');
-      expect(wrapper.emitted().click).toBeTruthy();
+    it('should emit a click event when clicked', async () => {
+      const { emitted } = renderComponent({ text: 'test' });
+
+      const button = screen.getByText('test');
```
AlexVelezLl: [Non-blocking] Just jumping in to note that the recommended query here is `screen.getByRole('button', {name: /test/i})`. See https://testing-library.com/docs/queries/about/#priority.

lib/buttons-and-links/KButton.vue
Pandaa007: Yes, the `data-testid` attribute is a widely used convention in frontend testing libraries based on the Testing Library (and its counterparts in other ecosystems, including Vue Testing Library). However, this isn't a strict convention enforced by Vue Testing Library or Testing Library in general—it’s more of a practical choice that's become popular.

lib/buttons-and-links/__tests__/KButton.spec.js
Pandaa007: Nice catch! Thanks for the suggestion. Have implemented it in the latest commit.

AlexVelezLl: Superr!! Thanks @Pandaa007! Could you please rebase on top of the latest develop so we can merge it :) thanks!

Pandaa007: Sure! will do it.

AlexVelezLl: Thanks! Please let us know once its rebased :)

AlexVelezLl: Thanks @Pandaa007! Please let us know once its rebased :)

Pandaa007: Hi @AlexVelezLl 
Due to an unavoidable emergency, I kindly request additional time to complete the task and appreciate your understanding.

AlexVelezLl: Oh no worries, its okay! No rush at all! :hugs: Please take your time :)

Pandaa007:
Hey! Thanks @AlexVelezLl for your kind patience!

I have rebased the PR appropriately and made the changes as well. There are two design designs that I have taken in this PR that might be of relevance: 
- I renamed the render function for visual testing from `renderComponent` to `renderComponentForVisualTest` as the same better conveys the intent for the use of the same, and also prevents the conflict of the name `renderComponent`, which according to the [LE Testing guidelines](https://kolibri-dev.readthedocs.io/en/develop/frontend_architecture/testing_layout.html) is the recommended name of the wrapper function that we should be using. This will help prevent more conflicts with naming as we move forward! The same also conveys the difference between these two types of rendering better IMHO.

- I have also reverted my change of adding `sass` package and restored the old `node-sass` package in KDS. The primary reason behind the same is that during local testing, I realized that with the most recent version of `sass`, when we run the application in `dev-only` mode, the console gets filled with multiple warnings about deprecated ways of using sass styling. This led me to believe that this might not be a drag-and-drop replacement as I was hoping it to be, and thus, it would be better not to club this change in this PR, and we should research it more before proceeding with it. 

The PR is now ready for review. Thank you all for working with me on this PR.

Pandaa007:
Hi! Just wanted to confirm if anything is required my side to get the PR reviewed?

MisRob:
Hi @Pandaa007, thank you for reaching out. I don't think anything else is needed from you right now, we just have lots of reviews to work through. We'll be in touch.

Pandaa007:
@MisRob @nucleogenesis I have updated the PR with the latest KDS content. Please let me know if there are any other blockers for this PR.

nucleogenesis:
Thank you @Pandaa007 !

PR #901, opened 2025-01-20
Integrate visual testing mechanism to the latest KDS version

KshitijThareja:
## Description
This PR aims to integrate the visual testing mechanism to the latest version of KDS and ensure its proper working by implementing necessary changes.

## Changelog

  - **Description:** Integrates visual testing setup to KDS.
  - **Products impact:** -
  - **Addresses:** -
  - **Components:** -
  - **Breaking:** -
  - **Impacts a11y:** -
  - **Guidance:** -

## Testing checklist

## Reviewer guidance


.github/workflows/visual_tests.yml
```diff
@@ -0,0 +1,109 @@
...
+          path: '**/node_modules'
+          key: ${{ runner.OS }}-node-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.OS }}-node-
+      - name: Install dependencies
+        run: |
+          yarn --frozen-lockfile
+          npm rebuild node-sass
+      - name: Download Chromium
+        run: npx puppeteer browsers install chrome
```
bjester, line 48: Curious if this is cached, and if not, if we should

jest.conf/visual.index.js
```diff
@@ -0,0 +1,121 @@
+const path = require('node:path');
+const http = require('http');
```
bjester: This should probably be `node:http`

lib/buttons-and-links/__tests__/KButton.spec.js
```diff
@@ -64,4 +65,106 @@ describe('KButton', () => {
       expect(wrapper.emitted().click).toBeTruthy();
     });
   });
+
+  describe.visual('KButton Visual Tests', () => {
```
bjester: I suppose `Visual` in the string is redundant with what `describe.visual` does

bjester (approved):
A few comments, but nothing blocking that I can see from a quick pass

.github/workflows/visual_tests.yml
KshitijThareja: image

From what I can infer from the test run, puppeteer does cache the browser after the first workflow run and uses the same for successive runs. Is there a need to add another step to the workflow for the same?

lib/buttons-and-links/__tests__/KButton.spec.js
KshitijThareja: I'll just make it `KButton Tests` in that case :+1:

lib/KThemePlugin.js
```diff
@@ -60,7 +60,7 @@ require('./grids/globalStyles.js'); // global grid styles
  * Also, set up global state, listeners, and styles.
  */
 export default function KThemePlugin(Vue) {
-  if (!isNuxtServerSideRendering()) {
+  if (!isNuxtServerSideRendering() && process.env.VISUAL_TESTING !== 'true') {
```
AlexVelezLl, line 63: Why wouldnt we install the KThemePlugin while running visual tests?

KshitijThareja: With the introduction of KLiveRegion and the changes in the KThemePlugin, if we try to  run visual tests, the variable 'document' will become undefined as for it to work as intended, we require a jsdom test environment. This will result in errors in visual test execution

AlexVelezLl: Alright, thanks! Lets merge it, let it run and then if we need to activate it, we can look for alternatives. Thanks!

AlexVelezLl:
Thanks a lot for all your hard work @KshitijThareja!!! Merging! :tada: :tada: :tada:.

AlexVelezLl:
Hey @KshitijThareja @bjester, we are getting this error on the execution. Do you have any idea why could it be happening? https://github.com/learningequality/kolibri-design-system/actions/runs/13038589590/job/36375439798

image

bjester:
@AlexVelezLl ChatGPT says to possibly try using `--no-sandbox` first, and then possibly add `--disable-setuid-sandbox`. Also, we should ensure all dependencies are installed in the action's container

KshitijThareja:
Hi @AlexVelezLl @bjester 
So I looked it up and found [this](https://chromium.googlesource.com/chromium/src/+/main/docs/security/apparmor-userns-restrictions.md). The easiest, and probably the best fix for now would be to just change the ubuntu version used for running visual test workflow to 22.04.
Should I go forward with making a PR for the same?

bjester:
> Hi @AlexVelezLl @bjester So I looked it up and found [this](https://chromium.googlesource.com/chromium/src/+/main/docs/security/apparmor-userns-restrictions.md). The easiest, and probably the best fix for now would be to just change the ubuntu version used for running visual test workflow to 22.04. Should I go forward with making a PR for the same?

Sounds like `--no-sandbox` is easier:
> What if I don‘t have root access to the machine and can’t install anything?
You will need to run developer builds with the --no-sandbox command line flag, but be aware that this disables critical security features of Chromium and should never be used when browsing the open web.

PR #873, opened 2024-12-13
Add DocsExample

chetan21122004:
Description:
I’ve added the DocsExample component that allows toggling between an example and its code.

Issue Addressed:
Fixes #826. Also references #836  for additional context on previous discussions related to the component.

Guidance:
The new component enhances the documentation UI by toggling code visibility.

Steps to Test:
1. Navigate to the KCardGrid documentation page.
2. Verify that the example and code toggle correctly with a slide-up/slide-down animation.

Implementation Notes:
- The DocsExample component uses a #code slot to accept code snippets in multiple languages.
- Slide toggle animation is applied when showing/hiding the code.

## Changelog

  - **Description:**  Implemented `DocsExample` component with toggle functionality for switching between an example and its code snippet in the KDS documentation.
  - **Products impact:** none
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/826
  - **Components:** -
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -

Issue #845, opened 2024-11-27
Further improve `DocsExample` visual output and API


MisRob:
**🌱 Are you new to the codebase? Welcome! Please see [the contributing guidelines](https://github.com/learningequality/kolibri-design-system/blob/develop/CONTRIBUTING.md).**

## Background

The [first version of `DocsExample`](https://github.com/learningequality/kolibri-design-system/issues/826) requires `DocsShow` being passed via `code` slot:

```
<DocsExample>

  // The default slot renders the example

  <template #code>
     // The code slot renders code snippet(s)
     <DocsShowCode language="html">
     
     </DocsShowCode>
     
     <DocsShowCode language="javascript">
     
     </DocsShowCode>
  </template>
</DocsExample>
```

Visually, all code snippets are displayed above the example.

## Addresses

This proposal aims at improving
- concise and better organized visual output
- simpler API for developers

## Proposal

**(1) Remove the need for passing `DocsShowCode` to `DocsExample`**

- instead provide `html`, `javascript`, and `scss` slots
- `DocsExample` will use `DocsShowCode` internally to render these slots

(2) **`DocsExample` shows a toolbar on the left with the option to switch the view of the code** between HTML, Javascript, and CSS.

```
<DocsExample>

  // The default slot renders the live example
  <KCardGrid ...>
    ...
  </KCardGrid>

  <template #html>
    <KCardGrid ...>
      ...
    </KCardGrid>
  </template>

 <template #javascript>
     export default { 
        ...
       }
  </template>
  
  <template #scss>
     .class {
        ...
      }
  </template>
</DocsExample>
```

would output very similar experience to Vuetify's:

![Image](https://github.com/user-attachments/assets/9b5a39cd-ee2d-4be1-ab2f-ae0840331dc9)

chetan21122004:
Hi @MisRob, I’ve added the tabs feature as suggested, allowing users to switch between HTML, JS, and SCSS. If there’s only one tab, the tab options won’t be shown. I’ve also attached a video for reference. Could you please assign this task to me? Looking forward to your feedback 
 https://github.com/user-attachments/assets/06514a0e-bf34-4fc7-a6dd-5ea58197baaa

MisRob:
[...]

PR #905, opened 2025-01-23
docs: allow `DocsShow` to automatically parse SFC's and display their source and output

### LINKED ISSUE #861: Configure prettier and eslint to treat DocsShowCode as code examples

## Product
KDS.

## Current behavior
When we want to show code examples in the KDS docs, we use the `DocsShowCode` component. We can show HTML, JavaScript, and CSS code with this component. For example:

```Vue
<DocsShowCode language="javascript">
  console.log("Hello world")
</DocsShowCode>
```

The problem is that eslint and prettier treat the code inside DocsShowCode as HTML code, so eslint raises exceptions for linting rules and prettier formats this code as html instead of treating this as a code example. So always that we have a css or javascript code examples, we need to write `<!-- eslint-disable -->` and `<!-- prettier-ignore -->` comments around it for them to avoid exceptions and unwanted formatting. So we end up with something like:

```javascript
<!-- eslint-disable -->
<!-- prettier-ignore -->
<DocsShowCode language="javascript">
  console.log("Hello world")
</DocsShowCode>
<!-- eslint-enable -->
```

Which adds unnecessary complexity to writing our code examples.

## Desired behavior

We should configure prettier and eslint to at least ignore the code inside the `DocsShowCode` component as it's just a code example, and there is no need to apply linting rules and HTML formatting.

At the end, we should be able to write code examples with just:

```Vue
<DocsShowCode language="javascript">
  console.log("Hello world")
</DocsShowCode>
```

without the need to write the `<!-- eslint-disable -->` and `<!-- prettier-ignore -->` comments.

## The Value Add

We will be able to write code examples more easily. And we won't have to worry about unwanted formatting in our code examples.

## Possible Tradeoffs
If prettier just ignores the code inside, it will be the responsibility of the developer to write the code example in a well-formatted way.

## Additional notes
@rtibbles tried it and it was difficult

EshaanAgg:
## Description
The PR modifies the `DocsShow` component to accept two new arguments: 
- `loadExample`: The path to the component that needs to be shown to the user. When passed, the component is loaded dynamically, and its source code is parsed into three separate `DocsShowCode` blocks. 
- `showOutput`: Controls whether the example component is to be rendered in the documentation. Defaults to `true`.

All the examples must be in the `/docs/examples` directory to be loaded correctly. `DocsShow` still accepts a default slot, so the user can use it to render a custom component with or without the example. This functionality should (probably) be migrated to the `DocsExample` component when ready. The `loadExample` prop is set to be `null` by default so that it does not cause any regressions or changes in it's existing usage across the documentation pages. 

I have also migrated the `KTable` documentation to use this new feature (as that is the component that I am most familiar with & believe could really benefit from this new API).

#### Issue addressed
Fixes #861 

## Changelog

  - **Description:** Added automatic parsing of examples from files for `DocsShow` component
  - **Products impact:** none
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/861
  - **Components:** -
  - **Breaking:** no
  - **Impacts a11y:** -
  - **Guidance:** -

Used the `raw-loader` plugin with Webpack to configure the loading of components.

No

## Testing checklist

## Reviewer guidance

### DIFF
--- a/docs/common/DocsShow.vue
+++ b/docs/common/DocsShow.vue
@@ -5,6 +5,24 @@
       class="show"
       :style="style"
     >
+      <!-- Code blocks display -->
+      <div v-if="loadExample && codeBlocks.length">
+        <DocsShowCode
+          v-for="(codeBlock, index) in codeBlocks"
+          :key="index"
+          :language="codeBlock.language"
+        >
+          {{ codeBlock.content }}
+        </DocsShowCode>
+      </div>
+
+      <!-- Rendered example output when showOutput is true -->
+      <component
+        :is="componentToRender"
+        v-if="showOutput && componentToRender"
+      />
+
+      <!-- Usual slots -->
       <slot></slot>
     </div>
   </div>
@@ -17,6 +35,26 @@
   export default {
     name: 'DocsShow',
     props: {
+      /**
+       * Path to the Vue component file to load and display.
+       * The path should be relative to the project root.
+       * @example 'examples/KComponent/FileName.vue'
+       */
+      loadExample: {
+        type: String,
+        required: false,
+        default: null,
+      },
+      /**
+       * Controls whether to render the actual component example.
+       * When true, displays both the code and the rendered output.
+       * When false, only displays the code.
+       */
+      showOutput: {
+        type: Boolean,
+        required: false,
+        default: true,
+      },
       block: {
         type: Boolean,
         default: false,
@@ -25,13 +63,16 @@
         type: Boolean,
         default: true,
       },
-      /**
-       * Toggles dark background
-       */
       dark: {
         type: Boolean,
         required: false,
       },
+    },
+    data() {
+      return {
+        codeBlocks: [],
+        componentToRender: null,
+      };
     },
     computed: {
       style() {
@@ -40,6 +81,63 @@
           padding: this.padding ? '8px 24px' : null,
           backgroundColor: this.dark ? this.$themePalette.grey.v_700 : undefined,
         };
+      },
+    },
+    async created() {
+      if (this.loadExample) {
+        await this.loadComponentContent();
+        await this.importComponent();
+      }
+    },
+    methods: {
+      /**
+       * Loads the content of the component file using raw-loader.
+       * Parses the content into separate code blocks for template, script, and style.
+       * @throws {Error} If the component content cannot be loaded
+       */
+      async loadComponentContent() {
+        try {
+          const content = await import(`!!raw-loader!@/${this.loadExample}?raw`);
+          this.codeBlocks = this.parseTemplate(content.default);
+        } catch (error) {
+          throw new Error(`Failed to load component content: ${error}`);
+        }
+      },
+      /**
+       * Dynamically imports the component for rendering
+       */
+      async importComponent() {
+        try {
+          const component = await import(`../${this.loadExample}`);
+          this.componentToRender = component.default;
+        } catch (error) {
+          throw new Error(`Failed to import component: ${error}`);
+        }
+      },
+      parseTemplate(content) {
+        const codeBlocks = [];
+        const template = content.match(/<template>([\s\S]*?)<\/template>/);
+        if (template) {
+          codeBlocks.push({
+            language: 'html',
+            content: template[1].trim(),
+          });
+        }
+        const script = content.match(/<script>([\s\S]*?)<\/script>/);
+        if (script) {
+          codeBlocks.push({
+            language: 'javascript',
+            content: script[1].trim(),
+          });
+        }
+        const style = content.match(/<style[\s\S]*?>([\s\S]*?)<\/style>/);
+        if (style) {
+          codeBlocks.push({
+            language: 'scss',
+            content: style[1].trim(),
+          });
+        }
+        return codeBlocks;
       },
     },
   };
--- a/docs/common/DocsShow.vue
+++ b/docs/common/DocsShow.vue
@@ -114,27 +114,47 @@
           throw new Error(`Failed to import component: ${error}`);
         }
       },
+      /*
+        * Parses the content of the component file into separate code blocks for template, script, and style.
+        * @param {string} content - The content of the component file
+        * @returns {Array} An array of code blocks
+      */
       parseTemplate(content) {
         const codeBlocks = [];
-        const template = content.match(/<template>([\s\S]*?)<\/template>/);
+
+        const applyRegex = (target) => {
+          const pattern = new RegExp(`(<${target}(\\s[^>]*)?>)([\\w\\W]*)(<\\/${target}>)`, 'g');
+          const parsed = pattern.exec(content);
+
+          if (!parsed) {
+            return null;
+          }
+
+          // Extract the content between the first opening and the last closing tags
+          return parsed[3].trim();
+        };
+
+        const template = applyRegex('template')
         if (template) {
           codeBlocks.push({
             language: 'html',
-            content: template[1].trim(),
+            content: template,
           });
         }
-        const script = content.match(/<script>([\s\S]*?)<\/script>/);
+
+        const script = applyRegex('script')
         if (script) {
           codeBlocks.push({
             language: 'javascript',
-            content: script[1].trim(),
+            content: script,
           });
         }
-        const style = content.match(/<style[\s\S]*?>([\s\S]*?)<\/style>/);
+
+        const style = applyRegex('style')
         if (style) {
           codeBlocks.push({
             language: 'scss',
-            content: style[1].trim(),
+            content: style[1],
           });
         }
         return codeBlocks;
--- a/docs/common/DocsShow.vue
+++ b/docs/common/DocsShow.vue
@@ -154,7 +154,7 @@
         if (style) {
           codeBlocks.push({
             language: 'scss',
-            content: style[1],
+            content: style,
           });
         }
         return codeBlocks;
--- a/docs/common/DocsShow.vue
+++ b/docs/common/DocsShow.vue
@@ -83,10 +83,10 @@
         };
       },
     },
-    async created() {
+    created() {
       if (this.loadExample) {
-        await this.loadComponentContent();
-        await this.importComponent();
+        this.loadComponentContent();
+        this.importComponent();
       }
     },
     methods: {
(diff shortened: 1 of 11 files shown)


MisRob:
Thanks a lot @EshaanAgg, I will review. Would you please clarify if and how we can override the code output, for example how to achieve code sample outputs like [here](https://design-system.learningequality.org/kcard#click-navigation)

Screenshot from 2025-01-28 19-54-05

It's something we do intentionally (and for larger components frequently) to prevent from overwhelming information. In the future, there will be a way to see the full via GitHub link, but documentation code samples are rather brief and to the point. During the course of this migration, I'd like to be sure we don't lose the flexibility in this area.

MisRob:
> This functionality should (probably) be migrated to the DocsExample component when ready

Yes thank you, definitely. Basically DocsExample's final interface is [planned to look like](https://github.com/learningequality/kolibri-design-system/issues/845)

```
<DocsExample>

  // The default slot renders the live example
  <KCardGrid ...>
    ...
  </KCardGrid>

  <template #html>
    <KCardGrid ...>
      ...
    </KCardGrid>
  </template>

 <template #javascript>
     export default { 
        ...
       }
  </template>
  
  <template #scss>
     .class {
        ...
      }
  </template>
</DocsExample>
```

I don't know when we will have it and I think it's fine to not be concerned as long as the approach take here is compatible enough so we can migrate it easily.

EshaanAgg:
> Would you please clarify if and how we can override the code output, for example, how to achieve code sample outputs like [here](https://design-system.learningequality.org/kcard#click-navigation)

Hi @MisRob! Using the new props introduced, we can't override the code sample as shown. As I see it, this should be used as an additional convenience helper: if we want to get the complete code for some components, we can migrate them to dedicated files and use the `loadExample` prop to fetch them dynamically. 

In case of custom behavior, we can continue using the documentation components as we currently do:

```vue
<DocsShow>
  <DocsShowCode language='html'>
    <p>...partial code that we want to show </p>
  </DocsShowCode>
</DocsShowCode>
```
 >  I'd like to be sure we don't lose the flexibility in this area.

Sure! That's why I have tried to only add some additional APIs, and not override any existing behaviour.

MisRob:
I see. Yeah I think it'd be fine, even though perhaps we can brainstorm a bit? Since this is something we do often. I think it wouldn't need to be complicated and quite aligned with your thinking and all work here, for example having

_examples/example1.vue_

```
<DocsExample>

</DocsExample>
```

and then loading it to the docs page


```
<DocsLoadExample name="example1">
```

All `examples/` could be ignored by lint.

How would something like that sound to you?

MisRob:
Of course now the way to imagine `example1` would be rather

```
<DocsShow/>
<DocsShowCode/>
```

Also I am not sure if we need another component `DocsLoadExample` just to load the example, perhaps there's more elegant way. I suggest it since it seems to be aligned with the approach you chose here.

I also want to keep thinking whether we can make all these perspectives work with auto-generation as you suggest originally in this PR. Every approach has benefits, and perhaps we can make it all work nicely together. Please let me know any thoughts.

MisRob:
I will return to this by the end of the week and try to think about these questions meanwhile too

EshaanAgg:
> Yes thank you, definitely. Basically, DocsExample's final interface is [planned to look like](https://github.com/learningequality/kolibri-design-system/issues/845)

Yes! If the API contract for the same is not fixed, I would like for it to support three usages:

1. Load and show the complete component from a file (complete code & example are shown to the user in a Vuetify-like experience):

```vue
<DocsExample component="/examples/KComponent/FileName" />
```

2. Load the component from a file, but show specific code when navigating the code slots: This is useful when we have long components whose complete examples are present in separate files, but for the sake of brevity or clarity, we want the users to see only a part of it when using the documentation site. We can make use of named slots as you proposed for the same.
```vue
<DocsExample component="/examples/KComponent/FileName">
 <template #html>
    <KCardGrid ...>
      ...
    </KCardGrid>
  </template>

 <template #javascript>
     export default { 
        ...
       }
  </template>
  
  <template #scss>
     .class {
        ...
      }
  </template>
 </DocsExample>
```

In both the examples above, we can configure it to link to the complete example files through a GitHub link in the navigation bar. The slots here would have the highest priority. 

3. If we want to show some code to the user without explicitly linking to any file, we can skip the `component` prop altogether!
```vue
<DocsExample>
 <template #html>
    <KCardGrid ...>
      ...
    </KCardGrid>
  </template>

 <template #javascript>
     export default { 
        ...
       }
  </template>
 </DocsExample>
```
The output would be similar to (2), but no link to any GitHub documentation file.

MisRob:
> Yes! If the API contract for the same is not fixed

Nono, no fixed :) I am opening this discussion because I am suspicious you may be interested in it and pretty confident we will come up with something nice :)

Thanks a lot for the comment, I will read it with fresh eyes on Thursday - I am just signing off and tomorrow am out for some errands.

EshaanAgg:
> I see. Yeah I think it'd be fine, even though perhaps we can brainstorm a bit? 

Most definitely :)

While designing this interface, my main idea was to place only complete SFCs as examples in the `examples` directory. Since we are assured that the examples in this directory are complete, we can dynamically load their template with props as described above. Keeping this in mind, 

1. I would prefer the example files not to have any `Docs*` tags present in them like: 

> _examples/example1.vue_
> 
> ```
> <DocsExample>
> 
> </DocsExample>
> ```

The same is more of a semantic choice for me. I would prefer the examples if they can be considered independent "drag & drop" SFCs that I can take inspiration from and use directly in my code from the GitHub files. Having these `Doc*` tags in them makes us aware of the internal implementation of the documentation and, in my opinion, defeats the purpose of moving the component to a separate file (as now the user can't simply copy and paste the same into their apps).

2. > All `examples/` could be ignored by lint.

I would like to have linting enabled for them as well, as it would ensure that the code we showcase in the documentation is properly formatted and "idiomatic". We can perhaps fine-tune the lining configuration for the `examples` directory.

A simple example of the same can be found in this PR itself: we have the `no-console-log` rule configured in the KDS source code as it makes sense from the POV of a library, but when writing examples, it might be completely justified to use `console.log` to convey intended behavior. 

https://github.com/learningequality/kolibri-design-system/blob/a6f7a8dc3e1d9da1ab905f3c671cead9508fdde4/docs/examples/KTable/DisableBuiltinSorting.vue#L35-L41

It would make sense to turn this rule off for this directory to eliminate the `eslint-disable` comment. We can add more such rules as needed, but switching off all lining seems like a bad default to me.

Under typical usage, we would only need to disable linting when we provide custom display slots for the templates. In such cases, the developer can either:
- Inline them in the template if the snippet is small
- For each section of a documentation page of considerable length, create a separate component in a new file (with an `eslint-disable` for the whole file), then render that component in the documentation page itself. The component should live close to the documentation page (preferably in the same directory as the same). 

3. >Also I am not sure if we need another component `DocsLoadExample` just to load the example.

I would also argue against it. I would prefer one component whose behavior I could customize to support all my use cases rather than having super-focused components for each related use case. The `DocsShow` component currently seems to be the best place to introduce it, though `DocsExample` is where it would be realized the best. 

4. > I also want to keep thinking whether we can make all these perspectives work with auto-generation as you suggest originally in this PR. 

Lost a bit of context here. Autogeneration as in that of the `eslint-disable` comments?

MisRob:
Hi @EshaanAgg, I really like the way you connected everything [here](https://github.com/learningequality/kolibri-design-system/pull/905#issuecomment-2619882993) and also the latest comment make sense. The interface you suggest is very well aligned with the experience we roughly imagined, and it feels very intuitive.

Regarding (3)

> I would also argue against it. I would prefer one component whose behavior I could customize to support all my use cases rather than having super-focused components for each related use case. The DocsShow component currently seems to be the best place to introduce it, though DocsExample is where it would be realized the best.

The motivation for this was that if we did this (together with disabling lint and prettier for _/examples_), we wouldn't need

```
<!-- eslint-disable -->
<!-- prettier-ignore -->
```

even in cases we are showing the customized code sample via those `#html`, `#javascript`, and `#scss` slots, which for some pages is quite often.

However, I agree that it's cumbersome and harder to use, compared to your proposal. And besides other benefits you mention, I think having usual Vue components, without `Docs..`, in _examples/_ will scale better, e.g. for automated QA, or whatever we need to.

Since what we discuss here is meant to be an additional improvement, and I don't really have any concrete ideas how to further improve it, I think we can move on [with this](https://github.com/learningequality/kolibri-design-system/pull/905#issuecomment-2619882993). The main thing is that we maintain the flexibility to override and this is satisfied. I discussed this work with the team recently and I believe it's aligned well with our thinking. Also, as soon as we have `DocsExample`, code samples won't be visible on first page load so if there's occasionally too much code, it won't be such a big deal like right now.

MisRob:
@EshaanAgg It's true this would be best done right in `DocsExample`. The feedback I left in https://github.com/learningequality/kolibri-design-system/pull/873 is non-blocking and I can coordinate with the contributor and ask them if it's fine we push them ourselves or they could resolve it in a follow-up PR. If it'd be helpful, let me know.

MisRob:
Also do you think you'd need to have https://github.com/learningequality/kolibri-design-system/issues/845 resolved? Since we're changing API here a bit, and that issue is not currently in progress, I don't mind coordinating some re-assignments if it makes this work easier. There's much more work coming in `DocsExample` waters that I can offer to other contributors instead.

You're welcome to suggest other strategies - I haven't yet looked into code so I am not sure what would make this smooth process.

EshaanAgg:
Thanks, @MisRob, for the kind comments! I do indeed feel that the best place to add this dynamic loading would be in the `DocsExample` component, as that is the principal component for which we finally intend to have this DX. I think the best way to proceed can be:
- Merge #873 into the repository as a proof of concept component of the UI & animated interactions that we want the component to have.
- I can then continue the work there and configure the dynamic loading of the examples from SFCs as I have done in this PR, which is directly into the same. 

After that, we can try rolling out the component on our documentation pages! I think that would be a great middle-ground where we can make sure that there in the most reuse of the work already done.

MisRob:
Great @EshaanAgg, thanks a lot. https://github.com/learningequality/kolibri-design-system/pull/873 is merged so feel free to build on top of it.

EshaanAgg:
Thanks a lot, @MisRob, for the heads up! I'll close this PR for now, and open a new PR as soon as possible!

MisRob:
Take your time @EshaanAgg

PR #913, opened 2025-01-29
Remove unnecessary scrollbar when using KCircularLoader (#210)

### LINKED ISSUE #210: KModal often displays a vertical scrollbar when used with KCircularLoader, even when there is enough space in the modal/on a page
## Product

Studio, Design System

## Expected behavior

`KModal` never displays a vertical scrollbar when used with `KCircularLoader` and when there is enough space in the modal/on a page.

## Actual behavior

`KModal` often displays a vertical scrollbar when used with `KCircularLoader`, even when there is enough space in the modal/on a page.

**This screenshot was taken in my Storybook in Chrome where I checked this problem in isolation for the following code** ([you can see the Storybook bundle uploaded here](https://frosty-leakey-98684a.netlify.app/?path=/story/kmodal--with-loader)):
```
<KModal title="Title"><KCircularLoader/></KModal>
```

modal-loader

**This seems to be the primary cause of an issue with vertical bar appearing and disappearing when moving a mouse cursor in Studio:**

There may be another problem, but the vertical scrollbar shouldn't be rendered at all in the first place.

Peek 2021-03-30 10-02

1. Use `KCircularLoader` in the content area of `KModal` (it occasionally happened to me I needed to reload the page to reproduce the bug)

## Additional information

Initial investigation shows that this may be caused by inaccurate calculation in [`updateContentSectionStyle`](https://github.com/learningequality/kolibri-design-system/blob/v0.2.x/lib/KModal.vue#L244).

When the modal is rendered for the first time, [`contentHeight` is updated](https://github.com/learningequality/kolibri-design-system/blob/v0.2.x/lib/KModal.vue#L247). That triggers the` update` lifecycle hook, which causes `updateContentSectionStyle` to be called again - but in this second run, the content area's `scrollHeight` is a couple of pixels larger than its `clientHeight` (usually it's 2 or 3 pixels though the value may vary) which [causes the scrollbar to be displayed](https://github.com/learningequality/kolibri-design-system/blob/v0.2.x/lib/KModal.vue#L266). I am not sure why `scrollHeight` slightly differs from `clientHeight` - I suspect it may have something to do with the loader overflow.

## Environment
- OS: Ubuntu
- Browser version: Chrome, Firefox

karankoder:
## Description  
Fixes #210 where `KModal` displays an unnecessary vertical scrollbar when used with `KCircularLoader`, even when there is enough space in the modal/page.  

#### Issue addressed  
Addresses #210   

### Demo Video 
(https://github.com/user-attachments/assets/f086ab15-6b47-4241-8cab-5b99d52ee4c6)  

## Changelog  
  

- **Description:** Prevents unwanted vertical scrollbar in `KModal` when using `KCircularLoader`.  
- **Products impact:** bugfix  
- **Addresses:** #210   
- **Components:**  `KCircularLoader`  
- **Breaking:** no  
- **Impacts a11y:** no  
- **Guidance:** No action required from consumers. The fix applies automatically.  

  

## Steps to test  

1. Open a `KModal` that contains a `KCircularLoader`.  
2. Ensure no vertical scrollbar appears when there is sufficient space.  
3. Verify modal content remains scrollable when needed.  

### How was this implemented?  
- Adjusted the styling of `KCircularLoader` to prevent forced overflow behavior.  

### Does this introduce any tech debt?  
No new tech debt introduced.  

## Testing Checklist  

- [ ] Contributor has tested the PR manually.  
- [ ] Verified that no vertical scrollbar appears unnecessarily.  
- [ ] Included before/after screenshots if applicable.  
- [ ] Ensured no regressions in modal scrolling behavior.  

## Reviewer Guidance  

- [ ] Does the fix properly prevent unwanted scrolling?  
- [ ] Is the modal still fully functional with varying content sizes?

### DIFF
--- a/lib/loaders/KCircularLoader.vue
+++ b/lib/loaders/KCircularLoader.vue
@@ -276,8 +276,9 @@
 
   .ui-progress-circular--transition-fade-enter-active,
   .ui-progress-circular--transition-fade-leave-active {
-    transition: opacity 0.3s ease;
-    transform: 0.3s ease;
+    transition:
+      opacity 0.3s ease,
+      transform 0.3s ease;
   }
 
   .ui-progress-circular--transition-fade-enter,
--- a/lib/loaders/KCircularLoader.vue
+++ b/lib/loaders/KCircularLoader.vue
@@ -235,9 +235,10 @@
   .ui-progress-circular {
     position: relative;
     margin: 0 auto;
+    overflow: hidden;
     animation: fade-in;
-    animation-fill-mode: backwards;
     animation-duration: 0s;
+    animation-fill-mode: backwards;
   }
 
   .ui-progress-circular-determinate {
@@ -275,9 +276,8 @@
 
   .ui-progress-circular--transition-fade-enter-active,
   .ui-progress-circular--transition-fade-leave-active {
-    transition:
-      opacity 0.3s ease,
-      transform 0.3s ease;
+    transition: opacity 0.3s ease;
+    transform: 0.3s ease;
   }
 
   .ui-progress-circular--transition-fade-enter,


AlexVelezLl (approved):
Thanks @karankoder! LGTM! an overflow: hidden in this context makes sense to me, and I have played a bit with it in Kolibri and had not observed any regression. Thanks!

PR #920, opened 2025-02-02
[KListWithOverflow]: Add unit tests

### LINKED ISSUE #833: [KListWithOverflow]: Add unit tests

## Product
KDS.

## Desired behavior

We should add unit tests to [KListWithOverflow](https://design-system.learningequality.org/klistwithoverflow/), since it has some complicated calculations and needs to be protected from regressions. So we need to test at least the following things:

* It should render the whole list items if it fits in the available space, even if there is just 1px of difference.
* It shouldnt render the `more` slot if the list items fits in the available space.
* It shouldnt render the items that are overflowed and dont fit in the available space even if there is just 1px of difference.
* It should render the `more` slot if there are overflowed items.
* It should pass the correct items in the `overflowItems` array passed to the `more` slot.
* It shouldnt render a divider as the last item in the visible list, even if it fits in the available space.

## Current behavior

We currently dont have unit tests for KListWithOverflow.

you-think-you-know-me:
## Description
We currently don't have unit tests for KListWithOverflow. It adds unit tests for KListWithOverflow.

#### Issue addressed
Closes #833

## Changelog

  - **Description:** Add unit tests for KListWithOverflow
  - **Products impact:** NA
  - **Addresses:**  #833 
  - **Components:** KListWithOverflow
  - **Breaking:** Will this change break something in a consumer? no
  - **Impacts a11y:** Does this change improve a11y or adds new features that can be used to improve it? no
  - **Guidance:** NA

## Testing checklist

## Reviewer guidance

## Comments
:smile:

### DIFF
--- a/lib/__tests__/KListWithOverflow.spec.js
+++ b/lib/__tests__/KListWithOverflow.spec.js
@@ -87,4 +87,38 @@ describe('KListWithOverflow.vue', () => {
     expect(listItems.at(1).element).toHaveStyle({ visibility: 'hidden', position: 'absolute' });
   });
 
+  it('passes correct overflow items to "more" slot', async () => {
+    const items = ['Item 1', 'Item 2'];
+
+    const wrapper = mount(KListWithOverflow, {
+      propsData: { items },
+      slots: {
+        item: '<div class="list-item">{{ item }}</div>',
+      },
+      scopedSlots: {
+        more(props) {
+          return this.$createElement('div', { class: 'more-slot' }, String(props.overflowItems));
+        },
+      },
+    });
+
+    await waitForNextTicks(wrapper);
+
+    setElementSize(wrapper.find('.list-wrapper').element, { width: LIST_WRAPPER_WIDTH, height: 30 });
+    setElementSize(wrapper.find('.more-button-wrapper').element, { width: MORE_BUTTON_WIDTH, height: 20 });
+
+    const listItems = wrapper.findAll('.list-item');
+    setElementSize(listItems.at(0).element, { width: 80, height: 20 });
+    setElementSize(listItems.at(1).element, { width: 25, height: 20 });
+
+    wrapper.vm.setOverflowItems();
+    await wrapper.vm.$nextTick();
+
+    expect(wrapper.vm.overflowItems).toEqual(['Item 2']);
+    expect(wrapper.vm.isMoreButtonVisible).toBe(true);
+
+    const moreSlot = wrapper.find('.more-slot');
+    expect(moreSlot.exists()).toBe(true);
+    expect(moreSlot.text()).toBe('Item 2');
+  });
 });
--- a/lib/__tests__/KListWithOverflow.spec.js
+++ b/lib/__tests__/KListWithOverflow.spec.js
@@ -26,7 +26,7 @@ async function waitForNextTicks(wrapper) {
 describe('KListWithOverflow.vue', () => {
   it('renders all items when they fit within available space', async () => {
     const items = ['Item 1'];
-    
+
     const wrapper = mount(KListWithOverflow, {
       propsData: { items },
       slots: {
@@ -38,8 +38,14 @@ describe('KListWithOverflow.vue', () => {
     await waitForNextTicks(wrapper);
 
     // Set dimensions for container and button
-    setElementSize(wrapper.find('.list-wrapper').element, { width: LIST_WRAPPER_WIDTH, height: 30 });
-    setElementSize(wrapper.find('.more-button-wrapper').element, { width: MORE_BUTTON_WIDTH, height: 20 });
+    setElementSize(wrapper.find('.list-wrapper').element, {
+      width: LIST_WRAPPER_WIDTH,
+      height: 30,
+    });
+    setElementSize(wrapper.find('.more-button-wrapper').element, {
+      width: MORE_BUTTON_WIDTH,
+      height: 20,
+    });
 
     // Simulate item size (should fit within available space)
     const listItem = wrapper.find('.list-item');
@@ -54,10 +60,10 @@ describe('KListWithOverflow.vue', () => {
     expect(wrapper.vm.isMoreButtonVisible).toBe(false);
     expect(listItem.element).toHaveStyle({ visibility: 'visible', position: 'unset' });
   });
-  
+
   it('hides overflow items and displays "more" button when items do not fit', async () => {
     const items = ['Item 1', 'Item 2'];
-    
+
     const wrapper = mount(KListWithOverflow, {
       propsData: { items },
       slots: {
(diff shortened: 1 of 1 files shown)


lib/__tests__/KListWithOverflow.spec.js
```diff
@@ -0,0 +1,182 @@
...
+    bottom: height,
+  });
+}
+
+// Constants for container and button dimensions
+const LIST_WRAPPER_WIDTH = 101;
+const MORE_BUTTON_WIDTH = 20;
+
+// Utility function to wait for two nextTicks after mounting
+async function waitForNextTicks(wrapper) {
```
AlexVelezLl, line 21: Is this method actually needed? I just removed it and all tests kept working fine.

lib/__tests__/KListWithOverflow.spec.js
```diff
@@ -0,0 +1,182 @@
...
+    expect(moreSlot.text()).toBe('Item 2');
+  });
+
+  it('divider is not the last visible item', async () => {
+    const items = ['Item 1', { type: 'divider', label: 'DIVIDER' }, 'Item 3'];
+
+    const wrapper = mount(KListWithOverflow, {
+      propsData: { items },
+      slots: {
+        item: '<div class="list-item">{{ typeof item === "object" ? item.label : item }}</div>',
```
AlexVelezLl, line 149: We dont need this ternary here, because when we have a divider we dont render the item slot.

lib/__tests__/KListWithOverflow.spec.js
```diff
@@ -0,0 +1,182 @@
...
+    setElementSize(listItems.at(0).element, { width: 60, height: 20 });
+    setElementSize(listItems.at(1).element, { width: 20, height: 20 });
+    setElementSize(listItems.at(2).element, { width: 30, height: 20 });
+
+    wrapper.vm.setOverflowItems();
+    await wrapper.vm.$nextTick();
+
+    expect(wrapper.vm.overflowItems).toEqual(['Item 3']);
+    expect(wrapper.vm.isMoreButtonVisible).toBe(true);
+    expect(listItems.at(1).element).toHaveStyle({ visibility: 'hidden' });
```
AlexVelezLl, line 180: We can be a little bit more explicit with the code so its easier to understand for other people. For example here, we can tweak this code extracting variables and adding comments to be more expressive with the intention. For example:

const divider = listItems.at(1).element;
// The divider its the last non-overflowed item, it should be hidden
expect(divider).toHaveStyle({ visibility: 'hidden' });

lib/__tests__/KListWithOverflow.spec.js
```diff
@@ -0,0 +1,182 @@
...
+  });
+});
```
AlexVelezLl, line 182: Lets add just a couple of tests more:
* It renders the divider slot if we pass a divider object.
* The first overflowed item shouldnt be a divider
* All items are shown if they fit if we dont render the more button
  * This means: If our `LIST_WRAPPER_WIDTH` is 101, we can still render two items of width 50 for example.
  * With this we will find an error because [in this line](https://github.com/you-think-you-know-me/kolibri-design-system/blob/218add7d6d98829517dd862ad5de9a25f3bff666/lib/KListWithOverflow.vue#L190) we set the item visibility to `visible` but we dont restore the position to `unset`. So if you can include this line too it would be nice.

AlexVelezLl (changes requested):
Thanks a lot @you-think-you-know-me! This looks super good! I have left just some minor comments, and requested a couple of additional test cases, but apart from that everything else looks good!

EshaanAgg:
Hi! I was passing by and felt that the way we are trying to test this component is a bit counter-intuitive: we are trying to set the sizes of the elements via JS, which is not the way the component would be used! Most probably, the user would have a pre-defined markup for the `item` slot, which will be passed to it, and the size of the `item` would be determined by it's styling! So shouldn't our tests also try to mimic the same?

Something like following perhaps?

```js
import KListWithOverflow from '../KListWithOverflow.vue';

// Constants for styling
const LIST_STYLES = `
  .list-item {
    display: inline-block;
    white-space: nowrap;
  }
  .list-item.fits {
    width: 80px;
    height: 20px;
  }
  .list-item.overflows {
    width: 25px;
    height: 20px;
  }
  .list-item.medium {
    width: 60px;
    height: 20px;
  }
  .list-item.small {
    width: 20px;
    height: 20px;
  }
`;

describe('KListWithOverflow.vue', () => {
  // Add styles to document before tests
  // TODO: Check if it actually works with Enzyme
  beforeAll(() => {
    const styleElement = document.createElement('style');
    styleElement.textContent = LIST_STYLES;
    document.head.appendChild(styleElement);
  });

  it('renders all items when they fit within available space', async () => {
    const items = ['Item 1'];

    const wrapper = mount(KListWithOverflow, {
      propsData: { items },
      slots: {
        // Can now just use the class on the list item to control its size
        // The class clearly indicates the size of the item and the intent of the test
        item: '<div class="list-item fits">{{ item }}</div>',
        more: '<button class="more-button">More</button>',
      },
    });

    await wrapper.vm.$nextTick();

    expect(wrapper.vm.overflowItems).toEqual([]);
    expect(wrapper.vm.isMoreButtonVisible).toBe(false);
    
    const listItem = wrapper.find('.list-item');
    expect(listItem.element).toHaveStyle({ 
      visibility: 'visible',
      position: 'unset'
    });
  });

  // Can easily write more tests by switching the class of the list item
});
```

This is just a brief idea, and I have never actually ran this code to ensure it works! but we probably should spend some time But seeing this PR, I thought that it might be worthwhile to look at this alternative approach. I would like to advocate for it as it does not make use of JS to control the sizing of the elements, and thus helps us to get rid of many "magic constants" that appear throughout the tests, or of having to manage the rendering ourselves from inside the tests.

What do you think about this @AlexVelezLl @you-think-you-know-me?

AlexVelezLl:
Hey @EshaanAgg, thanks for jumping in. Yes! This is a very good alternative too, although Im not a big fan of having to scroll up to the definition of `LIST_STYLES` and look for the specific width of a given item to understand the test, in that regard, i feel that having something like `setElementSize` express better the intention of the test by explicitly declaring the inputs and outputs within the same test case, and it makes it looks a bit more easy to read.

A cons of overwiting the `getBoundingClientRect` of the element is that if in the future we change this method to get the sizes, we will need to update this method too, although we will just only need to update it inside the `setElementSize` method as it is well encapsulated, but its nice to  have some alternatives.

I think we could implement something more complicated like defining unique classes dynamically in each test case and assigning them to the items inside a `setElementSize` method (or override the classes, although that could easily break the independency of the tests and one could affect the following ones), if this sounds good, @you-think-you-know-me can give it a try, but if we are spending too much time on it, we I think we can move forward with overriding the `getBoundingClientRect` method, at least for now, as the intention right now if making sure that the inner logic of `setOverflowItems` is working well. To have a better accuracy of the actual DOM rendering we can always use visual tests to complement these unit tests, although if in the future we found a better way to set these sizes, we can just change the `setElementStyle` method, and all the tests should keep working fine :).

EshaanAgg:
Thanks, @AlexVelezLl, for the detailed response. Even I do agree that having to scroll to the top of the document breaks the encapsulation of the test logic in the test itself approach and thus might be a bit counterintuitive. We can definitely ideate better ways to do the same later. Meanwhile, we will currently keep the main aim of this PR to introduce a test suite that covers all the edge cases! Thank a lot again :)

MisRob:
Hi @you-think-you-know-me, just checking in - are you still planning to come back to resolve few last notes from @AlexVelezLl's review?

AlexVelezLl:
Closing this as it has been over a month since there was any activity.

PR #923, opened 2025-02-03
Fix glob sync query

AlexVelezLl:
## Description

In #823, we removed the `globby` package in favor of `node-glob`, but it seems that due to a version mismatch we were using `const { GlobSync } = require('glob');` as the api, but it wasnt working well and when querying the files, it wasnt finding any, and the jsdocs result was an empty object that made our component docs to miss the API info:

image
 
This PR fixes this using `glob.sync` instead as suggested in [their docs](https://github.com/isaacs/node-glob?tab=readme-ov-file#globsyncpattern-string--string-options-globoptions--string--path) (and suggested by copilot :sweat_smile:).

## Changelog

  - **Description:** Fixes glob query in extrac api script.
  - **Addresses:** -.
  - **Components:** Docs.
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** .

## Steps to test

1. run `yarn dev` and check that the API sections (props, events, slots) are present in component docs

### DIFF
--- a/utils/extractApi.js
+++ b/utils/extractApi.js
@@ -1,7 +1,7 @@
 const path = require('node:path');
 const fs = require('node:fs');
 const docGenAPI = require('vue-docgen-api');
-const { GlobSync } = require('glob');
+const glob = require('glob');
 const consola = require('consola');
 
 /***********************************/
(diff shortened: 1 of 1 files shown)


MisRob:
For reference, I will link https://github.com/learningequality/kolibri-design-system/pull/823

MisRob (approved):
All good, thank you @AlexVelezLl for fixing KDS on so many fronts today :)!

AlexVelezLl:
Oh yes, it was #823 :laughing:, IDK why I linked another unrelated PR hahaha

PR #937, opened 2025-02-11
[Visual Testing] Allow object component rendering and unmount previoius component on render

AlexVelezLl:
## Description

This PR:
* Allow object components in the `renderComponent` method. This way we can render a custom test component withouth needing to register it in Vue.
* Clears the previous component that was rendered to clear context of previous test runs.
* Run visual tests on target branch after PR merge.
  * This is needed since we need the changes to run on the target branch, to have our baseline updated in Percy.
## Changelog

  - **Description:** Allow object component rendering and unmount previoius component on render
  - **Products impact:** internal
  - **Addresses:** .
  - **Components:** .
  - **Breaking:** 
  - **Impacts a11y:** 
  - **Guidance:**

### DIFF
--- a/.github/workflows/visual_tests.yml
+++ b/.github/workflows/visual_tests.yml
@@ -1,6 +1,12 @@
 name: Percy Visual Tests
 
-on: [pull_request_target]
+on:
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - closed
 
 jobs:
   pre_job:
@@ -15,20 +21,40 @@ jobs:
         with:
           github_token: ${{ github.token }}
           paths: '["lib/**", "jest.conf/visual*" ]'
+          # Avoid duplicate skipping so it can be run again on the target branch
+          skip_after_successful_duplicate: false
 
   visual_tests:
     name: Frontend Visual Tests
     needs: pre_job
     if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
     runs-on: ubuntu-latest
-    environment: percy_tests
+    strategy:
+      # Using matrix just to dynamically set the environment variable
+      matrix:
+        env:
+          # Protect deployments from non-merged PRs but allow merged PRs to run
+          # on the target branch
+          - ${{ github.event.pull_request.merged == false && 'percy_tests' || '' }}
+    environment: ${{ matrix.env }}
     outputs:
       percy_url: ${{ steps.extract-url.outputs.percy_url }}
     steps:
       - name: Checkout code from PR
+        if: github.event.pull_request.merged == false
         uses: actions/checkout@v4
         with:
           ref: ${{ github.event.pull_request.head.sha }}
+      - name: Checkout code from target branch
+        if: github.event.pull_request.merged == true
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.base.ref }}
+      - name: Set PERCY_BRANCH to target branch
+        if: github.event.pull_request.merged == true
+        # set the PERCY_BRANCH environment variable to the target branch for the Percy build
+        # to be associated with the target branch
+        run: echo "PERCY_BRANCH=${{ github.event.pull_request.base.ref }}" >> $GITHUB_ENV
       - name: Use Node.js
         uses: actions/setup-node@v4
         with:
@@ -61,7 +87,7 @@ jobs:
   comment:
     name: Comment Percy results
     needs: visual_tests
-    if: ${{ needs.visual_tests.result == 'success' }}
+    if: ${{ needs.visual_tests.result == 'success' && github.event.pull_request.merged == false }}
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code from PR
--- a/dev_docs/07_visual_testing_guide.md
+++ b/dev_docs/07_visual_testing_guide.md
@@ -88,14 +88,23 @@ KDS has a visual testing system that allows you to take snapshots of how KDS Com
 
           **Note:** Use `'default'` key for passing default slots, with the HTML content specified using `innerHTML` prop. Checkout [`KButton.spec.js`](../lib/buttons-and-links/__tests__/KButton.spec.js) for reference.
 
-      - **Example involving more complex component structures:** When dealing with more complex component structures, it's recommended to create a dedicated Vue component for visual testing purposes. This custom component can be placed in the same directory as the test file, inside a `components` folder. You can then import this component and render it using the `renderComponent` method.
+      - **Example involving more complex component structures:** When dealing with more complex component structures, it's recommended to create a dedicated test component for visual testing purposes.
 
-          ```javascript
-            import CustomVueComponent from './components/CustomVueComponent.vue';
+        * This custom component should be placed in the same directory as the test file, inside a `components` folder with a name following the pattern `K[TestCase]Test.vue`
+          * e.g.: `/lib/KImg/__tests__/components/KImgTest.vue`.
+        * Then you will need to declare this component in the the [visual.load-test-components.js](../jest.conf/visual.load-test-components.js) file so that its available for rendering in the visual testing playground.
+          * e.g.:
+            ```javascript
+            import KImgTest from '~~/lib/KImg/__tests__/components/KImgTest.vue';
 
             ...
 
-            await renderComponent(CustomVueComponent, { someProp: 'someValue' });
+            Vue.component('KImgTest', KImgTest);
+            ```
+        * Finally, you can then use the test component name in the `renderComponent` method.
+
+          ```javascript
+            await renderComponent('KImgTest', { someProp: 'someValue' });
           ```
 
           This approach ensures that all necessary child components and slots are correctly set up and rendered.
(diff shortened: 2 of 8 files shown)


bjester:
@AlexVelezLl Are there any gotchas with this approach? Has anyone else done anything similar?

AlexVelezLl:
I found this way of serializing functions here: https://stackoverflow.com/questions/7395686/how-can-i-serialize-a-function-in-javascript. I havent seen any gotcha with this approach yet, I tried rendering Custom components with different props and methods and it seems to be working fine. The alternative was to declare every custom component on Vue with Vue.component() in our nuxt plugins, but that would require one step more to use these custom components. What do you think?

bjester:
@AlexVelezLl The first potential problem that I see is that this won't account for function scope. Any of the functions could use variables or imports that were defined outside of the function scope, and therefore wouldn't necessarily have access to the same scope if the function was serialized to a string, then deserialized in a different context.

As the code is written currently, I also see these potential pitfalls:
```js
let x = y => y+1;
x.toString() === 'y => y+1' // true

let m = {a: 1, get b() { return this.a }};
structuredClone(m) // returns { a: 1, b: 1 }

let n = {a: 1, b() { return this.a }};
structuredClone(n) // Uncaught: DOMException [DataCloneError]: function() { return this.a } could not be cloned
```

EshaanAgg:
Just an alternate suggestion: 

Maybe we can refactor the `renderComponent` to do the following:
- If a string is passed, it will assume that the same is globally registered with `Vue` and will render it as it is. 
- If an object component is passed to the same, it would first register it with Vue (maybe with a unique name like `VisualTest_<RANDOM_ID>` and then pass this newly generated name in the message to the window.

We can also add a global cleanup function for all visual tests that de-registers all components starting with `VisualTest_` from the Vue instance to avoid unexpected bugs! This approach admittedly has a bit more of a redirection of logic, but it seems like a good way to handle the scope issues that @bjester talks about.

AlexVelezLl:
Hey @bjester! Yes, this is definitely a limitation of this approach. Should we accept this as a limitation in favour of the easy of use? If so, we should document it, something Its currently missing in the PR :sweat_smile:.

The limitation would be that we can use just use the `data()` and `props` properties in the options api of the test component. I honestly think that this should be enough for rendering the most complex examples, as at the end these test components are not the ones that will contain all the logic of our lib components, but for manage complex situations on the template that we cant currently manage with the current `{component, componentProps, slots}` (like having two anidated slots).

I  have tried to render something like this, for example, just simulating a case where we would want to render several KButtons based on an items prop, and were able to do it:

| component | percy result |
| --- | --- |
| image | image |

Another option can be that each time we create a new custom test component, we register the custom component in a new nuxt plugin just as we do with our docs components in [load-common-components.js](https://github.com/AlexVelezLl/kolibri-design-system/blob/5d7829543cc599937f2b489b60b72775ade3d513/docs/plugins/load-common-components.js#L1), is just an extra step that at this point I dont think is such a bad option either as it would reduce the complexity of the `renderComponent` method.

What do you think?

Thanks @EshaanAgg for the suggestion! The problem is that these tests live in two different contexts, the test runner and the nuxt server we run to render the components. Declaring the Vue component in the test runner wont make it be available in the nuxt server, so declaring it in the `renderComponent` method and then pass unique identifier to the postMessage wouldnt work :').

bjester:
@AlexVelezLl I'm okay with your approach, so if you're okay with the potential limitations, then it's fine with me! I was curious about the scope of the scope issue ( :sweat_smile: ) and so I searched for `lodash` imports which I think could hit the scope limitation:
```
lib/KDateRange/index.vue:  import get from 'lodash/get';
lib/KDateRange/index.vue:  import debounce from 'lodash/debounce';
lib/composables/useKShow/index.js:import has from 'lodash/has';
lib/cards/useGridLayout.js:import cloneDeep from 'lodash/cloneDeep';
lib/KModal.vue:  import debounce from 'lodash/debounce';
lib/KBreadcrumbs.vue:  import filter from 'lodash/filter';
lib/KBreadcrumbs.vue:  import startsWith from 'lodash/startsWith';
lib/KBreadcrumbs.vue:  import throttle from 'lodash/throttle';
lib/KBreadcrumbs.vue:  import every from 'lodash/every';
lib/KBreadcrumbs.vue:  import keys from 'lodash/keys';
lib/_Popper.vue:  import debounce from 'lodash/debounce';
lib/tabs/KTabsList.vue:  import merge from 'lodash/merge';
lib/keen/config.js:import merge from 'lodash/merge';
lib/KTable/useSorting/index.js:import _ from 'lodash';
lib/buttons-and-links/buttonMixin.js:import merge from 'lodash/merge';
lib/KTooltip/index.vue:  import isArray from 'lodash/isArray';
lib/KSelect/index.vue:  import has from 'lodash/has';
lib/KSelect/index.vue:  import isObject from 'lodash/isObject';
lib/KSelect/index.vue:  import throttle from 'lodash/throttle';
lib/KSelect/index.vue:  import startswith from 'lodash/startsWith';
lib/KSelect/index.vue:  import sortby from 'lodash/sortBy';
lib/KListWithOverflow.vue:  import throttle from 'lodash/throttle';
```

The other option, I'm not sure I'm understanding well, but maybe it's worth comparing the limitations of it?

Overall, this does seem challenging with our plan of combining multiple states in one render, to keep usage low. The only other idea I had was some sort of snapshot mechanism that captures the JSDOM output in some manner that could be passed as a self-contained DOM string (since I don't think interactivity is needed after render?), but IDK if that's even possible and if that would work with the combined-states-render we want.

MisRob:
@bjester @AlexVelezLl I don't understand all the context and discussion here, just glimpsed this note by Alex

> Another option can be that each time we create a new custom test component, we register the custom component in a new nuxt plugin just as we do with our docs components in [load-common-components.js](https://github.com/AlexVelezLl/kolibri-design-system/blob/5d7829543cc599937f2b489b60b72775ade3d513/docs/plugins/load-common-components.js#L1), is just an extra step that at this point I dont think is such a bad option either as it would reduce the complexity of the renderComponent method.

Besides simplicity and consistency with usual Vue code, using Vue components for testing could help with few other things:
- Tests of dependent components - e.g. it doesn't make much sense to test `KCard` on its own, we need to test it within `KCardGrid`
- Percy snapshots limitations - e.g. we could have `KCardGridTest` file that would contain several `KCardGrid`s with `KCard`s - I believe that would be one snapshot instead of potentially many?
- Over time could integrate well with the documentation examples that are now being refactored into single files (like [here](https://github.com/learningequality/kolibri-design-system/pull/956))

If I can understand this is already possible to do - just saying that  we may even welcome if it'd be the default test mechanism approach.

MisRob:
> we register the custom component in a new nuxt plugin just as we do with our docs components in [load-common-components.js](https://github.com/AlexVelezLl/kolibri-design-system/blob/5d7829543cc599937f2b489b60b72775ade3d513/docs/plugins/load-common-components.js#L1)

If we had one Vue test file/component for each component, then we wouldn't even need to do this frequently.

I hope it's not completely off as I haven't yet written a test suite myself and as mentioned, I don't quite get all the nuances of problems being discussed here. Take my comments as a high-level input - if it's not helpful, feel free to set it aside :)

AlexVelezLl:
Thanks @MisRob!

> If we had one Vue test file/component for each component, then we wouldn't even need to do this frequently.

Yes, that's true. I think the trade-off of having an extra step for defining a test component but avoiding any limitations or unexpected behaviors that could become complex to fix is worth it.

I'm gonna write an example test just for us to have a sense of how this alternative approach would look like :)

AlexVelezLl:
Okay, it was pretty simple to configure, I have just conditionally added this new plugin to load visual tests components just when we run the server for visual tests: https://github.com/learningequality/kolibri-design-system/pull/937/commits/34e9cc454ca05460621e155f5fe3c33815547dc3.

And an example of how we would use custom test components would be basically something like this: https://github.com/learningequality/kolibri-design-system/pull/937/commits/6e3757339bc1c18070f4fc673774d23e01a2ad32. This particular example is real example of something that we cant render using the current `renderComponent` props due to the complexity of the menu slot.

With this, we wont have any limitations, nor more complex rendering logic, just a matter of documenting it so people we need to define these tests components in this new plugin.

Couple of small decisions:
* I have defined this test component in the same component test folder `KButton/__test__` but inside another `components` folder, and have the component name as K[TestCase]Test.vue. These are some conventions we can encourage.
* I added this new load-test-components file in the `jest.conf` folder rather than in `docs/plugins` where the other two plugins live since I think its a more logical place to be.

What do you think? @bjester @MisRob 

(apologies for the late responses :sweat_smile:)

bjester:
@AlexVelezLl LGTM :+1:

AlexVelezLl:
Thanks @bjester! I have updated the docs and removed the old serializeComponent logic :^), I have left the KButton test example so we can have our [visual.load-test-components.js](https://github.com/learningequality/kolibri-design-system/pull/937/files#diff-ada10de317adf2ebcf0259662f30a85eccfe62bea0cf7197da57a0fa271e272b) already set up.

bjester (approved):
I checked out the Percy page and it seems to work?! So I think it's all good

MisRob:
I'm sorry for late response, did a quick skim now and it looks very nice and intuitive for me. I will soon finally write one for `KCardGrid` and `KCard` so I will play around with it and if I have some further ideas, I will open a a proposal issue. If you both are fine with it, I am for merge :) Thank you so much for taking care of this.

AlexVelezLl:
>  I will soon finally write one for KCardGrid and KCard so I will play around with it and if I have some further ideas, I will open a a proposal issue

Sounds good! Thanks @MisRob! Merging.

PR #942, opened 2025-02-25
Match the hover background color of flat-button KButton to that of KIconButton

AlexVelezLl:
## Description

When we have a KButton on top of non-white backgrounds, we see a gray background color on hover:

image

This PR fixes this by using the same black, semi transparent background color on hover that uses KIconButton:

image

image

Note:

This is changing a little bit the hover background color when the button is on top of white backgrounds. Is there any problem with this? Should we use a higher number for the background opacity?

| Before | After |
| --- | --- |
| image | image |

## Changelog

  - **Description:** Match the hover background color of flat-button KButton to that of KIconButton
  - **Products impact:** bugfix
  - **Addresses:** -
  - **Components:** KButton.
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:**

### DIFF
--- a/lib/buttons-and-links/KIconButton.vue
+++ b/lib/buttons-and-links/KIconButton.vue
@@ -108,8 +108,6 @@
     },
     computed: {
       appearanceOverrides() {
-        const hover =
-          this.appearance === 'flat-button' ? { backgroundColor: 'rgba(0,0,0,.1)' } : {};
         return {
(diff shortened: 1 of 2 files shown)


marcellamaki (approved):
Thank you @AlexVelezLl! Jessica confirmed on slack this is 👍 on the design side as well

PR #961, opened 2025-03-14
`KTable`: Migrate the test suite to `VTL` and add new tests

Pandaa007:
## Description
Migrated the test suite from enzyme to `VTL` for `KTable` so that we use the accessible roles and user-facing queries to test the behaviour of the component.
- Reduced the boilerplate associated with the exsiting tests to make the tests conscise and more expressive. 
- Added a stronger test suite for arrow key naviagation. 
- Added a new simple test suite as a proof of concept for tab and shift-tab navigation.
- Restructured the tests so that all of live under a directory.

## Changelog

  - **Description:** Enhanced the test coverage for KTable
  - **Products impact:** none
  - **Addresses:** -
  - **Components:** -
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:**  -
  

## Testing checklist

## Reviewer guidance


MisRob:
Thanks @Pandaa007! @BabyElias you're welcome to assign yourself as a reviewer if you'd like, or I can ask another reviewer later on.

lib/KTable/__tests__/index.spec.js
```diff
@@ -0,0 +1,168 @@
...
+    renderComponent();
+
+    const headerCells = getRenderedHeaders();
+    headerCells.forEach(headerCell => {
+      expect(headerCell).toHaveClass('sticky-header');
+    });
+
+    defaultRows.forEach(row => {
+      const firstColumnCell = screen.getByRole('gridcell', { name: row[0] });
+      expect(firstColumnCell).toHaveClass('sticky-column');
```
EshaanAgg: Can we move this class `sticky-column` to a separate constant at the top of the test file? It helps to avoid the problem of magic constants.

lib/KTable/__tests__/index.spec.js
```diff
@@ -0,0 +1,168 @@
...
+      expect(allCells[0][0]).toHaveFocus();
+
+      // Record the key presses and the expected next cell
+      const keyPresses = [
+        { key: 'ArrowDown', nextCell: allCells[1][0] },
+        { key: 'ArrowRight', nextCell: allCells[1][1] },
+        { key: 'ArrowUp', nextCell: allCells[0][1] },
+        { key: 'ArrowLeft', nextCell: allCells[0][0] },
+      ];
+      assertFocusWithKeyPresses(keyPresses);
```
EshaanAgg: Nice work on this utility here. I love how easy it has become to test the focus and keyboard now.

lib/KTable/__tests__/tab.spec.js
```diff
@@ -0,0 +1,51 @@
...
+      for (let rowIdx = defaultRowCount; rowIdx >= 0; rowIdx--) {
+        for (let colIdx = defaultColCount - 1; colIdx >= 0; colIdx--) {
+          // Press tab to move to the foucs to the current cell
+          await userEvent.tab({ shift: true });
+
+          expect(allCells[rowIdx][colIdx]).toHaveFocus();
+          assertActiveBackground(rowIdx, colIdx, allCells);
+        }
+      }
+    });
```
EshaanAgg, line 133: Can we add some more test cases here? A test case with some focusable elements inside some of the cells might be good to have. 

Also `KTable` has a hovering functionality, wherein if we use the pointer to hover over any cell, the whole row is highlighted. So it might be nice to have a test suite that:
- Checks this behaviour only when the pointer is used
- When pointer and tab focus are used on cells in
  - Different rows
  - Same row but different cells
  - Same cell

This test suite might be a bit too complex for the initial draft, and would help to safeguard a lot of regressions going forward in a component as nuanced as `KTable`. If you feel that this might be a but too much to implement for the scope of this PR, please feel free to drop a comment and we can try to address them in a new PR.

lib/KTable/test-utils.js
```diff
@@ -0,0 +1,153 @@
...
+ */
+export const assertFocusWithKeyPresses = async keyPresses => {
+  for (const { key, nextCell } of keyPresses) {
+    await nextCell.focus();
+    await userEvent.type(nextCell, key);
+    expect(nextCell).toHaveFocus();
+  }
+};
+
+const HIGHTLIGHT_BACKGROUND_COLOR = themePalette().grey[100];
```
EshaanAgg: Move all the magic constants to the top of the test file.

lib/KTable/test-utils.js
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * Simulates focus management by typing the specified keys and asserting that
+ * the expected cell has focus after each key press.
+ * @param {Array} keyPresses - An array of objects containing the key to press
+ * and the expected next cell to have focus
+ */
+export const assertFocusWithKeyPresses = async keyPresses => {
+  for (const { key, nextCell } of keyPresses) {
+    await nextCell.focus();
```
EshaanAgg: Seems like a bug. We shouldn't actually be calling `nextCell.focus()` but only using it as a reference for the assertion.

```suggestion
```

lib/KTable/test-utils.js
```diff
@@ -0,0 +1,153 @@
...
+    await userEvent.type(nextCell, key);
+    expect(nextCell).toHaveFocus();
+  }
+};
+
+const HIGHTLIGHT_BACKGROUND_COLOR = themePalette().grey[100];
+
+/**
+ * Asserts that the active cell has the expected background color and that
+ * all other cells do not have the background color.
```
EshaanAgg: Can you also update the comment to convey that all complete row of the active cell is also allowed to be highlighted? It would also be good to have `Tab` navigation explicitly mentioned in the description of this function.

EshaanAgg (commented):
Thanks @Pandaa007 for your contribution. Overall I am a huge fan of how expressive the test suite now has become, specially with the use of proper role selectors to check the rendered content instead of using DOM selectors as we had to earlier do with `enzyme`. The introduction of the multiple utilities is also much appreciated. 

I have left a few comments, most of them can be safely considered as nits. Feel free to address them as you have the time, or ignore them if they feel irrelevant. Thanks again.

lib/KTable/test-utils.js
Pandaa007: Renamed the function and added comments to it in the lastest commits for better clarity.

lib/KTable/test-utils.js
Pandaa007: Nice catch! Fixed in the latest commits.

lib/KTable/test-utils.js
Pandaa007: Done.

lib/KTable/__tests__/index.spec.js
Pandaa007: Done.

lib/KTable/__tests__/tab.spec.js
Pandaa007: Thanks for the suggestions @EshaanAgg. I have added some basic tests related to the pointer functionality as well as the focusable element functionality as you directed in the latest commits. I think so the test suite is already quite involed as of now, so I didn't add the tests composing the pointer and tab functionality together in this PR.

I would like to open a new PR for them in which I would add tests for:
- Multiple focusable elements in a cell
- Having disabled elements in a cell
- Checking that the navigation doesn't remain trapped inside or outside the table

I think that this functionality would be better suited there. Hope that is a right scale of balance for this PR and future work.

lib/KTable/test-utils.js
```diff
@@ -0,0 +1,181 @@
+import { render, screen } from '@testing-library/vue';
```
AlexVelezLl, line 1: Shouldnt this file be better placed inside the `__tests__` folder?

lib/KTable/test-utils.js
```diff
@@ -0,0 +1,181 @@
...
+import userEvent from '@testing-library/user-event';
+import { expect } from '@jest/globals';
```
AlexVelezLl, line 3: We dont actually need to import the expect method here. This is already a global variable and we have the convention of using these global variables instead of importing them from the jest package.

AlexVelezLl: Nvm I think its because of the linter not recognizing this as a test file, right? In that case I think its better just to have the import statement rather than renaming this file to `.spec.js` :sweat_smile:.

lib/KTable/__tests__/index.spec.js
```diff
@@ -0,0 +1,211 @@
...
+  assertNotRowHighlighted,
+  assertRowHighlighted,
+} from '../test-utils';
+
+const STICKY_HEADER_CLASS = 'sticky-header';
+const STICKY_COLUMN_CLASS = 'sticky-column';
+
+describe('KTable.vue', () => {
+  beforeEach(() => {
+    // Since our primary concern in this test suite is checking focus management rather
```
AlexVelezLl: We are not checking focus management here or yes?

lib/KTable/__tests__/index.spec.js
```diff
@@ -0,0 +1,211 @@
...
+
+    it('should sort the rows correctly in ascending order', () => {
+      renderComponent({
+        defaultSort: {
+          columnId: 'age',
+          direction: 'asc',
+        },
+      });
+
+      const sortedRows = defaultRows.slice().sort((a, b) => a[1] - b[1]);
```
AlexVelezLl, line 60: Lets try to be a little bit more verbose about this, perhaps with some comments about why are we using the index 1 here. Its a little bit difficut for a new reader to figure out what is happening here, and that this number is conected with the columnId specified above.

lib/KTable/__tests__/index.spec.js
```diff
@@ -0,0 +1,211 @@
...
+
+    it('should sort the rows correctly in descending order', () => {
+      renderComponent({
+        defaultSort: {
+          columnId: 'age',
+          direction: 'desc',
+        },
+      });
+
+      const sortedRows = defaultRows.slice().sort((a, b) => b[1] - a[1]);
```
AlexVelezLl, line 74: idem

lib/KTable/__tests__/index.spec.js
```diff
@@ -0,0 +1,211 @@
...
+        sortable: true,
+        disableBuiltinSorting: true,
+      });
+
+      const ageHeaderCell = screen.getByRole('columnheader', { name: /age/i });
+      expect(ageHeaderCell).toBeInTheDocument();
+      await userEvent.click(ageHeaderCell);
+
+      expect(emitted()).toHaveProperty('changeSort');
+      expect(emitted().changeSort[0]).toEqual([1, null]);
```
AlexVelezLl: Why should this equal to [1, null]? We can add some comments here or better variables names about why are we asserting against this value

lib/KTable/__tests__/index.spec.js
```diff
@@ -0,0 +1,211 @@
...
+
+      // Record the key presses and the expected next cell
+      const keyPresses = [
+        { key: 'ArrowDown', nextCell: allCells[1][0] },
+        { key: 'ArrowRight', nextCell: allCells[1][1] },
+        { key: 'ArrowUp', nextCell: allCells[0][1] },
+        { key: 'ArrowLeft', nextCell: allCells[0][0] },
+      ];
+      assertFocusWithKeyPresses(keyPresses);
+    });
```
AlexVelezLl, line 263: I think we should also add some tests with the expected behaviour when we are in the first row, and press arrow up, or if we are in the last row and press arrow down, the same with left/right too

lib/KTable/__tests__/tab.spec.js
AlexVelezLl: Hi @Pandaa007, Would you like to open a follow-up issue for this? We can assign that to you if you wish, but it's not strictly necessary; we can assign it to someone else, too :).

AlexVelezLl (changes requested):
Thanks @Pandaa007! This looks good! I have left some nitpicky comments, please let me know what you think :)

MisRob:
Hi @Pandaa007, are you still planning to resolve feedback?

Pandaa007:
> Hi @Pandaa007, are you still planning to resolve feedback?

Hey @MisRob! So sorry, the PR had completely skipped my radar. Will try to address the feedback today. Thanks for the patience.

lib/KTable/test-utils.js
Pandaa007: I had tried doing that, but if I place it inside the `__tests__` directory, then `jest` registers it as a test file, and on running `yarn test` we get the error that each test file must have atleast one test. Thus I kept it with the component files itself.

lib/KTable/test-utils.js
Pandaa007: Yes exactly!

lib/KTable/__tests__/index.spec.js
Pandaa007: Ahh yes. We are checking keyboard navigation, but not focus management. Will update the comment.

lib/KTable/__tests__/index.spec.js
Pandaa007: Sure. Will add the context via comments!

lib/KTable/__tests__/index.spec.js
Pandaa007: Addresses in the latest commit!

lib/KTable/__tests__/index.spec.js
Pandaa007: Got it! Will update in the latest push. Thanks!

MisRob:
Thanks a lot @Pandaa007, no problem

lib/KTable/__tests__/index.spec.js
Pandaa007: Ahh nice catch. Added 6 new test cases for navigation around the corner's in the latest push.

lib/KTable/__tests__/tab.spec.js
Pandaa007: Thanks for the confirmation @AlexVelezLl. I don't think I would be available to work on the follow-up issue now due to some time constraints, but I would love it if we can open a new issue for the same. Would love to see other remembers of the community pick it up!

lib/KTable/__tests__/index.spec.js
```diff
@@ -0,0 +1,293 @@
...
+        [
+          'should go to the first cell of the header when going right from the last cell',
+          {
+            startCellIdx: [3, 2],
+            key: 'ArrowRight',
+            nextCellIdx: [0, 0],
+          },
+        ],
+        [
+          'should go to the end of the previous row when going left from the first cell of the current row',
```
AlexVelezLl, line 229: Great!

AlexVelezLl (commented):
Thanks a lot for this @Pandaa007! Tests look great, thoroughly testing KTable's features, such as sorting, keyboard navigation, cell highlighting, etc. You have done great work here; thank you very much! This will definitely protect us against any regression we may have in KTable in the future.

Just one little final request, could you rebase your PR on top of the latest develop to resolve the conflict and proceed with the merge? If not, we can certainly take care of that; there is no problem :). Thanks again :open_hands:.

Pandaa007:
@AlexVelezLl Rebased the PR, and moved the visual tests to a new file so as to keep the context of tests seperate! PTAL.

AlexVelezLl (approved):
Thanks @Pandaa007! This looks great! :rocket:. Merging :100:.

PR #962, opened 2025-03-14
Migrate KImg documentation to use DocsExample component

### LINKED ISSUE #952: Migrate `KImg` documentation to use `DocsExample` component
### Please refer to the parent issue #949 for more detail

Component to work on: [KImg](https://design-system.learningequality.org/kimg/)

Pandaa007:
## Description
- Fixed the bug with the use of `scss` language in the `DocsShowCode` component
- Migrated the examples on the `KImg` page to use `DocsExample` wherever possible

#### Issue addressed
Fixes #952 

## Changelog

  - **Description:** Migrate KImg documentation to use DocsExample component
  - **Products impact:** -
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/952
  - **Components:** -
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -

## Implementation notes

While migrating the examples, I was unable to migrate two of the examples listed in the `KImg` documentation page as they relied on being rendered in an inline element like `span`, which by default the `DocsExample` component renders the same in a block `div` element. This led to different looking outputs for those particular examples.

It would be nice if the `DocsExample` component also support a `block` attribute much like `DocsShow`, so the end consumer can control out the output is to be shown to the user. I can make the required changes in this PR itself, or open a new PR to address the same. 

cc: @MisRob @EshaanAgg 

## Testing checklist

## Reviewer guidance


EshaanAgg (approved):
The changes look good to me @Pandaa007! The documentation page also looks much better than it previously did. Thanks for your work here. 

I will tag @MisRob for additional review, and to once get a second opinion on the before and after appearance of the documentation page.  Post that, we can merge this.

EshaanAgg:
> ## Implementation notes
> While migrating the examples, I was unable to migrate two of the examples listed in the `KImg` documentation page as they relied on being rendered in an inline element like `span`, which by default the `DocsExample` component renders the same in a block `div` element. This led to different looking outputs for those particular examples.
> 
> It would be nice if the `DocsExample` component also support a `block` attribute much like `DocsShow`, so the end consumer can control out the output is to be shown to the user. I can make the required changes in this PR itself, or open a new PR to address the same.
> 
> cc: @MisRob @EshaanAgg

Thanks for pointing out this limitation of the `DocsExample` component! I think it would be better to open a new issue for the same, and make those changes in a new PR to keep the scope of this PR limited and easier to review. We can migrate the leftover examples from `KImg` in the new PR addressing the `block` attribute issue as a sanity check for the changes made. The best way to address the same might be to ensure that we internally make use of `DocsShow` to wrap the rendering example, and "forward" the `block` prop from the `DocsExample` component to the `DocsShow` component.

What do you think about the same @MisRob?

MisRob:
@Pandaa007 @EshaanAgg Yes, I think adding `block` prop to `DocsExample` is needed.  Good catch, thank you. @Pandaa007 you're welcome to open a new PR - it should be a small internal change so I think we could merge it soon and then it will be available for all sub-tasks of this migration, including this PR. Much appreciated!

Pandaa007:
Thanks for the confirmation @MisRob. Will open a PR soon.

Pandaa007:
@EshaanAgg @MisRob I have updated all the examples post merging #965. The changes are now ready for review.

EshaanAgg (approved):
All of these changes look good to me. Thanks @Pandaa007 for your contributions. 

@MisRob It would be really helpful if you can check the deployment preview once to checkout the changes in the documentation! If all looks good, I think we can merge it :)

MisRob (approved):
It all looks very nice, and exactly as expected.

Thanks so much both of you @Pandaa007 and @EshaanAgg.

PR #968, opened 2025-03-18
KCheckbox: Use flex rather than table for display

nucleogenesis:
#### Before:

See how when the box is highlighted, there is space under it.

[Screencast_20250318_161152.webm](https://github.com/user-attachments/assets/3d2a1d67-41ee-4441-87c0-1cea754e3a39)

#### After

See how now the checkbox and its label are equal height & centered together

[Screencast_20250318_161241.webm](https://github.com/user-attachments/assets/276859e3-4142-42cd-8d17-f9835413ebc8)

#### Issue addressed

Using the table display value resulted in the checkbox icon having some unnecessary space beneath it which can make centering it bothersome when not using the label prop.

For example - [here in Kolibri](https://github.com/learningequality/kolibri/blob/280cb81aa1ebae0000cc88d70b39d0959eacc9a3/kolibri/plugins/coach/assets/src/views/quizzes/CreateExamPage/CreateQuizSection.vue#L633-L635) we have some /deep/ styling to mitigate this by moving the box itself down a bit like so:

slumpy-checkboxes

## Changelog

### DIFF
--- a/lib/KCheckbox.vue
+++ b/lib/KCheckbox.vue
@@ -5,7 +5,6 @@
     :class="{ 'k-checkbox-disabled': disabled }"
     @click="toggleCheck"
   >
-    <div class="tr">
       <div class="k-checkbox">
         <input
           :id="id"
@@ -64,7 +63,6 @@
           {{ description }}
         </div>
       </label>
-    </div>
   </div>
 
 </template>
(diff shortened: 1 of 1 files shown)


bjester:
Overall, this seems good. My only concern is that the checkbox container overall seems slightly narrower in height. If we match the height with some padding to compensate, I presume that will produce less regressions, but open to thoughts.

PR #972, opened 2025-03-19
Always load KImg lazily

MisRob:
## Description

Based on @rtibbles suggestion, this update makes `KImg` load lazily.

## Comments

Making this option configurable didn't seem necessary - as far as I can say from our use-cases and [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#lazy) this is what we'd want to do by default.

However, there's one possible issue with this approach:

Screenshot from 2025-03-19 16-59-08

I've considered adding documentation guidance in an attempt to address this:

Screenshot from 2025-03-19 17-56-02

However `KImg` API doesn't currently have the way to configure `height` and `width` attributes of ``. This needs to be discussed and then all places in Kolibri, Studio, and KDS that use `KImg` updated to use them.

[Slack thread](https://learningequality.slack.com/archives/CTG8XDTCL/p1742403875546469)

## Changelog

### DIFF
--- a/docs/pages/kimg.vue
+++ b/docs/pages/kimg.vue
@@ -39,6 +39,41 @@
         dimensions are 200×114 px.
       </p>
 
+      <h3>Alternative text</h3>
+
+      <p>
+        Unless an image is
+        <DocsExternalLink
+          text="decorative"
+          href="https://www.w3.org/WAI/tutorials/images/decorative/"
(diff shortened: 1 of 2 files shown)


MisRob:
I will write issue for this because it will have implications for `KCard` public API too

PR #985, opened 2025-04-07
KTable docs small improvements

MisRob:
## Description

- Adds "Caption" usage section with brief guidance
  - Updates examples to reflect the guidance
- Rephrases some props usage for more clarity and conciseness, and removes duplicate information
- Removes unnecessary code comments  

#### Issue addressed

Just noticed few places to improve as I was creating trivia.

## Changelog

  - **Description:** Smaller improvements of KTable documentation page
  - **Products impact:** none
  - **Addresses:** -
  - **Components:** -
  - **Breaking:** -
  - **Impacts a11y:** -
  - **Guidance:** -

## Steps to test

Have a look at the [preview build](https://deploy-preview-985--kolibri-design-system.netlify.app/ktable).


BabyElias (commented):
The changes look good!
Just a doubt, the fact that we are encouraging users to put translated strings as captions for the table - is there any way that this can be shown or referenced in the docs here?

MisRob:
Thank you, @BabyElias.

>  Just a doubt, the fact that we are encouraging users to put translated strings as captions for the table - is there any way that this can be shown or referenced in the docs here?

Ah yes good point. The thing is that KDS itself never handles translations, so I am not sure how to best communicate it in more explicit manner. Do you have any ideas?

docs/pages/ktable.vue
```diff
@@ -23,7 +23,18 @@
...
+      <h3>Table caption</h3>
+
+      <p>
+        A caption is required via the <code>caption</code> prop. Provide a concise and translated
+        string that accurately describes the table’s content. See
+        <a
+          href="https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption#table_with_caption"
+          target="_blank"
+        >MDN</a>
+        for more.
```
BabyElias, line 39: I was earlier thinking about adding reference to an example from Kolibri here that makes use of translated strings for caption, but that brought me to the observation, that as of now, none of the tables (Facility/Users) make use of translated strings for captions. Now, I'm doubtful, but is the use of translated strings even needed here, or can the corresponding screen readers handle them on their own?

docs/pages/ktable.vue
```diff
@@ -86,12 +94,13 @@
         block
       />
 
-      <!--Table with default sort-->
       <h3>Table with default sort</h3>
```
BabyElias, line 118: Also, should we add the section for `disableBuiltInSorting` before `defaultSort`, as we are already mentioning about the `disableBuiltInSorting` attribute here?

docs/pages/ktable.vue
MisRob: Ah yes we could reference Kolibri from here, that's a good idea. Or perhaps even better, our [Internationalization guide](https://kolibri-dev.readthedocs.io/en/develop/i18n.html#internationalization)? That link would be more stable than linking codebase, and it has examples + additional explanation.

I don't think screenreaders will translate captions on their own, and we need to use translated strings as for anything else. Nevertheless, I double-checked and Kolibri `KTables` have translated caption, such as [here](https://github.com/learningequality/kolibri/blob/cb022ad0dd6295365df48da2402643ac718db349/kolibri/plugins/facility/assets/src/views/UserPage/index.vue#L66). If I am missing some places, would you reference? We would need to fix that in Kolibri.

docs/pages/ktable.vue
MisRob: Thank you, yes, I will swap 'Table with default sort' with 'Disable built-in sorting' (even though the default sort will hopefully soon work no matter of `disableBuiltinSorting` :)!

docs/pages/ktable.vue
BabyElias: Ahh, my bad. When I did the KTable transition initially, I hadn't done this translation thing. Maybe, it was added in some later PRs. Apologies for the confusion caused :|
Also, adding the internationalization guide sounds like a great idea!

MisRob: No problem

MisRob:
@BabyElias Thank you for review, addressed. If all looks well, can you approve?

BabyElias (approved):

MisRob:
Thanks for review @BabyElias :)

Issue #980, opened 2025-03-27
KCardGrid: Should align items whether selectable or not if any are selectable


nucleogenesis:
## Expected behavior

When KCardGrid has a mix of selectable and non-selectable items, the non-selectable items do not have the space where their checkbox _would be_ left empty - but instead, the card extends the full width.

[See this code in Kolibri](https://github.com/nucleogenesis/kolibri/blob/49fbdd8b2881bc0363c3b3aa7e277eef71b4a02a/kolibri/plugins/coach/assets/src/views/common/resourceSelection/ContentCardList.vue#L54-L59) which was used to provide this padding for this particular case.

## Actual behavior

If any items are selectable in a KCardGrid, then all of them should be aligned and the same width.

PR #990, opened 2025-04-09
[Proposal] Add useGridMetrics to calculate common selectionControlWidth for KCard

AlexVelezLl:
## Description

This PR proposes a way to calculate common metrics in the KCards context. The idea is that on mount, each card provides a `getMetrics` function (name can be changed) that returns measurements of specific parts/sections of the card, e.g. the width of the selection control. With this, the KCardGrid component will calculate the greater selection control width among its children KCards, and will provide this value for the cards to use this measurement as min-width of its selection-control div, so they all can have the same measurement for that section of the card.

We can test how it goes a pattern liike this, and if it works well with the selection control, in the future we can expand the `getMetrics` method to share metrics like the height of the title/above title, and align these sections more proactevely.

For now, with this update, If I remove [this fallback](https://github.com/nucleogenesis/kolibri/blob/49fbdd8b2881bc0363c3b3aa7e277eef71b4a02a/kolibri/plugins/coach/assets/src/views/common/resourceSelection/ContentCardList.vue#L54-L59) element in the ContentCardList in Kolibri, this keeps working as expected:

https://github.com/user-attachments/assets/1b90c872-b9b5-440d-a996-2d4c71c16d4d

#### Issue addressed

Addresses #980

## Changelog

  - **Description:** Adds KCardGrid management for reserving space for selection control width in KCard to align cards that already selection controls and the ones who doesnt.
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/980.
  - **Components:** KCard
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** .

### DIFF
--- a/lib/cards/KCardGrid.vue
+++ b/lib/cards/KCardGrid.vue
@@ -1,44 +1,46 @@
 <template>
 
-  <div
-    v-if="showGrid"
-    class="k-card-grid"
-  >
-    <transition
-      name="fade"
-      mode="out-in"
-      appear
-    >
-      <ul
-        v-if="isLoading"
-        key="loading"
-        :style="gridStyle"
-      >
-        <SkeletonCard
-          v-for="i in skeletonCount"
-          :key="i"
-          :height="skeletonHeight"
-          :orientation="skeletonOrientation"
-          :thumbnailDisplay="skeletonThumbnailDisplay"
-          :thumbnailAlign="skeletonThumbnailAlign"
-        />
-      </ul>
-      <ul
-        v-else
-        key="loaded"
-        :style="gridStyle"
-      >
-        <!-- @slot Slot for `KCard`s -->
-        <slot></slot>
-      </ul>
-    </transition>
-
+  <div>
     <div
-      v-if="debug"
-      class="k-debug"
+      v-if="showGrid"
+      class="k-card-grid"
     >
-      <div>DEBUG</div>
-      <div>breakpoint: {{ windowBreakpoint }}</div>
+      <transition
+        name="fade"
+        mode="out-in"
+        appear
+      >
+        <ul
+          v-if="isLoading"
+          key="loading"
+          :style="gridStyle"
+        >
+          <SkeletonCard
+            v-for="i in skeletonCount"
+            :key="i"
+            :height="skeletonHeight"
+            :orientation="skeletonOrientation"
+            :thumbnailDisplay="skeletonThumbnailDisplay"
+            :thumbnailAlign="skeletonThumbnailAlign"
+          />
+        </ul>
+        <ul
+          v-else
+          key="loaded"
+          :style="gridStyle"
+        >
+          <!-- @slot Slot for `KCard`s -->
+          <slot></slot>
+        </ul>
+      </transition>
+
+      <div
+        v-if="debug"
+        class="k-debug"
+      >
+        <div>DEBUG</div>
+        <div>breakpoint: {{ windowBreakpoint }}</div>
+      </div>
     </div>
   </div>
 
@@ -52,6 +54,7 @@
   import { LAYOUT_1_1_1, LAYOUT_1_2_2, LAYOUT_1_2_3 } from './gridBaseLayouts';
   import useGridLayout from './useGridLayout';
   import useGridLoading from './useGridLoading';
+  import useGridMetrics from './useGridMetrics';
   import SkeletonCard from './SkeletonCard';
 
   /**
@@ -63,6 +66,7 @@
       SkeletonCard,
     },
     setup(props) {
+      useGridMetrics();
       const { currentBreakpointConfig, windowBreakpoint } = useGridLayout(props);
       const {
         showGrid,
--- a/lib/cards/KCardGrid.vue
+++ b/lib/cards/KCardGrid.vue
@@ -66,7 +66,9 @@
       SkeletonCard,
     },
     setup(props) {
-      useGridMetrics();
+      if (props.syncCardsMetrics) {
+        useGridMetrics();
+      }
       const { currentBreakpointConfig, windowBreakpoint } = useGridLayout(props);
       const {
         showGrid,
@@ -173,6 +175,17 @@
       debug: {
         type: Boolean,
         default: false,
+      },
+      /**
+       * Enables the `syncCardsMetrics` feature.
+       * This feature is used to synchronize the metrics of
+       * cards in the grid, e.g. when you have some cards with
+       * a select control and some without it, this will make that
+       * all cards have the same space for the select control.
+       */
+      syncCardsMetrics: {
+        type: Boolean,
+        default: true,
       },
     },
   };
--- a/lib/cards/useGridMetrics.js
+++ b/lib/cards/useGridMetrics.js
@@ -1,0 +1,87 @@
+import { throttle } from 'lodash';
+import { getCurrentInstance, inject, onBeforeUnmount, onMounted, provide, ref, watch } from 'vue';
+import useKResponsiveElement from '../composables/useKResponsiveElement';
+
+/**
+ * Composable to provide the context for calculating card metrics inside a grid.
+ * It allows cards to register and unregister their get metrics functions, and with
+ * this, we calculate the greatest metrics among all registered cards, so cards can then
+ * use this information to calculate their own metrics and have a consistent layout.
+ */
+export default function useGridMetrics() {
+  const registeredCards = ref({});
+
+  const registerCard = (uid, { getMetrics }) => {
+    registeredCards.value = {
+      ...registeredCards.value,
+      [uid]: {
+        getMetrics,
+      },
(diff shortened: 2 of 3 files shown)


lib/cards/KCardGrid.vue
```diff
@@ -1,44 +1,46 @@
...
 
-  <div
-    v-if="showGrid"
-    class="k-card-grid"
-  >
-    <transition
-      name="fade"
-      mode="out-in"
-      appear
+  <div>
```
AlexVelezLl, line 3: I had to add a div here so we can always have at least one element, even if `showGrid` is false, if on mount we dont have an element in the root of the component, `useKResponsiveElement` will fail.

lib/cards/useGridMetrics.js
```diff
@@ -0,0 +1,87 @@
...
+
+  const selectionControlWidth = ref(0);
+
+  const calculateGridMetrics = () => {
+    let newSelectionControlWidth = 0;
+    const cards = Object.values(registeredCards.value);
+    for (const card of cards) {
+      const metrics = card.getMetrics();
+      if (metrics.selectionControlWidth > newSelectionControlWidth) {
+        newSelectionControlWidth = metrics.selectionControlWidth;
```
nucleogenesis, line 36: I think you could add a `break;` here if I'm understanding correctly that, once `newSelectionControlWidth` is set, then we know we need to set the control-area width on all of the cards.

nucleogenesis (commented):
Gave it a thorough look and it's looking great. I'll come back to give it a manual test later on, but the code lgtm so far

MisRob:
Hi @AlexVelezLl, thanks a lot! I appreciate it's framed as proposal. Below are my high-level thoughts for your consideration. You're well familiar with card and performance areas and I'm happy for you to decide next steps.

- For communication between grid and cards, overall this looks like a very interesting and clever idea and I'd be curios to see it in action
- The only concerns I could think of were primarily performance, and  complexity a bit.
- `frame-throttle.throttle` is available in KDS - could be more performant than `lodash.throttle`
- This will have impact on grids where we don't use select controls at all, for example grids with many cards in the Kolibri Library. It could be useful to consider having an option to disable running all this logic when we know it's not needed.
- I'd recommend performance testing before/after on a page that has a very large amount of cards - with select controls, without, mixed.
- Generally with complexity and dimension calculations like this comes more potential for bugs and maintenance cost (e.g. never-ending problems with Kolibri scroll script). I couldn't think of any alternative myself, perhaps except of using the same strategy, but simplifying few parts - for example not reading the width of the select area, but simply detecting if the select slot is used and assuming its width since generally we know that number. On the other hand, the current approach feels more robust and precise since it doesn't make such assumptions and so we wouldn't need to be worried what would happen if we changed other KDS components. I would say if performance is good, why not keep it as is, and if there are some problems, we can consider optimizations?

lib/cards/KCardGrid.vue
MisRob: That sound fine, we will just need to test carefully if it doesn't cause some unexpected trouble with loading skeleton experience, particularly on slower networks. `showGrid` function has more details in comment about what issues it takes care of.

AlexVelezLl:
Thanks @MisRob! Yes, I framed this as a proposal, because I know we could have some little overhead in the performance, but would like to see if this is worth it vs the benefits of ease of development and making KCard smart enough to make sure these measurements are always followed, and not relying on the developer remembering to put something like the empty div to make room for the missing select controller (and the potential changes in the future).

I just pushed some changes adding a `syncCardsMetrics` prop to KCardGrid to enable this feature so we just dont have to enable this in places its not relevant. And have used the `frame-throttle.throttle` method :).

I added a console.log in the [calculateGridMetrics](https://github.com/AlexVelezLl/kolibri-design-system/blob/f93d96425a7ef9579fe294700f41e8828861b34d/lib/cards/useGridMetrics.js#L30) method, and it was just being executed twice each time the cards were updated:


https://github.com/user-attachments/assets/8158253e-8ebc-44f0-beae-a0e7f313d8c9


I have run a profiling session measuring rendering and painting times with and without this "styles sync" logic. I have tried this in the lessons side panel, loading multiple pages of resources, first until I loaded 72 resources, and then loading 156 cards (which is I think a relevant amount of a number of cards that we can find given that, for example, the number of public channels in studio is 152) and after 3 repetitions, I got these results:

Example loading 72 cards in a session of ~10s:

|           | With styles sync enabled |              | Without styles sync enabled |              |
|-----------|--------------------------|--------------|-----------------------------|--------------|
|           | Mean (ms)                | Std Dev (ms) | Mean (ms)                   | Std Dev (ms) |
| Painting  |                  **1589,48** |        35,54 |                     **1558,59** |        30,81 |
| Rendering |                  **1177,89** |        13,89 |                     **1177,15** |        30,09 | 

Example loading 156 cards in a session of approx 20s:

|           | With styles sync enabled |              | Without styles sync enabled |              |
|-----------|--------------------------|--------------|-----------------------------|--------------|
|           | Mean (ms)                | Std Dev (ms) | Mean (ms)                   | Std Dev (ms) |
| Painting  |                  **2479,07** |       162,86 |                     **2645,05** |       218,71 |
| Rendering |                  **2751,19** |        24,38 |                     **2801,76** |        81,40 |

So in this test, there's no significant difference in painting and rendering performance. For the 72-card test, the test with styles sync enabled took about 30ms longer to paint and 0.74ms longer to render compared to the test without styles sync. But for the 156-card test, the test with styles sync enabled took about 165ms less time to paint and 50ms less time to render compared to the test without styles sync. So these differences, besides being small, are likely due to external conditions, rather than any large overhead that the styles sync feature enabled may have.

And it makes sense that we dont have much difference in the painting times, because this method, in the way it is implemented, should not cause any reflow, because we are relying on the vue re-rendering system to read and change these nodes widths (rather than directly manipulating the DOM with something like `card.$el.style.minWidth = ...`)

I was also thinking that we can already cut a `release-v5` branch, and introduce this change for the next KDS major version rather than having this in v5, to not mess with the upcoming Kolibri's and Studio's releases and have more time to test this.

MisRob:
This is fantastic work and results @AlexVelezLl. I am happy it all turned out well and we have a solid basis for communication between card and grid.

It sounds that, if we add more styles to sync in the future, we'd only need to be careful to implement the new logic as efficiently as yours.

As for `syncCardsMetrics`, that looks good to me, even though now I wonder if it's needed since the difference in performance is basically none.

As mentioned, feel free to make final decisions here :)

lib/cards/useGridMetrics.js
AlexVelezLl: Oh a break is not needed here, because we are finding the greater width so we need to go through all cards.

AlexVelezLl:
> we'd only need to be careful to implement the new logic as efficiently as yours.

That wouldnt be too dificult to achieve! It will be mostly just return a new metric in the KCard `getMetrics` method, and add a new calculation in the [KCardGrid calculateGridMetrics method](https://github.com/AlexVelezLl/kolibri-design-system/blob/55dcd1ee858a73626a5f08ca8ceb520529e3cb23/lib/cards/useGridMetrics.js#L37) :)

> As for syncCardsMetrics, that looks good to me, even though now I wonder if it's needed since the difference in performance is basically none.

I think the better middle ground here will be to have this prop as true by default, so we dont need to explicitly set it, but we let the API open to disable it if needed.

Regarding the target branch. I just talked with Marcella and we agreed on having this on the 0.18 patch 1, so will keep targeting KDS v5 for this PR.

AlexVelezLl:
Just opened a PR in Kolibri to test this https://github.com/learningequality/kolibri/pull/13320

AlexVelezLl:
Hi @MisRob QA testing didn't find any regressions https://github.com/learningequality/kolibri/pull/13320#issuecomment-2880496198, should it be okay to proceed with this?

MisRob:
Yes @AlexVelezLl. I couldn't trigger re-run of Percy, do you know what would be best to do?

AlexVelezLl:
Ah apologies @MisRob, the failure of Percy's check was because 30 days had already passed since the deployment request was made and it seems that when this happens we cannot re-run the gh action in GitHub. I just force-pushed a dummy commit to re-trigger it :). Richard suggested that we can probably use tags instead to trigger these visual tests, which would avoid having to make these deployment requests, we'll see.

MisRob:
No problem @AlexVelezLl, thank you. Labels may be nice, even though I wouldn't mind force pushing occasionally. It sometimes happens with other checks in other repos too.

MisRob (approved):
I am leaving soon for today so I won't wait for Percy to finish, and I've just realized we actually don't yet have the visual tests for cards anyway.

I manually previewed in documentation and on the cards playground and haven't noticed any issues, and QA on Kolibri passes, so good to go! Nice work here, feel free to merge.

AlexVelezLl:
Thanks @MisRob!

PR #1003, opened 2025-04-25
Add unit tests for KListWithOverflow component

VincentChiang621:
This pull request introduces a new test suite for the KListWithOverflow.vue component. It includes:

A test to verify that a list of items is correctly rendered.

A test to ensure that an overflow indicator is displayed when the list exceeds a certain length.

These tests help ensure that the component behaves correctly under typical and edge-case usage.

### DIFF
--- a/lib/__tests__/KListWithOverflow.spec.js
+++ b/lib/__tests__/KListWithOverflow.spec.js
@@ -1,0 +1,26 @@
+// lib/__tests__/KListWithOverflow.spec.js
+
+import { mount } from '@vue/test-utils'
+import KListWithOverflow from '../KListWithOverflow.vue'
+
+describe('KListWithOverflow', () => {
+  it('renders a list of items', () => {
+    const items = ['One', 'Two', 'Three']
+    const wrapper = mount(KListWithOverflow, {
+      propsData: { items }
+    })
+
+    items.forEach(item => {
+      expect(wrapper.text()).toContain(item)
+    })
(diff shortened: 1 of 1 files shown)


MisRob:
Hi @VincentChiang621,

These tests are failing. Did you actually run the suite? Also, all test cases described in https://github.com/learningequality/kolibri-design-system/issues/833 need to be covered. @AlexVelezLl also asked you to re-use the previous pull request and review, and I don't see it happen here.

Feel free to give it another try, but first you will need to spend more time studying the component and thinking about how to write meaningful test suite. I also recommend to use Vue Testing Library that is available in the repository. And finally, if you're going to open a new pull request, please respect the pull request template and fill it in according to the guidance.

lib/__tests__/KListWithOverflow.spec.js
```diff
@@ -0,0 +1,26 @@
...
+    })
+  })
+
+  it('displays overflow indicator if list is too long', () => {
+    const items = Array.from({ length: 20 }, (_, i) => `Item ${i + 1}`)
+    const wrapper = mount(KListWithOverflow, {
+      propsData: { items }
+    })
+
+    expect(wrapper.text()).toContain('…') // or however overflow is rendered
```
MisRob, line 24: > _"or however overflow is rendered"_

It is _you_ who needs to know how overflow is rendered and write the test accordingly.

Issue #995, opened 2025-04-17
Add option to DocsExample to hide a tab


MisRob:
🌱 Are you new to the codebase? Welcome! Please see [the contributing guidelines](https://github.com/learningequality/kolibri-design-system/blob/develop/CONTRIBUTING.md).

## Current behavior

`DocsExample` doesn't have a configuration that would allow us to hide a tab.

This, in some cases, results in showing information that is unnecessary, e.g. examples in [KCard > Layout](https://design-system.learningequality.org/kcard#layout) have this "Style" tab that have no value for reader, and could even cause confusion since these styles are internal styles that don't need to necessarily correspond to our product pill styles.

![Image](https://github.com/user-attachments/assets/e6c5e937-ff39-479c-bdc8-969b1279206b)

## Desired behavior

`DocsExample` has new props `hideTemplate`, `hideScript`, and `hideStyle`. A truthy value hides the corresponding tab (in the example above, this means that only "Template" and "Script" tabs would be displayed.

This will support our attempts to keep the documentation concise and to the point, and allow more flexibility.

## Guidance

- [How to update the documentation website](https://github.com/learningequality/kolibri-design-system/blob/develop/dev_docs/04_how_to_update_docs.md#updating-content)

MisRob:
@EshaanAgg any feedback for this proposal? I will wait to hear your thoughts before I mark it as ready for work.

EshaanAgg:
Hi @MisRob! I am so sorry for being AFK for the past couple of weeks! 

I think that this issue would be a great addition to the API of the `DocExample` component, as it would allow the developers to hide the not-so-important parts of the examples! The idea of having three separate props seems the most flexible and clear way to me as well.

MisRob:
Hi @EshaanAgg, no problem, nothing pressing. Thanks for confirming. I will mark this as ready for work then.

yeshwanth235:
Hi @MisRob, I would like to work on this issue

I am proposing that we implement a single, dynamic prop, docTabs. This prop would be an array of strings (e.g., ['script', 'template', 'style']), and the tabs rendered would be based on the items within this array.

akolson:
HI @yeshwanth235! Yes sure! I have assigned you the issue. Please feel free to reach out incase you have any questions. Thank you!

MisRob:
Closed by https://github.com/learningequality/kolibri-design-system/pull/1021

PR #1020, opened 2025-05-08
Issue 995

yeshwanth235:
## Description

The docTabs prop is defined as an optional array (type: Array, required: false) that controls the displayed documentation tabs. If no value is provided, it defaults to showing 'template', 'script', and 'style' tabs.

### Before/after screenshots

Before
image

After
image

## Changelog

  - **Description:**
 The docTabs prop is defined as an optional array (type: Array, required: false) that controls the displayed documentation 
 tabs. If no value is provided, it defaults to showing 'template', 'script', and 'style' tabs.
  - **Addresses:** 
  https://github.com/learningequality/kolibri-design-system/issues/995
  - **Components:** 
  1. Add new prop to KCard component layout docsExample
  2. DocsExample component is updated with new prop
  - **Breaking:**  no

- [✓] Contributor has fully tested the PR manually
- [✓] If there are any front-end changes, before/after screenshots are included
- [✓] The change is described in the changelog section above

## Reviewer guidance

- [✓] Is the code clean and well-commented?

## Comments

Instead of using three different props as mentioned. Only one props to solve this issue. Let me know it we need it as three different props.

### DIFF
--- a/docs/common/DocsExample.vue
+++ b/docs/common/DocsExample.vue
@@ -123,6 +123,11 @@
         default: false,
         required: false,
       },
+      /**
+       * Its purpose is to dynamically control the set of documentation tabs rendered
+       * @type {Array}
+       * @example ['template', 'script', 'style']
+       */
       docTabs: {
         type: Array,
         required: false,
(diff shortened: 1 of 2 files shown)


yeshwanth235:
Will raise the PR again.

PR #1023, opened 2025-05-12
Add min-width: 0 for k-card-area

### LINKED ISSUE #958: KTextTruncator: Make custom styling easier/safer
## Desired behavior

During development on Kolibri, I ran into an issue where text passed to KTextTruncator (which was a long URL) resulted in a strange changes to the total width of the KTextTruncator's output.

The fix [here](https://github.com/learningequality/kolibri/pull/13060/commits/cf8264f5afd635151d9d563287e37884cba6c96d) was to set the `break-word` CSS property to `anywhere` for this particular case.

The /deep/ styles can be brittle if KTextTruncator changes, so it'd be ideal if KTextTruncator had a prop providing a way to pass custom styles.

#### Outcomes

- [ ] There is a way to avoid using `/deep/` styles to affect KTextTruncator
- [ ] An issue is created in the Kolibri repo to leverage the changes made to close this PR in the aforelinked commit.

AlexVelezLl:
## Description

This PR fixes the need for KTextTruncator to have an `overflow-wrap: anywhere` to solve an overflowed text issue in KCard. The main reason why we cannot use `overflow-wrap: anywhere` is that this is not supported by our browserlist (here the caniuse [link](https://caniuse.com/?search=overflow-wrap%3A%20anywhere)) while `overflow-wrap: break-word` is.

The main reason why `overflow-wrap: anywhere` was needed in the first place was because with `overflow-wrap: break-word` the cards with a long word/link (with no spaces within it) looked like this:

image

This happens because of a wonky interaction between `overflow-wrap` and horizontal flex containers. When we have an `overflow-wrap: break-word` with a long word inside a flex item, the `overflow-wrap: break-word` property makes the flex item grow as much as possible, causing all other flex items to overflow. This is solved by introducing a `min-width: 0` to the flex item containing the long words to force flex to shrink a child below its minimum content size as needed, see [Understanding the automatic minimum size of flex items](https://www.bigbinary.com/blog/understanding-the-automatic-minimum-size-of-flex-items) for a broader explanation.

After:

image

#### Issue addressed
Closes #958.

## Changelog

  - **Description:** Adds a `min-width: 0` to k card area to prevent text overflow issues within KCards with checkboxes.
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/958.
  - **Components:**  KCard.
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -.

### DIFF
--- a/lib/cards/KCard.vue
+++ b/lib/cards/KCard.vue
@@ -45,7 +45,7 @@
               This is to
                 - (1) prevent double navigation (the wrapping
                       <li> is supposed to take care of navigation)
-                - (2) together with the `draggable` disabled, 
+                - (2) together with the `draggable` disabled,
                       ensures that text selection works on
                       the title text (see the feature for allowing
                       selecting card's content in `onClick`)
@@ -61,7 +61,7 @@
               @blur.native="onTitleBlur"
             >
               <span aria-hidden="true">
-                <!-- @slot A scoped slot via which the `title` can be customized. 
+                <!-- @slot A scoped slot via which the `title` can be customized.
                  Provides the `titleText` attribute.-->
                 <slot
                   v-if="$scopedSlots.title"
@@ -92,7 +92,7 @@
               @blur="onTitleBlur"
             >
               <span aria-hidden="true">
-                <!-- @slot A scoped slot via which the `title` can be customized. 
+                <!-- @slot A scoped slot via which the `title` can be customized.
                  Provides the `titleText` attribute. -->
                 <slot
                   v-if="$scopedSlots.title"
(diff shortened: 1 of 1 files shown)


AlexVelezLl:
PR in Kolibri importing this change https://github.com/learningequality/kolibri/pull/13392

nucleogenesis (approved):
I tested the related Kolibri PR and things LGTM there, thanks Alex!

MisRob (approved):

MisRob:
@AlexVelezLl Thanks for explanation with visual references, that's helpful.

AlexVelezLl:
Thanks @MisRob, @nucleogenesis! QA tests have also confirmed there are no regressions with this change https://github.com/learningequality/kolibri/pull/13392#issuecomment-2890574998. Merging!

PR #1026, opened 2025-05-14
Wrap ResizeObserver callback with requestAnimationFrame

### LINKED ISSUE #960: ResizeObserver loop completed with undelivered notifications
## Product

Kolibri Studio

## Expected behavior

The error message `ResizeObserver loop completed with undelivered notifications` should not occur or occur very infrequently.

## Actual behavior

The error happens repeatedly and a fullscreen overlay takes over displaying the error, most likely because it connects with rendering somehow and triggers error handling.

## Recommendations
Looking at [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#observation_errors), it mentions this error. Currently, the KDS code uses a frame-throttle (aka `requestAnimationFrame`) at that the resizer target. Moving the frame-throttle to the `ResizeObserver` callback like the docs demonstrate could eliminate this issue, and that possibly makes sense since it says the resizer events are delivered before paint. This means the calls to the element's trigger will call the frame-throttled callback before paint, but that frame-throttled callback may not fire until after the element is destroyed or replaced during paint. Therefore, moving the frame-throttling higher up will align the logic to fire `entry.target._resizeListener` only after that has occurred, and overall, provide better performance by synchronizing all element callbacks to a single frame.

1. Start up Studio's devserver on `unstable`
2. Open a channel
3. Click to add/upload a file
4. Upload an EPUB file
5. Click 'Finish'
6. Observe many instances of this error

## Additional information

Image

## Environment
- OS: Chrome

AlexVelezLl:
## Description

Wraps ResizeObserver callback in `useKResponsiveElement` with requestAnimationFrame to avoid possible resize observer issues, see https://github.com/learningequality/kolibri-design-system/issues/960.

#### Issue addressed

Closes #960.

## Changelog

  - **Description:** Wraps ResizeObserver callback in `useKResponsiveElement` with requestAnimationFrame to avoid possible resize observer issues.
  - **Products impact:** none.
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/960.
  - **Components:** -.
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** .

### DIFF
--- a/lib/composables/useKResponsiveElement.js
+++ b/lib/composables/useKResponsiveElement.js
@@ -1,15 +1,16 @@
-import { throttle } from 'frame-throttle';
 import { onMounted, onBeforeUnmount, ref, getCurrentInstance } from 'vue';
 
 let _resizeObserver;
 
 if (typeof window !== 'undefined' && window.ResizeObserver) {
   _resizeObserver = new ResizeObserver(entries => {
-    for (const entry of entries) {
-      if (entry.target._resizeListener) {
-        entry.target._resizeListener();
+    requestAnimationFrame(() => {
+      for (const entry of entries) {
+        if (entry.target?._resizeListener) {
+          entry.target._resizeListener();
+        }
       }
-    }
+    });
   });
 }
 
@@ -28,8 +29,7 @@ export default function useKResponsiveElement() {
   onMounted(() => {
     updateEl();
     if (_resizeObserver) {
-      const throttledUpdateEl = throttle(updateEl);
-      instance.proxy.$el._resizeListener = throttledUpdateEl;
+      instance.proxy.$el._resizeListener = updateEl;
 
       _resizeObserver.observe(instance.proxy.$el);
     }


lib/composables/useKResponsiveElement.js
```diff
@@ -5,11 +5,13 @@ let _resizeObserver;
 
 if (typeof window !== 'undefined' && window.ResizeObserver) {
   _resizeObserver = new ResizeObserver(entries => {
-    for (const entry of entries) {
-      if (entry.target._resizeListener) {
-        entry.target._resizeListener();
+    requestAnimationFrame(() => {
+      for (const entry of entries) {
+        if (entry.target._resizeListener) {
```
bjester: Since there could be some time between when the resizer callback is fired and the animation frame callback is fired, I think it would be good to check the `entry.target`. In this case, the reference should still exist, but perhaps it isn't connected anymore: https://developer.mozilla.org/en-US/docs/Web/API/Node/isConnected 

Defensive code rarely hurts!

bjester (changes requested):
The changes so far look inline with the MDN docs. Adding the global animation frame means we should now also remove the use of `frame-throttle` in the same composable (line 33).

AlexVelezLl:
Thanks @bjester! I just pushed the suggested changes :open_hands:.

bjester (approved):

PR #1031, opened 2025-05-20
Kolibri 0.19 Bulk User Mgmt Icons

nucleogenesis:
## Description  

Adds icons from Figma:

- Admins
- Super Admins
- Coaches
- Learners
- Assign coaches (leftward circle arrow) 

| key | icon |
| admins  | admins  |
| allUsers  | all_users  |
| assignCoaches | assignCoaches |
| coaches | coaches | 
| learners | learners | 
| superAdmins | super_admins |

## Changelog

  - **Description:** Adds icons needed for feature development in Kolibri
  - **Products impact:** Kolibri
  - **Addresses:** n/a
  - **Components:** KIcon*
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** Use the given added icon key(s) to use the icons in Kolibri


custom-icons/admins.svg
```diff
@@ -0,0 +1,3 @@
...
...69217 3.34832 7.52793C2.40184 9.3637 2.06326 11.4528 2.38147 13.4935C2.69968 15.5342 3.65821 17.421 5.11866 18.8815C6.57911 20.3419 8.46592 21.3005 10.5067 21.6187C12.5474 21.9369 14.6364 21.5983 16.4722 20.6518C18.308 19.7053 19.7954 18.1999 20.7199 16.353C21.6443 14.506 21.9578 12.413 21.6151 10.3763C21.5823 10.1801 21.4729 10.005 21.311 9.88947C21.1491 9.77396 20.9479 9.7275 20.7517 9.76031ZM9.00015 11.25C9.00015 10.6567 9.17609 10.0766 9.50574 9.58329C9.83538 9.08994 10.3039 8.70542 10.8521 8.47836C11.4003 8.2513 12.0035 8.19189 12.5854 8.30764C13.1674 8.4234 13.7019 8.70912 14.1215 9.12868C14.541 9.54824 14.8267 10.0828 14.9425 10.6647C15.0583 11.2467 14.9988 11.8499 14.7718 12.3981C14.5447 12.9462 14.1602 13.4148 13.6669 13.7444C13.1735 14.0741 12.5935 14.25 12.0001 14.25C11.2045 14.25 10.4414 13.9339 9.87883 13.3713C9.31622 12.8087 9.00015 12.0457 9.00015 11.25ZM6.94515 18.5156C7.48771 17.6671 8.23515 16.9688 9.11857 16.4851C10.002 16.0013 10.993 15.7478 12.0001 15.7478C13.0073 15.7478 13.9983 16.0013 14.8817 16.4851C15.7651 16.9688 16.5126 17.6671 17.0551 18.5156C15.6098 19.6397 13.8311 20.2499 12.0001 20.2499C10.1692 20.2499 8.39046 19.6397 6.94515 18.5156Z" fill="black"/>
```
AlexVelezLl: Since we have this `fill="black"` at the end of the `<path>` tag, we can't override the icon color with KIcon. 
So, if I render all these icons with `color="green"`. It wont change the color of these new icons: (the first one is an existing one)
image

Which is a problem, because when we need to override their colors, like in a primary button, it wont do it:

image

Removing the `fill="black"` property should work for this :)

lib/KIcon/iconDefinitions.js
```diff
@@ -29,6 +29,25 @@ const KolibriIcons = {
     icon: require('./precompiled-icons/le/codeToggle.vue').default,
   },
   menu: { icon: require('./precompiled-icons/material-icons/menu/baseline.vue').default },
+  assignCoaches: {
```
AlexVelezLl, line 32: I just wonder if "assignCoaches" is the best identifier for this icon. There are no references to coaches in the icon itself. Perhaps just the action "assign" would fit better?

AlexVelezLl (changes requested):
Seems like we can have some potential bugs because of the `fill` property set in the raw svg icons, and just a nitpick comment about the naming of the `assignCoaches` icon

AlexVelezLl (approved):
Thanks @nucleogenesis! lgtm :)

Issue #1036, opened 2025-05-27
Insert horizontal separator lines beneath each row of the KTable


LianaHarris360:
## Description

Horizontal separator lines should be added below each row in KTable to improve the readability of content that overflows cell boundaries or spans multiple lines, as part of the Kolibri 0.19 Bulk User Management project.

![Image](https://github.com/user-attachments/assets/aa95d322-1639-4d6f-b1a0-bb00b136dabd)

#### Resources
- __[Link to Figma](https://www.figma.com/design/2aE0zEXmyysxMZ6qO0ufh8/Bulk-User-Management?node-id=98-8431&t=CRKVpZce9AIqRomZ-4)__

LianaHarris360:
Completed in #1043

Issue #1006, opened 2025-04-29
Add Visual Testing support across KDS components


AlexVelezLl:
<img height="20px" src="https://i.imgur.com/c7hUeb5.jpeg">

❌ **This issue is not open for contribution. Visit <a href="https://learningequality.org/contributing-to-our-open-code-base/" target="_blank">Contributing guidelines</a>** to learn about the contributing process and how to find suitable issues.

<img height="20px" src="https://i.imgur.com/c7hUeb5.jpeg">

## Overview
We recently introduced a new visual testing system in KDS to prevent visual regressions and ensure UI consistency, and we aim to gradually add visual tests to our KDS components.

This issue will serve as a tracking issue to organize and follow progress as we add visual tests through each component.

PR #1043, opened 2025-06-04
[KTable] : Added horizontal separator lines

BabyElias:
## Description
Added horizontal separator lines below each row, based on the Figma design provided.

#### Issue addressed

Addresses #1036

### Before/after screenshots

Before
before
After 
after

## Changelog

  - **Description:** Added horizontal separator lines
  - **Products impact:** None
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/1036
  - **Components:** KTable
  - **Breaking:** No
  - **Impacts a11y:** No
  - **Guidance:**  - 

## Testing checklist

## Reviewer guidance

### DIFF
--- a/lib/KTable/KTableGridItem.vue
+++ b/lib/KTable/KTableGridItem.vue
@@ -2,7 +2,7 @@
 
   <td
     :class="$computedClass(coreOutlineFocus)"
-    :style="{ textAlign: textAlign, minWidth: minWidth, width: width }"
+    :style="tdStyle"
     tabindex="0"
     data-focus="true"
     role="gridcell"
@@ -57,6 +57,14 @@
           },
         };
       },
+      tdStyle() {
+        return {
+          textAlign: this.textAlign,
+          minWidth: this.minWidth,
+          width: this.width,
+          borderBottom: `1px solid ${this.$themeTokens.fineLine}`,
+        };
+      },
     },
     methods: {
       onKeydown(event) {
@@ -83,6 +91,10 @@
   .cell-content {
     word-break: break-word;
     white-space: normal;
+  }
+
+  td {
+    border-bottom: 1px solid;
   }
 
 </style>
--- a/lib/KTable/KTableGridItem.vue
+++ b/lib/KTable/KTableGridItem.vue
@@ -93,8 +93,4 @@
     white-space: normal;
   }
 
-  td {
-    border-bottom: 1px solid;
-  }
-
 </style>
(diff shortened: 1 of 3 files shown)


MisRob:
Thank you @BabyElias! I will ask @LianaHarris360 and @nucleogenesis for review. Meanwhile, for new visual changes, it'd be great if you could add a related visual test. No need to cover all use-cases (this could be done later as part of https://github.com/learningequality/kolibri-design-system/issues/1006), but at least this basic display of the table. You can see [visual testing guide](https://github.com/learningequality/kolibri-design-system/blob/develop/dev_docs/07_visual_testing_guide.md). Depending on available time you can dedicate to this task - no problem to also open a follow-up.

BabyElias:
Okay @MisRob, I'll try adding a visual test within the scope of this PR itself.

lib/KTable/KTableGridItem.vue
```diff
@@ -85,4 +93,8 @@
     white-space: normal;
   }
 
+  td {
+    border-bottom: 1px solid;
```
LianaHarris360: The separation lines have been applied correctly. It looks good, thank you! 

Is there a particular reason why the border-bottom style is applied both here and on line 65? Just wondering if we can remove this second application since it gets overwritten.

BabyElias: My bad, we can - that was a trial application initially, and I totally missed out on cleaning it up :|

BabyElias:
Removed redundant code, and added visual test for the border-bottom style.

lib/__tests__/KTable.spec.js
```diff
@@ -1,5 +1,35 @@
 import { mount } from '@vue/test-utils';
 import KTable from '../KTable';
+import { renderComponentForVisualTest, takeSnapshot } from '../../jest.conf/visual.testUtils';
+
+describe.visual('KTable Border Visual Test', () => {
```
MisRob: Thanks for adding this, much appreciated. I think it'd be fine to simplify a bit and just frame it as 'sortable table' visual test? Because it actually visually tests for many more things than just the border - it shows the table when it's sortable.

Visual tests are a bit different from unit tests in this regard.

BabyElias: Makes sense, still new to understanding visual tests here. Made the required change :)

LianaHarris360 (approved):
This PR is good to go, all of the requested changes have been applied. Thank you @BabyElias!

Issue #1005, opened 2025-04-28
Update KTable to support dynamic column stickiness


LianaHarris360:
<img height="20px" src="https://i.imgur.com/c7hUeb5.jpeg">

❌ **This issue is not open for contribution. Visit <a href="https://learningequality.org/contributing-to-our-open-code-base/" target="_blank">Contributing guidelines</a>** to learn about the contributing process and how to find suitable issues.

<img height="20px" src="https://i.imgur.com/c7hUeb5.jpeg">

## Description
Currently, only the first left column in KTable remains fixed when scrolled horizontally. To support the updates introduced in the Kolibri 0.19 Bulk User Management project, KTable should be updated to support dynamically sticky columns. 

In 0.19, on the Facility > Users page in Kolibri, checkboxes will be added to the first column. These checkboxes, along with the Full Name column and the last column, will remain sticky during horizontal scrolling. Every KTable implementation will not require the first, second, and last columns to be designated as sticky, so the option to set a column as sticky should be optional.

![Image](https://github.com/user-attachments/assets/4ddcfceb-e5f3-4fc8-9450-b6b3e955ed93)

- [ ] Extend the headers definition to include an optional field `stickyPosition` that can be `"left"`, `"right"`, or `undefined` eg:
```
// How the headers would be passed to KTable
headers: [
  { label: 'Checkbox', dataType: '...', columnId: 'checkbox', width: '48px',  stickyPosition: 'left' },
  { label: 'Name', dataType: '...', columnId: 'name', width: '200px', stickyPosition: 'left', },
  // … all other middle columns
  { label: 'Options', dataType: '...', columnId: 'options', width: '48px',  stickyPosition: 'right' },
]
```
  1. Update the css class `sticky-column` to be assigned if a header or column has `stickyPosition === 'left'` OR if the index of that header or column is 0 to retain the original functionality of KTable.
  2. Create a new css class` sticky-column-right` that is assigned to every header or column that has `stickyPosition === 'right'`. This css class should be the same as `sticky-column`, except with `right:0` instead of left.
  3. Update `getCellStyle()` to set the background color of sticky cells to white if the column index is 0 OR `headers[colIndex].stickyPosition !== undefined `
- [ ] All sticky columns should have background colors set to ensure proper overlay. The background of all columns should still be gray when hovered over.
[...]

PR #1038, opened 2025-05-29
KImg test cases

yeshwanth235:
## Description
KImg Visual testcases

#### Issue addressed

Fixes #1009

### Before/after screenshots

None

## Changelog

  - **Description:** Adds visual tests for KImg.
  - **Products impact:** None
  - **Addresses:** [#1009](https://github.com/learningequality/kolibri-design-system/issues/1009)
  - **Components:** No
  - **Breaking:** No
  - **Impacts a11y:** No
  - **Guidance:** None

## Steps to test

1. Step 1
2. Step 2
3. ...

## Testing checklist

- [ ] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical and brittle code paths are covered by unit tests
- [ ] The change is described in the changelog section above

## Reviewer guidance

- [ ] Is the code clean and well-commented?
- [ ] Are there tests for this change?
- [ ] Are all UI components LTR and RTL compliant (if applicable)?
- [ ] _Add other things to check for here_

PR #1056, opened 2025-06-19
Adds configurable sticky columns to KTable

marcellamaki:
## Description
This PR allows for configurable stickiness for the first, first two, and last columns on a table

#### Issue addressed

Addresses [#1005](https://github.com/learningequality/kolibri-design-system/issues/1005)

### Before/after screenshots

default table with no content that overflows width; default table (i.e. no changes made by me to the documentation page for this example) when the content overflows the width; specific sticky column implementation

https://github.com/user-attachments/assets/6d9dbf95-6554-4efd-8be5-d849d92d135e

## Changelog

  - **Description:** This PR adds optional sticky columns for the first, firstTwo, and last columns. The default is 'first' only if nothing else is specified, and the stickiness is only apparent is the contents of the table is wider than the container
  - **Addresses:** [Link(s) to GH issue(s) addressed. Include KDS links as well as links to related issues in a consumer product repository too.](https://github.com/learningequality/kolibri-design-system/issues/1005). Part of 0.19 Kolibri work
  - **Components:** KTable
  - **Breaking:** No
  - **Impacts a11y:** No
  - **Guidance:** The "default" behavior of the first column being sticky will be introduced when KDS is updated after this is merged. Use of the additional options depends on the content of the table, and may need to be customized for mobile responsiveness

## Steps to test

1. Preview the additions to the KTable docs page built in the PR, specifically the sticky table example

No 

## Testing checklist

## Reviewer guidance

## Comments
Is the first column being sticky a reasonable default? 

I considered a more expansive API for additional column stickiness, but practically speaking, this is the only scenario we need. Since we are also having a broader conversation about some KTable refactors for a more dev friendly experience and reducing the use of column indices, etc. I decided to start with the most narrowly scoped approach to solve the problem at hand. I also can quite think of how a middle-of-the-table column stickiness would help anything. I also considered the value of using the indices vs. using a string prop with validation. I find the user experience of the 'first', 'firstTwo', 'last' to be easiest to follow, but would be interested in any other perspectives here.

### DIFF
--- a/lib/KTable/KTableGridItem.vue
+++ b/lib/KTable/KTableGridItem.vue
@@ -63,6 +63,8 @@
           minWidth: this.minWidth,
           width: this.width,
           borderBottom: `1px solid ${this.$themeTokens.fineLine}`,
+          minHeight: '40px',
+          height: 'auto',
         };
       },
     },
@@ -89,6 +91,9 @@
 <style scoped>
 
   .cell-content {
+    display: flex;
+    align-items: center;
+    min-height: 40px;
     word-break: break-word;
     white-space: normal;
   }
--- a/lib/KTable/KTableGridItem.vue
+++ b/lib/KTable/KTableGridItem.vue
@@ -91,9 +91,7 @@
 <style scoped>
 
   .cell-content {
-    display: flex;
-    align-items: center;
-    min-height: 40px;
+    justify-content: inherit;
     word-break: break-word;
     white-space: normal;
   }
--- a/lib/KTable/index.vue
+++ b/lib/KTable/index.vue
@@ -572,6 +572,7 @@
           const prevCell = this.getCell(prevRowIndex, prevColIndex);
           const prevFocusableElements = this.getFocusableElements(prevCell);
           prevFocusableElements[prevFocusableElements.length - 1].focus();
+          this.scrollCellIntoView(prevCell);
           this.updateFocusState(prevRowIndex, prevColIndex, false);
           event.preventDefault();
           return;
@@ -710,19 +711,30 @@
         }, []);
       },
       scrollCellIntoView(cell) {
-        if (cell) {
-          cell.scrollIntoView({ behavior: 'smooth', block: 'nearest', inline: 'nearest' });
+        if (!cell) return;
 
-          // Adjust scroll position to account for sticky headers
-          const stickyHeader = this.$refs.stickyHeader;
-          const stickyColumn = this.$refs['header-0'][0];
-          const tableWrapper = this.$refs.tableWrapper;
+        const tableWrapper = this.$refs.tableWrapper;
+        if (!tableWrapper) return;
+        const stickyHeader = this.$refs.stickyHeader;
+        const stickyHeaderHeight = stickyHeader ? stickyHeader.offsetHeight : 0;
+        let leftStickyWidth = 0;
+        let rightStickyWidth = 0;
 
-          const stickyHeaderHeight = stickyHeader ? stickyHeader.offsetHeight : 0;
-          const stickyColumnWidth = stickyColumn ? stickyColumn.offsetWidth : 0;
+        const stickyColumns = this.getStickyColumns();
+        stickyColumns.forEach(colIndex => {
+          const headerElement = this.$refs[`header-${colIndex}`];
+          if (headerElement && headerElement[0]) {
+            const width = headerElement[0].offsetWidth;
+            if (colIndex === this.headers.length - 1 && this.stickyColumns.includes('last')) {
+              rightStickyWidth += width;
+            } else {
+              leftStickyWidth += width;
+            }
+          }
+        });
 
-          const cellRect = cell.getBoundingClientRect();
-          const wrapperRect = tableWrapper.getBoundingClientRect();
+        const cellRect = cell.getBoundingClientRect();
+        const wrapperRect = tableWrapper.getBoundingClientRect();
 
         if (cellRect.top < wrapperRect.top + stickyHeaderHeight) {
           tableWrapper.scrollTop -= wrapperRect.top + stickyHeaderHeight - cellRect.top;
@@ -737,10 +749,14 @@
         if (cellRect.left < availableLeft) {
           const scrollAmount = availableLeft - cellRect.left;
           tableWrapper.scrollLeft -= scrollAmount;
-        } else if (cellRect.right > availableRight) {
+        }
+
+        else if (cellRect.right > availableRight) {
           const scrollAmount = cellRect.right - availableRight;
           tableWrapper.scrollLeft += scrollAmount;
-        } else if (cellRect.width > availableWidth && cellRect.left > availableLeft) {
+        }
+
+        else if (cellRect.width > availableWidth && cellRect.left > availableLeft) {
           const scrollAmount = cellRect.left - availableLeft;
           tableWrapper.scrollLeft += scrollAmount;
         }
--- a/lib/KTable/index.vue
+++ b/lib/KTable/index.vue
@@ -749,14 +749,10 @@
         if (cellRect.left < availableLeft) {
           const scrollAmount = availableLeft - cellRect.left;
           tableWrapper.scrollLeft -= scrollAmount;
-        }
-
-        else if (cellRect.right > availableRight) {
+        } else if (cellRect.right > availableRight) {
           const scrollAmount = cellRect.right - availableRight;
           tableWrapper.scrollLeft += scrollAmount;
-        }
-
-        else if (cellRect.width > availableWidth && cellRect.left > availableLeft) {
+        } else if (cellRect.width > availableWidth && cellRect.left > availableLeft) {
           const scrollAmount = cellRect.left - availableLeft;
           tableWrapper.scrollLeft += scrollAmount;
         }
--- a/lib/KTable/index.vue
+++ b/lib/KTable/index.vue
@@ -198,16 +198,26 @@
       const isTableScrollable = ref(false);
 
       const checkScrollable = () => {
-        if (!tableWrapper.value || !tableElement.value || !elementWidth.value) return;
+        if (!tableWrapper.value || !tableElement.value || !elementWidth.value) {
+          isTableScrollable.value = false;
+          return;
+        }
         isTableScrollable.value = tableElement.value.scrollWidth > elementWidth.value;
       };
+
+      const debouncedCheckScrollable = debounce(checkScrollable, 300);
 
       watch(
         () => props.rows,
         newRows => {
           rows.value = newRows;
+          nextTick(() => debouncedCheckScrollable());
         },
       );
+
+      watch(elementWidth, () => {
+        debouncedCheckScrollable();
+      });
 
       const handleSort = index => {
         if (headers.value[index].dataType === DATA_TYPE_OTHERS || !props.sortable) {
@@ -229,7 +239,7 @@
 
       onMounted(() => {
         nextTick(() => {
-          debounce(checkScrollable, 300)();
+          debouncedCheckScrollable();
         });
       });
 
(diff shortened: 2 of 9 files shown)


lib/KTable/index.vue
```diff
@@ -718,24 +856,46 @@
...
     z-index: auto;
-    padding: 8px;
+    padding: 0 8px;
   }
 
   .sticky-header {
     position: sticky;
     top: 0;
-    z-index: 2;
+    z-index: 10;
```
LianaHarris360: It looks like the increased z-index values are causing the headers and columns to be displayed over the AppBar in Kolibri, because the [ScrollingHeader](https://github.com/learningequality/kolibri/blob/6b54f83bade08d2a989df9f525457422b59ccba1/packages/kolibri/components/pages/ScrollingHeader.vue#L166) has a z-index of 4.

Screenshot 2025-06-26

lib/KTable/index.vue
```diff
@@ -182,6 +192,12 @@
 
       const isTableEmpty = computed(() => sortedRows.value.length === 0);
 
+      const isTableScrollable = computed(() => {
```
LianaHarris360: On the initial load, this value is false and the vertical lines beside the sticky columns don't appear, even though the table can be scrolled.

Screenshot 2025-06AM

LianaHarris360 (dismissed):
I agree that using 'first', 'firstTwo', and 'last' is the easiest user experience to follow and to validate against. I ran these changes in Kolibri and only noted two things, otherwise the rest of the code makes sense to me.

lib/KTable/index.vue
marcellamaki: ooo yikes! good catch - I'll update this

lib/KTable/index.vue
marcellamaki: The woes of testing things locally over and over! :D thanks

MisRob: I think we could optimize this method for performance. If I recall, `useKResponsiveElement` is already optimized well so that should be covered. For other logic, I can think of

- utilizing Vue `ref` should be more performant `querySelector`
- maybe we should debounce or even better `requestAnimationFrame` for reading `table.scrollWidth`?

@AlexVelezLl did quite detailed performance review of `KTable` some time ago. How does this sound or would you recommend something else? Also I see more places in code that read from DOM elements. Except this place, I haven't yet noticed anything obviously suspicious - if you'd like to check on it too, that'd be welcome.

lib/KTable/index.vue
```diff
@@ -685,6 +744,85 @@
           }
         });
       },
+      colIndexIsSticky(colIndex) {
+        const stickyIndices = this.getStickyColumns();
+        return stickyIndices.includes(colIndex);
+      },
+      getStickyColumnStyle(rowIndex, colIndex) {
```
MisRob, line 797: Looking at how `KTable` file grows, moving sticky logic like this, and related, to a composable would help.

MisRob (commented):
This looks very nice overall. Regarding API, within what we're able to achieve now, this feels simple and intuitive. In code, haven't dived into reviewing all calculations in detail, rather just brief skim.

I think this is a visual regression - on the latest `develop`, rows are higher than in this PR's preview.
| develop | PR |
| ----------- | ----- |
| 01-before | 01-after |

Seems to be the same issue that Percy has captured. 

I don't think all examples are covered with automated visual tests yet though, so would be good to compare [before](https://design-system.learningequality.org/ktable) / [after](https://deploy-preview-1056--kolibri-design-system.netlify.app/ktable) live documentation examples.

Relatedly, it'd be good to cover new sticky UI with visual tests.

MisRob:
Also, in the recording I see you added a nice new section to the documentation - seems you've just haven't yet commit that file?

By the way, after that, you could easily re-use the very same documentation example for a visual test. We've recently done the same for `KImg`. You can see [here](https://github.com/learningequality/kolibri-design-system/pull/1038/files#diff-ada10de317adf2ebcf0259662f30a85eccfe62bea0cf7197da57a0fa271e272bR22) and [here](https://github.com/learningequality/kolibri-design-system/pull/1038/files#diff-207a6b2bab9de60ed782444545a36bdc162ca91f8475fb123eeebe24db128ebcR104-R107). I think that two snapshots of this documentation example - one on mobile, and one on larger screen, would be sufficient.

lib/KTable/index.vue
AlexVelezLl: I agree, `useKResponsiveElement` doesn't do any kind of throttling nor debouncing when setting the sizes values so this computed property will be being recalculated a lot of times per second if the user resizes its window . I think throttling would be best here rather than debouncing to have a smoother UX if the user is resizing the window. 

I wonder @MisRob if would it worth to have a `throttle` prop for `useKResponsiveElement` that receives the number of ms to pass to the throttle function and wrap this [updateEl](https://github.com/marcellamaki/kolibri-design-system/blob/two-sticky-columns/lib/composables/useKResponsiveElement.js#L32) with a throttled function? I think throttling the watcher handler of the watch that observes the result of the `useKResposiveElement` is starting to be common in some places where we use this composable.

MisRob: @AlexVelezLl I think that `useKResponsiveElement` is optimized pretty well [here](https://github.com/learningequality/kolibri-design-system/blob/develop/lib/composables/useKResponsiveElement.js#L6-L14)? Since it uses (1) ResizeObserver and (2) requestAnimationFrame. I was more concerned with other parts such as `table.scrollWidth` and `querySelector`.

AlexVelezLl: Kindof, yes, but we still have so many recalculations that are perhaps not that necessary as we dont need that high reactivity. Just added a console.log in the computed property:

https://github.com/user-attachments/assets/3d7ff008-f19b-4a9a-821c-bfdcbbce5f4c

I think that's why, for example, in [KBreacrumbs](https://github.com/marcellamaki/kolibri-design-system/blob/1a370c62ebab0f838077018d5828b97366957499/lib/KBreadcrumbs.vue#L240) and [KListWithOverflow](https://github.com/marcellamaki/kolibri-design-system/blob/1a370c62ebab0f838077018d5828b97366957499/lib/KListWithOverflow.vue#L104) we throttle the watch handler function on top of the ResizeObserver and requestAnimationFrame that the useKResponsiveElement already provides.

And yes, the querySelector is the most concerning part here. I did a quick profile resizing the page, and in a 4s session, the scripting time was taking more than 1s, which is a lot. Not sure if reading the `table.scrollWidth` here would cause reflows, as this calculation is being made inside a computed property function, I would think that Vue aligns these calculations well in its render cycles, but not completely sure :sweat_smile:.

AlexVelezLl: I just realized that this is just to set the shadows, we wouldnt even need a throttle then, I dont think we need that level of reactivity for this :sweat_smile:, just wrapping this in a debounce and make this recalculation just once per resizing session will mitigate almost all of the performance impact that the calculation function would cause, as it will just be executed once.

marcellamaki: great thank you! I'll do that :)

MisRob: Ah yes that makes sense. Thanks for the measurement! And on top of that I think we can actually remove `querySelector` in favor of `ref`. Combining with debounce should then hopefully result in good performance overall.

I think this will help not only with resizing browser window but also anytime we change table width dynamically - e.g. when changing screen orientation on tablets or phones, where I would guess performance impact may be more apparent.

AlexVelezLl: Yes, I agree!

lib/KTable/index.vue
```diff
@@ -207,6 +227,12 @@
         return style;
       };
 
+      onMounted(() => {
+        nextTick(() => {
+          debounce(checkScrollable, 100);
```
LianaHarris360: The z-index issues have been resolved! I still see the `isTableScrollable` value as false on scrollable tables. After some investigation I realized that this debounced function needs to be assigned to a variable and then called here for it to eventually run `checkScrollable()`.

```
      const debouncedCheck = debounce(checkScrollable, 100);

      onMounted(() => {
        nextTick(() => {
          debouncedCheck();
        });
      });

```

marcellamaki:
Very belated, I have pushed my changes @LianaHarris360 @AlexVelezLl @MisRob. 

Particular focus areas/feedback requests for review: 

1. Have I actually addressed all of your feedback? Did I misunderstand any of your suggestions/is there anything I should further refine or fix?
2. I feel the least confidence about the visual tests. They pass, but I wrote them with the help of Claude. I did review them in comparison to the other visual tests we have but.... are they useful? Are they "right enough" for now? Are there important cases I'm missing? (an example of why I feel not so confident: in my Pr i accidentally deleted the overflow: auto behavior and the tests still passed, where that is clearly a visual regression. so, I may need to add more cases. does that seem worth doing now? in follow up?)
3. Any other feedback about the approach of the API? 
4. @MisRob regarding: 
> I think this is a visual regression - on the latest develop, rows are higher than in this PR's preview.

I have updated this to set a min-height for the row height, but looking at Tomiwa's most recent spec for the table in 0.19, his rows are a bit less "roomy" than the previous KTable rows, and the height here in this PR is more aligned with that (~40px) vs what it was previously (~60px). Does this feel like an appropriate change, or would you prefer I don't make any changes to the height at all. It isn't really in scope of this PR

lib/KTable/KTableGridItem.vue
```diff
@@ -89,6 +91,9 @@
 <style scoped>
 
   .cell-content {
+    display: flex;
```
LianaHarris360: I don't see it happening in Kolibri, but compared to the latest develop, the numbers for the `Age` column are no longer right-aligned.

KTable

lib/KTable/index.vue
```diff
@@ -207,6 +227,12 @@
         return style;
       };
 
+      onMounted(() => {
+        nextTick(() => {
+          debounce(checkScrollable, 300)();
```
LianaHarris360: Since `checkScrollable` is only called during the `onMounted` hook, if the table isn’t initially scrollable and is later resized to become scrollable, `isTableScrollable` doesn't gets set to true and the horizontal lines don’t appear. 

I wonder if, rather than setting the horizontal lines and table stickiness solely based on whether the table is scrollable, we should consider automatically applying the table stickiness and displaying the lines regardless of the current scrollability. Or, if it would be better to just watch the `elementWidth` and call `checkScrollable` whenever it changes.

LianaHarris360 (commented):
I think it's appropriate to set the min-height within this PR, since it aligns with our expectations for how KTable will eventually look. I also agree with Misha that moving the sticky logic into a composable could be beneficial, but that can be addressed separately, it doesn’t necessarily need to be included in this PR.

MisRob:
Hi @marcellamaki, thanks for all the updates.

> I’ve set a min-height for the row, but Tomiwa’s latest table spec for 0.19 uses less “roomy” rows than before.

Sounds like this is an intentional change for all `KTable`s. No problem at all then.

---

Regarding visual tests, my original suggestion was just to import `docs/examples/KTable/StickyColumns.vue` into the visual tests, rather than duplicate the example. KDS lets us share examples between docs and visual tests. I see you’ve covered more states—thank you! Just a heads-up: we’re preparing a refactor for visual tests, and the process will change a bit ([see guide](https://github.com/learningequality/kolibri-design-system/blob/3d7bad3933dcb7d45425fd9bd6f3c4a5ca048e14/dev_docs/07_visual_testing_guide.md)). We can still use work from this PR, but there will be some conflicts with https://github.com/learningequality/kolibri-design-system/pull/1107. Let’s chat about how to proceed—I’m sure we can resolve it without blocking this PR.

---

> Have I actually addressed all of your feedback?

Not sure if you saw my note about moving sticky logic to a composable. If you did and decided not to, a follow-up is fine as long as it doesn’t get delayed too long—the sooner, the easier for anyone's future work on `KTable`.

radinamatic:
Did a couple of passes on this documentation page, haven't noticed any regressions so far, but one point of improvement:  when the focus goes to a column that is **below the sticky one**, there should be a corresponding scroll of the columns in between the sticky ones, so a sighted keyboard user may actually be aware of which column is currently in focus. In the screenshot below, you can see that NVDA properly announces that the focus is on the **Favorite Color** column, which is located **below** the sticky **Options** one, and thus effectively hidden from the sighted keyboard navigation user. The best UX would be to always keep clearly visible whatever/whichever element or part of the page is currently in focus 🙂

radinamatic:
My previously reported issue is fixed: non-fixed columns are now properly scrolling and fully visible as they are being navigated by keyboard! 👏🏽 👏🏽 👏🏽 

One thing I did notice in the [documentation page](https://deploy-preview-1056--kolibri-design-system.netlify.app/ktable/) is that some content is repeated... Was that intentional? 🤔 

repeated

radinamatic (approved):
LGTM! 💯 👏🏽 :shipit:

marcellamaki:
Ah, no that looks to be an error with the documentation being duplicated that I missed. I'll update that

AlexVelezLl (approved):
Tanks @marcellamaki! The visual tests for the new KTable use cases can also be placed in the current `KTableVisualTest` component, but just opened a follow up to track this! https://github.com/learningequality/kolibri-design-system/issues/1127

PR #1087, opened 2025-08-04
[PROOF OF CONCEPT] Multi-select

GarvitSinghal47:
## Description

Do not merge. Will be moved to Studio. We're using KDS environment for easier development and testing.

### Part 2: Create public API and fine-tune - WIP @MisRob

https://github.com/learningequality/studio/issues/5429

### Part 1: Create skeleton - resolved by @GarvitSinghal47

This PR implements the initial accessible skeleton for an **autocomplete multi-select component**, as described in [#5133](https://github.com/learningequality/studio/issues/5133). It is part of the broader effort to remove Vuetify from Studio in favor of components built on the Kolibri Design System.

### Functionality Included

- ✅ **Accessible semantic markup** for:
  - Pills (selected items)
  - Dropdown list
  - Autocomplete input
- ✅ **Autocomplete search**
- ✅ **Selecting and deselecting items** via:
  - Clicking a list item
  - Clicking the `x` on a pill
  - Using the "Clear all" button
- ✅ **Keyboard interactions**
- ✅ **Screen reader support**
- ✅ **Toggle for `autocomplete` mode**


MisRob:
Hi @GarvitSinghal47, thanks a lot, I will start going through this week. We'll be in touch. Let me just convert it to draft stage to make a bit more clear to the rest of the team that we use KDS environment in this way only temporarily for development and QA.

MisRob:
Hi @GarvitSinghal47, thanks so much for all this work.  This is a solid foundation for keyboard navigation and screen reader support. I also noticed nice touches like highlighting searched text in the options 🙂

I understand some parts are perhaps still in progress. I’m focusing feedback on issue's main focus areas - key a11y behavior, markup structure, and the core architecture. I think it’d best to address the points below in this pull request, so there's no need to refactor core after even more code is added in the upcoming multi-select issues. None of it is urgent, and it’s totally up to you what you want to tackle, based on your time and priorities. I’m ready to jump in too.

After we've finished this round, I will invite our a11y expert to test.

lib/KMultiSelect.vue
MisRob, line 1: Observing `document.activeElement`, when I navigate with keyboard I see focus moving as I would expect 👍 However blue focus style is missing for some keyboard interactions, for example when I move between list options with Up / Down Arrow. Because of how our keyboard detection modality script works, you may need to apply `data-focus="true” `

lib/KMultiSelect.vue
```diff
@@ -0,0 +1,1084 @@
...
+          zIndex: '1000',
+          backgroundColor: $themeTokens.surface,
+          border: `1px solid ${$themeTokens.fineLine}`,
+          borderTop: 'none',
+          borderRadius: '0 0 4px 4px',
+          maxHeight: '400px',
+          overflowY: 'auto',
+        }"
+        role="region"
+        :aria-live="autocomplete ? 'polite' : 'off'"
```
MisRob: I noticed multiple `aria-live` in the file. Would you please remove them in favor of utilizing `useKLiveRegion` only? We're actually [hoping to remove](https://github.com/learningequality/kolibri/issues/12326) all such live regions from our codebases. [Here's](https://www.sarasoueidan.com/blog/accessible-notifications-with-aria-live-regions-part-2/#limit-the-number-of-live-regions-on-the-page) why.

lib/KMultiSelect.vue
```diff
@@ -0,0 +1,1084 @@
+<template>
+
+  <div class="autocomplete-multiselect">
+    <!-- Selected Pills Container -->
+    <div
```
MisRob: Would it be possible to make pills appear in the input? No need to spend much time on visual side of things, just to make sure that we have markup in place.

lib/KMultiSelect.vue
```diff
@@ -0,0 +1,1084 @@
...
+          :aria-describedby="ariaDescribedById"
+          tabindex="-1"
+        >
+          <!-- Select All Option -->
+          <li
+            v-if="showSelectAll"
+            :id="`select-all-${uid}`"
+            role="option"
+            :class="$computedClass(getSelectAllStyles())"
+            :aria-selected="false"
```
MisRob: Here it seems we should update this value, based on whether an option is selected or not.

Better to check screen reader output after that's done, but I think then you may not need `sendPoliteMessage(${option.label} selected);` and `sendPoliteMessage(${option.label} deselected);`. Live regions are [to be used only when there's no ARIA](https://design-system.learningequality.org/usekliveregion) alternative.

lib/KMultiSelect.vue
```diff
@@ -0,0 +1,1084 @@
...
+              gap: '8px',
+              minHeight: '40px',
+              backgroundColor: getOptionBackgroundColor(option),
+              outline: getOptionOutline(option),
+              outlineOffset: getOptionOutlineOffset(option),
+            }"
+            :tabindex="focusedOption?.id === option.id ? 0 : -1"
+            @click.stop="toggleOption(option)"
+            @mouseenter="setFocusedOption(option)"
+            @keydown="handleOptionKeydown($event, option, index)"
```
MisRob: Would it be possible to attach `@keydown` listener to the whole list `<ul>` instead of each `<li>` option? This component will be used with large datasets, and adding a listener to every option can hurt performance. Generally, it’s a good web development pattern to use event delegation - attaching a single even listener to the parent.

I used `@keydown` here as one example - it applies to as many events as possible.

lib/KMultiSelect.vue
```diff
@@ -0,0 +1,1084 @@
...
+          class="dropdown-list"
+          role="listbox"
+          aria-multiselectable="true"
+          :style="{
+            margin: '0',
+            padding: '0',
+            listStyle: 'none',
+            outline: 'none',
+          }"
+          :aria-labelledby="ariaLabelledby"
```
MisRob: When I navigated to the input, I didn't hear the screen reader to announce the context for what I’m actually choosing - 'Pick some fruits'. I wonder if it's just caused by `index.vue` example not providing `aria-labelledby`?

lib/KMultiSelect.vue
```diff
@@ -0,0 +1,1084 @@
...
+          borderTop: 'none',
+          borderRadius: '0 0 4px 4px',
+          maxHeight: '400px',
+          overflowY: 'auto',
+        }"
+        role="region"
+        :aria-live="autocomplete ? 'polite' : 'off'"
+        @mousedown.prevent
+      >
+        <ul
```
MisRob: When I open the dropdown, I hear all options announced one by one. That's not desired - imagine a screen reader reading 200 options aloud :) I just checked Alex's implementation that Radina (a11y lead) approved to be sure, and it doesn't do that either. If you'd like to hear the expected screen reader output for the list part, in Kolibri repository you can `git checkout abc1f0cc6acc1b45fcd68769728bf95a646d2e80` and then navigate to `http://localhost:8000/en/facility/#/classes` where there is a live example.

lib/KMultiSelect.vue
MisRob, line 1: Would you consider breaking the logic into more `.vue` components?

And possibly splitting `setup()` into multiple `.js` [composables](https://vuejs.org/guide/reusability/composables.html)—e.g., one for pills logic, one for list selection, and one for text highlighting (one example in our codebase [are composables in cards folder](https://github.com/learningequality/kolibri-design-system/tree/develop/lib/cards))

The file’s already quite large, and splitting early will make future work easier. It's based on experience with another complex component, `KTable`, where we waited a bit too long to do this step and later ran into issues - by then, there was a lot more code to move around, and it was difficult. It would’ve been much smoother if we had set up the internal structure earlier.

GarvitSinghal47:
Hey @MisRob, I’ve resolved all the comments. Please check when you’re back and let me know if anything else needs to be fixed.

MisRob:
Thank you @GarvitSinghal47. I will re-review everything - likely next week or week after that.

MisRob (commented):
@GarvitSinghal47 thanks a lot for resolving the feedback! I think this sets a wonderful structure that will make upcoming work easier, and I confirm that those foundational a11y pieces, such as keyboard and focus ring, work well.

In the future, we will definitely QA and fine-tune everything, even though I think it will actually be best to do in the final stage of the component.

As for the component's skeleton stage, this is already large and helpful chunk of work and I think we can consider https://github.com/learningequality/studio/issues/5133 complete :tada: 

I will take it from here for some time to play with the next steps in component's development - likely its public API - I don't yet have clarity on how exactly it will look like.

Even though we won't merge this particular PR because this code will be part of Studio at first, and KDS is only for easier testing, know that it'll all be actively used and built upon :)

MisRob:
Even though we may use some of the code from here, closing because I didn't get to complete as I planned and later we decided to open multi-select as a GSoC project and take a different strategy in building it.

PR #1104, opened 2025-08-18
Set containerAppearanceOverrides to ktextbox

AlexVelezLl:
## Description
* KTextbox has an inner style `.textbox` that sets its max-width to 400px and cannot be changed.
* For some designs we need this max-width to change. So adding a `containerAppearanceOverrides` props could achieve it in a general way.
* There are several instances in Kolibri that have handled this by using [deep styles](https://github.com/search?q=repo%3Alearningequality%2Fkolibri%20%2Fdeep%2F%20.textbox&type=code), so this new prop should solve this in a more robust way.

## Changelog

  - **Description:** Adds `containerAppearanceOverrides` prop to KTextbox to override its container styles without using deep selectors.
  - **Addresses:** Will help with https://github.com/learningequality/studio/issues/5209.
  - **Components:** KTextbox.
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** .

### DIFF
--- a/lib/KTextbox/index.vue
+++ b/lib/KTextbox/index.vue
@@ -16,7 +16,7 @@
       :maxlength="maxlength"
       :autocomplete="autocomplete"
       :autocapitalize="autocapitalize"
-      :style="changedOrFocused ? $coreOutline : {}"
+      :style="[changedOrFocused ? $coreOutline : {}, appearanceOverrides]"
       :type="type"
       :min="min"
       :max="max"
@@ -163,6 +163,13 @@
       clearable: {
         type: Boolean,
         default: false,
+      },
+      /**
+       * CSS styles to apply to the container
+       */
+      appearanceOverrides: {
+        type: Object,
+        default: null,
       },
       /**
        * @ignore


MisRob:
I realize the current limitations is likely related to Keen's `UiTextbox` usage under the hood, @AlexVelezLi . So if you encounter anything blocking - I'd suggest this would be a good candidate for the first KDS backglog session.

MisRob:
I realize the current limitations is likely related to Keen's `UiTextbox` usage under the hood, Alex. 

Scope of my suggestion is larger then what you need, since it may require some related updates in consumers and ideally removal of all those `/deeps/`. I think it'd be fine to use `/deep/` for that new use-case, and just make sure we plan some time to address this in a way that will help KDS to be scalable.

This would be a good candidate for first KDS backlog prioritization session.

MisRob:
Wouldn't it make more sense to stop limiting max-width to `400px`? It's one helpful general principle of any design system - to let width (and also outer spacing such as margins) be controlled from outside as often as possible. Rather than KDS components being responsible for it - which is destined to fail, because they cannot capture all contexts. Basically textbox would take ideally 100% width. And then we could simply:

`<KTextbox :style={ width: ... , maxWidth: ... }/>`, or let it spread by default to its parent container.

So I'd suggest we try fixing core of the issue - it will be helpful for any future work with `KTextbox`. I realize the current limitations is likely related to Keen's UiTextbox usage under the hood, Alex. Scope of my suggestion is larger than what you need, since it may require some related updates in consumers and ideally removal of all those `/deep/`s. I think it'd be fine to use `/deep/` for that new use-case, and just make sure we plan some time to address this in a way that will help KDS to be scalable and intuitive to use.

This would be a good candidate for first KDS backlog prioritization session.

MisRob:
To be clear, from pure technical perspective for this PR, above comment is not blocking. I wouldn't mind having this new prop for some time. Generally, `appearanceOverride` approach is definitely better than `/deep/`.  So I see that perspective.

It's just that in this particular case, I don't think we'd want to migrate all those other `/deep/`s to `appearanceOverride` but rather address the issue along the lines of above comment. So raising it here for consideration - it can inform how you think about next steps. Feel free to do the final decision :) The main thing I feel the need for is to chat about how to address it long term, and we will have space for it soon.

lib/KTextbox/index.vue
```diff
@@ -164,6 +164,13 @@
         type: Boolean,
         default: false,
       },
+      /**
+       * CSS styles to apply to the container
+       */
+      containerAppearanceOverrides: {
```
MisRob: I think we could go just with usual `appearanceOverrides`.

Since I wouldn't expect we will need to override two inner elements, and have need for another kind of `appearanceOverrides` (and if if we needed to override two or more inner elements, that'd be time to think about how to architecture the component better).

MisRob (approved):
Thank you @AlexVelezLl, I appreciated that you're taking steps to prevent from `/deep/` troubles/ Please before merging, see comments above. If you decided to go with this approach, I'm already approving so you can hit merge ;)

AlexVelezLl:
> Wouldn't it make more sense to stop limiting max-width to 400px? It's one helpful general principle of any design system - to let width (and also outer spacing such as margins) be controlled from outside as often as possible.

Yes, I also think it does not make any sense that this restriction is set within the KTextbox component. But removing this max-width would be a breaking change, since we would need to take actions on consumers that may rely on this max-width.

> This would be a good candidate for first KDS backlog prioritization session.

Yes, and we can see if there are more examples like this in other components! And we can wrap this for a new major release? We can also remove this [".mh" div](https://github.com/AlexVelezLl/kolibri-design-system/blob/93f855fe70b20712de1b180579195a20bbd3a752/lib/KTextbox/index.vue#L3) that makes these container styles overrides more difficult. Basically because of this div we right now cannot just do `<KTextbox :style={ width: ... , maxWidth: ... }/>` because instead of styling the container, we would be styling this div that can be just merged with the textbox container.

> Feel free to do the final decision :) The main thing I feel the need for is to chat about how to address it long term, and we will have space for it soon.

Yes! I agree! And seeing how many instances of these /deep/ selectors we have in Kolibri is a good indicator that this is a topic that should be address more robustly.

MisRob:
:+1:

lib/KTextbox/index.vue
AlexVelezLl: Thanks! Will go with this!

AlexVelezLl:
Thanks @MisRob!

PR #1130, opened 2025-09-17
Fix sticky headers styles

AlexVelezLl:
## Description

While preparing the new KDS release, I tested it in Kolibri before releasing it, and I noticed two little bugs with the new KTable sticky columns:

* The sticky headers background was transparent, and the rows were visible underneath if the table was vertically scrollable.
* The sticky shadows were stacking, and looked darker when the table was scrolled vertically.

### Before/after screenshots

Before:

https://github.com/user-attachments/assets/09f71682-cdd5-44f5-9407-f6881d617e90

After:

https://github.com/user-attachments/assets/21bcd813-d809-4928-a75c-f0acf90a5f67

## Changelog

  - **Description:** Fixes background color of sticky KTable headers, and sticky column shadows being stacked.
  - **Addresses:** -.
  - **Components:** KTable.
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** 

## Steps to test

1. Run Kolibri with this version of KDS
2. Go to facility > users and look at the users table

### DIFF
--- a/lib/KTable/index.vue
+++ b/lib/KTable/index.vue
@@ -795,18 +795,25 @@
         return stickyIndices.includes(colIndex);
       },
       getStickyColumnStyle(rowIndex, colIndex) {
-        if (!this.colIndexIsSticky(colIndex)) {
-          return {};
+        const styles = {};
+
+        // Headers are always sticky, so they need a background color "stack" properly
+        if (rowIndex === -1) {
+          styles.backgroundColor = this.$themeTokens.surface;
         }
 
-        const styles = {
+        if (!this.colIndexIsSticky(colIndex)) {
+          return styles
+        }
+
+        if (this.hoveredRowIndex === rowIndex || this.focusedRowIndex === rowIndex) {
           // Handle background color based on hover/focus state
           // ensure row hover state appears but that sticky columns "stack" properly
-          backgroundColor:
-            this.hoveredRowIndex === rowIndex || this.focusedRowIndex === rowIndex
-              ? this.$themePalette.grey.v_100
-              : this.$themeTokens.surface,
-        };
+          styles.backgroundColor = this.$themePalette.grey.v_100;
+        } else {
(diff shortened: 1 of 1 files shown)


lib/KTable/index.vue
```diff
@@ -923,28 +933,58 @@
...
-  .sticky-column-shadow-right::after {
+  /*
+    A common problem when we have sticky rows/columns with shadows or transparent backgrounds is
+    that the shadows or transparency stack up when there are multiple sticky rows/columns stacked.
+    This makes that when the table is scrolled, the shadows or transparency become darker because
+    they stack up.
+    A way to solve this is to have a ::before pseudo element that has the surface color as
+    background that covers the exact same area the transparent background or shadow is applied to,
+    This way, the ::before element acts as a "base" layer that covers the back of
+    the transparent color, and don't let it stack up.
```
nucleogenesis, line 944: Smart thinking!

lib/KTable/index.vue
```diff
@@ -923,28 +933,58 @@
...
-    position: absolute;
-    top: 0;
-    bottom: 0;
-    left: -3px;
-    z-index: 1;
-    width: 3px;
-    pointer-events: none;
-    content: '';
-    background: linear-gradient(to left, rgba(0, 0, 0, 0.2), transparent);
+  @mixin sticky-column-shadow($direction) {
```
nucleogenesis, line 964: Love that you used a SASS `@mixin` here to handle the rtl/ltr differentiation

nucleogenesis (approved):
Changes look good to me. I ran it locally and didn't see weirdness w/ the shadows.

lib/KTable/index.vue
AlexVelezLl: Yees, we'd need to repeat these styles four times if not!

AlexVelezLl:
Thanks @nucleogenesis!

PR #1136, opened 2025-10-07
KTable: Set min height during loading to prevent table height inconsistencies 

LianaHarris360:
## Description

This pull request introduces an improved loading state within KTable. These changes reduce the height inconsistency while the table is reloading and improve the user experience when the table updates very quickly.

#### Issue addressed

Addresses Kolibri issue [#13766](https://github.com/learningequality/kolibri/issues/13766)

### Before/after screenshots

Before:

https://github.com/user-attachments/assets/a07ef3be-8e4e-4cef-b088-be260e97ee55

After:

https://github.com/user-attachments/assets/ca65a5ae-ba51-47d5-8c99-366bc8666790

## Changelog

  - **Description:** Updates KTable's loading state, reducing height inconsistencies and improving user experience.
  - **Products impact:** -
  - **Addresses:** https://github.com/learningequality/kolibri/issues/13766
  - **Components:** KTable
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -

## Steps to test

1. Run Kolibri with these changes: https://github.com/learningequality/kolibri/pull/13795
2. Navigate to Facility > Users
3. Observe that the table height stays consistent while reloading the data

- Introduced new `useLoading` composable to KTable to manage the loading state and delay showing the circular loader if loading < 300ms
- On initial load, the minimum height (`minHeightPx`) is set to the pixel height of the table header, plus one row.
- After the table finishes loading, `ResizeObserver` is used to set and watch the height of the `tableWrapper`, allowing the table height to stay the same on subsequent loads.
- The `KCircularLoader` is now wrapped in a div and centered within the table.
- `useLoading` ensures that the loader remains visible for a certain amount of time (`minVisibleMs`) to prevent a jarring user experience, if the loader is displayed.
- The table has been updated to be displayed with `v-show` instead of `v-if` to offset the higher toggle costs when the table is sortable

## Testing checklist

## Reviewer guidance

### DIFF
--- a/lib/KTable/index.vue
+++ b/lib/KTable/index.vue
@@ -3,151 +3,148 @@
   <div
     ref="tableWrapper"
     class="k-table-wrapper"
+    :style="wrapperInlineStyle"
   >
-    <template v-if="dataLoading">
-      <p><KCircularLoader /></p>
-    </template>
-    <template v-else>
-      <table
-        v-if="!isTableEmpty"
-        ref="tableElement"
-        class="k-table"
-        role="grid"
+    <div
+      v-show="loaderVisible"
+      class="k-table-loader"
+    >
+      <KCircularLoader />
+    </div>
+    <table
+      v-show="!loaderVisible && !isTableEmpty"
+      ref="tableElement"
+      class="k-table"
+      role="grid"
+    >
+      <caption
+        v-if="caption"
+        class="visuallyhidden"
       >
-        <caption
-          v-if="caption"
-          class="visuallyhidden"
-        >
-          {{
-            caption
-          }}
-        </caption>
-        <thead>
-          <tr ref="stickyHeader">
-            <th
-              v-for="(header, index) in headers"
-              :ref="'header-' + index"
-              :key="index"
-              tabindex="0"
-              :aria-sort="isColumnSortable(index) ? getAriaSort(index) : null"
-              :class="{
-                [$computedClass(coreOutlineFocus)]: true,
-                sortable: isColumnSortable(index),
-                'sticky-header': true,
-                'sticky-column': colIndexIsSticky(index),
-                'sticky-column-shadow-right':
-                  isTableScrollable && isRightmostLeftStickyColumn(index),
-                'sticky-column-shadow-left': isTableScrollable && isLastStickyColumn(index),
-              }"
-              :style="[
-                getHeaderStyle(header),
-                getStickyColumnStyle(-1, index), // Use -1 for header row
-                isColumnSortActive(index)
-                  ? { color: $themeBrand.primary.v_500 }
-                  : { color: $themePalette.grey.v_800 },
-                isColumnFocused(index) ? { backgroundColor: $themePalette.grey.v_100 } : {},
-                { textAlign: getTextAlign(header.dataType) },
-                { borderBottom: `1px solid ${$themeTokens.fineLine}` },
-              ]"
-              role="columnheader"
-              data-focus="true"
-              :aria-colindex="index + 1"
-              @click="sortable ? handleSort(index) : null"
-              @keydown="handleKeydown($event, -1, index)"
-            >
-              <!--@slot Scoped slot for customizing the content of each header cell.
-               Provides a header object `header` and its column index `colIndex`.-->
-              <slot
-                name="header"
-                :header="header"
-                :colIndex="index"
-              >
-                {{ header.label }}
-              </slot>
-              <span
-                v-if="isColumnSortable(index)"
-                class="sort-icon"
-              >
-                <span v-if="isColumnSortActive(index) && sortOrder === SORT_ORDER_ASC"><KIcon
-                  icon="dropup"
-                  :color="
-                    isColumnSortActive(index)
-                      ? $themeBrand.primary.v_600
-                      : $themePalette.grey.v_800
-                  "
-                /></span>
-                <span v-else-if="isColumnSortActive(index) && sortOrder === SORT_ORDER_DESC"><KIcon
-                  icon="dropdown"
-                  :color="
-                    isColumnSortActive(index)
-                      ? $themeBrand.primary.v_600
-                      : $themePalette.grey.v_800
-                  "
-                /></span>
-                <span v-else><KIcon
-                  icon="sortColumn"
-                  :color="$themePalette.grey.v_800"
-                /></span>
-              </span>
-            </th>
-          </tr>
-        </thead>
-        <tbody>
-          <tr
-            v-for="(row, rowIndex) in finalRows"
-            :key="rowIndex"
-            :style="getRowStyle(rowIndex)"
-            @mouseover="handleRowMouseOver(rowIndex)"
-            @mouseleave="handleRowMouseLeave"
+        {{
+          caption
+        }}
+      </caption>
+      <thead>
+        <tr ref="stickyHeader">
+          <th
+            v-for="(header, index) in headers"
+            :ref="'header-' + index"
+            :key="index"
+            tabindex="0"
+            :aria-sort="isColumnSortable(index) ? getAriaSort(index) : null"
+            :class="{
+              [$computedClass(coreOutlineFocus)]: true,
+              sortable: isColumnSortable(index),
+              'sticky-header': true,
+              'sticky-column': colIndexIsSticky(index),
+              'sticky-column-shadow-right': isTableScrollable && isRightmostLeftStickyColumn(index),
+              'sticky-column-shadow-left': isTableScrollable && isLastStickyColumn(index),
+            }"
+            :style="[
+              getHeaderStyle(header),
+              getStickyColumnStyle(-1, index), // Use -1 for header row
+              isColumnSortActive(index)
+                ? { color: $themeBrand.primary.v_500 }
+                : { color: $themePalette.grey.v_800 },
+              isColumnFocused(index) ? { backgroundColor: $themePalette.grey.v_100 } : {},
+              { textAlign: getTextAlign(header.dataType) },
+              { borderBottom: `1px solid ${$themeTokens.fineLine}` },
+            ]"
+            role="columnheader"
+            data-focus="true"
+            :aria-colindex="index + 1"
+            @click="sortable ? handleSort(index) : null"
+            @keydown="handleKeydown($event, -1, index)"
           >
-            <KTableGridItem
-              v-for="(col, colIndex) in row"
-              :ref="'cell-' + rowIndex + '-' + colIndex"
-              :key="colIndex"
-              :content="col"
-              :dataType="headers[colIndex].dataType"
-              :minWidth="headers[colIndex].minWidth"
-              :width="headers[colIndex].width"
-              :rowIndex="rowIndex"
-              :colIndex="colIndex"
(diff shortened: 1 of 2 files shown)


lib/KTable/index.vue
```diff
@@ -195,6 +198,14 @@
 
       const isTableEmpty = computed(() => sortedRows.value.length === 0);
 
+      const loaderVisible = computed(() => show('KTableLoader', props.dataLoading, MIN_VISIBLE_MS));
```
AlexVelezLl: I am always unsure about when to use the useKShow composable. Is this a good use case? Is there a drawback about updating the table quicker if the api response is quicker? No hard preferences, just curious!

lib/KTable/index.vue
```diff
@@ -240,9 +251,24 @@
...
+            if (tableWrapper.value) resizeObserver.observe(tableWrapper.value);
+          }
         });
       });
 
+      onBeforeUnmount(() => {
+        if (resizeObserver) {
+          resizeObserver.disconnect();
+        }
+      });
```
AlexVelezLl: I was thinking that perhaps we can make the `useKResponsiveElement` composable more general to accept a custom element via props, so we can use it here, but that may add more complexities and overhead to this PR for just one use case, so dont think its worth. However, in that composable we wrapped the ResizeObserver callback in a `requestAnimationFrame` callback for some errors thrown in [Studio](https://github.com/learningequality/kolibri-design-system/issues/960), perhaps is a good idea to do it here too?

AlexVelezLl (commented):
Code changes looks good, and did not spot anything during manual QA, lgtm! Just a minor non-blocking comment 👐. Thanks @LianaHarris360!

lib/KTable/index.vue
LianaHarris360: Yes, I probably should've mentioned this directly in the PR description. The first updates only changed the minimum height of the table when it loads. However, the API would respond very quickly in some cases, and there would be a brief flash of the KCircularLoader right before the table data appears, which can be a sudden/jarring user experience, but this could also be a matter of preference.

I tried using the KCircularLoader props `delay` (the results weren't that much different), and `minVisibleTime` and `shouldShow` (the loader would show on top of the table rows during transitioning). I decided to use this composable because the KDS [documentation](https://design-system.learningequality.org/kcircularloader#related) suggested using useKShow when switching between KCircularLoader and more components.

I believe there's a good chance using the `delay` prop and a smaller `MIN_VISIBLE_MS` has the smoothest results.

https://github.com/user-attachments/assets/bdcf456c-9fee-4c65-8f0b-15b19df8dda2

lib/KTable/index.vue
LianaHarris360: I agree, eventually, it might be worth it to update `useKResponsiveElement`, but for this use case it might not be necessary. Adding the `requestAnimationFrame` callback looks like a good idea, good catch!

lib/KTable/index.vue
MisRob: I appreciate the consideration.

I'm not suggesting any changes, just noticed there were some questions around using delays and minimum visible times.

I think that discussing with designers is always helpful, and I also often refer to this [very good article](https://www.pencilandpaper.io/articles/ux-pattern-analysis-loading-feedback).

If that's helpful for reference, [these card skeleton loader values](https://design-system.learningequality.org/kcardgrid#loading-state) were agreed with designers:

> The loading skeletons won't be displayed for short loading times (< 1s)
> When the loading skeletons are displayed, they will be visible for at least 1s

It's quite a different use-case though.

Generally, a tricky thing about our use-cases is that API response can be very fast as well as very slow and anything in between, and we are not able to estimate. Perhaps just trying the experience on both slower and faster networks and feeling into a good compromise will do for now :)

LianaHarris360: Agreed, that is a great article. It touches on a pain point I encountered: when the wait is between 0.1 and 1 second, it is barely noticeable for the user and recommends that looped animations be avoided. However, in this specific case, the wait could be 0.1 seconds or it could take over 2 seconds. For users experiencing the 0.1-second wait, the looped animation quickly flashes and is distracting. So, I opted to implement `useKShow` to ensure the loader is visible for a minimum amount of time for all users.

However, after looking over the loading state for `KCardGrid`, I'm now wondering if it would be best to use a similar loading mechanism as `useGridLoading`, and not show the loader for loading times shorter than 1 second (or whatever timeframe makes sense for this use-case). And if the `KCircularLoader` is displayed, make it visible for at least 1s.

MisRob: I think that preventing the flash with `useKShow` is definitely a good step - it's aligned with what we agreed on in multiple use-cases in the past.

> not show the loader for loading times shorter than 1 second

Just FYI we observed one hiccup -  this means that we aren't able to display anything until that 1 second (since we need to wait 1s to see which approach to use) - which then has a observable impact in the case of a fast API response when rendering is delayed unnecessarily. With grids we evaluated it was a good compromise to begin with, with that we may need to adjust the exact value over time. I am a bit suspicious that in this paginated use-case, it would be even more observable - imagine 1s delay before displaying the loader or the table data. You're welcome to experiment of course, I just wanted to make sure I don't complicate things more than necessary - work in this PR is already a big improvement :)

Thinking about this discussion, and our doubts about loading states - I think it'd be best to discuss creating some guidance with the design team. I can raise it during KDS backlog planning. Then we could know more about suitable values, and consider abstracting some logic from `useGridLoading`, or adding `delay` to `useKShow`.

LianaHarris360: Agreed, a delay of 1s is too long before the table can display anything and just looks like a lag on the device. Experimenting with the network settings (between 4G and 3G data speeds), I found that any delay over 300ms caused a noticeable delay, and shorter values resulted in the loader appearing as a flash. 

I agree that it would be best to discuss the guidance around this with the design team to make sure the loading state is aligned with what we expect for a table. The best option moving forward would probably be a mixture of abstracting some logic from `useGridLoading` and adding `delay` to `useKShow`, but I implemented `useLoading` specifically for KTable before realizing this, because the file was becoming so large after the tracked loading and height additions.

But perhaps it can be used as an initial starting point for us to know what values need to be added to `useKShow` to adapt it for the same delayed loading usage in KTable.

MisRob: Definitely - and I hope it was clear from the outset that I was happy with the solution - the KDS backlog discussion for guidance is meant outside the scope of this PR too.

 No problem at all around useLoading - as long as it allows you to achieve the experience that seems to be best for users, given what we know now, all good.

MisRob: Thanks for exploring the values,  300ms may be eventually better threshold for card grid too. I will soon need to test grid loading in Studio so I will try this value un that use case as well. Perhaps we will have a good general basis for the guidance then.

MisRob:
@LianaHarris360 I gave the solution high-level look and it makes sense to me - thanks again for consideration about the loading experience and taking care to fix everything in KDS.

I will leave the final approve to @AlexVelezLl who has reviewed in more detail.

AlexVelezLl:
(This is totally subjective, and just want to share my thoughts, but it can be ignored if this is the desired behavior)

There's a slight difference from the behavior we have with cards. With the KCardGrid, the cards disappear as soon as the loading time starts. If it takes longer than 1 second, the spinner is displayed. In this implementation, the current items in the table don't disappear when loading starts, and it does so 300ms later. This is enough delay to give the perception of laggy behavior in the app, since the response time between pressing a button and having something happen in the application is noticeable, so much so that I can even click the options buttons before the loader is displayed and create UI inconsistent states.

I'm wondering if it would be better to replicate the exact behavior of the cards and not display any rows or loaders in the first 300ms of loading, so that there's a quicker reaction to user input and prevent any inconsistent behavior? It's like "okay, we're not showing the spinner yet, but be aware that something happened when you pressed that button".


https://github.com/user-attachments/assets/5db89e73-1bde-4287-9b52-1f666fd92bc3

LianaHarris360:
Thanks for pointing this out, I like this suggestion, it is a quick way to remove that perceived lag. I noticed the delay and initially tried to avoid it by lowering the `minVisibleMs`, but that didn't resolve the issue completely. This change will simply involve hiding the table if the `delayActive` value is true, which is simple to do.

lib/KTable/index.vue
```diff
@@ -193,6 +191,15 @@
         getAriaSort,
       } = useSorting(headers, rows, defaultSort, disableBuiltinSorting);
 
+      const { delayActive, loaderVisible, wrapperInlineStyle } = useLoading(props, {
+        wrapperRef: tableWrapper,
+        loadingDelay: 300, // delay showing loader by 300ms
+        minVisibleMs: 350, // keep loader visible for at least 350ms
+        minHeightPx: 120,
```
nucleogenesis, line 198: Since these are the defaults and there is no way to override them from outside, I'd remove these and do `useLoading(props)`.

However, if we want to give implementing products the ability to define these values themselves then we'd want to add props to KTable for the `useLoading` configuration values then pass them in here when present.

I don't think it's _necessary_ to make it something we can tweak while we use KTable in our products, but it might be a good idea in case we find that these specific values don't jive in particular contexts (ie, long loads vs short loads).

lib/KTable/useLoading.js
```diff
@@ -0,0 +1,104 @@
...
+          loaderVisible.value = false;
+          loadingStartTime = null;
+        }
+      }
+    },
+    { immediate: true },
+  );
+
+  onMounted(() => {
+    // Initialize ResizeObserver if browser supports it
```
nucleogenesis, line 80: What is the experience / outcome when there is no ResizeObserver supported? (Not a blocker just curious if this might cause issues somehow in obscure browsers/versions)

nucleogenesis (approved):
Comments are non-blocking - although I'd encourage you to think on if the props should be tweakable from within the products using KTable through prop-drilling the values for delay et al.

I tested this against my KTable design refactoring PR in Kolibri and things looked and worked great!

AlexVelezLl:
Code changes look good to me too! @LianaHarris360, could you please update the reference of https://github.com/learningequality/kolibri/pull/13795 to the last commit of this PR so we can have QA Testing it before merging this PR? Thanks!

lib/KTable/useLoading.js
LianaHarris360: In that case, the last height of the table with loaded data wouldn't be saved to `lastStableHeight` before loading, so when loading does start, the height will collapse to the minimum height (`120px`). It would result in a less smooth, but still slightly improved loading experience. [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#browser_compatibility) is generally well supported by modern browsers, I think the only case this would apply to is potentially Opera Mini, since we don't provide support Internet Explorer anymore if I remember correctly.

lib/KTable/index.vue
LianaHarris360: That's a good point, thanks for bringing that up. There might eventually be cases where the specific values need to be adjusted based on the context as `KTable` is added to more places in Kolibri. I am not particularly leaning towards one way or the other but feedback from Radina or Peter on the Kolibri PR might give insight into this as well.

AlexVelezLl (approved):
We got the ✅  from QA in https://github.com/learningequality/kolibri/pull/13795! Thanks @LianaHarris360! 🎉

PR #1140, opened 2025-10-13
Add v-model support to KCheckbox

### LINKED ISSUE #1135: Add v-model support to KCheckbox
🙂 Looking for an issue? Welcome! This issue is open for contribution. If this is the first time you’re requesting an issue, please:

- **Read <a href="https://learningequality.org/contributing-to-our-open-code-base/" target="_blank">Contributing guidelines</a>** carefully. **Pay extra attention to [Using generative AI](https://learningequality.org/contributing-to-our-open-code-base/#using-generative-ai)**. **Pull requests and comments that don’t follow the guidelines won’t be answered.**
- **Confirm that you’ve read the guidelines** in your comment.

Sub-issue of https://github.com/learningequality/studio/issues/5060.

**Complexity: Medium**

## Summary

**Add `v-model` support to `KCheckbox`.**

Currently, `KCheckbox` doesn't support `v-model` and so to dynamically update selected choices, there's a need to use the combination of `:checked` and `@changed`, for example:

```html
<KCheckbox
  v-for="group in groups"
  :label="group.name"
  :key="group.id"
  :checked="groupIsSelected(group)"
  @change="toggleGroup($event, group)"
/>
```

```js
export default {
  data() {
    return {
      groups: [
        { "id": "group-1", "name": "First group", },
        { "id": "group-2", "name": "Second group" },
        { "id": "group-3", "name": "Third group" }
      ],
      selectedGroupIds: [],
    };
  },
  methods: {
    groupIsSelected({ id }) {
      return this.selectedGroupIds.includes(id);
    },
    toggleGroup(isChecked, { id }) {
      if (isChecked) {
        this.selectedGroupIds = [...this.selectedGroupIds, id];
      } else {
        this.selectedGroupIds = this.selectedGroupIds.filter(
          groupId => groupId !== id,
        );
      }
    },
  },
};
```

The goal of this issue is to add logic to `KCheckbox` that will allow simplification of the above example into

```html
<KCheckbox
  v-for="group in groups"
  v-model="selectedGroupIds"
  :label="group.name"
  :key="group.id"
  :value="group.id"
/>
```

```js
export default {
  data() {
    return {
      groups: [
        { "id": "group-1", "name": "First group", },
        { "id": "group-2", "name": "Second group" },
        { "id": "group-3", "name": "Third group" }
      ],
      selectedGroupIds: [],
    };
  },
};
```

where `value` holds a value to be added/removed to/from the array referenced in `v-model` when the checkbox is checked/unchecked.

Furthemore, `v-model` should not only support arrays, but also the following types (typically used in scenarios with a single checkbox):

**Boolean**

If checked, `isSelected` will become `true`, otherwise `false`. Note that here, `value` is not needed.

```html
<KCheckbox
  v-model="isSelected"
/>
```

**Number**

If checked, `selectedNumber` will be set to `value`, otherwise `null`.

```html
<KCheckbox
  v-model="selectedNumber"
  :value="1"
/>
```

**String**

If checked, `selectedString` will be set to `value`, otherwise `null`.

```html
<KCheckbox
  v-model="selectedString"
  :value="'String'"
/>
```

**Object**

If checked, `selectedObject` will be set to `value`, otherwise `null`.

```html
<KCheckbox
  v-model="selectedObject"
  :value="{ id: 1, name: 'Name 1' }"
/>
```

## Value add

In support of https://github.com/learningequality/studio/issues/5060. Will allow removal of the temporary `Checkbox` wrapper in favour of the native `KCheckbox`.

## Guidance

- [Code from Studio's `Checkbox`](https://github.com/learningequality/studio/blob/eea9a8944d588601eb494f087352e5af471a516f/contentcuration/contentcuration/frontend/shared/views/form/Checkbox.vue), which has `v-model` logic added on top of `KCheckbox` can be re-used.
- Note how the indeterminate state currently overrides checked state. This behavior should be preserved when `KCheckbox` used with `v-model`. 
- [Getting started](https://github.com/learningequality/kolibri-design-system/blob/develop/dev_docs/01_getting_started.md)
- [How to update library](https://github.com/learningequality/kolibri-design-system/blob/develop/dev_docs/03_how_to_update_library.md)
- [Visual testing guide](https://github.com/learningequality/kolibri-design-system/blob/develop/dev_docs/07_visual_testing_guide.md)
- [How to update the documentation website](https://github.com/learningequality/kolibri-design-system/blob/develop/dev_docs/04_how_to_update_docs.md)
- [Writing guidelines](https://design-system.learningequality.org/writing)
- Refer to other component documentation pages, e.g. [KCard documentation](https://design-system.learningequality.org/kcard) ([source code](https://github.com/learningequality/kolibri-design-system/blob/develop/docs/pages/kcard.vue)) 

## Acceptance criteria

- [ ] There are unit tests. Do not use obsolete `@vue/test-utils` approach. Instead, use `@testing-library/vue` ([Vue Testing Library](https://kolibri-dev.readthedocs.io/en/develop/frontend_architecture/unit_testing.html)).
- [ ] There are visual tests
- [ ] `v-model` is documented on [`KCheckbox` documentation](https://design-system.learningequality.org/kcheckbox) page, including new examples. Documentation style follows the writing guidelines, is concise and consistent with the existing documentation.

WinnyChang:
## Description

- Add `v-model` support to `KCheckbox`.
- Update the `KCheckbox` documentation page.
- Update unit tests and visual tests to work with `KCheckbox`'s v-model. ([Percy build](https://percy.io/410c0afb/web/KDS-KCheckbox-5874ec59/builds/43811962/changed/2323378163?browser=edge&browser_ids=63%2C69%2C70%2C71&group_snapshots_by=similar_diff&subcategories=approved&viewLayout=side-by-side&viewMode=new&width=520&widths=300%2C375%2C400%2C520%2C600%2C768%2C800%2C900%2C1200%2C1380) for KCheckbox visual tests)

#### Issue addressed
Fixes #1135 

## Changelog

  - **Description:** Add v-model support to KCheckbox; update documentation page, unit tests, and visual tests.
  - **Products impact:** none
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/1135
  - **Components:** -
  - **Breaking:** -
  - **Impacts a11y:** -
  - **Guidance:** -

## Testing checklist

## Reviewer guidance

- [ ] There are unit tests, and they use `@testing-library/vue`.
- [ ] There are visual tests.
- [ ] v-model is documented on [KCheckbox documentation](https://design-system.learningequality.org/kcheckbox) page, including new examples.

### DIFF
--- a/docs/pages/kcheckbox.vue
+++ b/docs/pages/kcheckbox.vue
@@ -8,14 +8,12 @@
       <p>The checkbox is generally used to select one of two possible values in a form.</p>
       <DocsShow>
         <KCheckbox
+          v-model="checked1"
           label="Some label"
-          :checked="checked1"
-          @change="checked1 = !checked1"
         />
         <KCheckbox
+          v-model="checked2"
           label="Another label"
-          :checked="checked2"
-          @change="checked2 = !checked2"
         />
       </DocsShow>
     </DocsPageSection>
@@ -48,17 +46,17 @@
       title="States"
       anchor="#states"
     >
-      <p>The checked state represents an affirmative value.</p>
+      <p>The v-model checkbox state represents an affirmative value.</p>
       <p>
         Checkboxes can also have a "partially-checked" or "indeterminate" state used in cases where
         the value is neither true nor false, such as when a subset of a topic is selected:
       </p>
       <DocsShow>
         <KCheckbox
+          v-model="checked3"
           label="Topic is selected"
           :indeterminate="indeterminate3"
-          :checked="checked3"
-          @change="handle3"
+          @input="handle3"
         />
       </DocsShow>
       <p>
@@ -81,9 +79,8 @@
       <!-- eslint-enable -->
       <DocsShow>
         <KCheckbox
+          v-model="checked4"
           label="First lesson"
-          :checked="checked4"
-          @change="checked4 = !checked4"
         />
       </DocsShow>
 
@@ -102,10 +99,7 @@
       </DocsShowCode>
       <!-- eslint-enable -->
       <DocsShow>
-        <KCheckbox
-          :checked="checked5"
-          @change="checked5 = !checked5"
-        >
+        <KCheckbox v-model="checked5">
           <KLabeledIcon
             icon="lesson"
             label="First lesson"
@@ -137,6 +131,200 @@
         </template>
       </DocsDoNot>
     </DocsPageSection>
+
+    <DocsPageSection
+      title="Example: v-model"
+      anchor="#example-v-model"
+    >
+      <code>v-model</code> supports array, boolean, number, string, and object.
+
+      <h3>
+        Array
+        <DocsAnchorTarget anchor="#v-model-array" />
+      </h3>
+      <p>
+        Each <code>value</code> will be added/removed from the <code>selectedGroupIds</code> array
+        based on each checkbox state.
+      </p>
+      <!-- eslint-disable -->
+      <!-- prettier-ignore -->
+      <DocsShowCode language="html">
+        <KCheckbox
+          v-for="group in groups"
+          :key="group.id"
+          v-model="selectedGroupIds"
+          :label="group.name"
+          :value="group.id"
+        />
+      </DocsShowCode>
+      <!-- prettier-ignore -->
+      <DocsShowCode language="javascript">
+        export default {
+          data() {
+            return {
+              groups: [
+                { 'id': 'group-1', 'name': 'First group', },
+                { 'id': 'group-2', 'name': 'Second group' },
+                { 'id': 'group-3', 'name': 'Third group' }
+              ],
+              selectedGroupIds: ['group-1'],
+            };
+          },
+        };
+      </DocsShowCode>
+      <!-- eslint-enable -->
+      <DocsShow>
+        <KCheckbox
+          v-for="group in groups"
+          :key="group.id"
+          v-model="selectedGroupIds"
+          :label="group.name"
+          :value="group.id"
+        />
+      </DocsShow>
+
+      <h3>
+        Boolean
+        <DocsAnchorTarget anchor="#v-model-boolean" />
+      </h3>
+      <p>
+        If checked, <code>isSelected</code> will become <code>true</code>, otherwise
+        <code>false</code>. Note that here, <code>value</code> is not needed.
+      </p>
+      <!-- eslint-disable -->
+      <!-- prettier-ignore -->
+      <DocsShowCode language="html">
+        <KCheckbox
+          v-model="isSelected"
+          :label="`isSelected = ${isSelected}`"
+        />
+      </DocsShowCode>
+      <!-- prettier-ignore -->
+      <DocsShowCode language="javascript">
+        export default {
+          data() {
+            return {
+              isSelected: true,
+            };
+          },
+        };
+      </DocsShowCode>
+      <!-- eslint-enable -->
+      <DocsShow>
+        <KCheckbox
+          v-model="isSelected"
+          :label="`isSelected = ${isSelected}`"
+        />
+      </DocsShow>
+
+      <h3>
+        Number
+        <DocsAnchorTarget anchor="#v-model-number" />
+      </h3>
+      <p>
+        If checked, <code>selectedNumber</code> will be set to <code>value</code>, otherwise
+        <code>null</code>.
+      </p>
+      <!-- eslint-disable -->
+      <!-- prettier-ignore -->
+      <DocsShowCode language="html">
+        <KCheckbox
+          v-model="selectedNumber"
+          :label="`selectedNumber = ${selectedNumber}`"
+          :value="1"
+        />
+      </DocsShowCode>
+      <!-- prettier-ignore -->
+      <DocsShowCode language="javascript">
+        export default {
+          data() {
+            return {
+              selectedNumber: 1,
+            };
+          },
+        };
+      </DocsShowCode>
+      <!-- eslint-enable -->
+      <DocsShow>
+        <KCheckbox
+          v-model="selectedNumber"
+          :label="`selectedNumber = ${selectedNumber}`"
+          :value="1"
+        />
+      </DocsShow>
+
+      <h3>
+        String
+        <DocsAnchorTarget anchor="#v-model-string" />
+      </h3>
+      <p>
+        If checked, <code>selectedString</code> will be set to <code>value</code>, otherwise
+        <code>null</code>.
+      </p>
+      <!-- eslint-disable -->
+      <!-- prettier-ignore -->
+      <DocsShowCode language="html">
+        <KCheckbox
+          v-model="selectedString"
+          :label="`selectedString = ${selectedString}`"
+          :value="'String'"
+        />
+      </DocsShowCode>
+      <!-- prettier-ignore -->
+      <DocsShowCode language="javascript">
+        export default {
+          data() {
+            return {
+              selectedString: 'String',
+            };
+          },
+        };
+      </DocsShowCode>
+      <!-- eslint-enable -->
+      <DocsShow>
+        <KCheckbox
+          v-model="selectedString"
+          :label="`selectedString = ${selectedString}`"
+          :value="'String'"
+        />
+      </DocsShow>
+
+      <h3>
+        Object
+        <DocsAnchorTarget anchor="#v-model-object" />
+      </h3>
+      <p>
+        If checked, <code>selectedObject</code> will be set to <code>value</code>, otherwise
+        <code>null</code>.
+      </p>
+      <!-- eslint-disable -->
+      <!-- prettier-ignore -->
+      <DocsShowCode language="html">
+        <KCheckbox
+          v-model="selectedObject"
+          :label="`selectedObject = ${JSON.stringify(selectedObject)}`"
+          :value="{ id: 1, name: 'Name 1' }"
+        />
+      </DocsShowCode>
+      <!-- prettier-ignore -->
+      <DocsShowCode language="javascript">
+        export default {
+          data() {
+            return {
+              selectedObject: { id: 1, name: 'Name 1' },
+            };
+          },
+        };
+      </DocsShowCode>
+      <!-- eslint-enable -->
+      <DocsShow>
+        <KCheckbox
+          v-model="selectedObject"
+          :label="`selectedObject = ${JSON.stringify(selectedObject)}`"
+          :value="{ id: 1, name: 'Name 1' }"
+        />
+      </DocsShow>
+    </DocsPageSection>
   </DocsPageTemplate>
 
 </template>
@@ -153,11 +341,20 @@
         indeterminate3: true,
         checked4: false,
         checked5: false,
+        groups: [
+          { id: 'group-1', name: 'First group' },
+          { id: 'group-2', name: 'Second group' },
+          { id: 'group-3', name: 'Third group' },
+        ],
+        selectedGroupIds: ['group-1'],
+        isSelected: true,
+        selectedNumber: 1,
+        selectedString: 'String',
+        selectedObject: { id: 1, name: 'Name 1' },
       };
     },
     methods: {
       handle3() {
-        this.checked3 = !this.checked3;
         this.indeterminate3 = false;
       },
     },
--- a/docs/pages/kcheckbox.vue
+++ b/docs/pages/kcheckbox.vue
@@ -59,7 +59,7 @@
           v-model="checked3"
           label="Topic is selected"
           :indeterminate="indeterminate3"
-          @input="handle3"
+          @change="handle3"
         />
       </DocsShow>
       <p>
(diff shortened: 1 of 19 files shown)


MisRob:
Hi @WinnyChang, good to see you around and thanks a lot :) I will review documentation. For the feature itself, we will assign a reviewer next week.

WinnyChang:
@MisRob Got it. Thanks for the update!

docs/pages/kcheckbox.vue
```diff
@@ -137,6 +131,200 @@
...
+      </DocsShowCode>
+      <!-- eslint-enable -->
+      <DocsShow>
+        <KCheckbox
+          v-model="selectedObject"
+          :label="`selectedObject = ${JSON.stringify(selectedObject)}`"
+          :value="{ id: 1, name: 'Name 1' }"
+        />
+      </DocsShow>
+    </DocsPageSection>
```
AlexVelezLl: I think this may be a great time to use the `DocsExample` component and create/reuse example files showing this?

lib/KCheckbox/index.vue
```diff
@@ -147,8 +176,45 @@
...
+        get() {
+          if (Array.isArray(this.inputValue)) {
+            return this.inputValue.includes(this.value);
+          }
+
+          if (typeof this.inputValue === 'boolean') {
+            return this.inputValue;
+          }
+
+          return Boolean(this.inputValue);
```
AlexVelezLl: Not sure if this is even a possible use case, but... since we are allowing numbers and strings that will be emitted using the `value` prop, if for some reason we have a falsy value for them, e.g. value=0, then `inputValue` would be set to `0`, and that would mean the checkbox is selected, but this `Boolean(this.inputValue` would take it as not checked. To fully support this and prevent any unintended behavior, I think we can be more explicit here and say that it is checked only if it is different from Null or Undefined, which are the actual falsy values in these cases.

lib/KCheckbox/index.vue
```diff
@@ -104,19 +112,40 @@
         default: true,
       },
       /**
-       * Checked state
+       * Reactive v-model checkbox state. Updates with the `input` event.
        */
```
AlexVelezLl, line 114: Are these double comment blocks the intended result to only show the first one in the docs page?

lib/KCheckbox/index.vue
```diff
@@ -104,19 +112,40 @@
...
-       * Indeterminate state. Overrides `checked` state
+       * Indeterminate state. Overrides `v-model`.
        */
       indeterminate: {
         type: Boolean,
         default: false,
       },
+      /**
+       * Checkbox value
+       */
```
AlexVelezLl, line 144: I think it'd be great if we specify here that this is the value that will be set by the `v-model` when the checkbox is checked.

AlexVelezLl (changes requested):
Thanks a lot, @WinnyChang, this looks pretty good! However, this change was not intended to be a breaking change, i.e,. KDS consumers should not need to migrate all existing checkboxes to the new `v-model` api after this change is released; it should be an additional, optional api. So both APIs should be supported by KCheckbox: the `v-model` one and the `:checked/@change` API. Apart from that, this looks good!

MisRob:
> this change was not intended to be a breaking change, i.e,. KDS consumers should not need to migrate all existing checkboxes to the new v-model api after this change is released

Ah yes, I should have mentioned it explicitly in the issue - apologies @WinnyChang @AlexVelezLl.

Preventing a breaking change is an important aspect, and another reason is that even though `v-model` will be likely most commonly used, in some more nuanced cases, we may still need `:checked/@change`.

As for the documentation, I would suggest that the above is reflected:
- All examples updated to use to `v-model` since we want to encourage it whenever possible
- A new section that demonstrates how to use `:checked/@change` so developers are aware there's that possibility and how to use it

lib/KCheckbox/index.vue
WinnyChang: Yes, that’s intentional — only the first comment block is shown. The hidden ones are copied from [Studio's `Checkbox`](https://github.com/learningequality/studio/blob/eea9a8944d588601eb494f087352e5af471a516f/contentcuration/contentcuration/frontend/shared/views/form/Checkbox.vue).

I chose to show only the first ones since the comments are quite long, but I can include all of them in the docs page if you think that would be better.

docs/pages/kcheckbox.vue
WinnyChang: It is! The updated code with `DocsExample` looks much cleaner.

lib/KCheckbox/index.vue
WinnyChang: Good catch! Having 0 as the checkbox value is definitely possible. I've updated it to `return Boolean(this.inputValue !== null);`. Does [this line](https://github.com/learningequality/studio/blob/eea9a8944d588601eb494f087352e5af471a516f/contentcuration/contentcuration/frontend/shared/views/form/Checkbox.vue#L111) in Studio's `Checkbox.vue` also need to be updated as well?

lib/KCheckbox/index.vue
WinnyChang: Updated! Please let me know if it's clear enough or if it needs any adjustments.

lib/KCheckbox/index.vue
MisRob: > Does [this line](https://github.com/learningequality/studio/blob/eea9a8944d588601eb494f087352e5af471a516f/contentcuration/contentcuration/frontend/shared/views/form/Checkbox.vue#L111) in Studio's Checkbox.vue also need to be updated as well?

We will remove Studio's `Checkbox` in favor of this updated `KCheckbox`, so I wouldn't be worried :)

MisRob:
Just leaving a note that I skimmed quickly through the updated documentation and it felt clear to me, and as balanced amount of information. Very nice work, thank you.

lib/KCheckbox/index.vue
```diff
@@ -147,8 +186,50 @@
       isActive: false,
     }),
     computed: {
+      usingLegacyApi() {
+        return this.checked !== undefined;
+      },
```
AlexVelezLl, line 189: Just to be extra cautious, the old API assumed that if `checked` was `null` or `undefined`, the checkbox was unchecked, and there may be some consumers who rely on this too. So, to determine if the legacy api is being used, I think we can instead assert against the new `value` or `inputValue`, and if any of them are set, then we are not using the legacy API (which would 100% cover all examples of KCheckbox right now).

lib/KCheckbox/index.vue
```diff
@@ -179,16 +260,26 @@
...
+          }
         }
       },
+      updateInputValue(newValue) {
+        /**
+         * Emits input event with the new v-model checkbox state.
+         * Used by v-model to keep the checkbox state in sync.
+         */
+        this.$emit('input', newValue);
+      },
```
AlexVelezLl, line 274: I think we can centralize the emission of the input/change event in the `updateInputValue` and let it decide which event to emit. This way, we abstract the change of the `isChecked` computed property. If, for any reason, another method changes its value, the different methods should not have to determine whether to emit or not the `change` event. (Not sure if I made this clear, if not please let me know 😅)

lib/KCheckbox/index.vue
AlexVelezLl: From a brief skim, I found [a place in Kolibri](https://github.com/learningequality/kolibri/blob/beae6c5fee66b274c047036f7fa846fbb4340e70/packages/kolibri-common/components/SearchFiltersPanel/index.vue#L112) where we assume that `undefined` is the falsy value of  the checkbox, and this PR is introducing a regression on this checkbox:


https://github.com/user-attachments/assets/a29184f8-ce92-4e82-ac62-201cddb9514d

lib/KCheckbox/index.vue
AlexVelezLl: The other option may be to use the `@change` event as the input event for the `v-model` instead of `@input`, and just always emit this event for all values independently of the api that is being used. I think that may be the safest way to prevent any unintended behavior if, for any reason, a consumer sets an `undefined` value as a falsy value, but they intended to use the `v-model` api. What do you think @WinnyChang, @MisRob?

This will prevent any confusion if we ever, for any reason (not reading the documentation well 😅), declare something like:

```Vue
<template>
  <KCheckbox
    v-model="checkboxValue"
    ...
  />
</template>
<script>
  ...
  setup() {
    return {
      checkboxValue: ref(),
    }
  }
</script>
```

This would fail if we determine that this is not using the new api because both their value and input value are undefined, then we won't emit the `input` event, and would emit the `change` event instead. But if we used the same event for both, we wouldn't need to worry about cases like this.

AlexVelezLl (changes requested):
Thanks @WinnyChang! We're almost there! Just found a little regression in Kolibri because of an unrecognized legacy api usage, and reflected on a possible way this could be implemented safely.

lib/KCheckbox/index.vue
WinnyChang: > I think we can centralize the emission of the input/change event in the updateInputValue and let it decide which event to emit.

In my current code, I separate the mechanism of the legacy API and v-model, so the legacy API won't trigger `updateInputValue`. Is it okay to trigger `updateInputValue` when using the legacy API and put the `input`/`change` emissions together?

AlexVelezLl: Yes! That would be ideal! It should be the responsibility of the computed property to determine how to manage the changes to the `isChecked` property, in any case. This helps us centralize the property update mechanism. If, for some reason, we want to change the value of this variable in another method (without using the current toggle method) in the future, the legacy update mechanism will still work without duplicating code!

WinnyChang: > The other option may be to use the @change event as the input event for the v-model instead of @input, and just always emit this event for all values independently of the api that is being used.

This sounds like a good idea. The only concern I have is that `@change` is with a boolean, while `@input` is with array/boolean/number/string/object. I wonder if this will cause any issues in some cases.

I also thought about always emitting both events, no matter which API is used. But I wasn't sure if it follows the best practices. 😂

AlexVelezLl: Just coordinated with the team, and yes, we agreed on having the `change` event as the event for the `v-model` api @WinnyChang 😄. So we would just always emit the `@change` event

WinnyChang: Got it! I'll update accordingly.

MisRob:
Thanks @WinnyChang - we will re-review. We have more internal commitments this and next week so it may take longer than usual.

lib/KCheckbox/index.vue
```diff
@@ -145,10 +181,51 @@
...
+        get() {
+          if (this.usingLegacyApi) return this.checked;
+
+          if (Array.isArray(this.inputValue)) {
+            return this.inputValue.includes(this.value);
+          }
+          if (typeof this.inputValue === 'boolean') {
+            return this.inputValue;
+          }
+          return Boolean(this.inputValue !== null);
```
AlexVelezLl: (Nitpick) Since this.inputValue !== null is already a comparison, it returns a boolean value, and we no longer need to wrap it with the Boolean() type

lib/KCheckbox/index.vue
```diff
@@ -178,17 +255,23 @@
...
       toggleCheck(event) {
         if (!this.disabled) {
+          this.lastUserEvent = event || null;
           this.$refs.kCheckboxInput.focus();
-          /**
-           * Emits change event
-           */
-          this.$emit('change', !this.checked, event);
+          this.isChecked = !this.isChecked;
+          this.lastUserEvent = null;
```
AlexVelezLl: At first read, it is not apparent what the `lastUserEvent` is for here, or why it is set to null again. We should add a couple of comments to explain this better.

However, to play safe, I would suggest setting the `lastUserEvent` to null inside a [nextTick](https://vuejs.org/api/general#nexttick) call, just to make sure that this event is being set to null just after the change event has already been emitted, and not before. Another alternative to actually make sure that this is set to null just after the change event has been emitted is to set it to null within the `updateInputValue` method. Since it would mean "okay, I already used it, we don't need this anymore", and this will ensure that this value is removed just after the change event has been emitted.

AlexVelezLl (commented):
Just some final nitpicks! Everything else looks excellent! Thanks @WinnyChang!!

lib/KCheckbox/index.vue
WinnyChang: Updated!

lib/KCheckbox/index.vue
WinnyChang: Updated! (I decided to set `lastUserEvent` to null within the `updateInputValue`.)

AlexVelezLl (approved):
Thanks a lot for another great PR @WinnyChang!! Everything looks great! Code changes make sense, no modifications to the current KCheckbox usage, documentation looks :chef-kiss:, unit and visual tests are passing, and manual QA checks out! And I've also looked at all KCheckbox usages in Kolibri, and everything is still working fine! Thanks a lot!

MisRob:
Thank you @WinnyChang and thank you @AlexVelezLl :)

WinnyChang:
Thank you @MisRob and @AlexVelezLl, for reviewing and providing feedback! This issue was a little challenging for me, but I learned a lot! ☺️

PR #1150, opened 2025-10-30
adds provision for array to be passed in the value prop

AllanOXDi:
## Description
This PR adds an allowance for an array to be passed as KSelect value prop

#### Issue addressed

Addresses https://github.com/learningequality/kolibri/issues/13696

## Changelog

  - **Description:**  Adds an allowance for an array to be passed as KSelect value prop

  - **Products impact:** N/A
  - **Addresses:** https://github.com/learningequality/kolibri/issues/13696

  - **Components:** N/A
  - **Breaking:** N/A
  - **Impacts a11y:** N/A
  - **Guidance:** N/A

## Steps to test

1. Run Kolibri with this PR (`yarn run devserver-with-kds <kds-path> `)
2. Navigate to facility > Users > New User
3. Observe the console

## Testing checklist

## Reviewer guidance

### DIFF
--- a/lib/KSelect/index.vue
+++ b/lib/KSelect/index.vue
@@ -273,12 +273,10 @@
         required: true,
         validator(value) {
           if (Array.isArray(value)) {
-            return value.every(
-              item => item && typeof item === 'object' && 'value' in item
-            );
+            return value.every(item => item && typeof item === 'object' && 'value' in item);
           }
           return value && typeof value === 'object' && 'value' in value;
-        }
+        },
       },
       /**
        * Array of option objects { value, label, disabled }.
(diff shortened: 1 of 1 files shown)


lib/KSelect/index.vue
```diff
@@ -269,11 +269,8 @@
        * Object currently selected
        */
       value: {
-        type: Object,
+        type: [Object, Array],
         required: true,
```
AlexVelezLl, line 273: Could we avoid removing the validator and instead adapt it to the new value type? So we can always validate the value, just as it was before, but now supporting another possible value: an array.

AllanOXDi: Fixed!

AlexVelezLl (approved):
LGTM! Thanks @AllanOXDi!! 👐

PR #1153, opened 2025-11-04
Make KToolbar themeable, fix textColor validator check error in Kolibri, and remove obsolete Keen code

MisRob:
## Background

Kolibri expects `KToolbar` to accept theme color via `textColor`, such as [here](https://github.com/learningequality/kolibri/blob/ad8ebbdc43144f36d3d57d66d0e88f8c094d6853/packages/kolibri/components/pages/AppBarPage/internal/AppBar.vue#L17), but `KToolbar` was only able to work with two hardcoded colors `'black'` and `'white'`.

We recently added validators to `KToolbar` reflecting this restriction that now cause Kolibri console warnings:

Rather than reverting the validators, this PR addresses the issue by resolving the underlying problem of the unmet expectation for `KToolbar` to be themeable.

## Description

Makes `KToolbar` themeable and cleans up unused or not functional code (this was easier than trying to update all intertwined logic related to colors, and is aligned with Keen tech debt removal).

First, I audited our codebases with [this search](https://github.com/search?q=org%3Alearningequality+KToolbar&type=code), then update in such a way that should cause no breaking changes.  Some removed APIs are used from Kolibri, but they are all obsolete code that don't seem to take effect  - see review notes for details. 

Also adds visual tests and has documentation fixes and further improvements.

## Changelog

TBD

## Steps to test

1. Did I remove/change something that shouldn't be?
2. Preview [documentation](https://deploy-preview-1153--kolibri-design-system.netlify.app/ktoolbar)
3. Preview visual tests


lib/KToolbar/ProgressLinear.vue
```diff
@@ -0,0 +1,137 @@
...
+
+        return this.progress;
+      },
+    },
+  };
+
+</script>
+
+
+<style lang="scss">
```
MisRob: Should be `scoped`

lib/KToolbar.vue
```diff
@@ -1,328 +0,0 @@
...
-       * Icon name displayed in the navigation button when using the default icon slot
-       */
-      navIcon: {
-        type: String,
-        default: 'menu',
-      },
-      /**
-       * Whether to hide the navigation icon button completely
-       */
-      removeNavIcon: {
```
MisRob: Removed. Not being used from anywhere.

lib/KToolbar.vue
```diff
@@ -1,328 +0,0 @@
...
-       * Whether the toolbar has elevated shadow styling for visual hierarchy
-       */
-      raised: {
-        type: Boolean,
-        default: true,
-      },
-      /**
-       * Position of the loading progress bar indicator. Options are 'top' or 'bottom'.
-       */
-      progressPosition: {
```
MisRob: Not being used but but preserved - seems potentially useful?

lib/KToolbar.vue
```diff
@@ -1,328 +0,0 @@
...
-        type: String,
-        default: 'bottom',
-        validator(value) {
-          return ['top', 'bottom'].includes(value);
-        },
-      },
-      /**
-       * Whether to show the loading progress indicator
-       */
-      loading: {
```
MisRob: Not being used but but preserved - seems potentially useful?

lib/KToolbar.vue
```diff
@@ -1,328 +0,0 @@
...
-      },
-    },
-
-    methods: {
-      navIconClick() {
-        /**
-         * Emitted when the navigation icon is clicked
-         * @event nav-icon-click
-         */
-        this.$emit('nav-icon-click');
```
MisRob: Removed. Since `navIcon` and the default `UiIconButton` was removed, this will never be emitted. 

**Audit**

- Used in Kolibri from [ImmersiveToolbar](https://github.com/learningequality/kolibri/blob/ad8ebbdc43144f36d3d57d66d0e88f8c094d6853/packages/kolibri/components/pages/ImmersivePage/internal/ImmersiveToolbar.vue#L13) but doesn't take effect (since we don't use `navIcon`, but instead use the `#icon` slot and [emit `nav-icon-click` from `KIconButton`](https://github.com/learningequality/kolibri/blob/ad8ebbdc43144f36d3d57d66d0e88f8c094d6853/packages/kolibri/components/pages/ImmersivePage/internal/ImmersiveToolbar.vue#L38-L54))

lib/KToolbar.vue
```diff
@@ -1,328 +0,0 @@
...
-    components: {
-      UiIconButton,
-      UiProgressLinear,
-    },
-
-    props: {
-      /**
-       * Controls the toolbar appearance and styling. Options are:
-       * - 'default': Standard white background with shadow
-       * - 'colored': Uses theme primary color as background
```
MisRob, line 87: `'colored'` option removed. Background color can be customized via the `backgroundColor` prop if future need arises.

**Audit**

Not used from anywhere.

lib/KToolbar.vue
```diff
@@ -1,328 +0,0 @@
...
-        default: 'default',
-        validator(value) {
-          return ['default', 'colored', 'clear'].includes(value);
-        },
-      },
-      /**
-       * Controls the text color throughout the toolbar. Use 'white' for dark backgrounds
-       * and 'black' for light backgrounds to ensure proper contrast.
-       */
-      textColor: {
```
MisRob, line 101: `'black'` and `'white'` restriction removed. Now accepts all values, including our theme.

lib/KToolbar.vue
```diff
@@ -1,328 +0,0 @@
...
-       * Brand text displayed alongside the title in the left area
-       */
-      brand: {
-        type: String,
-        default: '',
-      },
-      /**
-       * Icon name displayed in the navigation button when using the default icon slot
-       */
-      navIcon: {
```
MisRob, line 125: Removed (utilized obsolete `UiIconButton`).

**Audit**
- Not used from anywhere (we always use the `#icon` slot since we need granular `KIconButton` configuration)

lib/KToolbar.vue
MisRob, line 132: Removed.

**Audit**
- Used in Kolibri from [AppBar](https://github.com/learningequality/kolibri/blob/ad8ebbdc43144f36d3d57d66d0e88f8c094d6853/packages/kolibri/components/pages/AppBarPage/internal/AppBar.vue#L15) but doesn't take effect (whether the navigation icon is displayed or not is configured in [the #icon slot condition](https://github.com/learningequality/kolibri/blob/ad8ebbdc43144f36d3d57d66d0e88f8c094d6853/packages/kolibri/components/pages/AppBarPage/internal/AppBar.vue#L31))

lib/KToolbar.vue
MisRob, line 146: Removed. If we need `KToolbar` to have API for linear loading progress in the future, it'd be better to utilize `KLinearLoader` and expose APIs that are meaningful. 

**Audit**
- Not used from anywhere

lib/KToolbar.vue
MisRob, line 156: Removed. If we need `KToolbar` to have API for linear loading progress in the future, it'd be better to utilize `KLinearLoader` and expose APIs that are meaningful. 

**Audit**
- Not used from anywhere

lib/KToolbar.vue
MisRob, line 188: Removed. Since `navIcon` and the default `UiIconButton` was removed, this will never be emitted. 

**Audit**

- Used in Kolibri from [ImmersiveToolbar](https://github.com/learningequality/kolibri/blob/ad8ebbdc43144f36d3d57d66d0e88f8c094d6853/packages/kolibri/components/pages/ImmersivePage/internal/ImmersiveToolbar.vue#L13) but doesn't take effect (since we don't use `navIcon`, but instead use the `#icon` slot and [emit `nav-icon-click` from `KIconButton`](https://github.com/learningequality/kolibri/blob/ad8ebbdc43144f36d3d57d66d0e88f8c094d6853/packages/kolibri/components/pages/ImmersivePage/internal/ImmersiveToolbar.vue#L38-L54))

MisRob:
Needs some decisions in regard to release planning ([Slack thread](https://learningequality.slack.com/archives/CTG8XDTCL/p1762344795953039))

docs/pages/ktoolbar.vue
```diff
@@ -6,44 +6,50 @@
...
-      <h3>Dark theme toolbar with router navigation</h3>
-      <p>
-        This example shows a dark-themed toolbar using
-        <code>$themeTokens.appBarDark</code> background and
-        <code>$themeTokens.textInverted</code> color, with a back navigation button wrapped in
-        router-link.
-      </p>
+      <h3>Dark theme</h3>
+
+      <p>Uses the dark app bar theme token.</p>
```
AlexVelezLl, line 30: (Non-blocking) In general, it feels weird to call these examples "Light theme" and "Dark theme" because it feels like they are somehow related to a general light/dark app theme, but they are not. They are both used in the same theming but with different purposes as described in the [appbars specs](https://deploy-preview-1153--kolibri-design-system.netlify.app/appbars). It seems the difference resides in "primary bars/Full-screen modal bars".

AlexVelezLl: Just looked at how these different styles are used, and in Kolibri, this concept only affects the `ImmersiveToolbar` components, and for `ImmersiveToolbar`, the differentiator factor is the [isFullscreen prop](https://github.com/learningequality/kolibri/blob/622b1d67ffc253cd4c548fd3ffb382a1e2183c43/packages/kolibri/components/pages/ImmersivePage/internal/ImmersiveToolbar.vue#L11), while on the `ImmersivePage` component, the differentiator factor is the [primary prop](https://github.com/learningequality/kolibri/blob/622b1d67ffc253cd4c548fd3ffb382a1e2183c43/packages/kolibri/components/pages/ImmersivePage/index.vue#L10).

This is definitely out of scope for this, but it may be good (or it may not be worth it, who knows! 😅) considering standardizing these concepts as we use this KToolbar component more in Studio. And this may be more of a conceptual discussion rather than changing anything on the KToolbar component, as right now it is implemented to be super flexible to match any layout.

lib/KToolbar/index.vue
```diff
@@ -0,0 +1,204 @@
...
+  export default {
+    name: 'KToolbar',
+
+    props: {
+      /**
+       * Controls the toolbar appearance and styling. Options:
+       * - `'default'`: `$themeTokens.surface` background with shadow
+       * - `'clear'`: Transparent background with shadow
+       */
+      type: {
```
AlexVelezLl, line 68: This `type` prop is a bit "tricky", because in essence the "primary" purpose of this prop could be achieved by just setting a `backgroundColor="transparent"`, and the result will be the same. e.g.:

---
image

```Vue
    <KToolbar
      title="Toolbar title"
      backgroundColor="transparent"
    />
```


---

image

```Vue
    <KToolbar
      type="clear"
      title="Toolbar title"
    />
```
---

But the only difference occurs when raised=false: if type="default", a bottom border is shown; if type="clear", it is not. Coincidentally (or not), the latter configuration is the one used in the main `AppBar` component in Kolibri [here](https://github.com/learningequality/kolibri/blob/622b1d67ffc253cd4c548fd3ffb382a1e2183c43/packages/kolibri/components/pages/AppBarPage/internal/AppBar.vue#L14), so better be careful with that, we could potentially just remove the `type` and manage both the bottom border and the box shadow just with the `raised` prop, but again, not sure if this will be worth the effort.

lib/KToolbar/index.vue
```diff
@@ -0,0 +1,204 @@
...
+        type: String,
+        default: null,
+      },
+      /**
+       * Background color
+       */
+      backgroundColor: {
+        type: String,
+        default: null,
+      },
```
AlexVelezLl, line 88: Just noting that we can have the same behavior if we use the `style` prop to set the color and background color. In some places in Kolibri, it seems this is the preferred way to do this.

Just reflecting on the tradeoff and whether we want to prefer these more general "props" like `style` to override these styles, or prefer the specific props so that we can have them better documented.

AlexVelezLl (commented):
Overall, this looks really nice; it feels like a lot of tech debt clean-up. In general, just left more reflective comments on how the design system concepts relate to how we have actually used this component in Kolibri, and if the efforts to standardize them would help in anything. Perhaps we will have a better idea when migrating the VToolbar component in Studio.

docs/pages/ktoolbar.vue
MisRob: Yes, I think it'd be meaningful for documentation to reflect how we think about toolbars. Alternatively, we may just simplify it and not use 'theme' language at all.

lib/KToolbar/index.vue
MisRob: Yes I agree that it'd be much clearer interface, and it'd be a change that would go well with other changes introduce here. I only didn't touch `type` in these areas to not cause a breaking change.

lib/KToolbar/index.vue
MisRob: I am not sure why I added it as a prop - it's possible that I didn't reflect on it, or there might be some dependencies to other props in connection with the limited scope of changes.

I agree that whenever we can use native

```
:style= { backgroundColor: ..., color: ... }
```

that'd be ideal.

So yes in long-term, I would prefer we drop `backgroundColor` and possibly `textColor` too.

MisRob:
Thank you @AlexVelezLl, those are good notes.

Based on what we discussed around releases, I will close this in favor of
- https://github.com/learningequality/kolibri-design-system/issues/1183
- https://github.com/learningequality/kolibri-design-system/pull/1182

PR #1191, opened 2026-01-13
Update lodash import to be selective.

rtibbles:
## Description
Fixes lodash import to import specific function rather than entire library (1-2KB vs ~500KB)

## Changelog

  - **Description:** Updates lodash import to use specific module path
  - **Products impact:** bugfix
  - **Addresses:** Bloated asset size due to import of entire lodash library
  - **Components:** KTable
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** N/A

## Steps to test

1. Ensure KTable sorting is still the same


AlexVelezLl (approved):
Thanks a lot @rtibbles!

Issue #1156, opened 2025-11-06
[KDropdownMenu] Click opens both context menus at once


MisRob:
<img height="20px" src="https://i.imgur.com/0ZZG9qx.jpeg">

🙂 Looking for an issue? Welcome! This issue is open for contribution. If this is the first time you’re requesting an issue, please:

- **Read <a href="https://learningequality.org/contributing-to-our-open-code-base/" target="_blank">Contributing guidelines</a>** carefully. **Pay extra attention to [Using generative AI](https://learningequality.org/contributing-to-our-open-code-base/#using-generative-ai)**. **Pull requests and comments that don’t follow the guidelines won’t be answered.**
- **Confirm that you’ve read the guidelines** in your comment.

<img height="20px" src="https://i.imgur.com/0ZZG9qx.jpeg">

## Observed behavior

In the context menu example on `KDropdownMenu` documentation, both context menus open at once.

<img width="400" height="auto" alt="Image" src="https://github.com/user-attachments/assets/d8995729-2768-4c32-a1dc-e9384f428b94" />

## Expected behavior

Only one menu should be open at a time.

## User-facing consequences

I don't know whether this is KDS website specific or if it may have impact on Kolibri apps.

## Steps to reproduce

- Go to https://design-system.learningequality.org/kdropdownmenu
- In the _Context menu_  section, righ-click on a paragraph

Abhishek-Punhani:
Hi @MisRob, I was looking at this issue and found that since the opening and closing of tippies are handled manually for the context menu, we can use `hideAll` from tippy.js, which will close all open tippies from other instances upon triggering `onShow`. This solves the issue locally, and I don't think it will affect other parts of the codebase. More specifically, we can use `hideAll` in `isContextMenuActive` to hide all tippies from other instances.
Let me know what you think.

MisRob:
Hi @Abhishek-Punhani. Thanks, that's good insights. I think we could consider such an approach - one area to think about would be how to achieve it while not breaking cases when there are more tippies at once intentionally. I believe that could happen - we don't use tippy only for context menus, but for other UI elements as well. And I think that generally it'd be more robust to not make any assumptions about surrounding context. Do you know what was the exact reason for the first context menu not being closed in this particular use-cases?

Abhishek-Punhani:
[...]

PR #1192, opened 2026-01-14
Bug : Click opens both context menus at once

Abhishek-Punhani:
## Description

- Fixes the opening of both context menus on a single click
- Fixes the UID reference.
#### Issue addressed

Addresses #1156 

### Before/after screenshots

https://github.com/user-attachments/assets/e85b7d67-6da7-4ff5-a2e4-3d1f7135c913

## Changelog

  - **Description:**  Fixes the UID reference and using global counter for uid
  - **Products impact:** KDS (KDropdownMenu)
  - **Addresses:**  #1156 
  - **Breaking:**  no
  - **Impacts a11y:** no
  - **Guidance:** 

## Steps to test

1. Go to `http://localhost:4000/kdropdownmenu`
2. In the Context menu section, right-click on a paragraph

## Testing checklist

## Reviewer guidance

### DIFF
--- a/lib/composables/_useKContextMenu.js
+++ b/lib/composables/_useKContextMenu.js
@@ -7,7 +7,7 @@ export default function useKContextMenu() {
   const clientY = ref(0);
 
   const instance = getCurrentInstance();
-  const id = instance.uid;
+  const id = instance.proxy._uid;
 
   const isActive = computed(() => activeMenu.value === id);
 
--- a/lib/composables/_useKContextMenu.js
+++ b/lib/composables/_useKContextMenu.js
@@ -1,13 +1,14 @@
 import { ref, computed, onMounted, onBeforeUnmount, getCurrentInstance } from 'vue';
 
 const activeMenu = ref(null);
+let uid = 0;
 
 export default function useKContextMenu() {
   const clientX = ref(0);
   const clientY = ref(0);
 
   const instance = getCurrentInstance();
-  const id = instance.proxy._uid;
+  const id = uid++;
 
   const isActive = computed(() => activeMenu.value === id);
 


AlexVelezLl (changes requested):
Thanks @Abhishek-Punhani! This implementation is correct and fixes the issue! However, it seems this `_uid` property is an internal Vue private property, and we shouldn't be relying on it since any changes to this property won't be externally announced (probably it was the same with the previous `instance.uid`, and it got unusable after a Vue update). So, since the goal of this property is to identify each component uniquely, we can instead use a global counter to uniquely identify each mounted component that uses the `_useKContextMenu`, similar to what we do for [KDateInput](https://github.com/rtibbles/kolibri-design-system/blob/e7fe99e90d3597188c00ce61d482a878129ca3e6/lib/KDateRange/KDateInput.vue#L41)

Abhishek-Punhani:
> Thanks @Abhishek-Punhani! This implementation is correct and fixes the issue! However, it seems this `_uid` property is an internal Vue private property, and we shouldn't be relying on it since any changes to this property won't be externally announced (probably it was the same with the previous `instance.uid`, and it got unusable after a Vue update). So, since the goal of this property is to identify each component uniquely, we can instead use a global counter to uniquely identify each mounted component that uses the `_useKContextMenu`, similar to what we do for [KDateInput](https://github.com/rtibbles/kolibri-design-system/blob/e7fe99e90d3597188c00ce61d482a878129ca3e6/lib/KDateRange/KDateInput.vue#L41)

@AlexVelezLl , I agree that vue internals can be unreliable and can break logics with version updates. I think we can use uuid each time generate a random uuid and link it to the tippie?

AlexVelezLl:
Don't recall the full details, but we have considered using UUIDs for these cases in other places in the past, but we ended up moving toward this more straightforward workaround!

AlexVelezLl (approved):
Thanks a lot @Abhishek-Punhani!

PR #1194, opened 2026-01-16
Add thumbnail slot to KCard

### LINKED ISSUE #1162: [KCard] Make thumbnail area ratio configurable

## Background

Due to several factors such as the thumbnail area taking 30% of the card's width, its square format, and surplus space on _My Channels_ and other Studio pages, `KCard` is rendered abnormally large. Furthermore, because Studio encourages and crops/adds transparent space to thumbnail images resulting in a 16:9 ratio, thumbnail images rarely fit into the square format nicely. Finally, due to the transparent space being added, the square format results in gray letterbox-like areas appearing on all sides of some images.

## Summary

Update `KCard` so that it will allow easy configuration of its thumbnail area ratio for use cases when the default behavior is not suitable. This will allow us to achieve:

## Proposal

Add a new slot `#thumbnail`, through which `KImg` can be inserted (`KImg` already supports ratio configuration).

```html
<KCard ...>
  <template #thumbnail>
    <KImg
      src="..."
      aspectRatio="16:9"
     isDecorative
     />
  </template>
</KCard>
```

The presence of the `#thumbnail` would override `thumbnailSrc` and `thumbnailScaleType`. Documentation should be updated to mention the need for `isDecorative`.

This way, we will not only be able to configure the ratio, but also anything else, should the need arise.

MisRob:
## Description

In support of migrating Studio channel cards to `KCard`, we need to achieve 16:9 thumbnail aspect ratio. This PR

- Adds the `thumbnail` slot to `KCard`
- Updates documentation and adds an example of how to achieve 16:9 thumbnail area
- Adds visual tests for the slot in combination with various card layout configurations

#### Issue addressed

Resolves https://github.com/learningequality/kolibri-design-system/issues/1162

## Changelog

  - **Description:** Add the `thumbnail` slot to `KCard`
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/1162
  - **Components:** `KCard`
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -

## Steps to test

- Preview the [new documentation section](https://deploy-preview-1194--kolibri-design-system.netlify.app/kcard#thumbnail-slot)
- Preview visual tests

### DIFF
--- a/docs/pages/kcard.vue
+++ b/docs/pages/kcard.vue
@@ -685,6 +685,48 @@
         />.
       </p>
 
+      <h4>
+        Thumbnail slot
+        <DocsAnchorTarget anchor="#thumbnail-slot" />
+      </h4>
+
+      <p>
+        When the thumbnail configuration options are insufficient, the <code>thumbnail</code> slot
+        can be utilized. It is typically used to gain full access to
+        <code>KImg</code> configurations, such as aspect ratio.
+      </p>
+
+      <p>
+        If providing <code>KImg</code>, set <code>isDecorative</code> to <code>true</code> on
+        <code>KImg</code> for correct accessibility.
+      </p>
+
+      <DocsExample
+        loadExample="cards/ThumbnailSlotSmall.vue"
+        exampleId="kcard-thumbnail-slot-small"
+        hideScript
+        block
+      >
+        <template #html>
+          <!-- eslint-disable -->
+          <DocsShowCode language="html">
+            <KCardGrid ...>
+              <KCard ...>
+                <template #thumbnail>
+                  <KImg
+                    :src="..."
+                    :style="{ width: '100%', height: '100%' }"
+                    aspectRatio="16:9"
+                    isDecorative
+                  />
+                </template>
+              </KCard>
+            </KCardGrid>
+          </DocsShowCode>
+          <!-- eslint-enable -->
+        </template>
+      </DocsExample>
+
       <h3>
         Interactive elements
         <DocsAnchorTarget anchor="#interactive-elements" />
--- a/examples/cards/Card.vue
+++ b/examples/cards/Card.vue
@@ -12,6 +12,12 @@
     :title="cardTitle"
   >
     <template
+      v-if="$slots.thumbnail"
+      #thumbnail
+    >
+      <slot name="thumbnail"></slot>
+    </template>
+    <template
       v-if="$slots.thumbnailPlaceholder"
       #thumbnailPlaceholder
     >
--- a/examples/cards/ThumbnailSlotLargeHorizontal.vue
+++ b/examples/cards/ThumbnailSlotLargeHorizontal.vue
@@ -1,0 +1,33 @@
+<template>
+
+  <KCardGrid layout="1-1-1">
+    <Card
+      :headingLevel="2"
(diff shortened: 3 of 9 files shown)


docs/pages/kcard.vue
```diff
@@ -685,6 +685,48 @@
...
 
+      <h4>
+        Thumbnail slot
+        <DocsAnchorTarget anchor="#thumbnail-slot" />
+      </h4>
+
+      <p>
+        When the thumbnail configuration options are insufficient, the <code>thumbnail</code> slot
+        can be utilized. It is typically used to gain full access to
+        <code>KImg</code> configurations, such as aspect ratio.
```
akolson, line 696: Is the slot only tied to the usage of `KImg` configurations as suggested by the documentation? What would happen if something else is used as the slot?

docs/pages/kcard.vue
```diff
@@ -685,6 +685,48 @@
...
+      </p>
+
+      <p>
+        If providing <code>KImg</code>, set <code>isDecorative</code> to <code>true</code> on
+        <code>KImg</code> for correct accessibility.
+      </p>
```
akolson, line 702: Not sure, maybe documentation more generic here. It seems to bias the slot's usage to `KImg`?

akolson (commented):
Code changes look correct to me. Thanks @MisRob ! I have left a few (non-blocking) thoughts on the documentation of the slot being more specific to `KImg` -- Is this intended? Does it mean we can't really pass anything else?

docs/pages/kcard.vue
MisRob: The `thumbnail` slot is meant to provide content to the `thumbnail` area. That would typically be an image. But as with any other slot, technically nothing prevents us from providing whatever we want.

At the same time, currently in our codebase we only have one use-case for this, and that is with `KImg` - that's why I use it as an example and provide concrete guidance. To make the documentation to the point, rather than abstract.

I've tried to capture this by using the phrase _It is typically used_ which was the best way I figured out so far to focus on our use-case, while not excluding other possible future options. If you have any suggestions how to re-phrase, I'm happy to hear.

MisRob:
Thank you @akolson - replied. Let me know if you have further ideas.

docs/pages/kcard.vue
MisRob: By starting the sentence with _If providing_ I wanted suggest a bit that it's not strictly the only option (even though as of now it is the only one we need)

docs/pages/kcard.vue
akolson: Ahh makes sense. Thanks for the clarification! In that case, no immediate changes are required. We can always make updates to the documentation when new use-cases come up.

docs/pages/kcard.vue
akolson: https://github.com/learningequality/kolibri-design-system/pull/1194/#discussion_r2731594838

akolson:
Hi @MisRob! No further comments from me! We should be good for the merge! Thanks

akolson (approved):
LGTM!

docs/pages/kcard.vue
MisRob: Alright

PR #1208, opened 2026-02-19
Fix KSelect absent value check

AlexVelezLl:
## Description

- Historically, the way we've had to represent that KSelect doesn't have any option selected is by passing an empty object as its `value`, right now the prop validator requires this `value` prop to have a `value` property defined, but it is not necessary when the KSelect doesn't have any value.

## Changelog

  - **Description:** Fixes KSelect `value` prop validator forcing a `value` key to be defined.
  - **Addresses:** -.
  - **Components:** KSelect.
  - **Breaking:** no
  - **Impacts a11y:**  no
  - **Guidance:** .

## Steps to test

Create a `KSelect` with an empty value (i.e., `{}`) and check that no error appears in the console.

### DIFF
--- a/lib/KSelect/index.vue
+++ b/lib/KSelect/index.vue
@@ -275,7 +275,7 @@
           if (Array.isArray(value)) {
             return value.every(item => item && typeof item === 'object' && 'value' in item);
           }
-          return value && typeof value === 'object' && 'value' in value;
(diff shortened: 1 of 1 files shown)


MisRob (approved):
Thank you @AlexVelezLl

Issue #1200, opened 2026-02-02
[KDropdownMenu] Enable options customization


MisRob:
## Overview

**Complexity:** Low

### Context

In several Studio use-cases, we need to customize content of `KDropdownMenu` items beyond what is possible via `options` array configuration. Typically icon placement, color, text size, ... Examples:

- [Main navigation dropdown](https://github.com/learningequality/studio/pull/5642#discussion_r2747273035)
- [Channel dropdown](https://github.com/learningequality/studio/pull/5668#issuecomment-3822969336)

`UiMenu` actually allows for this via this scoped `option` slot

https://github.com/learningequality/kolibri-design-system/blob/0318c091d146ea1c71b254cc98fdccdbf8b5d171/lib/keen/UiMenu.vue#L32

, we just yet didn't made the slot available in `KDropdownMenu` for public use.

### The Change

Expose `UiMenu`'s `option` slot in `KDropdownMenu` for public use like:

```
<KDropdownMenu :options="options" ... >
  <template #option="{ option }">
     <KIcon ... /> {{ option.label }}
  </template>
</KDropdownMenu>
```

- Update documentation with an example demonstrating the slot use
- Add a visual test

### Guidance

- [KDropdownMenu documentation page](https://design-system.learningequality.org/kdropdownmenu)
- [Getting started](https://github.com/learningequality/kolibri-design-system/blob/develop/dev_docs/01_getting_started.md)
- [How to update library](https://github.com/learningequality/kolibri-design-system/blob/develop/dev_docs/03_how_to_update_library.md)
- [How to update the documentation website](https://github.com/learningequality/kolibri-design-system/blob/develop/dev_docs/04_how_to_update_docs.md)
- [Visual testing guide](https://github.com/learningequality/kolibri-design-system/blob/develop/dev_docs/07_visual_testing_guide.md)
- [Writing guidelines](https://design-system.learningequality.org/writing)
- Refer to other component documentation pages, e.g. [KCard documentation](https://design-system.learningequality.org/kcard) ([source code](https://github.com/learningequality/kolibri-design-system/blob/develop/docs/pages/kcard.vue))

### Acceptance criteria

- [ ] The specification above is followed
- [ ] There are no regressions
- [ ] There is a visual test
- [ ] The documentation is updated
  - Follows writing guidelines
  - Is concise and consistent with other documentation pages

### References

Supports https://github.com/learningequality/studio/issues/5060

vtushar06:
[...]

PR #1210, opened 2026-02-22
Add custom option slot to KDropdownMenu and update documentation

vtushar06:
## Description

Exposes UiMenu's existing option scoped slot through KDropdownMenu, allowing consumers to customize the rendering of each menu item.

#### Issue addressed

Addresses #1200

### Before/after screenshots

Screenshot 2026-02-23 at 4 18 38 AM

## Changelog

  - **Description:**  New option scoped slot on KDropdownMenu. Use <template #option="{ option }"> to customize how each menu item renders (icon placement, color, text size, etc.).
  - **Addresses:** learningequality/kolibri-design-system#1200.
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:**Use #option when the options array config isn't sufficient. Existing behavior is unchanged when the slot is not used.

## Steps to test

1. yarn dev → go to KDropdownMenu docs page
2. Verify the new "Custom option slot" section appears with a working live example
3. Verify the API Slots table lists both header and option
4. Verify all existing examples still work (no regressions)
5. yarn test and yarn lint — both pass

## Testing checklist

## Reviewer guidance

### DIFF
--- a/docs/pages/kdropdownmenu.vue
+++ b/docs/pages/kdropdownmenu.vue
@@ -263,13 +263,13 @@
           <!-- prettier-ignore -->
           <DocsShowCode language="javascript">
               export default {
-                computed: {
-                  options() {
-                    return [
+                data() {
+                  return {
+                    options: [
                       { label: 'Edit', value: 'edit', icon: 'edit', color: null },
                       { label: 'Delete', value: 'delete', icon: 'delete', color: this.$themeTokens.error },
-                    ];
-                  },
+                    ],
+                  };
                 },
               };
             </DocsShowCode>
--- a/docs/pages/kdropdownmenu.vue
+++ b/docs/pages/kdropdownmenu.vue
@@ -244,12 +244,14 @@
                   ...
                 >
                   <template #option="{ option }">
-                    <KLabeledIcon
-                      :icon="option.icon"
-                      :label="option.label"
-                      :color="option.color"
-                      :style="{ color: option.color }"
-                    />
+                    <div style="display: flex; align-items: center; padding: 0 16px">
+                      <KLabeledIcon
+                        :icon="option.icon"
+                        :label="option.label"
+                        :color="option.color"
+                        :style="{ color: option.color }"
+                      />
+                    </div>
                   </template>
                 </KDropdownMenu>
               </template>
--- a/docs/pages/kdropdownmenu.vue
+++ b/docs/pages/kdropdownmenu.vue
@@ -228,6 +228,7 @@
         loadExample="KDropdownMenu/CustomOption.vue"
         exampleId="custom-option-slot"
         block
+        :hideOnClick="true"
         :openOnMount="false"
       >
         <template #html>
@@ -243,13 +244,12 @@
                   ...
                 >
                   <template #option="{ option }">
-                    <div style="display: flex; align-items: center; padding: 0 16px">
-                      <KIcon
-                        :icon="option.icon"
-                        :style="{ color: option.color, marginInlineEnd: '8px' }"
-                      />
-                      <span :style="{ color: option.color }">{{ option.label }}</span>
-                    </div>
+                    <KLabeledIcon
+                      :icon="option.icon"
+                      :label="option.label"
+                      :color="option.color"
+                      :style="{ color: option.color }"
+                    />
                   </template>
                 </KDropdownMenu>
               </template>
--- a/docs/pages/kdropdownmenu.vue
+++ b/docs/pages/kdropdownmenu.vue
@@ -25,6 +25,7 @@
           { text: 'Icon button with dropdown', href: '#icon-button-with-dropdown' },
           { text: 'Options with icons', href: '#options-with-icons' },
           { text: 'Context menu', href: '#context-menu' },
+          { text: 'Custom option slot', href: '#custom-option-slot' },
         ]"
       />
 
@@ -211,6 +212,23 @@
         loadExample="KDropdownMenu/ContextMenu.vue"
         exampleId="context-menu"
         block
+      />
+
+      <h3>
+        Custom option slot
+        <DocsAnchorTarget anchor="#custom-option-slot" />
+      </h3>
+
+      <p>
+        Use the <code>#option</code> scoped slot to customize the rendering of each menu item. The
+        slot receives <code>{ option }</code> — the option object — as a slot prop.
+      </p>
+
+      <DocsExample
+        loadExample="KDropdownMenu/CustomOption.vue"
+        exampleId="custom-option-slot"
+        block
+        :openOnMount="false"
       />
     </DocsPageSection>
   </DocsPageTemplate>
--- a/docs/pages/kdropdownmenu.vue
+++ b/docs/pages/kdropdownmenu.vue
@@ -220,8 +220,8 @@
       </h3>
 
       <p>
-        Use the <code>#option</code> scoped slot to customize the rendering of each menu item. The
-        slot receives <code>{ option }</code> — the option object — as a slot prop.
+        Use the <code>#option</code> scoped slot to customize each menu item. The slot provides the
+        option object via the <code>option</code> prop.
       </p>
 
       <DocsExample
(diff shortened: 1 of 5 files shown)


AlexVelezLl:
Thanks @vtushar06! Code changes look good, and visual tests show no regression. Just want to check-in something before merging:

@MisRob, There is a rather small problem with just exposing this option as is, the thing is that the node we pass through the slot cannot infer the 100% of the available height, for example, even if we add `height: 100%`, the slot does not take the whole height:

image

To get the full height, we will also need to set a min-height: 40px (or 2.5 rem) to the node that passes through the slot, and this will work fine unless we change the min-height of the UiMenuOption, in which case, we would have an unintended regression that would be difficult to detect. It seems a way to solve this could be by modifying [this](https://github.com/vtushar06/kolibri-design-system/blob/f47f06b9092ee2852d5635d934ce1e5d2193b61c/lib/keen/UiMenuOption.vue#L5) node to have a class like, so that any node we pass as slot gets the whole height without harcoding the min-height value:

```SCSS
  .ui-menu-option {
    display: flex;
    flex-direction: column;
    & > * {
      flex-grow: 1;
    }
  }
```

We might want to do this, or any other workaround in the scope of this PR, or if you consider this min height is something that might not change, then it is okay to leave it as it is right now.

vtushar06:
Thanks for catching this @AlexVelezLl, You're right our current example hardcodes `height: 40px` which is fragile. Your proposed CSS fix is the correct approach. I think we should update `.ui-menu-option` in `UiMenuOption.vue` to use `display: flex; flex-direction: column` with `& > * { flex-grow: 1 }`, and remove the hardcoded height from the example. This makes slot content fill the full height correctly without depending on a specific pixel value.

MisRob:
Thanks, that's good consideration. I'm fine with that solution @AlexVelezLl :) Within the Keen-related limitations we work with, it feels like the most robust solution we can do now.

docs/pages/kdropdownmenu.vue
```diff
@@ -212,6 +213,23 @@
...
         block
       />
+
+      <h3>
+        Custom option slot
+        <DocsAnchorTarget anchor="#custom-option-slot" />
+      </h3>
+
+      <p>
+        Use the <code>#option</code> scoped slot to customize the rendering of each menu item. The
```
MisRob: > Use the <code>#option</code> scoped slot to customize each menu item. The slot provides the option object via the `option` prop.

Nitpick. Feels a tiny bit better readable?

MisRob:
As for documentation example code, I recommend we keep it as concise and straightforward as possible, and we show only parts relevant to the new slot. The documentation component provides API to customize what template / script /css we show to a reader, e.g. for template (note the use of `...`) :

```
<KButton
  text="Options"
  hasDropdown
>
  <template #menu>
    <KDropdownMenu
      :options="options"
       ...
    >
      <template #option="{ option }">
        <div style="display: flex; align-items: center; height: 40px; padding: 0 16px">
          <KIcon
            :icon="option.icon"
            :style="{ color: option.color, marginInlineEnd: '8px' }"
          />
          <span :style="{ color: option.color }">{{ option.label }}</span>
        </div>
      </template>
    </KDropdownMenu>
  </template>
</KButton>
```

Script can be adjusted too.

Not that it'd be a big deal at this single place ~ it's just that if we don't consider it for new examples, documentation tends to grow a lot and in my opinion is then harder to navigate & understand. Readers can get full code by clicking GitHub icon if needed.

MisRob:
For the script we should just show:

```
export default {
  data() {
    return {
      options: [
        { label: 'Edit', value: 'edit', icon: 'edit', color: null },
        { label: 'Delete', value: 'delete', icon: 'delete', color: '#e74c3c' },
      ],
    };
  },
};
```

Other 80% of the current content would be irrelevant.

@vtushar06 to be clear, you don't need to change how the example component is implemented. This is only about what we show in the docs. You can search for examples how to achieve it or if it's not clear, @AlexVelezLl or I will help :) Thanks!

MisRob:
And finally, please don't use hardcoded `color: '#e74c3c`. Even in code examples, it's important to stick to KDS colors too ~ this is where new KDS users learn.

vtushar06:
Hi @MisRob, So I've made all the fixes based on the your feedback. The height thing now works way better with flexbox instead of hardcoding pixels, so it won't break if things change. I have switched to KDS color tokens instead of those random hex codes, which keeps everything looking consistent. And the docs example now shows the real way you'd actually use it.
Please let me know if  any further improvement needed.

docs/pages/kdropdownmenu.vue
vtushar06: updated..

docs/pages/kdropdownmenu.vue
```diff
@@ -212,6 +213,69 @@
...
+                  :options="options"
+                  ...
+                >
+                  <template #option="{ option }">
+                    <div style="display: flex; align-items: center; padding: 0 16px">
+                      <KIcon
+                        :icon="option.icon"
+                        :style="{ color: option.color, marginInlineEnd: '8px' }"
+                      />
+                      <span :style="{ color: option.color }">{{ option.label }}</span>
```
AlexVelezLl: Hey @MisRob, it seems that these `{{ options.label }}` expressions are being evaluated even though they are inside the `DocsShowCode` component, is this a known issue? For now, it seems that we can just let it evaluate a simple string, something like:

```Vue
<span :style="{ color: option.color }" > {{  "\{\{ option.label \}\}" }}</span>
```

And that would solve the issue.

However, in this case, I think it'd be better to use `KLabeledIcon` instead of aligning the icon and text ourselves. Something like
```
            <KLabeledIcon
              :icon="option.icon"
              :label="option.label"
              :color="option.color"
              :style="{ color: option.color }"
            />
```

I think this could be the recommended way to manage this. What do you think @MisRob?

docs/pages/kdropdownmenu.vue
```diff
@@ -212,6 +213,69 @@
...
+      <p>
+        Use the <code>#option</code> scoped slot to customize each menu item. The slot provides the
+        option object via the <code>option</code> prop.
+      </p>
+
+      <DocsExample
+        loadExample="KDropdownMenu/CustomOption.vue"
+        exampleId="custom-option-slot"
+        block
+        :openOnMount="false"
```
AlexVelezLl, line 232: Could you please also pass a `:hideOnClick="true"` prop here? Thanks!

AlexVelezLl (commented):
Thanks a lot @vtushar06! Code changes look good, visual tests show no visual regression, just a couple of observations before merging this PR 👐.

docs/pages/kdropdownmenu.vue
MisRob: Yes I agree with you @AlexVelezLl, thank you :)

docs/pages/kdropdownmenu.vue
vtushar06: done @AlexVelezLl .

docs/pages/kdropdownmenu.vue
vtushar06: okay proceeding with the approach.

vtushar06:
Hey @AlexVelezLl and @MisRob, Made all the latest fixes, swapped out the manual div + KIcon + span pattern for KLabeledIcon in both the example component and the docs snippet, which also takes care of the template expression evaluation issue. Also added :hideOnClick="true" to the DocsExample. Let me know if there's anything else to tweak!

AlexVelezLl:
Thanks @vtusharc06! Apologies for the late reply. Just one last thing. Could you add some padding to the example options? These examples are expected to serve as the point of reference for how people should use them, so it's important to keep the examples clean. Thanks!

image

vtushar06:
Hey @AlexVelezLl, I have added padding and also centered the content which makes examples clean,
have a look and please let me know if any more changes required.

AlexVelezLl (approved):
Thanks a lot @vtushar06! LGTM!

PR #1217, opened 2026-03-09
Documentation updates

MisRob:
## Description

- Align snackbar documentation with the latest guidance from designers
- Improves guidance for recently added `KCard` `thumbnail` slot
- Improves `useKLiveRegion` guidance

#### Issue addressed

- Updates obsolete specifications
- Improves and adds guidance in areas that I needed to explain few times

## Changelog

  - **Description:** Improve snackbar, `KCard`, and `useKLiveRegion` documentation pages
  - **Products impact:** Documentation
  - **Addresses:** -
  - **Components:** Snackbar, `KCard`, `useKLiveRegion`
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -

## Steps to test

Preview updated areas on the following pages:
- https://deploy-preview-1217--kolibri-design-system.netlify.app/snackbars
- https://deploy-preview-1217--kolibri-design-system.netlify.app/kcard#thumbnail-slot
- https://deploy-preview-1217--kolibri-design-system.netlify.app/usekliveregion

### DIFF
--- a/docs/pages/kcard.vue
+++ b/docs/pages/kcard.vue
@@ -696,10 +696,18 @@
         <code>KImg</code> configurations, such as aspect ratio.
       </p>
 
-      <p>
-        If providing <code>KImg</code>, set <code>isDecorative</code> to <code>true</code> on
-        <code>KImg</code> for correct accessibility.
-      </p>
+      <p>When using with <code>KImg</code>:</p>
+
+      <ul>
+        <li>Set <code>isDecorative</code> on <code>KImg</code> for correct accessibility</li>
+        <li>
+          Utilize <code>KImg</code>'s <code>placeholder</code> slot for placeholder content (<code>thumbnailPlaceholder</code>
+          slot doesn't take effect when using the <code>thumbnail</code> slot)
+        </li>
+        <li>
+          Depending on the card layout, you may need to set width and height on <code>KImg</code>
+        </li>
+      </ul>
 
       <DocsExample
         loadExample="cards/ThumbnailSlotSmall.vue"
@@ -718,7 +726,11 @@
                     :style="{ width: '100%', height: '100%' }"
                     aspectRatio="16:9"
                     isDecorative
-                  />
+                  >
+                    <template #placeholder>
+                      <KIcon ... />
+                    </template>
+                  </KImg>
                 </template>
               </KCard>
             </KCardGrid>
--- a/docs/pages/snackbars.vue
+++ b/docs/pages/snackbars.vue
@@ -57,10 +57,15 @@
         <li>Changes saved</li>
         <li>Resource added</li>
       </ul>
+
+      <p>For longer messages, consider a different pattern.</p>
+
       <p>
         Avoid including user-inputted text such as the name of a resource or lesson because we have
         less control over the length of this text.
       </p>
+
+      <p>Snackbar text should have a maximum of 2 lines (truncate with ellipsis).</p>
     </DocsPageSection>
 
     <DocsPageSection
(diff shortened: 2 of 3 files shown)


docs/pages/kcard.vue
```diff
@@ -696,10 +696,18 @@
...
-      <p>
-        If providing <code>KImg</code>, set <code>isDecorative</code> to <code>true</code> on
-        <code>KImg</code> for correct accessibility.
-      </p>
+      <p>When using with <code>KImg</code>:</p>
+
+      <ul>
+        <li>Set <code>isDecorative</code> on <code>KImg</code> for correct accessibility</li>
+        <li>
+          Utilize <code>KImg</code>'s <code>placeholder</code> slot for placeholder content (<code>thumbnailPlaceholder</code>
```
AlexVelezLl, line 704: I got a (little) bit confused because of the [thumbnail placeholder](https://deploy-preview-1217--kolibri-design-system.netlify.app/kcard/#thumbnail-placeholder) slot, thumbnail slot, and this KImg's placeholder slot. But I think I finally got it, the `thumbnail-placeholder` is only applicable for `thubmnailSrc`, right?

I think it is this paragraph that isn't entirely clear to me, that this is talking specifically about the thumbnailSrc and not about the thumbnail slot.

> When KCard is set to display the thumbnail, the thumbnail area acts as a placeholder if the image is missing, fails to load, or is still loading. In such cases, a light gray background is shown in place of the image.

Also, something that isn't clear to me by just reading this, if we set KCard to display the thumbnail using the thumbnail slot, we also get this gray background when the image is missing, fails to load or still loading? (im not completely profficient with KImg yet) 😅

docs/pages/snackbars.vue
```diff
@@ -57,26 +57,52 @@
...
 
     <DocsPageSection
       title="Specifications"
       anchor="#specifications"
     >
       <ul>
-        <li>Background color: $k-grey-900</li>
-        <li>Text: 14px white and bolded</li>
-        <li>Height: 48px</li>
+        <li>Timeout: 5 seconds</li>
```
AlexVelezLl, line 76: In the current KSnackbar PR, we are using 4000 ms as the default. Should we already update this in the scope of that PR, right? (i.e., are planning to do this in a follow-up?)

docs/pages/usekliveregion.vue
```diff
@@ -6,65 +6,78 @@
...
+          />.
+        </li>
+        <li>
+          <strong>Do not manually insert any live regions</strong> (e.g. via <code>aria-live</code>,
+          <code>role="alert"</code>, ...). Multiple live regions or live regions nested deep in the
+          DOM structure are problematic. Always use <code>useKLiveRegion</code>.
+        </li>
+        <li>
+          <strong>Use assertive messages only for critical information that requires immediate
+            attention.</strong>
```
AlexVelezLl: For me, as a developer, it is still a bit abstract what a "critical information that requires immediate attention" is; it'd be great if we had some examples here.

AlexVelezLl (approved):
Thanks @MisRob! Updates make sense to me, just a couple of comments/questions, but nothing blocking! :)

docs/pages/kcard.vue
MisRob: > When KCard is set to display the thumbnail

This part refers to card layouts that have thumbnail area. Does that help, or is the confusing part coming from somewhere else?

> If we set KCard to display the thumbnail using the thumbnail slot, we also get this gray background when the image is missing, fails to load or still loading?

Yes, this is taken care of by `KImg` (if something else than `KImg` would be used, then it wouldn't work ~ I was just thinking whether to document all possibilities, but I worry it would just become very complex, and as of now it'd be hypothetical)

Let me know if that helps or if you had any ideas on how to rephrase.

docs/pages/snackbars.vue
MisRob: Yes, thank you

docs/pages/usekliveregion.vue
MisRob: I changed it to use the same guidance as on MDN and also linked MDN docs for more detail. I can't think of any specific example, it sounds to me it needs to be considered carefully on case by case basis by asking whether the user absolutely needs to know it right now. Typically probably a very important error, but even that is questionable. Best to always talk to Radina :)

MisRob:
Thank you @AlexVelezLl, let me know your thoughts

docs/pages/usekliveregion.vue
AlexVelezLl: Sounds good!

docs/pages/kcard.vue
AlexVelezLl: Alright, thanks @MisRob! Yes, diving deeper into each case would make it very complex. I guess it was a bit abstract to me just reading the docs, but we usually read them while playing with the components, which may make them easier to follow!

AlexVelezLl:
Thanks @MisRob, looks good to me! Merging 🚀

PR #1221, opened 2026-03-20
Add picture login icon

### LINKED ISSUE #1215: Add new icons for picture login Kolibri feature

## Overview
Kolibri is adding a picture passcode login feature for pre-literate learners. A set of 24 icons, 12 in a standard line style and 12 in a colorful style, should be registered in KDS so they can be rendered as `KIcon`s across Kolibri's login, coach, and facility plugins.

**Complexity:** Medium

### Context

These icons are needed to support the Picture Login feature designs. Reference designs can be found in the [Figma file](https://www.figma.com/design/ZNXancdqXCbCAPXmf6r1zi/Picture-login?node-id=768-12210&t=cJFRZiICBc64hHUR-4)

- **Note:** **The new leaf icon is distinct from the current [pointsActive](https://design-system.learningequality.org/icons#icons:~:text=warningIncomplete-,Learning%20activities,-allActivities) icon**
- For guidance on adding icons, see:
     - [dev_docs/05_icons.md](https://github.com/learningequality/kolibri-design-system/blob/develop/dev_docs/05_icons.md) - Step-by-step process for adding new icons
     - [custom-icons/readme.md](https://github.com/learningequality/kolibri-design-system/blob/develop/custom-icons/readme.md) - Custom icons should follow [Material guidelines](https://material.io/design/iconography/system-icons.html) for geometry, keylines, and layout

### The Change

Add twenty-four new icons to support the Picture Passcode Login feature:
1. Register all 24 icons in the KDS icon registry following the existing process for adding new icons. Use a consistent naming convention that pairs each concept with its style variant. The colorful versions should use the bare concept name and the monochrome, standard line versions append a suffix. For example: `tree` and `tree_standard`, `moon` and `moon_standard`. This convention must be applied uniformly across all 12 concept pairs.
2. Update the KDS icon documentation at `docs/pages/icons/IconTable.vue` to include the new icons, grouped under a "Picture Passcodes" category.

### How to Get There

To view & verify that the new icons have been added correctly:
1. Run yarn dev to start the development server
2. Navigate to http://localhost:4000/icons#icons to see the icons list
3. After adding icons, run yarn run precompile-custom-svgs and yarn run pregenerate

### Acceptance Criteria

- [ ] 24 new SVG files added to custom-icons/ and the naming convention is consistent across all 12 pairs
- [ ] Icons follow [Material Design guidelines](https://material.io/design/iconography/system-icons.html) for geometry, keylines, and layout
- [ ] Each icon registered in `lib/KIcon/iconDefinitions.js` with the `picturePasscode: true` property
- [ ] Icons added to docs/pages/icons/IconTable.vue & appear correctly in KDS documentation under a "Picture Passcodes" category
- [ ] `yarn run precompile-custom-svgs` runs successfully
- [ ] `yarn run pregenerate` runs successfully

### Testing

- Ensure that the colorful SVGs use only their designed fixed palette and do not inherit or respond to any KDS theme color tokens. They should render identically regardless of KDS theme context.

### References

- Figma: https://www.figma.com/design/ZNXancdqXCbCAPXmf6r1zi/Picture-login?node-id=768-12208&t=cJFRZiICBc64hHUR-4
- Related Kolibri issues:
       - https://github.com/learningequality/kolibri/issues/14021 - Parent epic
- Material Design icon guidelines: https://material.io/design/iconography/system-icons.html

## AI usage
🤖 This issue was written with AI assistance, under supervision, review and final edits by [@LianaHarris360 ] 🤖

AllanOXDi:
## Description
This PR adds set of 24 icons to support picture login work.

#### Issue addressed

#1215

### Before/after screenshots
image

## Changelog

  - **Description:** adds set of 24 icons to support picture login work.
  - **Products impact:** Kolibri
  - **Addresses:** #1215
  - **Components:** 
  - **Breaking:**N/A
  - **Impacts a11y:** NO
  - **Guidance:** Test this icon on Kolibri by running  `pnpm devserver-with-kds /path` to verify if they render correctly 

## Testing checklist

## Reviewer guidance


lib/KIcon/iconDefinitions.js
```diff
@@ -655,6 +655,122 @@ const KolibriIcons = {
...
+  },
+  fishStandard: {
+    icon: require('./precompiled-icons/le/fishStandard.vue').default,
+    picturePasscode: true,
+    defaultSize: '4em',
+  },
+  dogStandard: {
+    icon: require('./precompiled-icons/le/dogStandard.vue').default,
+    picturePasscode: true,
+    defaultSize: '4em',
```
AllanOXDi: Added default sizing to the standard icon because their label text at the bottom occupies part of the 84px viewBox, making the illustration portion look smaller.

LianaHarris360: Is this `defaultSize` property still needed, even though the label text isn't there anymore?

docs/pages/icons/IconTable.vue
```diff
@@ -45,6 +45,16 @@
         class="icon-block"
       />
     </div>
+
+    <h3 v-if="picturePasscodes.length">Picture Passcodes</h3>
```
LianaHarris360: Can you rename this "Picture Passcodes" heading to "Picture Login Icons" instead? This is correct as stated in the issue, but in Kolibri these icons are referenced as Picture Login/Password icons, so, in hindsight, I think the naming should be consistent in KDS as well. Same with the definitions in the `iconDefinitions.js `file, let's change "picturePasscode" to "pictureLogin"

LianaHarris360 (changes requested):
The icons have been correctly added to the KDS Icons documentation page, but the standard icons are missing from the `custom-icons/` folder, they should be added there as well. I'm not sure how, but they are correctly referenced within the `precompiled-icons` folder. Also requested a minor name change for the grouped picture login icons.

lib/KIcon/iconDefinitions.js
AllanOXDi: Yes! It's looking wired without a `defaultSize`

lib/KIcon/index.vue
```diff
@@ -39,6 +39,14 @@
...
         default: null,
       },
+      /**
+       * Size override for the icon (e.g. '2em', '48px').
+       *  Falls back to the icon's defaultSize, then the CSS default.
+       */
+      size: {
+        type: String,
+        default: null,
+      },
```
AlexVelezLl: The reason why we don't have a `size` prop for KIcon is that we should define the size using CSS properties. Please take a look at the KIcon docs: https://design-system.learningequality.org/kicon/#size.

lib/KIcon/iconDefinitions.js
```diff
@@ -655,6 +655,122 @@ const KolibriIcons = {
...
+  bird: {
+    icon: require('./precompiled-icons/le/bird.vue').default,
+    pictureLogin: true,
+    defaultSize: '3.5em',
+  },
+  birdStandard: {
+    icon: require('./precompiled-icons/le/birdStandard.vue').default,
+    pictureLogin: true,
+    defaultSize: '3.5em',
+  },
```
AlexVelezLl, line 766: We do not need a `defaultSize` here. It is true that icons look weird without this, but the reason is that the exported icon is too small. Look at the difference between the 32px x 32px available space and the space the icon occupies.

Image

This is because the SVG icon itself has too much padding.

Look at other icons that do not have that much padding as an example:

Image

So the problem is how the icons were exported, rather than a need for a "defaultSize" prop on the iconDefinitions.

AlexVelezLl (changes requested):

lib/KIcon/index.vue
LianaHarris360: Yes, this pr shouldn't need to update `KIcon` at all, thanks for the additional context @AlexVelezLl

lib/KIcon/iconDefinitions.js
LianaHarris360: @AllanOXDi It appears that the first set of colorful picture login icons haven't been removed and replaced with the ones Jessica resized. Once that's done, the commands `yarn run precompile-custom-svgs` and `yarn run pregenerate` have to be re-ran as well. Then the new icon sizes should be more consistent and `defaultSize` can be removed.

AllanOXDi: It has been fix

docs/pages/icons/IconTable.vue
```diff
@@ -45,6 +45,16 @@
         class="icon-block"
       />
     </div>
+
+    <h3 v-if="pictureLoginIcons.length">Picture Login Icons</h3>
+    <div class="icon-table">
```
LianaHarris360: Right now, on the Icons documentation page, the new icons are all mixed together. Could we organize them to be side by side instead? Adding a new class to this div and giving it a set width could be one way to do it!

| Mixed | Side by Side|
|  mixed | sideBySide |

The icons are still a little bit smaller than the rest so also adjusting the icon height and width with a `::v-deep` on the icon class would help fill the space.
beeIcon

lib/KIcon/iconDefinitions.js
```diff
@@ -655,6 +655,99 @@ const KolibriIcons = {
   github: {
     icon: require('./precompiled-icons/le/github.vue').default,
   },
+  // Picture Login Icons
+  bee: {
```
LianaHarris360: Could we also set `fixedColor: true` on all of the non-standard picture login icons? To ensure the `KIcon` color property is never applied on them.

custom-icons/starColorful.svg
LianaHarris360, line 1: Just realized that this icon is not displayed on the documentation page because there is already a `star` [icon](https://design-system.learningequality.org/icons#token-star)! Since having a consistent naming convention for all of the picture login icons is really important to how we will map them in Kolibri, let's add "Colorful" to all of the colorful, kid friendly icons instead. So we'll name this one `starColorful`, and the rest will change to `beeColorful`, `birdColorful`, `moonColorful`, etc. The names for the standard icons can be left as is.

LianaHarris360 (changes requested):
Thanks for the updates Allan, this PR is almost ready to be merged! I realized the name of the colorful icons will have to change, so there's just a few more tweaks to make.

AllanOXDi:
Fixed 
image

lib/KIcon/iconDefinitions.js
```diff
@@ -655,6 +655,115 @@ const KolibriIcons = {
   github: {
     icon: require('./precompiled-icons/le/github.vue').default,
   },
+  // Picture Login Icons
+  beeColorful: {
+    icon: require('./precompiled-icons/le/bee.vue').default,
```
LianaHarris360: Let's change the file names as well, to keep this entire change consistent.

docs/pages/icons/IconTable.vue
LianaHarris360: The sizing of the icons has definitely improved. But when the screen size changes, the icon display goes right back to being mixed:

Image

Earlier I suggested changing the width of the `div` with the class `icon-table` (but only for the picture login icons), because a `width` or `max-width` of `600px` keeps the icons in two columns, and ensures the icon styles never mix, even when the page size is adjusted. For example:

Image

docs/pages/icons/IconTable.vue
```diff
@@ -173,4 +189,8 @@
     margin: 8px;
   }
 
+  .picture-login-table {
+    max-width: 600px;
```
LianaHarris360: Let's set this as just the `width` instead, so that at the smallest screen size where the sidenav is still visible, the icons stay in two columns.

Image


I also think your previous solution of shortening the width of the IconBlock and increasing the size of the icons would be useful here as well. Instead of adding another class on the `IconBlock`, it's original class could be placed within this css block as:
```
.picture-login-table {
	width: 600px;
	.icon-block {
		....
	}
}
```
I would also suggest adding a comment above it, explaining that this code keeps the two distinct picture login icon styles  in separate columns.

custom-icons/beeColorful.svg
LianaHarris360, line 1: We should make sure the names of the files in the `custom-icons/` folder match the new naming convention for colorful icons as well, to keep it consistent all the way and avoid any future collisions.

LianaHarris360 (approved):
Thanks for all the updates Allan, everything looks great!

AlexVelezLl (approved):
lgtm!

Issue #1158, opened 2025-11-06
Finalize `useKFloatingPosition`, `useKFloatingInteraction`, and `KTooltip/next`


MisRob:
<img height="20px" src="https://i.imgur.com/c7hUeb5.jpeg">

❌ **This issue is not open for contribution. Visit <a href="https://learningequality.org/contributing-to-our-open-code-base/" target="_blank">Contributing guidelines</a>** to learn about the contributing process and how to find suitable issues.

<img height="20px" src="https://i.imgur.com/c7hUeb5.jpeg">

## Summary

Finalize `useKFloatingPosition`, `useKFloatingInteraction`, and `KTooltip/next` from https://github.com/learningequality/kolibri-design-system/pull/1114.

MisRob:
Superseded by https://github.com/learningequality/kolibri-design-system/issues/1311

PR #1240, opened 2026-04-21
["floating" feature branch] Add useKFloatingPosition

MisRob:
## Description

The first PR in the series that aims to prepare selected parts of https://github.com/learningequality/kolibri-design-system/pull/1114 for production environment. Targets `floating-production` feature branch. The next PR will be for `useKFloatingInteraction`.

Adds `useKFloatingPosition` composable with new documentation pages
- [Browser support](https://deploy-preview-1240--kolibri-design-system.netlify.app/browsersupport)
- [Floating elements](https://deploy-preview-1240--kolibri-design-system.netlify.app/floatingelements)
- [useKFloatingPosition](https://deploy-preview-1240--kolibri-design-system.netlify.app/usekfloatingposition)

Note that Floating UI may be replaced by Popper, depending on results of browser testing in Kolibri & dev discussion (the structure of the composable and documentation will be the same for both, except few differences).

#### Issue addressed

Part of https://github.com/learningequality/kolibri-design-system/issues/1158

## Changelog

Skipped. Will prepare the changelog before merging the feature branch to `develop`.

## AI usage

I used Claude for draft and then further assistance when iterating on it. There's not a place that I wouldn't tweak or see ;) Also self-reviewed the final diff fully.

### DIFF
--- a/lib/composables/useKFloatingPosition/__tests__/index.spec.js
+++ b/lib/composables/useKFloatingPosition/__tests__/index.spec.js
@@ -1,7 +1,7 @@
 import useKFloatingPosition, { _instances } from '../index.js';
 
 jest.mock('@floating-ui/dom', () => ({
-  computePosition: jest.fn().mockResolvedValue({}),
+  computePosition: jest.fn().mockResolvedValue({ x: 10, y: 20, strategy: 'absolute' }),
   autoUpdate: jest.fn(() => jest.fn()),
 }));
 
@@ -29,7 +29,7 @@ describe('useKFloatingPosition', () => {
   });
 
   describe('initPosition', () => {
-    it('stores the instance & sets up auto-updating', () => {
+    it('stores the instance, sets up auto-updating, and updates the position', async () => {
       initPosition('floating-1', floatingEl, anchorEl, { placement: 'bottom' });
 
       expect(mockAutoUpdate).toHaveBeenCalledWith(anchorEl, floatingEl, expect.any(Function));
@@ -39,11 +39,14 @@ describe('useKFloatingPosition', () => {
       });
 
       const updateFn = mockAutoUpdate.mock.calls[0][2];
-      updateFn();
+      await updateFn();
 
       expect(mockComputePosition).toHaveBeenCalledWith(anchorEl, floatingEl, {
         placement: 'bottom',
       });
+      expect(floatingEl).toHaveStyle({ position: 'absolute' });
+      expect(floatingEl).toHaveStyle({ left: '10px' });
+      expect(floatingEl).toHaveStyle({ top: '20px' });
     });
 
     it('cleans up any existing instance before reinitializing', () => {
--- a/lib/composables/useKFloatingPosition/__tests__/index.spec.js
+++ b/lib/composables/useKFloatingPosition/__tests__/index.spec.js
@@ -1,0 +1,69 @@
+import useKFloatingPosition, { _instances } from '../index.js';
+
+jest.mock('@floating-ui/dom', () => ({
+  computePosition: jest.fn().mockResolvedValue({}),
+  autoUpdate: jest.fn(() => jest.fn()),
+}));
+
+const {
+  initPosition,
+  destroyPosition,
+  autoUpdate: mockAutoUpdate,
+  computePosition: mockComputePosition,
+} = useKFloatingPosition();
+
+describe('useKFloatingPosition', () => {
+  let anchorEl, floatingEl;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    anchorEl = document.createElement('div');
+    floatingEl = document.createElement('div');
+    document.body.appendChild(anchorEl);
+    document.body.appendChild(floatingEl);
+  });
+
+  afterEach(() => {
+    destroyPosition('floating-1');
+    document.body.innerHTML = '';
+  });
+
+  describe('initPosition', () => {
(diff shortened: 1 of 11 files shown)


rtibblesbot (commented):
Clean, focused addition of `useKFloatingPosition` with solid documentation.

CI passing. Verified the three new/updated docs pages (useKFloatingPosition, floatingelements, browsersupport) and the interactive tooltip example against the Netlify preview — all render correctly, the DocsWarning component is visually consistent, and the tooltip example works as expected. No design spec to compare against; color-contrast and html-lang a11y violations are pre-existing site-wide issues.

- **suggestion:** missing `.catch()` on the `computePosition` promise — see inline
- **suggestion:** test mock resolves with `{}` but no assertion covers the style update — see inline
- **nitpick:** partial mock leaves middleware exports undefined — see inline
- **praise:** defensive cleanup before re-init, verified by tests — see inline

---

Reviewed the pull request diff checking for:
- **Correctness:** bugs, edge cases, undocumented behavior, resource leaks, hardcoded values
- **Design:** unnecessary complexity, naming, readability, comment accuracy, redundant state
- **Architecture:** duplicated concerns, minimal interfaces, composition over inheritance
- **Testing:** behavior-based assertions, mocks only at hard boundaries, accurate coverage
- **Completeness:** missing dependencies, unupdated usages, i18n, accessibility, security
- **Principles:** DRY (same reason to change), SRP, Rule of Three (no premature abstraction)
- For UI changes: inspected screenshots for layout, visual completeness, and consistency

</details>

lib/composables/useKFloatingPosition/index.js
```diff
@@ -0,0 +1,99 @@
...
+   * @param {Element} floatingEl - Floating DOM element (e.g. tooltip)
+   * @param {Element} anchorEl - Anchor DOM element
+   * @param {Object} options - Floating UI options
+   *   (https://floating-ui.com/docs/computePosition#options)
+   */
+  function initPosition(floatingId, floatingEl, anchorEl, options = {}) {
+    destroyPosition(floatingId);
+
+    function updatePosition() {
+      computePosition(anchorEl, floatingEl, options).then(({ x, y, strategy }) => {
```
rtibblesbot: **suggestion:** `computePosition(...).then()` has no `.catch()`. If `computePosition` rejects — e.g. when the anchor or floating element is removed from the DOM before the promise resolves — the error is silently discarded and the element is left with stale or missing positioning. Adding even a minimal handler (`.catch(() => {})` with a `console.warn` in development) would make these failures visible during debugging.

lib/composables/useKFloatingPosition/__tests__/index.spec.js
```diff
@@ -0,0 +1,69 @@
+import useKFloatingPosition, { _instances } from '../index.js';
+
+jest.mock('@floating-ui/dom', () => ({
+  computePosition: jest.fn().mockResolvedValue({}),
```
rtibblesbot: **suggestion:** The mock resolves with `{}`, so `x`, `y`, and `strategy` inside `updatePosition` are all `undefined`. The test verifies that `computePosition` was called, but no assertion checks that `floatingEl.style.left`, `.top`, or `.position` were actually set. Adding a check like `expect(floatingEl.style.left).toBe('undefinedpx')` would at least detect regressions in the style-assignment path — or better, update the mock to resolve with `{ x: 10, y: 20, strategy: 'absolute' }` and assert the correct style values.

lib/composables/useKFloatingPosition/__tests__/index.spec.js
```diff
@@ -0,0 +1,69 @@
...
+  autoUpdate: jest.fn(() => jest.fn()),
+}));
```
rtibblesbot, line 6: **nitpick:** The mock only covers `computePosition` and `autoUpdate`; the middleware re-exports (`offset`, `shift`, `flip`, etc.) will be `undefined` in this test environment. The current tests don't exercise them directly so it's fine, but a brief comment like `// middleware re-exports are not exercised by these tests` would save future test authors from wondering why the mocked module is partial.

lib/composables/useKFloatingPosition/index.js
```diff
@@ -0,0 +1,99 @@
...
+   * problems.
+   *
+   * @param {String} floatingId - Unique ID of the floating element
+   * @param {Element} floatingEl - Floating DOM element (e.g. tooltip)
+   * @param {Element} anchorEl - Anchor DOM element
+   * @param {Object} options - Floating UI options
+   *   (https://floating-ui.com/docs/computePosition#options)
+   */
+  function initPosition(floatingId, floatingEl, anchorEl, options = {}) {
+    destroyPosition(floatingId);
```
rtibblesbot, line 46: **praise:** Calling `destroyPosition` unconditionally at the top of `initPosition` is good defensive coding — it prevents resource leaks when the same ID is reused without an explicit destroy in between. The test "cleans up any existing instance before reinitializing" verifies this precisely.

lib/composables/useKFloatingPosition/__tests__/index.spec.js
MisRob: Doesn't feel that necessary

rtibblesbot: Fair enough — withdrawing this nitpick.

rtibbles (approved):
All changes make sense to me.

PR #1247, opened 2026-04-29
Add KListbox and setup KDS for candidate components

MisRob:
## Description

Moves our implementation of APG listbox pattern from Kolibri's [SelectableList.vue](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/facility/frontend/views/common/SelectableList.vue#L27-L90) to `candidate/KListbox` and `candidate/KListboxOption`. Part of this work is refactoring `SelectableList` to use `KListbox` (https://github.com/learningequality/kolibri/pull/14695) 

Additionally
- New documentation pages & componnets (banners, badges) to set up KDS for candidate components
- Cleans up unnecessary global Jest router setup and router imports in tests (it seems that the global setup caused having to import router in components that didn't even use routing)

### Issue addressed

- Encapsulates APG `listbox` pattern
- Prepares KDS for multi-select dropdown implementation - multi-select dropdown is a combination of `combobox` and `listbox` pattern. Abstracting `listbox` pattern into a standalone component prevents from having to develop/QA/maintain the very same (and rather nuanced) a11y pattern multiple times. `KListbox` + `KListboxOption` will be used as a standalone `listbox` (as in the original Kolibri use-case) as well as under the hood of the new multi-select dropdown.

## Steps to test

**Visual tests**

Should look same as docs examples. Please preview & confirm, I can't access the build.

**Preview new documentation pages & examples**
- https://deploy-preview-1247--kolibri-design-system.netlify.app/klistbox
- https://deploy-preview-1247--kolibri-design-system.netlify.app/klistboxoption
- https://deploy-preview-1247--kolibri-design-system.netlify.app/candidate-components

**Do QA in Kolibri**

https://github.com/learningequality/kolibri/pull/14695

## AI usage

I had Claude to generate a draft according to a detailed specification I wrote and the original Kolibri's `SelectableList.vue` implementation. Then went through all places and refined in collaboration with Claude. Did final self-review of the whole diff.

## Changelog

  - **Description:** Add KListbox and KListboxOption
  - **Products impact:** New component
  - **Addresses:**-
  - **Components:** KListbox, KListboxOption
  - **Breaking:** No
  - **Impacts a11y:** yes
  - **Guidance:** -

  - **Description:**  Cleans up unnecessary global Jest router setup and router imports in tests
  - **Products impact:** None
  - **Addresses:**-
  - **Components:** -
  - **Breaking:** No
  - **Impacts a11y:** -
  - **Guidance:** -

### DIFF
--- a/docs/assets/definitions.scss
+++ b/docs/assets/definitions.scss
@@ -6,4 +6,4 @@ $link-hover-color: #26614d;
 $selection-color: #badbd2;
 $basic-transition: color 0.25s ease,
   fill 0.25s ease;
-$nav-width: 260px;
+$nav-width: 280px;
--- a/docs/common/DocsPageSection.vue
+++ b/docs/common/DocsPageSection.vue
@@ -42,7 +42,7 @@
     },
     computed: {
       style() {
-        return this.fullwidth ? {} : { maxWidth: '900px' };
+        return this.fullwidth ? {} : { maxWidth: '1000px' };
       },
     },
     mounted() {
--- a/docs/common/DocsPageTemplate/BadgeCandidate.vue
+++ b/docs/common/DocsPageTemplate/BadgeCandidate.vue
@@ -1,0 +1,48 @@
+<template>
+
+  <span
+    class="badge"
+    :class="{ small }"
+    :style="{
+      backgroundColor: $themePalette.lightblue.v_50,
+      color: $themePalette.lightblue.v_600,
+    }"
+  >
+    Candidate
+  </span>
+
+</template>
+
+
+<script>
+
+  export default {
+    name: 'BadgeCandidate',
+    props: {
+      small: {
+        type: Boolean,
+        default: false,
+      },
+    },
+  };
+
+</script>
+
+
+<style lang="scss" scoped>
+
+  .badge {
+    display: inline-block;
+    padding: 2px 8px;
+    margin-left: 12px;
+    font-size: 14px;
+    border-radius: 6px;
+  }
+
+  .small {
+    padding: 2px 6px;
+    margin-left: 6px;
+    font-size: 12px;
+  }
+
+</style>
--- a/docs/common/DocsPageTemplate/index.vue
+++ b/docs/common/DocsPageTemplate/index.vue
@@ -16,6 +16,7 @@
       :sections="pageSections"
       :title="page.title"
       :codeStyle="page.isCode"
+      :candidate="page.candidate"
       class="floating-header"
       @update-side-nav="handleSideNavUpdate"
     />
@@ -26,10 +27,26 @@
         :sections="pageSections"
         :title="page.title"
         :codeStyle="page.isCode"
+        :candidate="page.candidate"
         style="visibility: hidden"
       />
 
       <div class="content">
+        <DocsPageSection>
+          <DocsBanner v-if="page.candidate">
+            <p style="margin: 0">
+              This component is a
+              <DocsInternalLink
+                href="/candidate-components"
+                text="candidate"
+              />.
+            </p>
+            <div v-if="$slots.developerNotes">
+              <h2 class="developer-notes-title">Developer notes</h2>
+              <slot name="developerNotes"></slot>
+            </div>
+          </DocsBanner>
+        </DocsPageSection>
         <slot></slot>
         <DocsPageSection
           v-for="(section, i) in apiSections"
@@ -242,6 +259,13 @@
       padding: 0 16px 200px;
       overflow-x: hidden;
     }
+  }
+
+  .developer-notes-title {
+    margin: 18px 0 4px;
+    font-size: inherit;
+    font-weight: normal;
+    text-decoration: underline;
   }
 
   .overlay {
--- a/docs/pages/candidate-components.vue
+++ b/docs/pages/candidate-components.vue
@@ -62,6 +62,19 @@
         <li>It is highly specific to a single application</li>
       </ul>
     </DocsPageSection>
+
+    <DocsPageSection
+      title="Note"
+      anchor="#note"
+    >
+      <p>
+        See additional details and examples on the candidate process in this
+        <DocsExternalLink
+          href="https://www.figma.com/design/nGLCIW2iZBsY1AZcyjQdDE/KDS-Candidate"
+          text="Figma file"
+        />.
+      </p>
+    </DocsPageSection>
   </DocsPageTemplate>
 
 </template>
--- a/docs/pages/candidate-components.vue
+++ b/docs/pages/candidate-components.vue
@@ -1,0 +1,76 @@
+<template>
+
+  <DocsPageTemplate>
+    <DocsPageSection
+      title="Overview"
+      anchor="#overview"
+    >
+      <p>
+        A <strong>candidate component</strong> is a component whose accessibility, visual design,
+        and interactions are largely established, and which has been identified as useful to include
+        in the design system, but whose user experience or implementation details need more time to
+        mature. This means that the following are expected to be refined:
+      </p>
+      <ul>
+        <li>Visual and interaction details</li>
+        <li>Props, slots, events, sub-components</li>
+        <li>What parts belong in the design system versus consuming applications</li>
+      </ul>
+
+      <p>
+        Candidate components are <strong>in active use</strong> in consuming applications.
+        Therefore, they need to follow the
+        <strong>same release planning as regular components</strong>, particularly regarding
+        breaking changes. Thorough unit and visual test coverage is expected to allow for smooth
+        iteration.
+      </p>
+    </DocsPageSection>
+
+    <DocsPageSection
+      title="Criteria"
+      anchor="#criteria"
+    >
+      <p>
+        A component
+        <strong>can be considered for candidacy when one or more of the following are true:</strong>
+      </p>
+      <ul>
+        <li>
+          It maps to a well-established accessibility pattern or is a collection of such patterns
+          (e.g.
+          <DocsExternalLink
+            href="https://www.w3.org/WAI/ARIA/apg/patterns/"
+            text="APG patterns"
+          />)
+        </li>
+        <li>The overall user experience is unlikely to change significantly</li>
+      </ul>
+
+      <p>
+        For example, consider a modal side panel component. Its accessibility requirements (focus
+        management, ARIA roles, and keyboard handling) apply uniformly to all use cases, and the
+        main parts of its visual layout are well-defined. Even with some visual or implementation
+        details still unresolved, a single component ensures a more reliable and consistent user
+        experience, avoids duplicated work, and lets real use cases strengthen its robustness across
+        a wide variety of contexts.
+      </p>
+
+      <p>A component <strong>is unlikely to become a candidate when:</strong></p>
+      <ul>
+        <li>The accessibility pattern is still being explored</li>
+        <li>The user experience is expected to change significantly</li>
+        <li>It is highly specific to a single application</li>
+      </ul>
+    </DocsPageSection>
+  </DocsPageTemplate>
+
+</template>
+
+
+<script>
+
+  export default {
+    name: 'CandidateComponents',
+  };
+
+</script>
--- a/docs/pages/candidate-components.vue
+++ b/docs/pages/candidate-components.vue
@@ -18,11 +18,8 @@
       </ul>
 
       <p>
-        Candidate components are <strong>in active use</strong> in consuming applications.
-        Therefore, they need to follow the
-        <strong>same release planning as regular components</strong>, particularly regarding
-        breaking changes. Thorough unit and visual test coverage is expected to allow for smooth
-        iteration.
+        Candidate components are <strong>in active use</strong> in consuming applications. Thorough
+        unit and visual test coverage is expected to allow for smooth iterations.
       </p>
     </DocsPageSection>
 
--- a/docs/pages/klistbox.vue
+++ b/docs/pages/klistbox.vue
@@ -21,6 +21,11 @@
           listbox pattern and use cases vary. If a frequent pattern emerges, it may motivate
           dedicated search/filter components that integrate well with <code>KListbox</code>.
         </li>
+        <li>
+          Currently KListbox doesn't support options reodering, so bear in mind that if the list
+          changes due to a filter, the filtered list should preserve the order of the original list
+          to avoid unexpected keyboard navigation.
+        </li>
       </ul>
     </template>
 
--- a/docs/pages/klistbox.vue
+++ b/docs/pages/klistbox.vue
@@ -1,0 +1,247 @@
+<template>
+
+  <DocsPageTemplate apiDocs>
+    <template #developerNotes>
+      <ul :style="{ margin: 0 }">
+        <li>
+          Grouped options aren't supported yet, as we haven't needed them. They could be added by
+          extending our listbox components according to
+          <DocsExternalLink
+            text="grouped options of APG's listbox pattern"
+            href="https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-grouped/"
+          />
+          (likely adding <code>KListboxGroup</code>)
+        </li>
+        <li>
+          Reflecting our first use-cases, multi selection is the default behavior. Single selection
+          mode could be added by a few adjustments in line with the same APG listbox pattern.
+        </li>
+        <li>
+          Listboxes are often combined with searching or filtering, but that's not part of the
+          listbox pattern and use cases vary. If a frequent pattern emerges, it may motivate
+          dedicated search/filter components that integrate well with <code>KListbox</code>.
+        </li>
+      </ul>
+    </template>
+
+    <DocsPageSection
+      title="Overview"
+      anchor="#overview"
+    >
+      <p>
+        <code>KListbox</code> is a representation of the
+        <DocsExternalLink
+          text="APG listbox pattern"
+          href="https://www.w3.org/WAI/ARIA/apg/patterns/listbox/"
+        />
+        with several additional enhancements. It can be used as a standalone listbox or within a
+        composite widget such as a dropdown or combobox.
+      </p>
+
+      <p>Its options are provided via <DocsLibraryLink component="KListboxOption" />.</p>
+
+      <p>
+        It currently defaults to multi selection and does not yet support single selection or
+        grouped options (see developer notes above).
+      </p>
+    </DocsPageSection>
+
+    <DocsPageSection
+      title="Messages"
+      anchor="#messages"
+    >
+      <p>
+        <code>KListbox</code> requires a <code>messages</code> object. Each entry must be a
+        translated string that is short and focused.
+      </p>
+      <KTable
+        :headers="messageHeaders"
+        :rows="tableRows"
+        caption="KListbox messages"
+        :stickyColumns="['first']"
+      >
+        <template #cell="{ content, colIndex }">
+          <code v-if="colIndex === 0">{{ content }}</code>
+          <i v-else-if="colIndex === 3">{{ content }}</i>
+          <template v-else>{{ content }}</template>
+        </template>
+      </KTable>
+    </DocsPageSection>
+
+    <DocsPageSection
+      title="Usage"
+      anchor="#usage"
+    >
+      <DocsSubNav
+        :items="[
+          { text: 'Default', href: '#default' },
+          { text: 'Select all', href: '#select-all' },
+          { text: 'Customized', href: '#customized' },
+          { text: 'Scrollable', href: '#scrollable' },
+        ]"
+      />
+
+      <h3>
+        Default
+        <DocsAnchorTarget anchor="#default" />
+      </h3>
+
+      <p><code>KListboxOption</code> is a direct child of <code>KListbox</code>.</p>
+
+      <DocsBanner a11y>
+        If <code>KListbox</code> is not part of another widget, such as a combobox, add either a
+        visible label referenced by <code>aria-labelledby</code> or a value specified for
+        <code>aria-label</code>. Use concise and descriptive value.
+      </DocsBanner>
+
+      <DocsExample
+        block
+        exampleId="klistbox-default"
+        loadExample="KListbox/Default.vue"
+      />
+
+      <h3>
+        Select all
+        <DocsAnchorTarget anchor="#select-all" />
+      </h3>
+
+      <p>
+        Use the <code>#selectAll</code> scoped slot together with a select control, typically
+        <code>KCheckbox</code>, to implement the select all functionality.
+      </p>
+
+      <DocsBanner a11y>
+        Set <code>aria-controls</code> on the control to the listbox <code>id</code> to associate
+        the two.
+      </DocsBanner>
+
+      <DocsExample
+        block
+        exampleId="klistbox-select-all"
+        loadExample="KListbox/SelectAll.vue"
+      />
+
+      <h3>
+        Customized
+        <DocsAnchorTarget anchor="#customized" />
+      </h3>
+      <p>
+        Apply style on the listbox and its rows to add more space, borders, or other appearance
+        customizations. Use slots to show rich content.
+      </p>
+
+      <DocsBanner a11y>
+        Ensure custom content is accessible and the final implementation works as a whole. For
+        example here, setting both the row and checkbox width to 100% is needed to make the entire
+        <i>All users</i> row clickable.
+      </DocsBanner>
+
+      <DocsExample
+        block
+        exampleId="klistbox-customized"
+        loadExample="KListbox/Customized.vue"
+      />
+
+      <h3>
+        Scrollable
+        <DocsAnchorTarget anchor="#scrollable" />
+      </h3>
+      <p>
+        To limit the space a long list takes up, apply a maximum height, and the list will become
+        scrollable.
+      </p>
+      <DocsExample
+        block
+        exampleId="klistbox-scrollable"
+        loadExample="KListbox/Scrollable.vue"
+        hideStyle
+        hideScript
+      >
+        <template #html>
+          <!-- eslint-disable -->
+          <DocsShowCode language="html">
+            <KListbox
+              :style="{ maxHeight: '200px' }"
+              ...
+            >
+              ...
+            </KListbox>
+          </DocsShowCode>
+          <!-- eslint-enable -->
+        </template>
+      </DocsExample>
+    </DocsPageSection>
+
+    <DocsPageSection
+      title="Related"
+      anchor="#related"
+    >
+      <ul>
+        <li>
+          <DocsLibraryLink component="KListboxOption" /> is a single option inside
+          <code>KListbox</code>
+        </li>
+        <li>
+          <DocsExternalLink
+            text="APG listbox"
+            href="https://www.w3.org/WAI/ARIA/apg/patterns/listbox/"
+          />
+          specifies the related accessibility pattern
+        </li>
+      </ul>
+    </DocsPageSection>
+  </DocsPageTemplate>
+
+</template>
+
+
+<script>
+
+  export default {
+    data() {
+      return {
+        messageHeaders: [
+          { label: 'Name', dataType: 'string', columnId: 'name', minWidth: '220px' },
+          { label: 'Required', dataType: 'string', columnId: 'required', minWidth: '140px' },
+          { label: 'Description', dataType: 'string', columnId: 'description' },
+          { label: 'Examples', dataType: 'string', columnId: 'example', minWidth: '200px' },
+        ],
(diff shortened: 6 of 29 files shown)


MisRob:
cc @Prashant-thakur77 (more in #design-system channel on Slack)

docs/assets/definitions.scss
```diff
@@ -6,4 +6,4 @@ $link-hover-color: #26614d;
 $selection-color: #badbd2;
 $basic-transition: color 0.25s ease,
   fill 0.25s ease;
-$nav-width: 260px;
+$nav-width: 280px;
```
MisRob, line 9: This change was to create more room for the candidate badges

docs/common/DocsPageSection.vue
```diff
@@ -42,7 +42,7 @@
     },
     computed: {
       style() {
-        return this.fullwidth ? {} : { maxWidth: '900px' };
+        return this.fullwidth ? {} : { maxWidth: '1000px' };
```
MisRob, line 45: This change was to create more room for wider tables, such as the one I added for `KListbox` messages & it aligns maximum width with our convention in other codebases.

MisRob:
@LianaHarris360 @AlexVelezLl Both KDS and Kolibri PR are ready.

cc @Prashant-thakur77 please familiarize yourself.

cc @marcellamaki you can preview documentation pages, especially the one for candidate process. 

Will be good to get both PRs merged soon, or at least KDS one, Prashant & Liana should then have easier work :) I am not sure which release to target (this is not breaking), please retarget as needed anyone. Alex will provide guidance.

MisRob:
`KListbox` is implemented to be as lest opinionated as possible (such as as minimal spacing around options as possible) in its default look to allow for easy customization from outside, useful in general & needed in all our current-use cases.

This includes lack of space next to the left of checkbox ~ should be helpful particularly in GSoC multi-select nesting that are visually distinguished by left padding.

In support of GSoC multi-select, where nesting will be needed, I recommend to (1) not make any assumptions about number of nesting levels (I think I saw even 3 levels somewhere in Studio), (2) adding another component `KListboxGroup` to represent APG listbox grouped options (see more in developer notes at the top of KListbox docs)

GSoC multi-select doesn't need to have the same public interface as `KListbox` and related components, however it should use it under the hood. `KListbox` can be adjusted if needed, however it needs to be aware of standalone listbox uses cases mentioned here https://github.com/learningequality/kolibri/pull/14695. Best to coordinate with Alex and Marcella on releases and breaking changes if any arise.

docs/common/DocsPageTemplate/BadgeCandidate.vue
```diff
@@ -0,0 +1,48 @@
+<template>
+
+  <span
+    class="badge"
+    :class="{ small }"
+    :style="{
+      backgroundColor: $themePalette.lightblue.v_50,
+      color: $themePalette.lightblue.v_600,
```
rtibblesbot, line 8: **suggestion:** axe-core reports this badge fails WCAG AA contrast. Computed styles confirm `rgb(0, 121, 194)` on `rgb(229, 245, 253)` ≈ 4.1:1, below the 4.5:1 threshold for 14 px text. Switching to `$themePalette.lightblue.v_700` (or darker) should cross the threshold — worth checking with a WCAG contrast calculator before landing.

lib/candidate/listbox/KListbox/__tests__/index.spec.js
```diff
@@ -0,0 +1,285 @@
...
+
+function renderComponent({ props = {}, slots = {}, ...rest } = {}) {
+  return render(KListbox, {
+    props: { id: LISTBOX_ID, value: [], messages: MESSAGES, ...props },
+    slots: { default: makeListboxOptions(), ...slots },
+    ...rest,
+  });
+}
+
+describe('KListbox', () => {
```
rtibblesbot, line 36: **suggestion:** The `#selectAll` slot API has no direct test coverage. `changeSelectAll` is exercised via Ctrl+A, but none of the tests:
- render the component with a `#selectAll` slot to confirm `allSelected`/`someSelected`/`toggle` slot props are wired correctly
- exercise the container's `@click.self` handler (clicking the row outside the checkbox toggles all — this is easy to break silently)

Consider adding a `describe('#selectAll slot')` block that mounts with a minimal slot and verifies the slot props reflect selection state and that calling `toggle` drives `allSelected`.

lib/candidate/listbox/KListbox/index.vue
```diff
@@ -0,0 +1,384 @@
...
+        unregisterOption,
+        isSelected,
+        isFocused,
+        isFocusVisible,
+        toggleOption,
+      });
+
+      onMounted(() => {
+        // Warn if any KListboxOption is not a direct child of the listbox element
+        if (process.env.NODE_ENV !== 'production') {
```
rtibblesbot, line 280: **praise:** Good dev-mode guard — catching indirect options at mount time with a clear warning message is exactly the kind of developer-ergonomics detail that prevents subtle a11y bugs silently.

rtibblesbot (commented):
Clean, well-structured addition of `KListbox` and `KListboxOption` — the APG pattern is faithfully implemented and the test suite is one of the more thorough I've seen in this repo.

CI passing. Verified via Netlify deploy preview: pages render correctly, the `Candidate` badge and developer notes display as expected, Candidate entries show in sidebar nav. axe-core flagged one new contrast issue (see inline).

- **suggestion**: `BadgeCandidate` badge text fails WCAG AA contrast — see inline on `BadgeCandidate.vue:8`
- **suggestion**: `#selectAll` slot API not covered by tests — see inline on `index.spec.js:36`
- **praise**: dev-mode warning for indirect options (see inline on `KListbox/index.vue:284`)
- **praise**: Keyboard test coverage is thorough — all APG key bindings tested with wrap navigation and the tricky dynamic-selection-preservation case

---

Reviewed the pull request diff checking for:
- **Correctness:** bugs, edge cases, undocumented behavior, resource leaks, hardcoded values
- **Design:** unnecessary complexity, naming, readability, comment accuracy, redundant state
- **Architecture:** duplicated concerns, minimal interfaces, composition over inheritance
- **Testing:** behavior-based assertions, mocks only at hard boundaries, accurate coverage
- **Completeness:** missing dependencies, unupdated usages, i18n, accessibility, security
- **Principles:** DRY (same reason to change), SRP, Rule of Three (no premature abstraction)
- For UI changes: inspected screenshots for layout, visual completeness, and consistency

</details>

docs/common/DocsPageTemplate/BadgeCandidate.vue
MisRob: This concerns just the new KDS docs candidate badge and we don't have darker `v_700` available. I will skip it now but if anyone wants to play with its appearance, feel free to.

lib/candidate/listbox/KListbox/__tests__/index.spec.js
MisRob: I don't think there's need to test integration with `KCheckbox`, but yes, sounds useful to test that props passed from the slot update properly. Nothing blocking though.

docs/pages/candidate-components.vue
```diff
@@ -0,0 +1,89 @@
...
+      <ul>
+        <li>Visual and interaction details</li>
+        <li>Props, slots, events, sub-components</li>
+        <li>What parts belong in the design system versus consuming applications</li>
+      </ul>
+
+      <p>
+        Candidate components are <strong>in active use</strong> in consuming applications.
+        Therefore, they need to follow the
+        <strong>same release planning as regular components</strong>, particularly regarding
```
AlexVelezLl: Given that it is expected that these components may have "Props, slots, events, sub-components" to be refined, I'd say we need to be more flexible with these components, so that we have room to play, because if not, we will end up having to plan breaking changes on these components, and that would limit what we want to achieve with them of "not having everything resolved" on their first implementation. So we can say we are more flexible about introducing breaking changes in a minor release, but make sure the change is well aligned with current component usage, and that the person opening the PR is responsible for updating all component usages accordingly before merging the KDS PR.

docs/pages/candidate-components.vue
AlexVelezLl, line 1: Reflecting on the question "would it be okay to use a candidate component on a non-candidate component," because of the upcoming work with multi-select, (I think we shouldn't allow it), but that led me to realize that we can also potentially consider "candidate" components to draft components that we are building through multiple PRs, and we end up having a draft state on a KDS release branch but still arent fully usable. Like, for example, KCard or KTable. Not that we should care a lot about the wording, I know that this isn't the exact same case, but it still shares some aspects, and having something like this candidate flag would be a good note to inform that the component isn't yet finalized.

docs/common/DocsPageTemplate/index.vue
```diff
@@ -26,10 +27,26 @@
...
+          <DocsBanner v-if="page.candidate">
+            <p style="margin: 0">
+              This component is a
+              <DocsInternalLink
+                href="/candidate-components"
+                text="candidate"
+              />.
+            </p>
+            <div v-if="$slots.developerNotes">
+              <h2 class="developer-notes-title">Developer notes</h2>
```
AlexVelezLl, line 45: The developer notes are something like documenting the current plan/decisions about the component that may impact future implementation?

docs/pages/klistbox.vue
```diff
@@ -0,0 +1,247 @@
...
+        caption="KListbox messages"
+        :stickyColumns="['first']"
+      >
+        <template #cell="{ content, colIndex }">
+          <code v-if="colIndex === 0">{{ content }}</code>
+          <i v-else-if="colIndex === 3">{{ content }}</i>
+          <template v-else>{{ content }}</template>
+        </template>
+      </KTable>
+    </DocsPageSection>
```
AlexVelezLl, line 80: Like it! I'd like to see this pattern more broadly across KDS components that may need messages on it. Do you think it may be a good idea to standardize this even if the component only needs one message?

The only other thing I was thinking about is what if we ever need custom data on the message, then we would need the string function here, but we can always just extend this current pattern to accept functions, too. So, don't think we should worry too much about that/

lib/candidate/listbox/KListbox/index.vue
```diff
@@ -0,0 +1,384 @@
...
+        name="selectAll"
+        :allSelected="allSelected"
+        :someSelected="someSelected"
+        :toggle="changeSelectAll"
+      ></slot>
+    </div>
+
+    <ul
+      v-show="hasOptions"
+      v-bind="$attrs"
```
AlexVelezLl: The only thing I don't like about `v-bind="$attrs"` is that we do not document on the `props` table of the docs page that we should pass the aria attributes. Usually, we (I 😅 ) just go and see the `props` table because it's like a summary of the page, and a quick reference of the most "actionable" info of the component. Just wondering if we should just accept these values as props to be clearer about what we expect here?

MisRob:
Kolibri QA done & passes

docs/common/DocsPageTemplate/BadgeCandidate.vue
rtibblesbot: Understood — makes sense to skip it given the palette constraint. Withdrawing the finding.

lib/candidate/listbox/KListbox/__tests__/index.spec.js
rtibblesbot: Agreed — slot prop wiring tests would be the useful part. Happy to leave it as a non-blocking follow-up.

lib/__tests__/KCheckbox.spec.js
```diff
@@ -1,12 +1,10 @@
 import { render, screen } from '@testing-library/vue';
 import userEvent from '@testing-library/user-event';
-import VueRouter from 'vue-router';
 import KCheckbox from '../KCheckbox';
 
 const renderComponent = (props = {}, slots = {}) =>
   render(KCheckbox, {
     props,
-    routes: new VueRouter(),
```
AlexVelezLl, line 9: I recall wanting to remove this, but it was throwing errors, I think. Did you do anything special to be able to remove this? Was it due to the `Vue.use(VueRouter);` on the jest setup file?

lib/candidate/listbox/KListbox/__tests__/index.spec.js
```diff
@@ -0,0 +1,285 @@
...
+    expect(warn).toHaveBeenCalledWith('[KListboxOption] must be a direct child of KListbox.');
+    warn.mockRestore();
+  });
+
+  it('listbox has correct role and attributes', async () => {
+    renderComponent();
+    const listbox = await screen.findByRole('listbox');
+    expect(listbox).toHaveAttribute('aria-multiselectable', 'true');
+    expect(listbox).toHaveAttribute('tabindex', '0');
+    expect(listbox).toHaveAttribute('data-focus', 'true');
```
AlexVelezLl, line 67: (Nitpick) Do we need the `data-focus` on the listbox element? We'd never show the focus outline on the node itself; it rather redirects the focus to the options. Just wondering, is there the possibility of having a flash of the focus outline jumping from the listbox element to an option on low-resource devices?

lib/candidate/listbox/KListbox/index.vue
```diff
@@ -0,0 +1,384 @@
...
+      :class="$computedClass(selectAllStyles)"
+      :style="{ backgroundColor: $themeTokens.surface }"
+      @click.self="changeSelectAll(!allSelected)"
+    >
+      <!-- @slot Scoped slot for implementing select all functionality -->
+      <slot
+        name="selectAll"
+        :allSelected="allSelected"
+        :someSelected="someSelected"
+        :toggle="changeSelectAll"
```
AlexVelezLl: Just noting that the `toggle` name implies that it just toggles the allSelected state without needing any prop, but it does receive a prop, and is what it is set to `allSelected.` This works well with KCheckbox because it automatically sends the new state to the `@change` handler, but in case we attach this to another component, this may not be clear.

IMO, we can either turn the `changeSelectAll` into an actual toggle and just not receive any prop, or we can change the `toggle` name to something like `setAllSelected`.

lib/candidate/listbox/KListbox/index.vue
```diff
@@ -0,0 +1,384 @@
...
+  export default {
+    name: 'KListbox',
+    inheritAttrs: false,
+    setup(props, { emit }) {
+      const { sendPoliteMessage } = useKLiveRegion();
+      const instance = getCurrentInstance();
+      const ariaDescribedById = `${props.id}-description`;
+      const options = ref([]);
+      const focusedValue = ref(null);
+      const isKeyboardFocus = ref(false);
```
AlexVelezLl: Tracked down the usages of this ref, and it seems it is used to set the focus outline styles on the option, but in the option component, we already use `$coreOutline` and this already takes care of showing the outline only if we are keyboard-navigating. Is there any edge case that this solves? Or should we just remove the logic related to this `isKeyboardFocus` and rely on the `$coreOutline`, which is how we usually handle this?

lib/candidate/listbox/KListbox/index.vue
```diff
@@ -0,0 +1,384 @@
...
+      function registerOption(option) {
+        options.value.push(option);
+        // After an option mounts, reorder options state to match DOM order
+        // so that keyboard navigation and focus behave as expected
+        nextTick(() => {
+          const orderedIds = Array.from(listEl.value.querySelectorAll('[role="option"]')).map(
+            el => el.id,
+          );
+          options.value.sort((a, b) => orderedIds.indexOf(a.id) - orderedIds.indexOf(b.id));
+        });
```
AlexVelezLl, line 100: Wow nice catch, hadn't thought about it.

AlexVelezLl: I've been thinking that querying the DOM on every option registration and then sorting these values may not be the most efficient way if we have too many options. And more so, because we only need to run this once after items have finished rendering (i.e., just one call to `nextTick`, not N times). So i was thinking that we can either wrap this on a debound method, or just use a simple lock so that the callback on `nextTick` is only called once.

```JavaScript
// We can use this `isSortPending` as a "lock" to prevent queueing more sorts if we have already done so with the first value. Then, after Vue finishes rendering, and the sort has happened, it will be reset.
let isSortPending = false;

function registerOption(option) {
  options.value.push(option);

  // If a sort is already scheduled, don't do anything else
  if (isSortPending) return;

  isSortPending = true;

  // After an option mounts, reorder options state to match DOM order
  // so that keyboard navigation and focus behave as expected
  nextTick(() => {
    const orderedIds = Array.from(listEl.value.querySelectorAll('[role="option"]'))
      .map(el => el.id);

    options.value.sort((a, b) => orderedIds.indexOf(a.id) - orderedIds.indexOf(b.id));

    // Reset the flag so future dynamic additions (like search) can trigger a resort
    isSortPending = false;
  });
}
```

lib/candidate/listbox/KListbox/index.vue
```diff
@@ -0,0 +1,384 @@
...
+        }
+
+        event.preventDefault();
+      }
+
+      provide('klistbox', {
+        registerOption,
+        unregisterOption,
+        isSelected,
+        isFocused,
```
AlexVelezLl, line 274: Seems like we are not using this `isFocused` method.

lib/candidate/listbox/KListbox/index.vue
```diff
@@ -0,0 +1,384 @@
...
+          case 'a':
+          case 'A':
+            if (!event.ctrlKey && !event.metaKey) {
+              return;
+            }
+            if (allSelected.value) {
+              deselectAllOptions();
+            } else {
+              selectAllOptions();
+            }
```
AlexVelezLl, line 259: I recall making a very intentional decision around not deselecting when pressing Ctrl+A. I think it was because this is how it behaves on some OS's, where on a list of files on your file explorer, Ctrl A only selects, but does not deselect. But just noticed that the APG guidelines allow it, so it's fine.

lib/candidate/listbox/KListbox/index.vue
AlexVelezLl: Ah, just saw we are not using `$coreOutine`, but... should we?

AlexVelezLl (changes requested):
Thanks, Misha! This is looking great! I'm really liking the `candidate` components. I think the most "worrying" comment is about the options registration efficiency. Everything else is pretty much optional/questions :).

MisRob:
Thank you @AlexVelezLl, good notes as always. I have time to and would prefer to respond & resolve all at once while in this PR and should post not later than next Monday, so if that works well, we can do it that way :)

lib/candidate/listbox/KListbox/index.vue
MisRob: Ah okay, I don't think I thought about this at all. I'm glad to let you do the final decisions.

One thing to consider ~ while it's true that CTRL+A doesn't deselect all files in a file explorer there is a simple way to deselect all files by a single keyboard action ~ e.g. to press key up/down, etc. However listbox is not file explorer & if someone selects 20 options with CTR+A, and realizes that was a mistake they will need to circle through all of them with Arrow Up/Down + Space. Or is there another way to deselect all at once?

lib/candidate/listbox/KListbox/index.vue
MisRob: Ah that's a good catch ~ yes, this will definitely be better

AlexVelezLl:
That'd work perfectly, thanks!

lib/candidate/listbox/KListbox/index.vue
AlexVelezLl: > in a file explorer there is a simple way to deselect all files by a single keyboard action

That's a good point, yes! 😅 I feel like this would be a good reason to keep it :)

docs/pages/candidate-components.vue
MisRob: My original idea was that candidate components won't bypas the usual release process and follow the same semantic versioning, however the idea behind it was really only about balancing communication, planning, and flexibility. There are many ways to formalize the flow, perhaps you can just observe and see over time :)? And in this PR I would cleanup notes that are either too specific or tied into how I tried to imagine it?

docs/pages/candidate-components.vue
MisRob: Yes that was too my thinking, and I actually thought `KTable` would definitely qualify. I planned to raise labeling it as candidate. Same for GSoC multi-select. Not that sure about `KCard`, feels fairly stable?

docs/common/DocsPageTemplate/index.vue
MisRob: Yes. I wanted to write some notes that felt important to share but putting them into the documentation itself felt like unnecessarily 'polluting' it ~ they are not so much about how to use it.

docs/pages/klistbox.vue
MisRob: Yes, had exactly the same thought :)

lib/candidate/listbox/KListbox/index.vue
MisRob: Okay that's fine, I don't have strong preference. Will change :)

lib/__tests__/KCheckbox.spec.js
MisRob: Yes it was related to the global Vue Router setup

lib/candidate/listbox/KListbox/index.vue
MisRob: Alright. In any case I see value in consistency of CTRL+A across different interfaces, so I think it was good consideration.

docs/pages/candidate-components.vue
AlexVelezLl: > There are many ways to formalize the flow, perhaps you can just observe and see over time :)?

Yeah, sounds good! We can keep it as is right now, and then evaluate it over time.

docs/pages/candidate-components.vue
AlexVelezLl: Ah, no, no, I meant, it would have made sense for KCard to have a `candidate` label when they were still being implemented 😅. I think it may be fine for now. And yes, it'd make sense to flag Multi-Select as a candidate during the development process until we have a stable implementation, cc: @LianaHarris360.

lib/__tests__/KCheckbox.spec.js
AlexVelezLl: great!

docs/pages/candidate-components.vue
MisRob: Ah okay, yes, that would too make sense. I think all mentioned above qualifies :)

docs/pages/candidate-components.vue
MisRob: Alright, I've just removed the note on the release process to not confuse things

lib/candidate/listbox/KListbox/__tests__/index.spec.js
MisRob: Yes, it's needed ~ without it, focus ring on options doesn't work. If I recall I took this from the original implementation. I've just tried to play a bit with few options, but there's lots of interconnected pieces such as `tabindex=0`, `activedescendant` strategy, and at some point I felt lost :) Didn't figure out an alternative that would seem to work well.

lib/candidate/listbox/KListbox/index.vue
MisRob: Thank you, good note. I changed to `setAllSelected` and preserved Boolean parameter ~ it's more explicit, just in case it's handy.

lib/candidate/listbox/KListbox/index.vue
MisRob: Ah yes. If I recall at some point I observed that focus ring was displayed on mouse click unexpectedly so I added `isKeyboardFocus` detection to deal with it. Forgot that `$coreOutline` (now used in `KListboxOption`) detects keyboard modality. Cleaned up.

lib/candidate/listbox/KListbox/index.vue
MisRob: Done

lib/candidate/listbox/KListbox/index.vue
MisRob: It's used by `KListboxOption`

MisRob:
@AlexVelezLl I think all feedback is resolved or responded. I haven't previewed updated version in Kolibri, but haven't noticed any regressions in documentation examples so wouldn't expect trouble. In any case, after KDS is released and installed to the companion Kolibri PR, then it may be good to do final QA anyway.

lib/candidate/listbox/KListboxOption/index.vue
```diff
@@ -0,0 +1,104 @@
+<template>
+
+  <li
+    :id="optionId"
+    role="option"
+    class="k-listbox-option"
+    :class="$computedClass(rowStyles)"
+    :style="isFocused ? { ...$coreOutline, outlineOffset: '-3px' } : {}"
```
rtibblesbot, line 8: **praise:** Delegating the keyboard/mouse outline decision to `$coreOutline` removes a tracking ref (`isKeyboardFocus`) that was serving as a local proxy for something already tracked globally. `$coreOutline` is a reactive computed on `Vue.observable(globalThemeState)`, so it updates on modality changes exactly as the old manual tracking did — same behavior, less state.

lib/composables/useNextTickOnce.js
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Returns an `once(fn)` function that defers `fn`
+ * to the next Vue tick. If called again before
+ * the tick fires, the call is ignored. After the
+ * tick, the lock resets so future calls can schedule
+ * again. Useful for preventing from multiple calls
+ * of the same function in the same tick.
+ */
+export default function useNextTickOnce() {
```
rtibblesbot, line 11: **praise:** Clean encapsulation of a non-obvious scheduling invariant. The important detail — that only the *first* registered callback fires per tick, but the callback reads live `ref` values at execution time so it correctly sees all registrations made before the tick fires — is a subtle correctness property that would otherwise need a comment in `registerOption`. Pulling it into a named composable makes the intent at the call site (`onceSort`) self-documenting.

rtibblesbot (approved):
Good iteration — the new commits clean up focus-tracking state, make the aria-label API explicit and self-documenting, and batch DOM sorting without unnecessary ticks.

### Prior findings

All 2 prior findings remain acknowledged — no changes since last review.

CI: All standard checks passing. Percy Visual Tests shows "waiting" status, which is the expected pending-human-approval state, not a failure.

---

- Retrieved prior bot reviews deterministically via the GitHub API
- Reviewed the pull request diff checking for correctness, design, architecture, testing, completeness, and adherence to DRY/SRP principles

</details>

docs/pages/klistbox.vue
```diff
@@ -0,0 +1,270 @@
...
+        <li>
+          Listboxes are often combined with searching or filtering, but that's not part of the
+          listbox pattern and use cases vary. If a frequent pattern emerges, it may motivate
+          dedicated search/filter components that integrate well with <code>KListbox</code>.
+        </li>
+        <li>
+          Currently KListbox doesn't support options reodering, so bear in mind that if the list changes
+          due to a filter, the filtered list should be preserve the order of the original list to avoid
+          unexpected keyboard navigation.
+        </li>
```
AlexVelezLl, line 28: While testing this on Kolibri, where we use Fuse for filtering, I noticed that it reorders the list based on an internal score, and that reordering doesn't trigger the KListbox internal options sorting, which causes some issues with keyboard navigation. We may want to dive a bit deeper depending on whether we want the listbox to support reordering scenarios. For now, in Kolibri, I just reordered the filtered list again because, in general, that is what makes more sense there (to preserve the alphabetical order). cc: @LianaHarris360 @radinamatic @Prashant-thakur77.

AlexVelezLl (approved):
Found a little non-blocking edge case, but just added a note on the docs to disclose this. LGTM! Thanks @MisRob!

docs/pages/klistbox.vue
MisRob: @AlexVelezLl thanks for catching that Alex, yes I think just having a note in the docs is good :)

Minor note ~ since there is already a paragraph about filtering (in the Overview), it may make sense to have that information together. However for the Overview section it could be too much info. Maybe it'd help to add "Filtering and pagination" sub-section to "Usage" and have all related information there

> Filtering and pagination
>
> KListbox supports dynamic option sets to allow for integration with filtering, pagination, and similar. Currently rendered options are referred to as visible options. Selection state of hidden options is preserved. 
>
> If the list changes due to a filter, ensure that the filtered list preserves the order of the original list to avoid unexpected keyboard navigation.

MisRob: Also I know the border between doc info and developer notes can be gray area, as for me I would consider the note you added important as an alert for anyone using `KListbox` (rather than just those developing it) so having information organized in this way should help a bit to notice it.

PR #1265, opened 2026-05-26
Fix KPageContainer noPadding on mobile devices

AlexVelezLl:
## Description
* Fixes KPageContainer `noPadding` prop not working on mobile devices.
  * Given that `.page-container.small` was being defined after `.page-container.no-padding`, the last class was not taking effect if the window was small, even if the `noPadding` prop was true, which is an incorrect behavior.

**Comment:** In general, we can also achieve this by setting the `style` attribute directly on the component tag, but it's still a bug in the KPageContainer implementation that needs to be fixed until we agree on a better implementation for these layout components.

#### Issue addressed

Found while testing https://github.com/learningequality/studio/pull/5924.

### Before/after screenshots

Before:
image

After:

image

## Changelog

  - **Description:** Fixes KPageContainer `noPadding` prop not working on mobile devices.
  - **Addresses:** https://github.com/learningequality/studio/pull/5924.
  - **Components:** KPageContainer.
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** -.

### DIFF
--- a/lib/KPageContainer.vue
+++ b/lib/KPageContainer.vue
@@ -66,13 +66,13 @@
     border-radius: 4px;
   }
 
-  .no-padding {
-    padding: 0;
-  }
-
   .page-container.small {
     padding: 8px 16px 16px;
     margin-top: 0;
+  }
+
+  .page-container.no-padding {
+    padding: 0;
   }
 
 </style>


akolson (approved):
Changes make sense to me! Thanks @AlexVelezLl

AlexVelezLl:
Thanks @akolson!

PR #1272, opened 2026-06-03
Implement the useMultiSelectList composable and its tests for KMultiSelect

### LINKED ISSUE #1267: Implement the useMultiSelectList composable and its tests for KMultiSelect

## Overview
This issue covers building useMultiSelectList.js to handle option normalization, search filtering, chip data shaping
Set up useMultiSelectList.js to handle option normalization (flat and tree formats), search filtering, chip data shaping, and remove/clear helpers
Write the respective test file for the component too consisting of all needed test cases.

### Acceptance Criteria
- Options are correctly normalized from both primitive and object formats.
- removeOption and clearAll emit correct events with accessible live region messages

Prashant-thakur77:
## Description
Implements the `useMultiSelectList` composable — the core data and filtering layer for the new `KMultiSelect` candidate component. This composable handles option normalization (supporting both primitive and object arrays), multi-field fuzzy search filtering, selected option resolution, chip removal, and clear-all actions.

#### Issue addressed
Part of the KMultiSelect implementation series #1267 

## Changelog

  - **Description:** Add `useMultiSelectList` composable — core data and filtering layer for the new `KMultiSelect` candidate component
  - **Addresses:** #1267 
  - **Components:** `KMultiSelect` (candidate)
  - **Breaking:**  no
  - **Impacts a11y:** yes 
  - **Guidance:** Emits polite screen reader announcements via `useKLiveRegion` when options are removed or cleared

## Steps to test

1. Run unit tests to verify the composable's functionality and if all cases are covered,some cases might be covered in the test file of main component directly
2) Verify all tests in the repository pass

Implemented useMultiSelectList as a Vue composable exporting:

- normalizedOptions: Computed mapping of primitives to { value, label } shapes, leaving object options unchanged.
- displayedOptions: Filtered options based on searchText, supporting matching on itemText and optional searchKeys (which can be strings or arrays of strings).
- selectedOptionsData: Resolves selected value array or single value back to original option objects to support chip and text rendering.
- removeOption / clearAll: Emits input changes to the parent while utilizing useKLiveRegion to send screen reader messages.

## Testing checklist

## Reviewer guidance

## Comments

Used Claude to verify the composable and helped to summarixe in order to write the pr description.

### DIFF
--- a/lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
+++ b/lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
@@ -93,7 +93,7 @@ export default function useMultiSelectList(props, emit, { searchText }) {
 
     emit('input', newValue);
     if (option) {
-      const text = getOptionText(option) || option.label || '';
+      const text = getOptionText(option) || '';
       sendPoliteMessage(`${text} removed, ${newValue.length} items selected`);
     }
   }
--- a/lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
+++ b/lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
@@ -2,17 +2,15 @@ import { computed } from 'vue';
 import useKLiveRegion from '../../../composables/useKLiveRegion';
 
 // Resolves a live-region announcement from the messages object.
-// Supports two formats per key:
-//   Static string:  messages.removed  = 'Selection removed'
-//   Translator fn:  messages.removed$ = ({ label, count }) => `${label} removed`
-// The $ suffix signals a translator function. Function variant takes priority.
-// If neither is provided the announcement is silently skipped (messages is optional).
+// All message values must be functions — even for static strings, wrap
+// them as: removed: () => 'Selection removed'.
+// If the key is missing or not a function, the
+// announcement is silently skipped (messages is optional).
 function getAnnouncement(messages, key, variables = {}) {
-  const fnKey = key + '$';
-  if (typeof messages[fnKey] === 'function') {
-    return messages[fnKey](variables);
+  if (typeof messages[key] === 'function') {
+    return messages[key](variables);
   }
-  return messages[key] || null;
+  return null;
 }
 
 export default function useMultiSelectList(props, emit, { searchText, messages = {} }) {
@@ -79,30 +77,42 @@ export default function useMultiSelectList(props, emit, { searchText, messages =
   });
 
   function removeOption(optionValue) {
+    const option = normalizedOptions.value.find(o => getOptionValue(o) === optionValue);
+    const label = option ? getOptionText(option) : null;
+
     if (!props.multiple) {
       emit('input', null);
-      const msg = getAnnouncement(messages, 'removed', {});
+      const msg = getAnnouncement(messages, 'removed', { label, count: 0 });
       if (msg) sendPoliteMessage(msg);
       return;
     }
 
-    const option = normalizedOptions.value.find(o => getOptionValue(o) === optionValue);
     const currentArray = Array.isArray(props.value) ? props.value : [];
     const newValue = currentArray.filter(v => v !== optionValue);
 
     emit('input', newValue);
-    if (option) {
-      const label = getOptionText(option);
-      const count = newValue.length;
-      const msg = getAnnouncement(messages, 'removed', { label, count });
-      if (msg) sendPoliteMessage(msg);
-    }
+    const count = newValue.length;
+    const msg = getAnnouncement(messages, 'removed', { label, count });
+    if (msg) sendPoliteMessage(msg);
   }
 
   function clearAll() {
-    emit('input', props.multiple ? [] : null);
-    const msg = getAnnouncement(messages, 'cleared', {});
-    if (msg) sendPoliteMessage(msg);
+    if (!props.multiple) {
+      // Single mode: resolve the label of the currently selected option
+      // so the translator can announce e.g. "English cleared".
+      const option = normalizedOptions.value.find(o => getOptionValue(o) === props.value);
+      const label = option ? getOptionText(option) : null;
+      emit('input', null);
+      const msg = getAnnouncement(messages, 'cleared', { label });
+      if (msg) sendPoliteMessage(msg);
+    } else {
+      // Multi mode: pass count of items being cleared so the translator
+      // can announce e.g. "3 selections cleared" or "All selections cleared".
+      const count = Array.isArray(props.value) ? props.value.length : 0;
+      emit('input', []);
+      const msg = getAnnouncement(messages, 'cleared', { count });
+      if (msg) sendPoliteMessage(msg);
+    }
   }
 
   return {
--- a/lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
+++ b/lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
@@ -1,8 +1,6 @@
 import { computed } from 'vue';
-import useKLiveRegion from '../../../composables/useKLiveRegion';
 
-export default function useMultiSelectList(props, emit, { searchText }) {
-  const { sendPoliteMessage } = useKLiveRegion();
+export default function useMultiSelectList(props, emit, { searchText, announce }) {
 
   const normalizedOptions = computed(() => {
     return props.options
@@ -67,7 +65,9 @@ export default function useMultiSelectList(props, emit, { searchText }) {
   function removeOption(optionValue) {
     if (!props.multiple) {
       emit('input', null);
-      sendPoliteMessage('Selection removed');
+      if (typeof announce === 'function') {
+        announce({ type: 'removed' });
+      }
       return;
     }
 
@@ -77,14 +77,18 @@ export default function useMultiSelectList(props, emit, { searchText }) {
 
     emit('input', newValue);
     if (option) {
-      const text = getOptionText(option) || '';
-      sendPoliteMessage(`${text} removed, ${newValue.length} items selected`);
+      const label = getOptionText(option) || '';
+      if (typeof announce === 'function') {
+        announce({ type: 'removed', label, count: newValue.length });
+      }
     }
   }
 
   function clearAll() {
     emit('input', props.multiple ? [] : null);
-    sendPoliteMessage('All selections cleared');
+    if (typeof announce === 'function') {
+      announce({ type: 'cleared' });
+    }
   }
 
   return {
--- a/lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
+++ b/lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
@@ -1,6 +1,20 @@
 import { computed } from 'vue';
 import useKLiveRegion from '../../../composables/useKLiveRegion';
 
+// Resolves a live-region announcement from the messages object.
+// Supports two formats per key:
+//   Static string:  messages.removed  = 'Selection removed'
+//   Translator fn:  messages.removed$ = ({ label, count }) => `${label} removed`
+// The $ suffix signals a translator function. Function variant takes priority.
+// If neither is provided the announcement is silently skipped (messages is optional).
+function getAnnouncement(messages, key, variables = {}) {
+  const fnKey = key + '$';
+  if (typeof messages[fnKey] === 'function') {
+    return messages[fnKey](variables);
+  }
+  return messages[key] || null;
+}
+
 export default function useMultiSelectList(props, emit, { searchText, messages = {} }) {
   const { sendPoliteMessage } = useKLiveRegion();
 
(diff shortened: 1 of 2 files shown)


lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
```diff
@@ -0,0 +1,97 @@
...
+    return props.options.map(option => {
+      if (typeof option !== 'object' || option === null) {
+        return { value: option, label: String(option) };
+      }
+      return option;
+    });
+  });
+
+  function getOptionText(option) {
+    return option[props.itemText];
```
LianaHarris360, line 31: Looking at `LanguageDropdown.vue` in Studio, which also uses `VAutocomplete`, it passes `:itemText="languageText"`, with `languageText` as a method. 

Even though the usage in `CategoryOptions` only needs to handle strings, what would happen if someone passed a function as `itemText`? Is there a change we could have in `getOptionText` so that it could handle both cases?

lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
```diff
@@ -0,0 +1,97 @@
...
+
+  function getOptionValue(option) {
+    return option[props.itemValue];
+  }
+
+  const displayedOptions = computed(() => {
+    const query = searchText.value ? searchText.value.toLowerCase() : '';
+    const filtered = !query
+      ? normalizedOptions.value
+      : normalizedOptions.value.filter(option => {
```
LianaHarris360, line 42: I see how `normalizedOptions` handles primitive and object arrays, but `useMultiSelectList` should handle tree-shaped data as well, right?

E.g. if an option has a `filterText` field (such as "Mathematics > Arithmetic"), should `displayedOptions` search against that field too, or only against `itemText`?

lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
```diff
@@ -0,0 +1,97 @@
+import { computed } from 'vue';
+import useKLiveRegion from '../../../composables/useKLiveRegion';
+
+export default function useMultiSelectList(props, emit, { searchText }) {
+  const { sendPoliteMessage } = useKLiveRegion();
+
+  const normalizedOptions = computed(() => {
+    return props.options.map(option => {
+      if (typeof option !== 'object' || option === null) {
```
LianaHarris360: Because `typeof null` returns `object`, if someone passed in `[null, 'Apple']`, the code would create `{ value: null, label: 'null' }`. Based on our use cases, should we filter out `null` or `undefined` entries from the options array before processing them? Or wrap them the same way other primitives are wrapped?

lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
```diff
@@ -0,0 +1,97 @@
...
+    const currentArray = Array.isArray(props.value) ? props.value : [];
+    return currentArray
+      .map(v => normalizedOptions.value.find(o => getOptionValue(o) === v))
+      .filter(Boolean);
+  });
+
+  function removeOption(optionValue) {
+    if (!props.multiple) {
+      emit('input', null);
+      sendPoliteMessage('Selection removed');
```
LianaHarris360: Since Studio and Kolibri are translated into other languages, we avoid using hard coded strings directly. We have a few options we could go with:
1. The composable could emit a removed or cleared event with payload data that the parent component turns into a translated sentence. 
2. Accept a function like `announce({ type, label, count })`, letting the parent decide how it should be phrased.
3. Return objects instead of strings, for example `{ kind: "selection-removed", label, selectedCount }`.

LianaHarris360 (changes requested):

lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
Prashant-thakur77: This would currently be handled through the `searchKeys` prop of `KMultiSelect`. For example:

```html
<KMultiSelect
  :options="allCategories"
  :searchKeys="['filterText']"
/>
```

Since `displayedOptions` already iterates through the keys provided in `searchKeys` and checks their values against the search query, an option with `filterText: "Mathematics > Arithmetic"` would be matched when searching for either "Mathematics" or "Arithmetic".

The reason I kept this behavior explicit rather than automatically searching a `filterText` field is that this is intended to be a generic component. Different consumers may use different field names (`filterText`, `path`, `breadcrumb`, `fullLabel`, etc.), so I didn't want to introduce a reserved field name or special-case logic for a specific use case.

lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
Prashant-thakur77: I think we should filter them out.
The component already uses null internally to represent "nothing selected" in single-select mode (clearAll and removeOption emit null), so allowing an option with value: null would create an ambiguity between a valid option and an unselected state.
Also, none of our current use cases (primitive arrays, object arrays, or tree arrays) intentionally include null or undefined options, so filtering them out is the safer behavior.
I'll update normalizedOptions to filter out null/undefined values before mapping and add a test to cover it.

lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
Prashant-thakur77: Thanks, that makes sense.
i think we can go ahead with the option 2 but just in case, I checked how similar accessibility announcements are handled elsewhere in KDS. Both `KListbox` and `KSelect` avoid hardcoded strings by receiving translated text from the parent:
* `KListbox` uses a `messages` object prop (`messages.optionDeselected`, `messages.allOptionsSelected`, etc.).
* `KSelect` uses individual translated string props such as `clearText` and `noResultsText`.

to stay consistent do you think we should have a `announcements` (or `messages`) object prop on `KMultiSelect`, pass it into the composable, and use those translated strings instead of hardcoded text.
Should we go with that approach, or should we  keep the dynamic label/count information via Option 2?

lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
LianaHarris360: Ah okay, I understand. You make a good point about keeping the component more generic and not introducing a reserved field name, plus there's a good chance that automatically checking `filterText` whenever it exists might make things more difficult to debug and document later. Let's keep the approach you have planned.

lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
Prashant-thakur77: I initially looked into this because of LanguageDropdown.vue, where VAutocomplete is passed a languageText method via :itemText="languageText".
My first thought was that we could support the same pattern here. However, after digging deeper, I found that function-type props don't appear to be used anywhere else in KDS .
I think we can achieve the same result more cleanly by normalizing the options in the parent before passing them to `KMultiSelect`.
For example, instead of passing a function, the parent can precompute the display text:

```js
formattedLanguageDropdown() {
  return this.allLanguages.map(opt => ({
    ...opt,
    displayLabel: `${opt.native_name.split(',')[0].trim()} (${opt.id})`,
  }));
}
```
and then pass:

```html
<KMultiSelect
  :options="formattedLanguageDropdown"
  itemText="displayLabel"
/>
```
The end result is identical the user still sees `"हिन्दी (hi)"` but `itemText` remains a simple string key and `getOptionText` can stay as:

```js
function getOptionText(option) {
  return option[props.itemText];
}
```

what are your thoughts about it as i might be overcomplicating the things here?

LianaHarris360: I don't think it's overcomplicating things at all, normalizing the options in the parent could definitely be done this way instead and would keep the code here more consistent/predictable. This will just be an intentional decision (for me) to keep note of.

lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
LianaHarris360: I think either approach works, but lean towards the `announcements` or `messages` object prop, passing it into the composable, and using those translated strings instead.

Prashant-thakur77: One thing I realized after looking into it a bit more: Option 2 solves the composable side, but KMultiSelect would still need a way to turn { type, label, count } into a translated message.
That means we'd still need an announcements API on KMultiSelect, and I think there are two directions we could take:

Static announcements/messages (similar to KListbox)  this keeps things consistent with existing KDS patterns, but we'd probably simplify messages to something like "Selection removed" and lose the dynamic label/count information.
Dynamic announcements : announcements prop would contain functions that receive { label, count } and return the translated message at runtime. This preserves the richer screen reader feedback, but introduces function values in props, which we know isnt done in KDS normally.
What do you think? 🙂

Prashant-thakur77: > I think either approach works, but lean towards the `announcements` or `messages` object prop, passing it into the composable, and using those translated strings instead.

yes ,i agree with this.

Prashant-thakur77: > One thing I realized after looking into it a bit more: Option 2 solves the composable side, but KMultiSelect would still need a way to turn { type, label, count } into a translated message. That means we'd still need an announcements API on KMultiSelect, and I think there are two directions we could take:
> 
> Static announcements/messages (similar to KListbox) this keeps things consistent with existing KDS patterns, but we'd probably simplify messages to something like "Selection removed" and lose the dynamic label/count information. Dynamic announcements : announcements prop would contain functions that receive { label, count } and return the translated message at runtime. This preserves the richer screen reader feedback, but introduces function values in props, which we know isnt done in KDS normally. What do you think? 🙂

@LianaHarris360 this is just something i found,we would need a prop anyway so i think this is resolved,but One thing I'm still a bit unclear on though: if announcements prop is just an object of strings(like in klistbox ), I don't think we can include dynamic values like the label or selected count. It seems like we'd need function values in the prop to generate those messages at runtime.
static: "Selection removed"
Dynamic: "Apple removed, 3 items selected"
And as we discussed above that we tend to not use functions,so what should we do in this case 🙂

lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
Prashant-thakur77: done

lib/candidate/multiselect/KMultiSelect/useMultiSelectList.js
LianaHarris360: The consensus is that function props are indeed appropriate for this use case. @AlexVelezLl  suggestion is the approach we can move forward with: use a single `messages` prop that can accept both static strings and translator functions.

To make the API clear, we'll differentiate translator functions from static strings by appending a $ at the end of the key, to signal that the key expects a function rather than a string, for example:
```
messages: {
  messageLabel: 'Message',
  selectedLabel$: selectedLabel$, // destructured from a translator, document what values it should receive
}
```
The messages can be optional (and the announcement skipped if not provided) or required (and logged as a warning when missing), depending on the context. Since there's no existing example beyond the current `KListbox` documentation, if you have doubts about the exact implementation, we can work on it together through pair programming 🙂

Prashant-thakur77: Thanks @LianaHarris360! I've implemented the `$` suffix approach based on your guidance and Alex suggestion.

* Added a `getAnnouncement` helper that checks `messages[key + '$']` first (translator function), then falls back to `messages[key]` (static string).
* Supported keys:

  * `removed` / `removed$` → `removed$` receives `{ label, count }`
  * `cleared` / `cleared$`
* If neither variant is provided, the announcement is skipped.

For now I went with optional messages rather than requiring them. A number of our current Studio usages are `multiple=false` without a clear button, so `removeOption` and `clearAll` are never triggered there. Making the messages required would mean passing values that are never used, while multi-select usages that need announcements can provide the relevant keys.

I've also added tests for static and dynamic announcements
Happy to adjust it according to the review further.

LianaHarris360: @Prashant-thakur77 Got it, thanks for updating this! I agree with your approach to make the messages optional given the Studio usages. Separate question: when the translator function is provided in single mode or `clearAll`, shouldn't it receive useful context like the previous label or count? Right now in `removeOption`, it gets `{}`, which means a component like `LanguageDropdown` couldn't announce "English removed" if it wanted to.

Prashant-thakur77: Resolved this by updating the removeoption and the clearall to handle single select with label and count too.

Prashant-thakur77:
@LianaHarris360 I've also updated the messages to be functions-only  as we discussed.Misha also mentioned to update the klistbox implementation,I think i will raise that pr after this one is verified to be the approach we are choosing:)

LianaHarris360 (approved):

PR #1273, opened 2026-06-03
Implement the useMultiSelectCascade composable and its tests for KMultiSelect

### LINKED ISSUE #1268: Implement the useMultiSelectCascade composable and its tests for KMultiSelect

## Overview

This issue covers building useMultiSelectCascade.js to manage tree-aware selection: downward cascade (deselecting a parent deselects all descendants), upward cascade (selecting a child selects all ancestors), and indeterminateValues for partial selection states.
Write the respective test file too consisting of all needed test cases.
### Acceptance Criteria
- Downward and upward cascade work correctly at any tree depth.
- indeterminateValues correctly reflects partial descendant selection.
- Tests cover all cascade scenarios following KDS isolated composable testing patterns.

Prashant-thakur77:
## Description
Adds the `useMultiSelectCascade` Vue composable to manage downward cascade selection logic, intermediate checkbox states, and accessible announcements for the upcoming `KMultiSelect` component.

#### Issue addressed

Addresses #1268 

## Changelog

  - **Description:** Add `useMultiSelectcascade` composable — core data and filtering layer for the new `KMultiSelect` candidate component
  - **Addresses:** #1268 
  - **Components:** `KMultiSelect` (candidate)
  - **Breaking:**  no
  - **Impacts a11y:** yes 
  - **Guidance:** Emits polite screen reader announcements via `useKLiveRegion` when options are removed or cleared

## Steps to test
All tests should be present and pass accordingly.
## Testing checklist

## Reviewer guidance

### DIFF
--- a/lib/candidate/multiselect/KMultiSelect/__tests__/useMultiSelectCascade.spec.js
+++ b/lib/candidate/multiselect/KMultiSelect/__tests__/useMultiSelectCascade.spec.js
@@ -1,0 +1,57 @@
+import { ref, reactive } from 'vue';
+import useMultiSelectCascade from '../useMultiSelectCascade';
+
+describe('useMultiSelectCascade', () => {
+  let props, emit, normalizedOptions, searchText;
+
+  beforeEach(() => {
+    props = reactive({
+      options: [
+        { value: 'parent', level: 1 },
+        { value: 'child1', level: 2 },
+        { value: 'child2', level: 2 },
+      ],
+      value: [],
+      multiple: true,
+    });
+    emit = jest.fn();
+    normalizedOptions = ref(props.options);
+    searchText = ref('');
+  });
+
+  it('computes indeterminate state when some but not all children are selected', () => {
+    props.value = ['child1'];
+    const { indeterminateValues } = useMultiSelectCascade(props, emit, {
+      normalizedOptions,
+      getOptionValue: o => o.value,
+      searchText,
+      closeDropdown: jest.fn(),
+    });
+
+    expect(Array.from(indeterminateValues.value)).toEqual(['parent']);
+  });
+
+  it('does not set indeterminate when all children are selected', () => {
+    props.value = ['child1', 'child2', 'parent'];
+    const { indeterminateValues } = useMultiSelectCascade(props, emit, {
+      normalizedOptions,
+      getOptionValue: o => o.value,
+      searchText,
+      closeDropdown: jest.fn(),
+    });
+
+    expect(Array.from(indeterminateValues.value)).toEqual([]);
+  });
+
+  it('cascades correctly on listbox input', () => {
+    const { onListboxInput } = useMultiSelectCascade(props, emit, {
+      normalizedOptions,
+      getOptionValue: o => o.value,
+      searchText,
+      closeDropdown: jest.fn(),
+    });
+
+    onListboxInput(['child1']);
+    expect(emit).toHaveBeenCalledWith('input', ['child1', 'parent']);
+  });
+});
--- a/lib/candidate/multiselect/KMultiSelect/__tests__/useMultiSelectCascade.spec.js
+++ b/lib/candidate/multiselect/KMultiSelect/__tests__/useMultiSelectCascade.spec.js
@@ -2,82 +2,87 @@ import { ref, reactive } from 'vue';
 import useMultiSelectCascade from '../useMultiSelectCascade';
 
 describe('useMultiSelectCascade', () => {
-  let props, emit, normalizedOptions, searchText;
+  let props, emit, normalizedOptions;
 
   beforeEach(() => {
     props = reactive({
       options: [
-        { value: 'parent', level: 1 },
-        { value: 'child1', level: 2 },
-        { value: 'child2', level: 2 },
+        { value: 'parent', label: 'Parent', level: 1 },
+        { value: 'child1', label: 'Child One', level: 2 },
+        { value: 'child2', label: 'Child Two', level: 2 },
       ],
       value: [],
       multiple: true,
     });
     emit = jest.fn();
     normalizedOptions = ref(props.options);
-    searchText = ref('');
   });
+
+  function makeCascade(overrides = {}) {
+    return useMultiSelectCascade(props, emit, {
+      normalizedOptions,
+      getOptionValue: o => o.value,
+      getOptionText: o => o.label,
+      setSearchText: jest.fn(),
+      closeDropdown: jest.fn(),
+      defaultRemoveOption: jest.fn(),
+      ...overrides,
+    });
+  }
 
   it('computes indeterminate state when some but not all children are selected', () => {
     props.value = ['child1'];
-    const { indeterminateValues } = useMultiSelectCascade(props, emit, {
-      normalizedOptions,
-      getOptionValue: o => o.value,
-      searchText,
-      closeDropdown: jest.fn(),
-    });
-
+    const { indeterminateValues } = makeCascade();
     expect(Array.from(indeterminateValues.value)).toEqual(['parent']);
   });
 
-  it('does not set indeterminate when all children are selected', () => {
+  it('does not set indeterminate when all children and parent are explicitly selected', () => {
     props.value = ['child1', 'child2', 'parent'];
-    const { indeterminateValues } = useMultiSelectCascade(props, emit, {
-      normalizedOptions,
-      getOptionValue: o => o.value,
-      searchText,
-      closeDropdown: jest.fn(),
-    });
-
+    const { indeterminateValues } = makeCascade();
     expect(Array.from(indeterminateValues.value)).toEqual([]);
   });
 
-  it('cascades correctly on listbox input', () => {
-    const { onListboxInput } = useMultiSelectCascade(props, emit, {
-      normalizedOptions,
-      getOptionValue: o => o.value,
-      searchText,
-      closeDropdown: jest.fn(),
-    });
+  it('marks parent as indeterminate when all children are selected but parent is not', () => {
+    props.value = ['child1', 'child2'];
+    const { indeterminateValues } = makeCascade();
+    expect(indeterminateValues.value.has('parent')).toBe(true);
+  });
 
+  it('indeterminateValues returns empty set in single mode', () => {
+    props.multiple = false;
+    props.value = 'child1';
+    const { indeterminateValues } = makeCascade();
+    expect(indeterminateValues.value.size).toBe(0);
+  });
+
+  it('emits only the child when a child is selected without its parent', () => {
+    const { onListboxInput } = makeCascade();
     onListboxInput(['child1']);
-    expect(emit).toHaveBeenCalledWith('input', ['child1', 'parent']);
+    expect(emit).toHaveBeenCalledWith('input', ['child1']);
   });
 
   it('removes descendants when a parent is deselected', () => {
     props.value = ['parent', 'child1', 'child2'];
-    const { onListboxInput } = useMultiSelectCascade(props, emit, {
-      normalizedOptions,
-      getOptionValue: o => o.value,
-      searchText,
-      closeDropdown: jest.fn(),
-    });
-
+    const { onListboxInput } = makeCascade();
     onListboxInput(['child1', 'child2']);
     expect(emit).toHaveBeenCalledWith('input', []);
+  });
+
+  it('emits the new values unchanged for a flat option list', () => {
+    normalizedOptions.value = [
+      { value: 'a', label: 'A' },
+      { value: 'b', label: 'B' },
+    ];
+    props.value = ['a'];
+    const { onListboxInput } = makeCascade();
+    onListboxInput(['a', 'b']);
+    expect(emit).toHaveBeenCalledWith('input', ['a', 'b']);
   });
 
   it('emits a scalar value and closes the dropdown in single mode', () => {
     props.multiple = false;
     const closeDropdown = jest.fn();
-    const { onListboxInput } = useMultiSelectCascade(props, emit, {
-      normalizedOptions,
-      getOptionValue: o => o.value,
-      searchText,
-      closeDropdown,
-    });
-
+    const { onListboxInput } = makeCascade({ closeDropdown });
     onListboxInput(['child1']);
     expect(emit).toHaveBeenCalledWith('input', 'child1');
     expect(closeDropdown).toHaveBeenCalled();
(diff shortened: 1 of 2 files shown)


lib/candidate/multiselect/KMultiSelect/__tests__/useMultiSelectCascade.spec.js
```diff
@@ -0,0 +1,161 @@
...
+    const { indeterminateValues } = makeCascade();
+    expect(indeterminateValues.value.size).toBe(0);
+  });
+
+  it('removeOption announces singular item when one item remains', () => {
+    props.value = ['parent', 'child1'];
+    const { removeOption } = makeCascade();
+    removeOption('child1');
+    expect(getPoliteRegion()).toHaveTextContent('Child One removed, 1 item selected');
+  });
```
LianaHarris360, line 218: So far, the tests only cover one level (the parent + child), but categories in Studio can go 3+ levels deep. Can you add a test to include grandparent + parent + child, showing what happens with the parent and child if the grandparent is deselected?

lib/candidate/multiselect/KMultiSelect/useMultiSelectCascade.js
```diff
@@ -0,0 +1,131 @@
...
+  const isHierarchical = computed(() =>
+    normalizedOptions.value.some(o => typeof o.level === 'number' && o.level > 0),
+  );
+
+  function getDescendants(value, flatOptions) {
+    const idx = optionIndexMap.value.get(value);
+    if (idx === undefined) return [];
+    const nodeLevel = flatOptions[idx].level ?? 0;
+    const result = [];
+    for (let i = idx + 1; i < flatOptions.length; i++) {
```
LianaHarris360, line 46: Since `getDescendants()` walks the flat array starting at `idx + 1` and stops when it hits a sibling or higher level, what happens if the options are given in the wrong order? E.g:
```
      options: [
        { value: 'child1', label: 'Child One', level: 2 },
        { value: 'parent', label: 'Parent', level: 1 },
        { value: 'child2', label: 'Child Two', level: 2 },
      ],
```
Perhaps we should validate the order? Categories in Studio are pre-sorted through `getSortedCategories()` so at the very least, the assumption should be documented with a comment like "Assumes options are in depth-first tree order" for future debugging.

lib/candidate/multiselect/KMultiSelect/useMultiSelectCascade.js
```diff
@@ -0,0 +1,131 @@
...
+
+  function onListboxInput(newValues, { silent = false } = {}) {
+    if (!props.multiple) {
+      const result = newValues.length > 0 ? newValues[newValues.length - 1] : null;
+      emit('input', result);
+
+      const prev = Array.isArray(props.value) ? props.value : props.value ? [props.value] : [];
+      const added = result && !prev.includes(result);
+      if (!silent && added) {
+        sendPoliteMessage('Selected');
```
LianaHarris360: Here, to avoid hardcoded text, I wonder if announcing the selected value could be useful instead? Or duplicating the decision you move forward with for `useMultiSelectList.js`

lib/candidate/multiselect/KMultiSelect/__tests__/useMultiSelectCascade.spec.js
```diff
@@ -0,0 +1,161 @@
...
+    props.multiple = false;
+    props.value = 'child1';
+    const { indeterminateValues } = makeCascade();
+    expect(indeterminateValues.value.size).toBe(0);
+  });
+
+  it('emits only the child when a child is selected without its parent', () => {
+    const { onListboxInput } = makeCascade();
+    onListboxInput(['child1']);
+    expect(emit).toHaveBeenCalledWith('input', ['child1']);
```
LianaHarris360, line 70: Might be useful to add a comment and/or `expect` call to confirm that the parent is not selected, but is indeterminate in this case?

LianaHarris360 (changes requested):
Regarding the `autoselectparent` boolean to handle both implicit and explicit selection of the parent, building the composable with that hook in mind now (even if the prop defaults to false) might be cleaner than adding it later.

lib/candidate/multiselect/KMultiSelect/__tests__/useMultiSelectCascade.spec.js
Prashant-thakur77: done

lib/candidate/multiselect/KMultiSelect/useMultiSelectCascade.js
Prashant-thakur77: yes,going forward with the decision made for the useMultiSelectList.js.

lib/candidate/multiselect/KMultiSelect/__tests__/useMultiSelectCascade.spec.js
Prashant-thakur77: done

Prashant-thakur77:
@LianaHarris360 I have done the changes that you asked,also added the autoselectparent functionality that we discussed with misha and Blaine ,just to state it it was that 

When false: Selecting a child leaves the parent unselected (turning the parent indeterminate).
When true: Selecting a child automatically forces all of its parent ancestors to become selected as well. and the default is false:).
Also the question about the messages/announcement prop I asked in the usemultiselectlist composable about the dynamic and the static announcement that i have here too. Currently i have used the exact same approach here that I did in the useMultiSelectList composable. I am passing down a static string object (similar to how it's handled in KListbox) rather than dynamic strings, because dynamic strings would require passing down an object of functions for translations.

lib/candidate/multiselect/KMultiSelect/useMultiSelectCascade.js
Prashant-thakur77: added the comment:)

Prashant-thakur77:
I have updated this branch too according to our discussion in the usemultiselectlist composable about message announcement dynamic and static.

lib/candidate/multiselect/KMultiSelect/useMultiSelectCascade.js
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  function getAncestors(value, flatOptions) {
+    const idx = optionIndexMap.value.get(value);
+    if (idx === undefined) return [];
+    const nodeLevel = flatOptions[idx].level ?? 0;
+    const result = [];
+    for (let i = idx - 1; i >= 0; i--) {
+      const l = flatOptions[i].level ?? 0;
+      if (l < nodeLevel) {
```
rtibblesbot: **blocking:** `getAncestors` collects wrong ancestors in multi-branch trees. The loop compares every visited node's level against the original `nodeLevel` and never narrows the target as ancestors are found. In a tree with sibling subtrees this includes unrelated nodes.

Concrete example:
```
A (level 1)
  B (level 2)   ← sibling branch, unrelated to D
C (level 1)
  D (level 2)   ← target: getAncestors('D', …)
```
Walking back from D: hits C (l=1 < 2 ✓), then B (l=2 — skip), then A (l=1 < 2 ✓ — **wrongly added**). A is not an ancestor of D.

Fix: narrow the target level each time a real ancestor is found:
```js
function getAncestors(value, flatOptions) {
  const idx = optionIndexMap.value.get(value);
  if (idx === undefined) return [];
  let currentLevel = flatOptions[idx].level ?? 0;
  const result = [];
  for (let i = idx - 1; i >= 0; i--) {
    const l = flatOptions[i].level ?? 0;
    if (l < currentLevel) {
      result.push(getOptionValue(flatOptions[i]));
      currentLevel = l;   // next ancestor must be above this one
      if (currentLevel <= 0) break;
    }
  }
  return result;
}
```

lib/candidate/multiselect/KMultiSelect/useMultiSelectCascade.js
```diff
@@ -0,0 +1,192 @@
...
+        result.push(getOptionValue(flatOptions[i]));
+        if (l === 0) break;
```
rtibblesbot: **nitpick:** This `break` is unreachable under the current level convention. `isHierarchical` requires `level > 0`, so all hierarchical nodes have `level >= 1`; flat nodes (`level ?? 0`) are handled in the non-hierarchical branch before `getAncestors` is ever called. No node will have `l === 0` here. The fixed version of `getAncestors` above naturally reintroduces this guard as `if (currentLevel <= 0) break`, where it becomes meaningful.

lib/candidate/multiselect/KMultiSelect/__tests__/useMultiSelectCascade.spec.js
```diff
@@ -0,0 +1,246 @@
...
+  it('removeOption sends the messages.removed string to the live region', () => {
+    props.value = ['parent', 'child1'];
+    const { removeOption } = makeCascade({
+      messages: { removed: 'Selection removed' },
+    });
+    removeOption('child1');
+    expect(getPoliteRegion()).toHaveTextContent('Selection removed');
+  });
+
+  it('autoSelectParent: selecting a child also includes its parent in emitted values', () => {
```
rtibblesbot, line 192: **blocking:** Both `autoSelectParent` tests use a single-branch fixture (`parent → [child1, child2]` or a linear chain), which never exposes the `getAncestors` bug above. A tree with sibling subtrees is the missing case:

```js
it('autoSelectParent: selecting a child only selects its own ancestor, not a sibling-branch parent', () => {
  normalizedOptions.value = [
    { value: 'a', label: 'A', level: 1 },
    { value: 'b', label: 'B', level: 2 },
    { value: 'c', label: 'C', level: 1 },
    { value: 'd', label: 'D', level: 2 },
  ];
  props.autoSelectParent = true;
  const { onListboxInput } = makeCascade();
  onListboxInput(['d']);
  // Should emit ['d', 'c'] — NOT ['d', 'c', 'a']
  expect(emit).toHaveBeenCalledWith('input', ['d', 'c']);
});
```

This test currently fails against the unpatched `getAncestors`, making it the right regression anchor.

lib/candidate/multiselect/KMultiSelect/__tests__/useMultiSelectCascade.spec.js
```diff
@@ -0,0 +1,246 @@
+import { ref, reactive } from 'vue';
+import useMultiSelectCascade from '../useMultiSelectCascade';
+
+const getPoliteRegion = () => document.querySelector('#k-live-region [aria-live="polite"]');
```
rtibblesbot: **suggestion:** Several tests assert live-region content by querying `#k-live-region [aria-live="polite"]` directly. This couples the spec to `useKLiveRegion`'s internal DOM structure — if that composable ever changes its ID or aria-live placement, these assertions will break for reasons entirely unrelated to `useMultiSelectCascade`.

The standard unit-test approach is to mock the module boundary and assert the public call:
```js
const mockSendPoliteMessage = jest.fn();
jest.mock('../../../composables/useKLiveRegion', () => ({
  __esModule: true,
  default: () => ({ sendPoliteMessage: mockSendPoliteMessage }),
}));
// …
expect(mockSendPoliteMessage).toHaveBeenCalledWith('Child One removed, 2 selected');
```
This also removes the implicit requirement that `useKLiveRegion` initialise its DOM node before assertions run.

lib/candidate/multiselect/KMultiSelect/useMultiSelectCascade.js
```diff
@@ -0,0 +1,192 @@
...
+      const firstOpt = firstIdx !== undefined ? normalizedOptions.value[firstIdx] : null;
+      const label = firstOpt ? getOptionText(firstOpt) : '';
+      const msg = getAnnouncement(messages, 'selected', { label, count: result.length });
+      if (msg) sendPoliteMessage(msg);
+    }
+
+    return result;
+  }
+
+  function cascadeRemoveOption(optionValue) {
```
rtibblesbot, line 162: **praise:** Clean double-announcement prevention here — `cascadeRemoveOption` calls `onListboxInput` with `{ silent: true }` and then fires its own `'removed'` announcement with the final post-cascade count. This avoids a spurious "N selected" announcement when removing a chip also cascades off descendants, and the pattern is easy to reason about and well-exercised by the `removeOption sends only the removed message` test.

rtibblesbot (changes requested):
One correctness bug in `getAncestors` causes it to collect wrong ancestors in trees with sibling branches, and the `autoSelectParent` tests only exercise single-branch trees so neither the bug nor the gap is currently visible. Both need to be fixed before merge.

CI status: pending.

- 2 **blocking** findings (inline)
- 1 **suggestion** (inline)
- 1 **nitpick** (inline)
- 1 **praise** (inline)

---

</details>

lib/candidate/multiselect/KMultiSelect/useMultiSelectCascade.js
Prashant-thakur77: DONE

lib/candidate/multiselect/KMultiSelect/useMultiSelectCascade.js
Prashant-thakur77: Updated

lib/candidate/multiselect/KMultiSelect/__tests__/useMultiSelectCascade.spec.js
Prashant-thakur77: done

lib/candidate/multiselect/KMultiSelect/__tests__/useMultiSelectCascade.spec.js
Prashant-thakur77: Updated

Prashant-thakur77:
@LianaHarris360 I have here also updated the code related to message prop as a function here too and also did the updates suggested by the ribblebot:)

lib/candidate/multiselect/KMultiSelect/__tests__/useMultiSelectCascade.spec.js
Prashant-thakur77: @LianaHarris360 should i use this approach in the useMultiSelectlist composable too?

rtibblesbot: Yes, the same reasoning applies to `useMultiSelectList` (or any composable that calls `sendPoliteMessage`): if those tests query the live-region DOM directly, they're tied to `useKLiveRegion`'s internals just as much as the cascade tests were. Mocking the module there keeps each spec focused on its own unit.

LianaHarris360: Yes, I agree. Can you apply these same changes to the test for `useMultiSelectList` as well?

rtibblesbot: That's noted — I've already flagged it in [my reply above](https://github.com/learningequality/kolibri-design-system/pull/1273#discussion_r3423847649): the same `jest.mock` pattern should be applied to the `useMultiSelectList` tests wherever they query the live-region DOM directly. @Prashant-thakur77 can apply it there alongside the cascade tests.

Prashant-thakur77: > Yes, I agree. Can you apply these same changes to the test for `useMultiSelectList` as well?

done

LianaHarris360 (approved):

Issue #1290, opened 2026-06-22
KListbox: Support functions (in addition to strings) for the messages prop


Prashant-thakur77:
<img height="20px" src="https://i.imgur.com/c7hUeb5.jpeg">

❌ **This issue is not open for contribution. Visit <a href="https://learningequality.org/contributing-to-our-open-code-base/" target="_blank">Contributing guidelines</a>** to learn about the contributing process and how to find suitable issues.

<img height="20px" src="https://i.imgur.com/c7hUeb5.jpeg">

## Overview

KListbox previously only accepted plain strings for its messages prop. This change updates the prop to also accept functions, making it compatible with KMultiSelect (which passes functions) without breaking existing consumers like SelectableList.vue in Kolibri.Later on this will be completly updated to use function message prop for simplicity.

### The Change
Update KListbox messages prop to accept both strings and functions:

### Out of Scope
Out of Scope
- Migrating existing string-based usages to functions (tracked separately)
- Any changes to SelectableList.vue in Kolibri

### References
Sub issue of #1259

LianaHarris360:
Closed by #1291

PR #1291, opened 2026-06-22
KListbox: Support functions (in addition to strings) for the messages prop

Prashant-thakur77:
## Description
Updates KListbox to accept both strings and functions for its messages prop, enabling KMultiSelect to pass its messages as functions without breaking existing consumers that still pass plain strings.

#### Issue addressed

Addresses #1290 

## Changelog

  - **Description:** KListbox messages prop now accepts both plain strings and functions. A getMessage() helper resolves the value at runtime by calling the function if needed, or returning the string directly.
  - **Addresses:** #1290 
  - **Components:** Klistbox
  - **Breaking:** no
  - **Impacts a11y:** yes
  - **Guidance:** No changes required for existing consumers passing plain strings

## Steps to test

check if the desired functionality is achieved.

## Testing checklist

## Reviewer guidance

### DIFF
--- a/lib/candidate/listbox/KListbox/index.vue
+++ b/lib/candidate/listbox/KListbox/index.vue
@@ -5,7 +5,7 @@
       :id="ariaDescribedById"
       class="visuallyhidden"
     >
-      {{ messages.clickable }}
+      {{ getMessage('clickable') }}
     </p>
 
     <!--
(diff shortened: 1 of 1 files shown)


Prashant-thakur77:
@LianaHarris360 currenlty i have only updated the main index of Klistbox to have the functionality to have both string and function as message prop ,I haven't updated the docs and the examples,so they are using the strings as before.do tell if they need to be updated for this pr or not:)Would be happy to do update them if needed.

Prashant-thakur77:
And i think we should merge this pr before the last pr that i made today as i will then rebase these new changes to that pr for simplicity:)

LianaHarris360 (approved):
I tested this change locally, running Kolibri with the changes in this PR and can confirm that there is no behavior change for the current implementation of `KListbox` in the Kolibri component `SelectableList` and the messages prop correctly supports string functions. Thank you for updating this @Prashant-thakur77

PR #1289, opened 2026-06-22
Implement the KMultiSelectNode and KMultiSelectInput component for KMultiSelect

Prashant-thakur77:
## Description

Implements `KMultiSelectInput.vue` (the ARIA 1.2 combobox input row) and
`KMultiSelectNode.vue` (the recursive option renderer) as the two foundational
internal sub-components of the candidate `KMultiSelect` component. Also includes
two small backward-compatible extensions to `KListbox` and `KListboxOption`.

#### Issue addressed

Addresses #1266 

## Changelog

  - **Description:** Add `KMultiSelectInput` and `KMultiSelectNode` as internal
    sub-components of the candidate `KMultiSelect`. Extend `KListbox` with a
    `multiple` prop and `active-descendant-change` event. Extend `KListboxOption`
    with a `showCheckbox` prop.
  - **Products impact:** none
  - **Addresses:** #1266 
  - **Components:** KMultiSelect
  - **Breaking:** no
  - **Impacts a11y:** yes
  - **Guidance:** These are internal sub-components consumed only by the candidate
    `KMultiSelect` wrapper. `KListbox` and `KListboxOption` extensions are
    backward-compatible (all existing defaults preserved). Implements the full
    ARIA 1.2 combobox pattern and RTL-safe indentation via `paddingInlineStart`.

## Steps to test
1)All test should pass.

### At a high level, how did you implement this?

KMultiSelectInput is a stateless presentational component state flows in as props, interactions flow out as events. All ARIA combobox attributes are computed in a single inputAriaAttrs property and spread onto the native <input>.

KMultiSelectNode is a recursive component that renders a KListboxGroup for nodes with children and a KListboxOption for leaf nodes, with RTL-safe depth-based indentation via paddingInlineStart.

**`KListbox` / `KListboxOption` extensions:**
- `KListbox`: added `multiple` prop (`default: true`) so `aria-multiselectable`
  is dynamic. Added a `watch` on `focusedOptionId` to emit
  `active-descendant-change` for `KMultiSelectInput` to sync
  `aria-activedescendant`.
- `KListboxOption`: added `showCheckbox` prop (`default: true`) with `v-if/v-else`
  so single-select mode can hide the checkbox while `aria-selected` on the `<li>`
  still conveys state to screen readers. All defaults preserve existing behavior.

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [x] Critical and brittle code paths are covered by unit tests
- [x] The change is described in the changelog section above

## Reviewer guidance

[...]

PR #1293, opened 2026-06-26
[KListbox] Add comprehensive Single-Select mode support

### LINKED ISSUE #1292: [KListbox] Add comprehensive Single-Select mode support

## Overview

Update `KListbox` to natively support strict single-selection. This is a foundational requirement for the upcoming `KMultiSelect` component, which needs to be able to operate in a single-select mode without checkboxes.

**Target branch:** release-v5

### Context

When `KListbox` was originally implemented, it defaulted to multi-selection and did not support single selection or grouped options. As we build out `KMultiSelect` (which relies heavily on `KListbox` internally), we need the ability to pass `multiple="false"` to enforce single-selection rules structurally and visually.

### The Change

This task implements the single-select behavior directly inside `KListbox` and `KListboxOption`:
- **`KListbox` logic:** Added the `multiple` prop (defaults to true). When `multiple="false"`, the listbox structurally enforces single-select (clicking an option replaces the array instead of appending, re-clicking does not deselect, and `Ctrl+A` "select all" is blocked).
- **`KListboxOption` styling:** Added the `showCheckbox` prop (defaults to true). When `showCheckbox="false"`, the option hides the checkbox .
- **Visual Highlighting:** In single-select mode (no checkbox), selected items are styled with the primary blue text color, a bold `500` font weight, and a subtle gray background (`themePalette().grey.v_100`) on hover/selection to clearly indicate active state.Inpired from KSelect.

#### General
- [ ] Passing `multiple="false"` to `KListbox` prevents multi-selection, replacing the selected array instead of toggling.
- [ ] Passing `multiple="false"` blocks the `Ctrl+A` keyboard shortcut.
- [ ] Passing `showCheckbox="false"` to `KListboxOption` hides the checkbox and falls back to a plain text label.
- [ ] Plain text options apply blue text and gray backgrounds when selected.

#### Documentation & Testing
- [ ] Add a `SingleSelect.vue` example to the `KListbox` docs page.
- [ ] Update the `KListbox` documentation page to remove outdated claims that single-select is unsupported.
- [ ] Add the single-select example to the `KListboxVisualTest.vue` suite.
- [ ] Add unit tests in `index.spec.js` to verify the single-select logic guards (replacement, anti-deselection, and Ctrl+A block).

## AI usage

I used Gemini to help draft this issue from my implementation notes. I reviewed all generated logic and edited this description for accuracy.

Prashant-thakur77:
## Description
This PR updates `KListbox` to natively support strict single-selection. This is a foundational requirement for the upcoming `KMultiSelect` component, which needs to be able to operate in a single-select mode without checkboxes. It introduces a functional `multiple="false"` guard to the listbox and a `showCheckbox="false"` visual fallback to the options.

https://github.com/user-attachments/assets/7c8c268c-1b7e-4548-8346-54c6768817c3

#### Issue addressed
Addresses #1292 

## Changelog

  - **Description:** Added native single-select support to `KListbox` and `KListboxOption`.
  - **Products impact:** none
  - **Addresses:** #1292 
  - **Components:** KListbox, KListboxOption
  - **Breaking:** no
  - **Impacts a11y:** yes
  - **Guidance:** Set `multiple="false"` on `KListbox` and `showCheckbox="false"` on `KListboxOption` to enforce single selection rules (disables Ctrl+A, replaces value instead of appending) while dropping the visual checkbox for plain text rendering with selection highlighting.

## Steps to test

1. Navigate to the `KListbox` component page in the local documentation.
2. Scroll down to the new **Single Select** example.
3. Verify that clicking options correctly replaces the active selection (rather than selecting multiple).
4. Verify that the `Ctrl+A` "select all" keyboard shortcut is blocked.
5. Verify that selected options highlight with the primary blue text and a subtle gray background.

- Added `multiple` prop to `KListbox`. When false, clicking an option replaces the array and ignores `Ctrl+A`.
- Added `showCheckbox` prop to `KListboxOption`. When false, the `KCheckbox` wrapper is removed and replaced with a plain `<span class="k-listbox-option-label">`.
- Migrated inline styles to the `<style>` block and strictly maintained the `36px` minimum row height and `24px` line height to ensure absolute visual parity with the checkbox rows.

## Testing checklist

## Reviewer guidance

## Comments
**Note to reviewers:** This branch (`feature/klistbox-single-select`) was branched off of my previous PR for `feature/kmultiselect-components-input-node`. Therefore, the git diff currently shows the `KMultiSelectInput` and `KMultiSelectNode` files as well. 

If my previous PR (#1289) gets merged first, I can rebase this branch to clean up the diff. Otherwise, you can focus your review on the `KListbox` and `KListboxOption` directories!

### DIFF
--- a/docs/pages/klistbox.vue
+++ b/docs/pages/klistbox.vue
@@ -4,22 +4,12 @@
     <template #developerNotes>
       <ul :style="{ margin: 0 }">
         <li>
-          Grouped options aren't supported yet, as we haven't needed them. They could be added by
-          extending our listbox components according to
-          <DocsExternalLink
-            text="grouped options of APG's listbox pattern"
-            href="https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-grouped/"
-          />
-          (likely adding <code>KListboxGroup</code>)
-        </li>
-
-        <li>
           Listboxes are often combined with searching or filtering, but that's not part of the
           listbox pattern and use cases vary. If a frequent pattern emerges, it may motivate
           dedicated search/filter components that integrate well with <code>KListbox</code>.
         </li>
         <li>
-          Currently KListbox doesn't support options reodering, so bear in mind that if the list
+          Currently KListbox doesn't support options reordering, so bear in mind that if the list
           changes due to a filter, the filtered list should preserve the order of the original list
           to avoid unexpected keyboard navigation.
         </li>
@@ -43,8 +33,8 @@
       <p>Its options are provided via <DocsLibraryLink component="KListboxOption" />.</p>
 
       <p>
-        It currently defaults to multi selection and does not yet support grouped options (see
-        developer notes above).
+        It supports both multi selection (default, <code>multiple: true</code>) and single selection
+        (<code>multiple: false</code>).
       </p>
 
       <p>
@@ -228,6 +218,10 @@
       <ul>
         <li>
           <DocsLibraryLink component="KListboxOption" /> is a single option inside
+          <code>KListbox</code>
+        </li>
+        <li>
+          <DocsLibraryLink component="KListboxGroup" /> groups options inside
           <code>KListbox</code>
         </li>
         <li>
--- a/docs/pages/klistbox.vue
+++ b/docs/pages/klistbox.vue
@@ -123,10 +123,10 @@
       </ul>
 
       <DocsBanner a11y>
-        Link the control to the listbox <code>id</code> using <code>aria-controls</code>. Build the
-        label for the checkbox in a way that makes it clear it affects the list below (e.g.,
-        <code>'Select all users'</code> instead of <code>'All users'</code>) because
-        <code>aria-controls</code> does not change what the screen reader says.
+        Set <code>aria-controls</code> on the select all control to the listbox <code>id</code> to
+        associate the two. Build the label for the checkbox in a way that makes it clear it affects
+        the list below (e.g., <code>'Select all users'</code> instead of <code>'All users'</code>)
+        because <code>aria-controls</code> does not change what the screen reader says.
       </DocsBanner>
 
       <DocsExample
--- a/docs/pages/klistbox.vue
+++ b/docs/pages/klistbox.vue
@@ -123,10 +123,10 @@
       </ul>
 
       <DocsBanner a11y>
-        Link the control to the listbox <code>id</code> using <code>aria-controls</code>. Provide an
-        explicit label like <code>'Select all users'</code> instead of
-        <code>'All users'</code> because <code>aria-controls</code> does not change what the screen
-        reader says.
+        Link the control to the listbox <code>id</code> using <code>aria-controls</code>. Build the
+        label for the checkbox in a way that makes it clear it affects the list below (e.g.,
+        <code>'Select all users'</code> instead of <code>'All users'</code>) because
+        <code>aria-controls</code> does not change what the screen reader says.
       </DocsBanner>
 
       <DocsExample
--- a/docs/pages/klistbox.vue
+++ b/docs/pages/klistbox.vue
@@ -123,7 +123,10 @@
       </ul>
 
       <DocsBanner a11y>
-        Link the control to the listbox <code>id</code> using <code>aria-controls</code>. Provide an explicit label like <code>'Select all users'</code> instead of <code>'All users'</code> because <code>aria-controls</code> does not change what the screen reader says.
+        Link the control to the listbox <code>id</code> using <code>aria-controls</code>. Provide an
+        explicit label like <code>'Select all users'</code> instead of
+        <code>'All users'</code> because <code>aria-controls</code> does not change what the screen
+        reader says.
       </DocsBanner>
 
       <DocsExample
--- a/docs/pages/klistbox.vue
+++ b/docs/pages/klistbox.vue
@@ -12,10 +12,7 @@
           />
           (likely adding <code>KListboxGroup</code>)
         </li>
-        <li>
-          Reflecting our first use-cases, multi selection is the default behavior. Single selection
-          mode could be added by a few adjustments in line with the same APG listbox pattern.
-        </li>
+
         <li>
           Listboxes are often combined with searching or filtering, but that's not part of the
           listbox pattern and use cases vary. If a frequent pattern emerges, it may motivate
@@ -46,8 +43,8 @@
       <p>Its options are provided via <DocsLibraryLink component="KListboxOption" />.</p>
 
       <p>
-        It currently defaults to multi selection and does not yet support single selection or
-        grouped options (see developer notes above).
+        It currently defaults to multi selection and does not yet support grouped options (see
+        developer notes above).
       </p>
 
       <p>
@@ -87,6 +84,7 @@
         :items="[
           { text: 'Default', href: '#default' },
           { text: 'Select all', href: '#select-all' },
+          { text: 'Single Select', href: '#single-select' },
           { text: 'Customized', href: '#customized' },
           { text: 'Scrollable', href: '#scrollable' },
           { text: 'Grouped', href: '#grouped' },
@@ -143,6 +141,22 @@
         block
         exampleId="klistbox-select-all"
         loadExample="KListbox/SelectAll.vue"
+      />
+
+      <h3>
+        Single Select
+        <DocsAnchorTarget anchor="#single-select" />
+      </h3>
+
+      <p>
+        Set <code>multiple="false"</code> on the listbox and <code>showCheckbox="false"</code> on
+        the options to render plain text options for single selection.
+      </p>
+
+      <DocsExample
+        block
+        exampleId="klistbox-single-select"
+        loadExample="KListbox/SingleSelect.vue"
       />
 
       <h3>
--- a/docs/pages/klistbox.vue
+++ b/docs/pages/klistbox.vue
@@ -74,7 +74,7 @@
         :items="[
           { text: 'Default', href: '#default' },
           { text: 'Select all', href: '#select-all' },
-          { text: 'Single Select', href: '#single-select' },
+          { text: 'Single select', href: '#single-select' },
           { text: 'Customized', href: '#customized' },
           { text: 'Scrollable', href: '#scrollable' },
           { text: 'Grouped', href: '#grouped' },
@@ -91,7 +91,7 @@
       <DocsBanner a11y>
         If <code>KListbox</code> is not part of another widget, such as a combobox, add either a
         visible label referenced by <code>ariaLabelledBy</code> or a value specified for
-        <code>ariaLabel</code>. Use concise and descriptive value.
+        <code>ariaLabel</code>. Use a concise, descriptive value.
       </DocsBanner>
 
       <DocsExample
@@ -123,8 +123,7 @@
       </ul>
 
       <DocsBanner a11y>
-        Set <code>aria-controls</code> on the select all control to the listbox <code>id</code> to
-        associate the two.
+        Link the control to the listbox <code>id</code> using <code>aria-controls</code>. Provide an explicit label like <code>'Select all users'</code> instead of <code>'All users'</code> because <code>aria-controls</code> does not change what the screen reader says.
       </DocsBanner>
 
       <DocsExample
@@ -134,13 +133,14 @@
       />
 
       <h3>
-        Single Select
+        Single select
         <DocsAnchorTarget anchor="#single-select" />
       </h3>
 
       <p>
-        Set <code>multiple="false"</code> on the listbox and <code>showCheckbox="false"</code> on
-        the options to render plain text options for single selection.
+        Set <code>multiple="false"</code> on the listbox. Options automatically hide the checkbox
+        selector in single-select mode. Use the <code>showSelector</code> prop on
+        <DocsLibraryLink component="KListboxOption" /> to override this behaviour explicitly.
       </p>
 
       <DocsExample
@@ -276,6 +276,13 @@
             required: 'Yes',
             description: 'Announced via live region when an option is deselected.',
             example: 'Deselected',
+          },
+          {
+            name: 'partiallySelected',
+            required: 'No',
+            description:
+              'Renders a visually-hidden text inside indeterminate options so screen readers announce the partial selection state. Use with grouped listboxes that have parent checkboxes.',
+            example: 'Partially selected',
           },
         ],
       };
--- a/docs/pages/klistbox.vue
+++ b/docs/pages/klistbox.vue
@@ -32,10 +32,7 @@
 
       <p>Its options are provided via <DocsLibraryLink component="KListboxOption" />.</p>
 
-      <p>
-        It supports both multi selection (default, <code>multiple: true</code>) and single selection
-        (<code>multiple: false</code>).
-      </p>
+      <p>It supports both multi selection and single selection.</p>
 
       <p>
         <code>KListbox</code> supports dynamic option sets (useful for integration with filtering,
--- a/examples/KListbox/Customized.vue
+++ b/examples/KListbox/Customized.vue
@@ -20,7 +20,7 @@
             than necessary and aligns its correctly within the row
           -->
           <KCheckbox
-            label="All users"
+            label="Select all users"
             class="select-all-checkbox"
             :aria-controls="listboxId"
             :disabled="isDisabled"
@@ -29,7 +29,7 @@
             :style="{ marginTop: '6px', marginBottom: '0' }"
             @change="setAllSelected"
           >
-            <KLabeledIcon label="All users">
+            <KLabeledIcon label="Select all users">
               <template #icon>
                 <KIcon
                   icon="allUsers"
--- a/examples/KListbox/Customized.vue
+++ b/examples/KListbox/Customized.vue
@@ -114,6 +114,10 @@
     border-bottom: 1px solid;
   }
 
+  .option {
+    padding-inline-start: 16px;
+  }
+
   .option:last-child {
     border-bottom: 0;
   }
--- a/examples/KListbox/Customized.vue
+++ b/examples/KListbox/Customized.vue
@@ -115,7 +115,7 @@
   }
 
   .option {
-    padding-inline-start: 16px;
+    padding-left: 16px;
   }
 
   .option:last-child {
--- a/examples/KListbox/Scrollable.vue
+++ b/examples/KListbox/Scrollable.vue
@@ -93,6 +93,10 @@
     border-bottom: 1px solid;
   }
 
+  .option {
+    padding-inline-start: 10px;
+  }
+
   .option:last-child {
     border-bottom: 0;
   }
--- a/examples/KListbox/Scrollable.vue
+++ b/examples/KListbox/Scrollable.vue
@@ -94,7 +94,7 @@
   }
 
   .option {
-    padding-inline-start: 10px;
+    padding-left: 10px;
   }
 
   .option:last-child {
--- a/examples/KListbox/Scrollable.vue
+++ b/examples/KListbox/Scrollable.vue
@@ -16,7 +16,7 @@
           :style="{ borderColor: $themeTokens.fineLine }"
         >
           <KCheckbox
-            label="All classes"
+            label="Select all classes"
             class="select-all-checkbox"
             :aria-controls="listboxId"
             :disabled="!options.length"
--- a/examples/KListbox/SingleSelect.vue
+++ b/examples/KListbox/SingleSelect.vue
@@ -1,0 +1,75 @@
+<template>
+
+  <div>
+    <h4 :id="classLabel">Class (Single Select)</h4>
+    <KListbox
+      :id="listboxId"
+      v-model="selected"
+      class="listbox"
+      :ariaLabelledBy="classLabel"
+      :messages="messages"
+      :multiple="false"
+      :style="{ borderColor: $themeTokens.fineLine }"
+    >
+      <KListboxOption
+        v-for="option in options"
+        :key="option.id"
+        :value="option.id"
+        :label="option.label"
+        :showCheckbox="false"
+        class="option"
+        :style="{ borderColor: $themeTokens.fineLine }"
+      />
+    </KListbox>
+    <p>
+      Selected: <code>{{ JSON.stringify(selected) }}</code>
+    </p>
+  </div>
+
+</template>
+
+
+<script>
+
+  import { ref } from 'vue';
+
+  export default {
+    setup() {
+      const classLabel = 'single-select-class';
+      const listboxId = 'single-select-listbox';
+      const selected = ref(['literature']);
+      const options = [
+        { id: 'biology', label: 'Biology' },
+        { id: 'literature', label: 'Literature' },
+        { id: 'physics', label: 'Physics' },
+      ];
+      const messages = {
+        clickable: 'Options are clickable',
+        allOptionsSelected: 'All options selected',
+        allOptionsDeselected: 'No options selected',
+        optionDeselected: 'Deselected',
+      };
+      return { classLabel, listboxId, selected, options, messages };
+    },
+  };
+
+</script>
+
+
+<style lang="scss" scoped>
+
+  .listbox {
+    border: 1px solid;
+    border-radius: 4px;
+  }
+
+  .option {
+    padding: 8px 12px;
+    border-bottom: 1px solid;
+  }
+
+  .option:last-child {
+    border-bottom: 0;
+  }
+
+</style>
--- a/examples/KListbox/SingleSelect.vue
+++ b/examples/KListbox/SingleSelect.vue
@@ -1,7 +1,7 @@
 <template>
 
   <div>
-    <h4 :id="classLabel">Class (Single Select)</h4>
+    <h4 :id="classLabel">Class</h4>
     <KListbox
       :id="listboxId"
       v-model="selected"
@@ -16,7 +16,6 @@
         :key="option.id"
         :value="option.id"
         :label="option.label"
-        :showCheckbox="false"
         class="option"
         :style="{ borderColor: $themeTokens.fineLine }"
       />
@@ -64,7 +63,7 @@
   }
 
   .option {
-    padding: 8px 12px;
+    padding: 8px 6px;
     border-bottom: 1px solid;
   }
 
--- a/lib/candidate/listbox/KListbox/index.vue
+++ b/lib/candidate/listbox/KListbox/index.vue
@@ -438,6 +438,7 @@
   .k-listbox-select-all {
     display: flex;
     align-items: center;
+    padding-inline-start: 4px;
   }
 
   .k-listbox-list {
--- a/lib/candidate/listbox/KListbox/index.vue
+++ b/lib/candidate/listbox/KListbox/index.vue
@@ -438,7 +438,7 @@
   .k-listbox-select-all {
     display: flex;
     align-items: center;
-    padding-inline-start: 4px;
+    padding-left: 4px;
   }
 
   .k-listbox-list {
--- a/lib/candidate/listbox/KListbox/index.vue
+++ b/lib/candidate/listbox/KListbox/index.vue
@@ -297,7 +297,9 @@
         isSelected,
         isFocused,
         toggleOption,
-        multiple: props.multiple,
+        get multiple() {
+          return props.multiple;
+        },
       });
 
       onMounted(() => {
--- a/lib/candidate/listbox/KListbox/index.vue
+++ b/lib/candidate/listbox/KListbox/index.vue
@@ -13,7 +13,7 @@
       checkbox is clicked (otherwise click would cause double-toggle, resulting in no change))
     -->
     <div
-      v-if="$scopedSlots.selectAll"
+      v-if="multiple && $scopedSlots.selectAll"
       class="k-listbox-select-all"
       :class="$computedClass(selectAllStyles)"
       :style="{ backgroundColor: $themeTokens.surface }"
@@ -173,6 +173,7 @@
 
       function changeSelectAll(checked) {
         if (!hasOptions.value) return;
+        if (!props.multiple) return; // Prevent "select all" in single-select mode
         if (checked) {
           selectAllOptions();
         } else {
@@ -271,6 +272,9 @@
           case 'A':
             if (!event.ctrlKey && !event.metaKey) {
               return;
+            }
+            if (!props.multiple) {
+              return; // Do not allow select-all in single-select mode
             }
             if (allSelected.value) {
               deselectAllOptions();
@@ -396,7 +400,10 @@
       },
       /**
        * Whether the listbox allows multiple selections.
-       * Controls the aria-multiselectable attribute.
+       * Controls both the aria-multiselectable ARIA attribute and the
+       * actual selection behavior: when false, re-clicking an already
+       * selected option does nothing, and selecting a new option replaces
+       * the current selection instead of adding to it.
        */
       multiple: {
         type: Boolean,
--- a/lib/candidate/listbox/KListbox/index.vue
+++ b/lib/candidate/listbox/KListbox/index.vue
@@ -297,6 +297,7 @@
         isSelected,
         isFocused,
         toggleOption,
+        getMessage,
         get multiple() {
           return props.multiple;
         },
--- a/lib/candidate/listbox/KListbox/index.vue
+++ b/lib/candidate/listbox/KListbox/index.vue
@@ -354,7 +354,8 @@
         required: true,
       },
       /**
-       * Array of currently selected option values
+       * Array of selected option values. Always an Array in both single and
+       * multi-select modes for a consistent v-model API.
        */
       value: {
         type: Array,
--- a/lib/candidate/listbox/KListbox/index.vue
+++ b/lib/candidate/listbox/KListbox/index.vue
@@ -297,6 +297,7 @@
         isSelected,
         isFocused,
         toggleOption,
+        multiple: props.multiple,
       });
 
       onMounted(() => {
--- a/lib/candidate/listbox/KListboxOption/index.vue
+++ b/lib/candidate/listbox/KListboxOption/index.vue
@@ -41,10 +41,11 @@
       }"
     >
       <KIcon
-        v-if="isSelected"
         icon="done"
         aria-hidden="true"
         class="k-option-check-icon"
+        :color="$themeTokens.primary"
+        :style="{ visibility: isSelected ? 'visible' : 'hidden' }"
       />
       <slot>{{ label }}</slot>
     </span>
@@ -167,7 +168,7 @@
 
   .k-option-check-icon {
     flex-shrink: 0;
-    font-size: 18px;
+    font-size: 16px;
   }
 
 </style>
--- a/lib/candidate/listbox/KListboxOption/index.vue
+++ b/lib/candidate/listbox/KListboxOption/index.vue
@@ -42,10 +42,9 @@
     >
       <KIcon
         icon="done"
-        aria-hidden="true"
         class="k-option-check-icon"
         :color="$themeTokens.primary"
-        :style="{ visibility: isSelected ? 'visible' : 'hidden' }"
+        :style="{ visibility: isSelected ? 'visible' : 'hidden', top: '0' }"
       />
       <slot>{{ label }}</slot>
     </span>
@@ -156,7 +155,7 @@
     display: flex;
     align-items: center;
     min-height: 36px;
-    padding-inline-start: 4px;
+    padding-left: 4px;
     cursor: pointer;
   }
 
--- a/lib/candidate/listbox/KListboxOption/index.vue
+++ b/lib/candidate/listbox/KListboxOption/index.vue
@@ -61,7 +61,15 @@
       const uid = optionCount++;
       const optionId = `klistbox-option-${uid}`;
 
-      onMounted(() => listbox.registerOption({ id: optionId, value: props.value }));
+      onMounted(() => {
+        listbox.registerOption({ id: optionId, value: props.value });
+        if (process.env.NODE_ENV !== 'production' && props.showCheckbox && !listbox.multiple) {
+          // eslint-disable-next-line no-console
+          console.warn(
+            "[KListboxOption] 'showCheckbox' is true but the parent KListbox is in single-select mode ('multiple: false'). Set 'showCheckbox: false' to hide the checkbox.",
+          );
+        }
+      });
 
       onBeforeUnmount(() => listbox.unregisterOption({ id: optionId, value: props.value }));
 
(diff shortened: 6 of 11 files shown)


Prashant-thakur77:
I have rebased this branch with the latest updates on the release v-5 branch :)

lib/candidate/listbox/KListboxOption/index.vue
```diff
@@ -68,7 +70,8 @@
       const isFocused = computed(() => listbox.isFocused(props.value));
 
       const rowStyles = computed(() => ({
-        backgroundColor: themeTokens().surface,
+        backgroundColor:
+          !props.showCheckbox && isSelected.value ? 'rgba(0, 0, 0, 0.05)' : themeTokens().surface,
```
rtibblesbot: ✅ **Resolved** — addressed in the current code.


**blocking:** This selected-state background, combined with the `$themeTokens.primary` text color set two lines up (line 35), fails WCAG AA contrast. Measured via axe-core on the live single-select example: foreground `#4368f5` on background `rgba(0,0,0,0.05)` over white (≈`#f2f2f2`) = 4.13:1, below the 4.5:1 required for this 16px/500-weight text.

This also hardcodes a literal color instead of using the theme tokens already imported in this file (`themePalette().grey.v_100` is used for `:hover` right below). Routing the selected-background through a verified-AA token pair (darker background and/or a less-saturated primary shade) would fix both the contrast failure and the literal-vs-token inconsistency in one change.

rtibblesbot (changes requested):
Implements all stated acceptance criteria from #1292 with focused, well-scoped changes and matching unit tests. One blocking accessibility issue in the new selected-option styling.

- **blocking**: selected-option text/background combo fails WCAG AA contrast (inline comment).
- Same root cause also bypasses the design-token pattern used elsewhere in the file (folded into the inline comment).

CI: mixed/pending at time of review — please confirm it's green before merge.

Manual QA (live docs-page session):
- Click replaces selection instead of appending — pass
- Already-selected option doesn't toggle off — pass
- Ctrl+A is blocked in single-select mode — pass
- Keyboard focus/navigation works, no focus trap — pass
- RTL layout mirrors correctly — pass
- Mobile viewport (375px) has no overflow/clipping — pass
- Selected-option color contrast — fail, see inline comment

---

</details>

lib/candidate/listbox/KListboxOption/index.vue
Prashant-thakur77: The rgba(0, 0, 0, 0.05) selected-state background is a direct replication of the existing KSelectOption pattern (KSelect/KSelectOption.vue).
https://github.com/learningequality/kolibri-design-system/blob/1b9ce481b2697a3d331e28404296b3e25dffdd9d/lib/KSelect/KSelectOption.vue#L170-L173

@LianaHarris360 should we change here as the bot suggest for the compliance with WCAG AA contrast?

rtibblesbot: Confirmed — I checked KSelectOption.vue at that commit, and it has the same combination (`color: this.$themeTokens.primary` on `.ui-select-option` text, `background-color: rgba(black, 0.05)` on `.is-selected`), so this does appear to be a pre-existing contrast gap being carried over rather than something new to this PR. That said, it doesn't change the contrast math for this file — 4.13:1 is still below the 4.5:1 AA threshold here. Deferring to @LianaHarris360 on whether to fix it now in KListboxOption or track it separately alongside KSelectOption.

Prashant-thakur77: Currently it is like:
image
I tried 2 implementation if we decided to change it here:
1)No background, primary text + bold
The selected state is now indicated by $themeTokens.primary text color + fontWeight: 500 on a plain themeTokens().surface background.
image
2)grey.v_100 tint + bold
Fixed by replacing rgba(0, 0, 0, 0.05) with the themePalette().grey.v_100 token,fontWeight: 500 
Screenshot From 2026-07-02 20-49-54
In this  we can increase font weigth if we need that:
image

LianaHarris360: I checked the single selection use case in Studio,  and there is no background color when a value is selected:

Image

Looks like it only uses the` $themeTokens.primary` text color with the plain `themeTokens().surface` background to indicate what value is selected. As for the `KSelectOption` contrast gap, it should be changed also, but not in this PR. Instead, I can address it in a separate issue.

docs/pages/klistbox.vue
```diff
@@ -46,8 +43,8 @@
       <p>Its options are provided via <DocsLibraryLink component="KListboxOption" />.</p>
 
       <p>
-        It currently defaults to multi selection and does not yet support single selection or
-        grouped options (see developer notes above).
+        It currently defaults to multi selection and does not yet support grouped options (see
+        developer notes above).
```
LianaHarris360: Since `KListbox` now supports grouped options with `KListboxGroup`, it's a good time to remove the part that says "and does not yet support grouped options (see developer notes above)," along with this earlier sentence:

> Grouped options aren't supported yet, as we haven't needed them. They could be added by extending our listbox components according to [grouped options of APG's listbox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-grouped/) (likely adding KListboxGroup)

lib/candidate/listbox/KListboxOption/index.vue
Prashant-thakur77: Done:)
image

rtibblesbot: Verified — confirmed against the current branch: the selected-state background was removed, so `$themeTokens.primary` text now sits on a plain `themeTokens().surface` (`#FFFFFF`) instead of the tinted `rgba(0,0,0,0.05)` background. That's a contrast ratio of ~4.63:1, which clears the 4.5:1 AA threshold (up from 4.13:1). This also resolves the literal-color concern since the background is now just the surface token. Withdrawing this as blocking.

@LianaHarris360 agreed that the same gap in `KSelectOption` is out of scope here and worth tracking separately.

lib/candidate/listbox/KListboxOption/index.vue
```diff
@@ -109,10 +112,8 @@
...
       },
       /**
-       * When true (default), wraps the option label in a KCheckbox for
-       * multi-select use cases. When false, renders a plain label — used
-       * when checkboxes are not meaningful (e.g. single-select, or
-       * hideSelected mode where selected items are already removed from the list).
+       * When true, wraps the option label in a KCheckbox for multi-select use cases.
+       * When false, renders a plain label for single-select use cases.
        */
       showCheckbox: {
```
LianaHarris360: If `multiple="false"`, `showCheckbox` still defaults to true, resulting in checkboxes displayed inside the single-select listbox. We might want to either guard against allowing a true `showCheckbox` to be passed with a false `multiple` or, similar to the [Missing ariaLabel console warning](https://github.com/learningequality/kolibri-design-system/blob/fa34442ef0ea6db14aec2c27df5e3936b4d83a0b/lib/candidate/listbox/KListbox/index.vue#L320) in `KListbox`, display a console warning if there is a misconfiguration.

LianaHarris360 (dismissed):

docs/pages/klistbox.vue
Prashant-thakur77: done,also referenced KlistboxGroup in the docs:)

lib/candidate/listbox/KListboxOption/index.vue
Prashant-thakur77: done,went with the warning:

https://github.com/user-attachments/assets/b281fc4e-005a-46be-ae26-cf48b3eef2d2

lib/candidate/listbox/KListboxOption/index.vue
```diff
@@ -59,7 +61,15 @@
...
-      onMounted(() => listbox.registerOption({ id: optionId, value: props.value }));
+      onMounted(() => {
+        listbox.registerOption({ id: optionId, value: props.value });
+        if (process.env.NODE_ENV !== 'production' && props.showCheckbox && !listbox.multiple) {
+          // eslint-disable-next-line no-console
+          console.warn(
+            "[KListboxOption] 'showCheckbox' is true but the parent KListbox is in single-select mode ('multiple: false'). Set 'showCheckbox: false' to hide the checkbox.",
+          );
+        }
+      });
```
AlexVelezLl, line 76: Wonder if we should have something like "showSelector" and switch between a checkbox and a radio button based on whether multiple is true or false. Is there already a use case for single select?

lib/candidate/listbox/KListboxOption/index.vue
```diff
@@ -31,8 +31,10 @@
     -->
     <span
       v-else
-      class="k-listbox-option-label"
-      :style="{ lineHeight: '24px' }"
+      :style="{
+        color: isSelected ? $themeTokens.primary : 'inherit',
+        fontWeight: isSelected ? '500' : 'normal',
+      }"
```
AlexVelezLl, line 41: I wonder if KDS should be less opinionated about what the style should look like, since we can use KListbox with `multiple=false` for layouts like this, where the "selected" state is just a different background color.

Image

A potential solution may be to use these styles by default and provide a prop to override them.

lib/candidate/listbox/KListbox/index.vue
```diff
@@ -293,6 +297,7 @@
         isSelected,
         isFocused,
         toggleOption,
+        multiple: props.multiple,
```
AlexVelezLl: Not completely sure, but it's very likely that this won't be reactive. If for some reason `props.multiple` changes from true to false, the children won't get the updates.

examples/KListbox/SingleSelect.vue
```diff
@@ -0,0 +1,75 @@
...
+
+<script>
+
+  import { ref } from 'vue';
+
+  export default {
+    setup() {
+      const classLabel = 'single-select-class';
+      const listboxId = 'single-select-listbox';
+      const selected = ref(['literature']);
```
AlexVelezLl, line 39: So, even if it's just a single select, should we still pass an array? Not saying it's wrong, although it may potentially be a bit counterintuitive.

AlexVelezLl (commented):
Just some thoughts!

lib/candidate/listbox/KListboxOption/index.vue
LianaHarris360: There's a couple of single selection use cases in Studio. For the `LanguageDropdown` and the `CountryField` in the usersTable, there is no visual selected state displayed.

LanguageDropdown

usersTable

But do we need to prevent the checkbox from being displayed? In both Studio cases, `showCheckbox` and `multiple` can be set to false to achieve the same result, while still allowing it to be flexible for future implementations.

AlexVelezLl: Makes sense 😅. It's only that if we have a future implementation with a `radioButton`, then we will need to add a `showRadioButton` prop to allow showing the radio button, because reusing the `showCheckbox` wouldn't apply (and then we would need to add another check for showRadioButton=true and multiple=true).

lib/candidate/listbox/KListboxOption/index.vue
AlexVelezLl: Now that I saw https://github.com/learningequality/kolibri-design-system/pull/1293#discussion_r3516059012. I wonder whether the default styles should be just `$themeTokens.primary`, or if we should use the gray-ish background KSelect uses as well?  (since KSelect is already used in many places and this would make everything feel more consistent)

image

Ultra nitpicky, and I think this can be decided in the future 😅, but just flagging it.

Prashant-thakur77: Yes,we did go with the same styles as used in Kselect before but it fails WCAG AA contrast .You can see this in this [comment](https://github.com/learningequality/kolibri-design-system/pull/1293#discussion_r3514085503). And that is why we decided to use the single select selected option with a $themeTokens.primary only

examples/KListbox/SingleSelect.vue
Prashant-thakur77: Yes, the v-model value is always an array regardless of the multiple mode. This was intentional  from my end as it keeps the API consistent so consumers don't need to handle different value types based on the mode.So i thought it was a valid trade-off but yes it is a bit counterintuitive.Should i update it to handle modes seperatively as it is done in Kselect array for multiple and object for single select?

lib/candidate/listbox/KListbox/index.vue
Prashant-thakur77: Fixed it using a getter:

get multiple() {
  return props.multiple;
}

examples/KListbox/SingleSelect.vue
AlexVelezLl: Not really, just making sure it is a thoughtful decision and it is well documented :), would be nice to have thoughts from others too.

Another example of something kind of similar is the KCheckbox `v-model`, if we pass it an array, it adds the value to the array, but if not, it will just unset the checkbox value to the `v-model` (i.e., setting `value` to `null` directly to `inputValue` instead of `[value]`/`[]`)

lib/candidate/listbox/KListboxOption/index.vue
```diff
@@ -31,8 +31,10 @@
     -->
     <span
       v-else
-      class="k-listbox-option-label"
-      :style="{ lineHeight: '24px' }"
+      :style="{
+        color: isSelected ? $themeTokens.primary : 'inherit',
+        fontWeight: isSelected ? '500' : 'normal',
```
rtibblesbot, line 40: ✅ **Resolved** — addressed in the current code.


**blocking:**
- Commit `48400dfd0` deleted the flagged `rgba(0,0,0,0.05)` background instead of replacing it — this block now only varies `color`/`fontWeight`, no background on selection.
- Issue #1292 requires "blue text and gray backgrounds when selected."
- Suggest `themePalette().grey.v_100` (already used for `:hover` below) as the selected-state background — same token, already AA-safe with primary text.

lib/candidate/listbox/KListbox/index.vue
```diff
@@ -293,6 +297,9 @@
         isSelected,
         isFocused,
         toggleOption,
+        get multiple() {
```
rtibblesbot, line 301: ✅ **Resolved** — addressed in the current code.


**praise:** Getter avoids snapshotting `props.multiple` at `setup()` time — keeps `KListboxOption`'s single-select warning check correct if the parent toggles the prop reactively.

rtibblesbot (changes requested):
1 of 2 prior findings resolved; 1 still open (see below).

- **[blocking]** `lib/candidate/listbox/KListboxOption/index.vue:34-37` — selected-state background was removed rather than replaced, leaving an acceptance criterion unimplemented. See inline comment.

CI: pending.

RESOLVED — lib/candidate/listbox/KListboxOption/index.vue:32-40 — WCAG AA contrast failure on selected-state background (`rgba(0,0,0,0.05)`) (see new finding above)
RESOLVED — lib/candidate/listbox/KListbox/index.vue:300-302 — injected `multiple` flag not reactive — fixed via getter

</details>

---

Ran an automatic code-only delta review triggered by new commits on a previously reviewed PR:
- Core review pass only — specialized frontend/backend lenses and manual QA run when a review is explicitly requested

</details>

lib/candidate/listbox/KListboxOption/index.vue
Prashant-thakur77: I tested grey.v_100 (#F5F5F5) actually makes the contrast worse, not better:
Primary #4368F5 on grey.v_100 #F5F5F5 → 4.25:1 which fails WCAG AA
Primary #4368F5 on white #FFFFFF → 4.63:1 passes WCAG AA
Adding the grey background drops us below the 4.5:1 threshold.

rtibblesbot: Withdrawing this. I re-checked the contrast math myself and your numbers are right — primary on `grey.v_100` is ~4.25:1 (fails AA), primary on white is ~4.63:1 (passes) — so `grey.v_100` isn't a viable fix either, it'd trade one AA failure for another.

More importantly, I shouldn't have re-raised this at all: this exact background-vs-no-background question was already settled earlier in this thread — I withdrew the original blocking finding once the background was removed, and @LianaHarris360 confirmed that matches the actual Studio usage (primary text only, no background). Citing the issue's literal "gray background" wording to reopen it was a mistake on my part given that prior resolution. Apologies for the noise.

AlexVelezLl: We also have a grey.v_50. Would that work better? 😅

Prashant-thakur77: I tried grey.v_50 (#FAFAFA),it is close at 4.43:1 but still just below the 4.5:1 AA threshold.😅

rtibblesbot: Thanks both — that confirms neither `grey.v_100` (4.25:1) nor `grey.v_50` (4.43:1) clears the 4.5:1 AA threshold with primary text, so the no-background resolution already in place (matching actual Studio usage per @LianaHarris360) is the right call. Nothing further needed from me here.

rtibblesbot (commented):
2 of 2 prior findings resolved; no new findings on the current diff.

CI: mostly green, Percy visual tests still WAITING. Manual QA was required but did not complete this round, so this is a COMMENT rather than an approval.

RESOLVED — lib/candidate/listbox/KListboxOption/index.vue:34-37 — selected-state background/contrast: code at this location is unchanged from the `upstream/release-v5` baseline (no gray/rgba background either way) and has not regressed since it was marked resolved; out of scope for this PR's delta.
RESOLVED — lib/candidate/listbox/KListboxOption/index.vue:121-128 — `showCheckbox` prop docstring wording cleanup.
ACKNOWLEDGED — lib/candidate/listbox/KListbox/index.vue:300 — praise: `get multiple()` getter avoids snapshotting `props.multiple` at setup time.

</details>

---

Ran an automatic code-only delta review triggered by new commits on a previously reviewed PR:
- Core review pass only — specialized frontend/backend lenses and manual QA run when a review is explicitly requested

</details>

radinamatic:
Apologies for late reply, nothing like a forest 🔥 near where one lives to make everything else fall behind... 😅 

Great work in this improved component implementation, @Prashant-thakur77, looks awesome! 👏🏽 

I may have more to say after testing the actual implementation in Kolibri and Studio, but after testing the preview with NVDA, for now I have just a couple of comments:

1. In the `Select all` example, I believe a stronger output about the connection between the checkbox and the list that follows it, would be beneficial for the screen reader user.  In the KDS example it seems obvious, and it may not require any code changes, but I would recommend to add a recommendation in the documentation to build the label for the checkbox in a way that it's clear that the checkbox affects the list below. `aria-controls` enables the proper interaction, but does not add anything to the screen reader output.

   Win11 2606 (start)  Running  - Oracle VirtualBox_001

   Another thing that might need some CSS adjustment is to make the visual focus outline not be _glued_ to the checkbox on the left. 😉 

2. In the `Single select` (btw, the `Select` should not be capitalized 🙂) with `showCheckbox="false"` example, I'm concerned that the sole slight difference in color between the selected and non selected options is not enough to communicate those states to the user (**Don't use the color ALONE to communicate the message!** principle), especially in the case of low vision and color vision deficiencies. Can we add a SVG checkmark (can be hidden from the screen readers, since they already give a proper state output) to reinforce the message?

   Win11 2606 (start)  Running  - Oracle VirtualBox_002
   
3. In the `Grouped` example, could we add a `Partially selected` output somewhere for that state? Not sure for other screen readers, but NVDA interprets the _half checked_ state as `not selected` in this example, and it would be good to have it consistent with the state of the checkbox when it is _half checked_.

   Win11 2606 (start)  Running  - Oracle VirtualBox_004

lib/candidate/listbox/KListboxOption/index.vue
MisRob: Have you considered having a `#select` slot or similar paths? With props tied too much to multi-select use case, the listbox component may not scale that well.

```
<KListboxOption>

  <template #select={ ... }>
  // I can put nothing here or just empty span (= resulting no select control needed in this single-select use-case), or some custom checkbox, or radio button..
  </template>

</KListboxOption>
```

Similarly to the existing `<template #selectAll="{ allSelected, someSelected, setAllSelected }">` slot on `KListbox`.

docs/pages/klistbox.vue
```diff
@@ -3,26 +3,13 @@
   <DocsPageTemplate apiDocs>
     <template #developerNotes>
       <ul :style="{ margin: 0 }">
-        <li>
```
MisRob, line 6: Thanks for updating this section, and good new additions to `KListbox` :)!

examples/KListbox/SingleSelect.vue
```diff
@@ -0,0 +1,75 @@
+<template>
+
+  <div>
+    <h4 :id="classLabel">Class (Single Select)</h4>
```
MisRob: @Prashant-thakur77 I've noticed couple of times you often use _Title Case_. The vast majority of documentation follows [this guidance](https://deploy-preview-1293--kolibri-design-system.netlify.app/writing#punctuation) and is _Sentence case_. Generally when writing docs, it'd be nice to keep it consistent. It helps some people with sensitive eye, like myself, from cognitive overwhelm related to different styles being mixed. 

Also in this particular example, it'd be actually enough to just use _Class_. Examples are meant to resemble real-use cases. This will also remove duplicate information; single select is already mentioned in the section title.

MisRob:
@radinamatic  @Prashant-thakur77 @LianaHarris360 @AlexVelezLl 

> Another thing that might need some CSS adjustment is to make the visual focus outline not be glued to the checkbox on the left.

That was originally intentional so that in the consuming apps, it's easier to configure it precisely without the need to calculate what KDS does already (I had multi-select nesting use-case in mind). It's possible I took it too far & some reasonable default would be good. I'd just recommend it's as small value as possible to give a little bit of space, but not too much - so it's easy to control from each use-case without the need for override tricks.

lib/candidate/listbox/KListboxOption/index.vue
MisRob: I also agreed with @AlexVelezLl's idea about, at the very least, not coupling the prop name to checkbox. Instead using `showSelect` or `showSelector`. 

In the long run, maybe both `showSelect` and `#select` would be meaningful ~ having to use `#select` slot to just put nothing into it could be unusual, and that's where `showSelect` is useful. And if we need to show it and customize it, then `#select` slot (you probably don't need it at this point though). ?

MisRob: Finally, since `KListbox` already communicates with `KListboxOption` via `provide/inject`, you may also consider hiding chechkboxes by default in `KListboxOption` when `KListbox` has `multiple` falsy prop value. This is alternative to having `showSelect` prop` and I think it'd be okay - I doubt that we'd ever want to show checkboxes in any single-select pattern due to their association with everything that is not single-select :)

(with that, in the future, this could still be further overriden by the presence of `#select` slot in case there's need to show something after all)

Prashant-thakur77:
> 1. In the `Select all` example, I believe a stronger output about the connection between the checkbox and the list that follows it, would be beneficial for the screen reader user.  In the KDS example it seems obvious, and it may not require any code changes, but I would recommend to add a recommendation in the documentation to build the label for the checkbox in a way that it's clear that the checkbox affects the list below. `aria-controls` enables the proper interaction, but does not add anything to the screen reader output.
>    Win11 2606 (start)  Running  - Oracle VirtualBox_001
>    Another thing that might need some CSS adjustment is to make the visual focus outline not be _glued_ to the checkbox on the left. 😉

Thank you so much for the thorough review and testing with NVDA:)
 I have added a new documentation explicitly recommending this, explaining that aria-controls isn't read aloud, and advising developers to build labels in a way that it's clear that the checkbox affects the list below . I also updated all of our examples to follow this recommendation.

As for the focus outline I added a padding-inline-start: 4px to the listbox options:
image

Prashant-thakur77:
> 2\. In the `Single select` (btw, the `Select` should not be capitalized 🙂) with `showCheckbox="false"` example, I'm concerned that the sole slight difference in color between the selected and non selected options is not enough to communicate those states to the user (**Don't use the color ALONE to communicate the message!** principle), especially in the case of low vision and color vision deficiencies. Can we add a SVG checkmark (can be hidden from the screen readers, since they already give a proper state output) to reinforce the message?

I have added an SVG checkmark that renders left to the label in single-select mode.
Oh, and I also went through all the documentation pages and updated the headings and examples to strictly use Sentence case (e.g. "Single select" instead of "Single Select"),will keep this in mind while doing documentation in future too:)

https://github.com/user-attachments/assets/fcd0ddd7-ccb5-4220-bee6-800b30765726

Prashant-thakur77:
> 3\. In the `Grouped` example, could we add a `Partially selected` output somewhere for that state? Not sure for other screen readers, but NVDA interprets the _half checked_ state as `not selected` in this example, and it would be good to have it consistent with the state of the checkbox when it is _half checked_.

To fix this, I extended the usage of the existing messages prop in KListbox. Consumers of the component can now pass an optional partiallySelected string inside the messages object.
Currenlty it is announcing: "Physics, Partially selected, not selected",Is it good now?the not selected part is being announced because of aria-selected="false".
image

examples/KListbox/SingleSelect.vue
Prashant-thakur77: I will keep this in mind and i also looked into the writing guidance properly:).I have also updated it to just class.

lib/candidate/listbox/KListboxOption/index.vue
Prashant-thakur77: @LianaHarris360  @MisRob @AlexVelezLl I have made updates regarding this and wanted to share if this is good:

1) I renamed showCheckbox to showSelector. As Alex pointed out, this decouples the prop name from checkboxes so it can scale to radio buttons or other controls in the future.
2) I leveraged the existing provide/inject connection, as MisRob suggested. KListboxOption now automatically checks listbox.multiple. If multiple is falsy, the option automatically defaults to hiding the selector without the developer needing to manually pass showSelector="false" to every single option.
And it has true or false to override it.
I did not implement a #select slot just yet since we don't have a use-case for overriding the control with a custom component right now,so this can be added in future improvements:)

Prashant-thakur77:
@radinamatic @LianaHarris360 @MisRob @AlexVelezLl  ,I have done the changes and updates needed to be handled based on the review and also mentioned here what those changes are.Do tell if any thing i missed or needs improvement or reconsideration:)

lib/candidate/listbox/KListboxOption/index.vue
MisRob: Yes I think the combination of these is good balance. It's a straightforward default behavior, while still having a choice to override with `showSelector` if needed.

And yes I don't think that the `#select` slot is needed right now with the above in place ~ and with this overall direction, it should work nicely as soon it's needed with what is already there.

examples/KListbox/SingleSelect.vue
MisRob: Thanks a lot

docs/pages/klistbox.vue
```diff
@@ -46,8 +33,8 @@
       <p>Its options are provided via <DocsLibraryLink component="KListboxOption" />.</p>
 
       <p>
-        It currently defaults to multi selection and does not yet support single selection or
-        grouped options (see developer notes above).
+        It supports both multi selection (default, <code>multiple: true</code>) and single selection
```
MisRob:
```suggestion
        It supports both multi selection and single selection.
```

Just a nitpick, I think there's no need to be as specific in the section that is just introduction to the component.

docs/pages/klistbox.vue
```diff
@@ -46,8 +33,8 @@
...
+        (<code>multiple: false</code>).
```
MisRob:
```suggestion

```

Related to the above.

MisRob:
> I have added a new documentation explicitly recommending this, explaining that aria-controls isn't read aloud, and advising developers to build labels in a way that it's clear that the checkbox affects the list below . I also updated all of our examples to follow this recommendation.

Very nice, thanks a lot for the updates of examples, I am sure it will be useful.

> As for the focus outline I added a padding-inline-start: 4px to the listbox options:

Looks good!

One small issue to address - now the select all checkbox is slightly misaligned. Imagine a vertical line. If I recall Kolibri use-cases, all checkboxes, including the select all are aligned precisely, and I think it'd be reasonable default (again to allow for easier customization)

screenshot


Also @AlexVelezLl, not sure if the 4px change will have impact on releases. 4px is not a huge difference, on the other hand Kolibri will likely need some updates in response.

> I have added an SVG checkmark that renders left to the label in single-select mode.

I agree that using color alone is not enough to indicate selected state. @radinamatic @LianaHarris360 just wanted to check-in that the green checkmark is agreed with design - while I think it's a good direction, there will be some considerations as for its color, alignment, etc. Or perhaps more generally about the overall appearance of the selected state in a single-select mode, should even be a plain checkmark? As for the initial multi-select mode, I followed existing implementation that was based on our designs, whereas single-select is something new so if it's not happening yet, it'd be good to loop in Jess :)

AlexVelezLl:
> Also @AlexVelezLl, not sure if the 4px change will have impact on releases. 4px is not a huge difference, on the other hand Kolibri will likely need some updates in response.

Thanks @MisRob! Will check in a bit :).

> I agree that using color alone is not enough to indicate selected state. @radinamatic @LianaHarris360 just wanted to check-in that the green checkmark is agreed with design - while I think it's a good direction, there will be some considerations as for its color, alignment, etc. Or perhaps more generally about the overall appearance of the selected state in a single-select mode, should even be a plain checkmark? As for the initial multi-select mode, I followed existing implementation that was based on our designs, whereas single-select is something new so if it's not happening yet, it'd be good to loop in Jess :)

I was just going to say the same 😅. We should align this with Jessica, and see what the expected behavior should be having in mind the color contrast now.

radinamatic:
Thank you @Prashant-thakur77 for addressing all my comments! ❤️ 

> Currenlty it is announcing: "Physics, Partially selected, not selected",Is it good now?the not selected part is being announced because of aria-selected="false".

Unless there's a way to completely remove the ``not selected`` from the screen reader output, this might be slightly confusing for the user, but still closer to the _truth_ than just ``not selected``. Let's keep it for now and reassess down the line 👍🏽 

Regarding the discussion around the type of checkmark icon/graphic, I leave that totally to the KDS and design teams to choose... 😁 

The main thing is to have one visible so it reinforces the label color change to deliver a clear message that the option has been selected. 🙏🏽

radinamatic (approved):
Great work, thank you @Prashant-thakur77! 👏🏽👏🏽👏🏽

LGTM! 💯 :shipit: 🚀

MisRob:
> Unless there's a way to completely remove the not selected from the screen reader output

@radinamatic Technically it will be easy to remove `aria-selected` attribute conditionally in cases when _Partially selected_ is announced, I think Prashant here

> the not selected part is being announced because of aria-selected="false".

is trying to ask about if it's fine to not apply `aria-selected` attribute.

@Prashant-thakur77  I think when exactly none or exactly all child options are selected, there is a value in using `aria-selected` on the group. However I wouldn't apply the attribute on the group as long as it's in an indeterminate state - in such a condition both `true` and `false` values are not accurate and only confusing. Could you conditionally remove `aria-selected` on the group option when it's in an indeterminate state? That should be straightforward to do and I would expect it will resolve Radina's note.

I haven't found multi-select + grouped example in the APG guide, but they have [single-select + grouped](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-grouped/) and the group option doesn't have `aria-selected` at all (it's different from cases discuss above, but it makes me fairly confident that it's not a strict expectation to have `aria-selected` applied on the group in all cases)

Prashant-thakur77:
> > Unless there's a way to completely remove the not selected from the screen reader output
> 
> @radinamatic Technically it will be easy to remove `aria-selected` attribute conditionally in cases when _Partially selected_ is announced, I think Prashant here
> 
> > the not selected part is being announced because of aria-selected="false".
> 
> is trying to ask about if it's fine to not apply `aria-selected` attribute.
> 
> @Prashant-thakur77 I think when exactly none or exactly all child options are selected, there is a value in using `aria-selected` on the group. However I wouldn't apply the attribute on the group as long as it's in an indeterminate state - in such a condition both `true` and `false` values are not accurate and only confusing. Could you conditionally remove `aria-selected` on the group option when it's in an indeterminate state? That should be straightforward to do and I would expect it will resolve Radina's note.
> 
> I haven't found multi-select + grouped example in the APG guide, but they have [single-select + grouped](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-grouped/) and the group option doesn't have `aria-selected` at all (it's different from cases discuss above, but it makes me fairly confident that it's not a strict expectation to have `aria-selected` applied on the group in all cases)

Thanks for clearing this up :) I checked the solution, and it works smoothly now without the unnecessary 'not selected' announcement for indeterminate options.I was initially worried about violating the ARIA guidelines by completely omitting the attribute, but your point about the APG grouped single-select example makes perfect sense:)

MisRob:
:tada: :relaxed: :pray:

MisRob:
> I was initially worried about violating the ARIA guidelines by completely omitting the attribute, but your point about the APG grouped single-select example makes perfect sense:)

Yeah I thought so :)

This is not the case, but more generally, from time to time we even had to deviate a bit from an official guidance or come up with something of our own. It's best to follow the patterns, but they (1) don't cover everything and (2) are generalizations. Thankfully Radina can try everything on the actual device :) Not saying it's not good to follow the patterns as strictly as possible ~ I think it's still important for ensuring good-enough and standard experience across many various assistive technologies, but the actual screenreader output is as close as we can get to the actual experience, so very very important factor :)

Prashant-thakur77:
> One small issue to address - now the select all checkbox is slightly misaligned. Imagine a vertical line. If I recall Kolibri use-cases, all checkboxes, including the select all are aligned precisely, and I think it'd be reasonable default (again to allow for easier customization)

Added 4px padding to the select all too.and updated the documentation use cases to be aligned properly,specifically where customization were done:
Select all Default:
Screenshot From 2026-07-09 20-57-40
Scrollable:
Screenshot From 2026-07-09 20-57-54
Customized
Screenshot From 2026-07-09 20-58-01
Grouped:
image
As @MisRob already pointed there would be a little change in the selectable list in kolibri too to get this aligned later on;)

docs/pages/klistbox.vue
Prashant-thakur77: done

docs/pages/klistbox.vue
Prashant-thakur77: updated:)

Prashant-thakur77:
updated the design of the single select selected state according to the design pref given:)

https://github.com/user-attachments/assets/256c2f48-33cd-4f14-b6ce-7790e5ca2e76

examples/KListbox/Customized.vue
```diff
@@ -114,6 +114,10 @@
     border-bottom: 1px solid;
   }
 
+  .option {
+    padding-inline-start: 16px;
```
AlexVelezLl: Just flagging that padding-inline-start is not supported by our browser list https://caniuse.com/?search=padding-inline-start, so best to keep the examples aligned

examples/KListbox/Scrollable.vue
```diff
@@ -93,6 +93,10 @@
     border-bottom: 1px solid;
   }
 
+  .option {
+    padding-inline-start: 10px;
```
AlexVelezLl: idem

lib/candidate/listbox/KListbox/index.vue
```diff
@@ -426,6 +438,7 @@
   .k-listbox-select-all {
     display: flex;
     align-items: center;
+    padding-inline-start: 4px;
```
AlexVelezLl: idem. Let's just use padding-left and leave the bundler the responsibility for RTL alignment :).

lib/candidate/listbox/KListboxOption/index.vue
```diff
@@ -6,34 +6,47 @@
...
-      :style="{ lineHeight: '24px' }"
+      class="k-option-plain-label"
+      :style="{
+        color: isSelected ? $themeTokens.primary : 'inherit',
+        fontWeight: isSelected ? '500' : 'normal',
+      }"
     >
+      <KIcon
+        icon="done"
+        aria-hidden="true"
```
AlexVelezLl: Hmm, KIcon already renders the icon as decorative; is this aria-hidden really needed?

lib/candidate/listbox/KListboxOption/index.vue
```diff
@@ -130,7 +156,19 @@
     display: flex;
     align-items: center;
     min-height: 36px;
+    padding-inline-start: 4px;
```
AlexVelezLl: idem

lib/candidate/listbox/KListboxOption/index.vue
```diff
@@ -6,34 +6,47 @@
...
-      class="k-listbox-option-label"
-      :style="{ lineHeight: '24px' }"
+      class="k-option-plain-label"
+      :style="{
+        color: isSelected ? $themeTokens.primary : 'inherit',
+        fontWeight: isSelected ? '500' : 'normal',
+      }"
     >
+      <KIcon
+        icon="done"
```
AlexVelezLl, line 44: Ah, we have a little big problem with KIcon 😅, it automatically sets a given `top` value that makes some layouts look a bit misaligned (aligned more to the bottom rather than being in the middle):

Image

Setting a top: 0 style fixes it:

Image

AlexVelezLl (changes requested):
Looks great! Just found a couple of css properties that are not supported by our browserlist, and a small request on our KIcon positioning :).

lib/candidate/listbox/KListboxOption/index.vue
Prashant-thakur77: updated:)

lib/candidate/listbox/KListboxOption/index.vue
Prashant-thakur77: done

lib/candidate/listbox/KListboxOption/index.vue
Prashant-thakur77: done,removed

lib/candidate/listbox/KListbox/index.vue
Prashant-thakur77: done

examples/KListbox/Customized.vue
Prashant-thakur77: done

examples/KListbox/Scrollable.vue
Prashant-thakur77: done

AlexVelezLl (approved):
Code changes look good to me. Thanks, @Prashant-thakur77! I checked the Kolibri build, and yes, we will need to change a small style, but can do it after the next release :)

PR #1296, opened 2026-07-01
Implement the KMultiSelect wrapper and Dropdown components for KMultiSelect

### LINKED ISSUE #1295: Implement the KMultiSelect wrapper and Dropdown components for KMultiSelect

## Overview

We need to implement the final `KMultiSelect` wrapper component and its associated dropdown menu (`KMultiSelectDropdown`). This will integrate the previously built input nodes and data composables into a fully functional, consumer-ready multiselect component.

### Context

Currently, we have successfully done all the foundational prerequisites for this component: the data logic composables (`useMultiSelectList`, `useMultiSelectCascade`, and `useMultiSelectDropdown`), the UI components for the chips (`KMultiSelectInput` and `KMultiSelectNode`), and the core single-select enhancements for `KListbox`. However, despite these pieces existing in the codebase, developers cannot actually use a multiselect component yet. This is because the main `KMultiSelect` wrapper that orchestrates the overall state, and the visual dropdown menu that lists the options for the user to select from, do not yet exist. 

### The Change

We need to build out the core `KMultiSelect` functionality:
- **`KMultiSelect` Wrapper:** Create an `index.vue` wrapper that orchestrates the input node and the dropdown. It should manage the overall component state, handle `v-model` binding (for both single and multiple selection), use the data composables, and proxy events between the input and the dropdown.
- **Dropdown Component:** Create a `KMultiSelectDropdown.vue` component that renders a visually elevated, absolutely positioned menu below the input. It should utilize `KListbox` under the hood, handle its own scrolling constraints, and render an empty state (e.g., "No results found") when a search query yields no matches.
- **Single-Select UX Enhancements:** The composables should be updated so that when a single-select dropdown opens and pre-fills the input with the current selection, it does not immediately filter out the rest of the list.
- **Documentation:** Create the official `docs/pages/kmultiselect.vue` page with comprehensive interactive examples .

#### General
- [ ] The `KMultiSelect` wrapper successfully passes down props and handles `v-model` binding for both single and multiple selection modes.
- [ ] The `KMultiSelectDropdown` opens when the input is focused/clicked, and closes when clicking outside or pressing Escape.
- [ ] Official documentation page is added with interactive examples for all major features.

#### Testing
- [ ] Unit tests added covering wrapper behavior, focus events, and value binding.
- [ ] Unit tests added covering dropdown rendering, filter

## AI usage

I used Claude to help outline the architecture and requirements for the final multiselect components. I reviewed the generated acceptance criteria to ensure they aligned with Kolibri's accessibility standards and the ongoing work from previous PRs, and manually edited the context and scope to fit our issue templates.

Prashant-thakur77:
## Description
This PR introduces the final `KMultiSelect` wrapper component and its associated `KMultiSelectDropdown` menu. This completes the multiselect architecture by integrating the previously merged data composables and input node into a fully functional, consumer-ready component.

#### Issue addressed
Addresses #1295 

### Before/after screenshots

https://github.com/user-attachments/assets/7a7faf3e-a7bf-4758-9858-11b444bc18e2

## Changelog

  - **Description:** Implement KMultiSelect wrapper and Dropdown menu UI.
  - **Addresses:** Addresses #1295 
  - **Components:** KMultiSelect
  - **Breaking:** no
  - **Impacts a11y:** yes
  - **Guidance:** KMultiSelect is now fully usable for consumers. It introduces robust keyboard navigation and aria-live region announcements that handle complex multiselect filtering and state changes natively.

## Steps to test

Check all test cases are covered and passes.

## Testing checklist

## Reviewer guidance

## Comments
This PR relies on the single-select `KListbox` enhancements which are currently in another branch. Once that branch is merged, this branch will cleanly sit on top of it.


Prashant-thakur77:
@LianaHarris360 I think we might need to reconsider our initial approach of keeping the message prop(In Kmultiselect) as optional .Now while completing the final component i do think keeping it required might be better option ,making it similar to as it is done in Klistbox.What do you think?

lib/candidate/multiselect/KMultiSelect/useMultiSelectDropdown.js
```diff
@@ -12,16 +19,16 @@ export default function useMultiSelectDropdown(
     if (!props.multiple && props.value != null && props.value !== '') {
       const opt = normalizedOptions.value.find(o => getOptionValue(o) === props.value);
       if (opt) {
+        if (suppressFilter) suppressFilter.value = true;
```
rtibblesbot: **blocking:** No test anywhere in the diff exercises `suppressFilter`. This is the mechanism behind the issue's explicitly-called-out enhancement: "when a single-select dropdown opens and pre-fills the input with the current selection, it does not immediately filter out the rest of the list." `useMultiSelectList.spec.js`/`useMultiSelectCascade.spec.js` are untouched. `index.spec.js`'s single-mode block only checks that the dropdown closes after selecting. There's no dedicated spec for this composable. The issue's testing acceptance criteria explicitly require "dropdown rendering, filter[ing]" tests. Add a test that opens a single-select `KMultiSelect` with an existing `value`, and asserts the other options are still present (not filtered down) until the user types.

examples/KMultiSelect/MultipleSelection.vue
```diff
@@ -0,0 +1,71 @@
+<template>
+
+  <div>
+    <KMultiSelect
+      v-model="selected"
+      noResultsText="No results found"
+      :messages="messages"
+      :options="allCountries"
+      itemText="name"
```
rtibblesbot, line 9: **blocking:** `itemText`/`itemValue` are swapped relative to the data (`{ id: 'Afghanistan', name: 'AF' }`). The dropdown/chips show ISO codes (`AF`, `AL`, ...) instead of country names. Meanwhile `v-model` binds out the full country name. Confirmed live: selecting the first two options renders chip text `AF` while the page's own "Selected" debug output reads `["Afghanistan"]`. This contradicts the section's own description — "replicating the CountryField used globally in Studio" — since a real CountryField shows names, not codes. Swap the data (`{ id: 'AF', name: 'Afghanistan' }`) or swap the props.

examples/KMultiSelect/SingleSelection.vue
```diff
@@ -0,0 +1,71 @@
+<template>
+
+  <div>
+    <KMultiSelect
+      v-model="selected"
+      noResultsText="No results found"
+      :messages="messages"
+      :options="allCountries"
+      itemText="name"
```
rtibblesbot, line 9: **blocking:** Same `itemText`/`itemValue` swap as `CountryFieldMultiple.vue` — same `allCountries` data, copy-pasted. Dropdown/chip shows ISO codes instead of country names.

docs/pages/kmultiselect.vue
```diff
@@ -1,77 +1,199 @@
...
+    </DocsPageSection>
+
+    <DocsPageSection
+      title="Messages"
+      anchor="#messages"
+    >
       <p>
-        For now, this page demonstrates <code>KChip</code>, which is used to render the individual
-        selected items.
+        <code>KMultiSelect</code> requires a <code>messages</code> object. Each entry must be a
```
rtibblesbot: **suggestion:** This says a `messages` entry "must be a translated string or a function." But `useMultiSelectList.js`'s `getAnnouncement` (and the `useMultiSelectCascade.js` equivalent) only accept functions. A plain string silently produces no announcement, with no error. This PR is flagged as impacting a11y, so it's worth tightening. Either relax `getAnnouncement` to accept strings too, matching `KListbox`'s own `getMessage`, or fix this wording to say functions are required.

docs/pages/kmultiselect.vue
```diff
@@ -1,77 +1,199 @@
...
+      />
+
+      <h3>
+        CountryField (Single)
+        <DocsAnchorTarget anchor="#country-single" />
+      </h3>
+      <p>
+        Demonstrates a single country selection, replicating Studio's
+        <code>RequestForm.vue</code> usage. Uses
+        <code>appearanceOverrides="{ maxWidth: '500px' }"</code> to lock width like KTextbox.
```
rtibblesbot: **nitpick:** Says `appearanceOverrides="{ maxWidth: '500px' }"`, but `CountryFieldSingle.vue` actually uses `350px`. Minor doc/code drift — a reader who opens "view code" will see a different number.

lib/candidate/multiselect/KMultiSelect/index.vue
```diff
@@ -0,0 +1,590 @@
...
+        if (newVal && !props.multiple && suppressFilter.value) {
+          nextTick(() => {
+            inputRef.value?.select();
+          });
+        }
+      });
+
+      function onInputBlur() {
+        inputFocused.value = false;
+        setTimeout(() => {
```
rtibblesbot: **suggestion:** This 150ms `setTimeout` isn't cancelled on unmount. If the component unmounts within the window, the callback still fires `closeDropdown()`/`emit('blur')`. Low risk since it's one-shot, not recurring. Consider storing the id and clearing it in `onBeforeUnmount`, mirroring the listener cleanup already done in `useMultiSelectDropdown.js`.

lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue
```diff
@@ -0,0 +1,278 @@
...
+        const key = props.itemValue;
+        return opt => opt[key];
+      });
+
+      const optionTree = computed(() =>
+        buildTree(props.options, resolveLabel.value, resolveValue.value),
+      );
+
+      const filteredOptionTree = computed(() => {
+        if (!props.hideSelected) return optionTree.value;
```
rtibblesbot, line 130: **nitpick:** `filteredOptionTree` only filters root-level `optionTree.value`, not `node.children`. That's correct for the documented flat-list use case. If `hideSelected` were ever combined with hierarchical options, a selected child would remain visible nested under its parent. No test/example combines the two, so flagging this as a known limitation rather than a bug.

lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue
```diff
@@ -0,0 +1,278 @@
...
+            v-bind="slotProps"
+          ></slot>
+        </template>
+      </KMultiSelectNode>
+    </KListbox>
+
+    <div
+      v-if="decoratedOptionTree.length === 0"
+      class="kmselect-dropdown-empty"
+      role="status"
```
rtibblesbot: ✅ **Resolved** — addressed in the current code.


**praise:** `role="status"` on the empty-results container gets an implicit `aria-live="polite"` region for free. Screen reader users learn a search yielded no matches without extra live-region wiring.

rtibblesbot (changes requested):
PR #1296 review — synthesized from core, frontend, and manual-QA passes. I also re-checked the four commits pushed after those passes ran (`0be9b1ea2`..`e86f7286`). Those commits add KListbox single-select guards, a dev-only `showCheckbox` warning, and klistbox.vue docs cleanup. They look solid and introduce no new issues. Notably, the frontend pass's finding about a hardcoded `rgba(0,0,0,0.05)` selected-row background in `KListboxOption` is already resolved in the current head — it's back to `themeTokens().surface`. That finding is dropped.

CI: lint passes. Frontend unit tests pass. Percy visual tests are pending, which is expected for a new component's baselines.

**Blocking**
- Zero test coverage anywhere for the single-select "don't filter on pre-fill" behavior. This is the specific enhancement the linked issue calls out, and the acceptance criteria explicitly require dropdown filtering tests (see inline comment).
- `CountryFieldMultiple.vue`/`CountryFieldSingle.vue` have `itemText`/`itemValue` swapped relative to the data. The flagship "replicates Studio's CountryField" example shows ISO codes instead of country names, confirmed live (see inline comments).

**Suggestions**
- Docs claim `messages` entries can be plain strings. `getAnnouncement` only accepts functions. A consumer following the docs literally gets a silently-dropped a11y announcement (inline).
- `onInputBlur`'s 150ms `setTimeout` isn't cancelled on unmount (inline).
- `hideSelected` is a real, tested prop with no interactive example in the new docs page. Worth a short example given the issue calls for "interactive examples for all major features."

**Nitpicks**
- `hideSelected` filtering only checks root-level nodes, not `children`. Fine for the documented flat-list use case — flagging as a known limitation (inline).
- Docs prose says `CountryFieldSingle` uses `maxWidth: '500px'`. The example actually uses `350px` (inline).

Manual QA (from the earlier pass): core interactions all work correctly — open/close, single & multi select, cascading, filtering, chip removal, keyboard nav via `aria-activedescendant`, live-region announcements, RTL mirroring. No new axe violations were attributed to this PR.

---

</details>

lib/candidate/listbox/KListboxOption/index.vue
```diff
@@ -59,7 +61,15 @@
       const uid = optionCount++;
       const optionId = `klistbox-option-${uid}`;
 
-      onMounted(() => listbox.registerOption({ id: optionId, value: props.value }));
+      onMounted(() => {
+        listbox.registerOption({ id: optionId, value: props.value });
+        if (process.env.NODE_ENV !== 'production' && props.showCheckbox && !listbox.multiple) {
```
rtibblesbot: **blocking:** No `__tests__` directory exists for `KListboxOption` at all, so neither of this diff's two new behaviors is covered: the dev-mode `console.warn` when `showCheckbox` is true under single-select mode, and the `isSelected`-driven color/fontWeight styling (replacing the old static `lineHeight`). Add a test asserting the warning fires under `multiple: false` + `showCheckbox: true`, and one asserting the selected option picks up the themed color/weight, so this can't regress silently.

lib/candidate/multiselect/KMultiSelect/index.vue
```diff
@@ -0,0 +1,590 @@
...
+            event.preventDefault();
+            openDropdown();
+            return;
+          }
+          event.preventDefault();
+          dropdownComponent.value?.forwardKeydown(event);
+          return;
+        }
+
+        switch (event.key) {
```
rtibblesbot, line 341: **blocking:** No `case ' ':` here or in the `default` branch forwards Space to the listbox—only `Enter` does (via the `fwd(' ')` call in the `Enter` case above). With the dropdown open and an option highlighted via arrow keys, pressing Space types a literal space into the input and re-filters the list instead of toggling the highlighted option, breaking keyboard-only selection despite the visible checkboxes suggesting it should work. Repro: open a dropdown, ArrowDown to highlight an option, press Space — input value becomes `" "`, list filters to no-results, option never selected. This directly contradicts the PR's "robust keyboard navigation" claim.

docs/pages/kmultiselect.vue
```diff
@@ -1,77 +1,199 @@
...
+        exampleId="kmultiselect-language-filter"
+        loadExample="KMultiSelect/LanguageFilter.vue"
+      />
+
+      <h3>
+        CategoryOptions (Dropdown Tree)
+        <DocsAnchorTarget anchor="#category-options" />
+      </h3>
+      <p>
+        Demonstrates a hierarchical tree structure with multi-select and cascading behavior,
```
rtibblesbot: **suggestion:** This line promises "cascading behavior," but manual QA on the CategoryOptionsTree example shows selecting a parent doesn't select its children, and selecting all of a node's children doesn't roll up to check the parent—only deselect cascades downward. The underlying `useMultiSelectCascade` composable predates this PR (merged in #1273), so this isn't a regression to block on, but the new doc copy overclaims what the demo actually does. Consider softening this line's wording or filing a follow-up issue for the composable gap.

lib/candidate/multiselect/KMultiSelect/index.vue
```diff
@@ -0,0 +1,590 @@
+<template>
+
+  <div
+    class="kmselect"
+    :style="appearanceOverrides"
```
rtibblesbot, line 5: **suggestion:** `.kmselect` defaults to `width: 100%`, but every doc example passes a fixed-px `appearanceOverrides` width (e.g. 750px) through this inline `:style`, which wins with no `max-width: 100%` fallback. At a 375px viewport this pushes `scrollWidth` to 437px, causing sideways page scroll on the documented usage pattern. Consider adding `max-width: 100%` to `.kmselect` so a fixed width still respects the viewport.

rtibblesbot (changes requested):
PR #1296 — delta re-review of `e86f7286` (byte-identical to the commit behind my prior CHANGES_REQUESTED review; no fixes have landed since).

0 of 7 prior substantive findings resolved; all 7 still open (see collapsed block). This round's frontend/QA passes also surfaced 2 new blocking issues and 2 new suggestions.

**Still open from prior review:**
- **blocking** `useMultiSelectDropdown.js:22` — no test exercises `suppressFilter` (single-select pre-fill not-immediately-filtering), despite the issue's explicit testing acceptance criteria.
- **blocking** `CountryFieldMultiple.vue:9` / `CountryFieldSingle.vue:9` — `itemText`/`itemValue` swapped; the "CountryField" demo shows ISO codes, not country names.
- **suggestion** `kmultiselect.vue:35` — docs say `messages` accepts plain strings; `getAnnouncement` only ever calls it as a function.
- **suggestion** `index.vue:323` — blur-close `setTimeout` isn't cleared on unmount.
- **nitpick** `kmultiselect.vue:90` — doc says `maxWidth: 500px`, example uses `350px`.
- **nitpick** `KMultiSelectDropdown.vue:126` — `hideSelected` only filters root-level nodes, not children.

**New this round** (inline below):
- blocking: `KListboxOption/index.vue` — zero test coverage for the new dev-warning and selected-state styling.
- blocking: `KMultiSelect/index.vue` — Space key doesn't select the highlighted option; it types into the search box instead.
- suggestion: `kmultiselect.vue:145` — new docs promise full cascading behavior; manual QA shows select-direction cascade/roll-up isn't implemented (root composable predates this PR).
- suggestion: `KMultiSelect/index.vue:5` — fixed-width `appearanceOverrides` (used by every doc example) has no max-width fallback, causing horizontal page overflow on mobile viewports.

CI: unit tests/lint passing; Percy visual tests still pending.

UNADDRESSED — lib/candidate/multiselect/KMultiSelect/useMultiSelectDropdown.js:22 — no suppressFilter test coverage
UNADDRESSED — examples/KMultiSelect/CountryFieldMultiple.vue:9 — itemText/itemValue swapped
UNADDRESSED — examples/KMultiSelect/CountryFieldSingle.vue:9 — itemText/itemValue swapped
UNADDRESSED — docs/pages/kmultiselect.vue:35 — messages docs claim plain strings work
UNADDRESSED — docs/pages/kmultiselect.vue:90 — maxWidth doc says 500px vs 350px actual
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/index.vue:323 — blur setTimeout not cancelled on unmount
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue:126 — hideSelected only filters root-level nodes
ACKNOWLEDGED — lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue:42 — praise: role=status live region (no action needed)

</details>

---

</details>

MisRob:
@Prashant-thakur77 @LianaHarris360 I think I will spare you (and keep enough time for my meditation efforts :)) & so I won't be reviewing here and just keep chiming only in the listbox PR. But wanted to leave a note that I saw the preview and it looks just beyond all the expectations & awesome, wonderful work both :)!

lib/candidate/multiselect/KMultiSelect/useMultiSelectDropdown.js
Prashant-thakur77: Fixed. Added two tests in index.spec.js under KMultiSelect single mode pre-fill (suppressFilter):)

examples/KMultiSelect/MultipleSelection.vue
Prashant-thakur77: Fixed.

examples/KMultiSelect/SingleSelection.vue
Prashant-thakur77: Fixed.

docs/pages/kmultiselect.vue
Prashant-thakur77: updated

docs/pages/kmultiselect.vue
Prashant-thakur77: done

lib/candidate/multiselect/KMultiSelect/index.vue
Prashant-thakur77: done

lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue
Prashant-thakur77: hideSelected is intentionally designed for flat-list use cases only.as there are not such use of hideselected in hierarchical options in studio.

lib/candidate/listbox/KListboxOption/index.vue
Prashant-thakur77: @LianaHarris360 should i add this test in the index.spec.js file of klistbox?

lib/candidate/multiselect/KMultiSelect/index.vue
Prashant-thakur77: This is intentional and correct per the [ARIA APG Combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/). As space is a Printable Character, it must type into the search input, not select an option. The Enter key is the correct selection key for this.

docs/pages/kmultiselect.vue
Prashant-thakur77: fixed

lib/candidate/multiselect/KMultiSelect/index.vue
Prashant-thakur77: done

lib/candidate/multiselect/KMultiSelect/index.vue
rtibblesbot: You're right — for an editable combobox the ARIA APG pattern treats Space as a printable character that belongs in the text input, with Enter as the selection key. I was wrong to expect Space to toggle the highlighted option. Withdrawing this finding.

lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue
rtibblesbot: Makes sense — flagged this as a known limitation rather than a bug, and confirming it's flat-list-only by design is exactly the right call. Thanks.

lib/candidate/listbox/KListboxOption/index.vue
rtibblesbot: On placement — I'll defer to @LianaHarris360's preference, but either works: a co-located KListboxOption/__tests__/index.spec.js keeps the two new behaviors (the single-select + showCheckbox console.warn, and the isSelected color/fontWeight styling) close to the component, while folding them into KListbox's index.spec.js is fine too if you'd rather keep listbox specs together.

LianaHarris360: Yes, it would be fine to add this test to index.spec.js within `KListbox`

Prashant-thakur77:
Hlo @LianaHarris360 I have a little doubt related to the design as the current klistbox implementation allows us to have this:And this is also after i have added the https://github.com/learningequality/kolibri-design-system/pull/1296/changes#diff-067f16fea4419103da0d5de5c195056e18080c9f3097fb06d042c0b34c06c621R278-R282 
Code:
```js
// Without this, if the user navigates to the component without tabbing into it,
// arrow keys do move through the dropdown options but no visual focus marker
// is shown on the active option.
globalThemeState.inputModality = 'keyboard';
```
without this addition there was no focus or scrolling of the dropdown:

https://github.com/user-attachments/assets/2f18af7e-7708-4b3e-b9f1-f8fe6d7b423c

But if we want to make it similar to studio then we have to update a little in klistbox and the klistbox option file and the result would be like this(This would mak e some little changes in klistbox use cases desgin too in the focused state)

https://github.com/user-attachments/assets/6656ddf9-cd9b-487b-940d-4bc1310c97f4


How it is in studio:

https://github.com/user-attachments/assets/2a27352b-f403-4fa4-b275-9f876d8eac5a

LianaHarris360:
I'm not completely sure I understand the difference. Is it that one way shows the focus outline when moving through the options with the keyboard, and the other way does not show the outline? If that's the case, I think the preferred experience is for the focus outline to always be visible when navigating using the keyboard, rather than relying only on the darkened background, but Radina might have the best insight on this. cc @radinamatic

Prashant-thakur77:
Also this is how currently we are showing the error state:

https://github.com/user-attachments/assets/993b0913-7c59-4a6f-8c1b-2b8c6039638f

I f you want i can add this example in the kmutlsectl if we want to check it out more and also the hideselected one(That is going to be used with Vcombobox where we can create custom chips but still i have checked if it is functioning well):
This is when noresulttext is passed:

https://github.com/user-attachments/assets/c5547bd6-6ce4-4afb-ae57-3203e872fed4

And if not:

https://github.com/user-attachments/assets/560fbfde-3df3-4883-af4c-8e1f3b522bf5

Prashant-thakur77:
> Is it that one way shows the focus outline when moving through the options with the keyboard, and the other way does not show the outline?

yes

LianaHarris360:
The error state example looks great! For your example when `noresulttext` is passed, the text appears after all of the selections have been made. In the Studio implementation, it seems like the no results text should only appear if a user types something but no matching results are found:

https://github.com/user-attachments/assets/2be83265-f10a-4031-8ef9-17877259eac0

Prashant-thakur77:
> The error state example looks great! For your example when `noresulttext` is passed, the text appears after all of the selections have been made. In the Studio implementation, it seems like the no results text should only appear if a user types something but no matching results are found:

Updated accordingly:now it only shows the noresult text on typing and just not show the dropdown at all when all the options are selected.

https://github.com/user-attachments/assets/b89f94b6-061f-422a-9f03-7c14366dd956

lib/candidate/listbox/KListboxOption/index.vue
Prashant-thakur77: done

rtibblesbot: Confirmed — the new `KListboxOption` describe block in `KListbox/index.spec.js` asserts the selected option picks up the themed font-weight and the visible checkmark, so the `isSelected`-driven styling is now covered against regression. Thanks for adding it. Resolving.

lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue
MisRob: For live regions functionality please use https://design-system.learningequality.org/usekliveregion (guidance has more information on why)

rtibblesbot: Thanks for the pointer to `useKLiveRegion` — noted. I'll defer to that as the canonical approach for live-region behavior in KDS and reference the guidance going forward.

Prashant-thakur77: done

lib/candidate/multiselect/KMultiSelect/index.vue
```diff
@@ -0,0 +1,725 @@
...
+
+          case 'Enter':
+            event.preventDefault();
+            event.stopPropagation();
+            if (!isOpen.value) {
+              openDropdown();
+            } else {
+              const fwd = key =>
+                dropdownComponent.value?.forwardKeydown(
+                  new KeyboardEvent('keydown', { key, bubbles: true, cancelable: true }),
```
LianaHarris360: This keyboard event is created so that the `moveFocusBy` and `toggleOption` functions inside `KListbox` are called? Maybe we wouldn't need to forward the key if those functions were exposed by `KListbox`, and were directly called in this file instead?

lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue
```diff
@@ -0,0 +1,349 @@
...
+          }
+        },
+      );
+
+      const showEmptyState = computed(() => {
+        return (
+          props.isOpen &&
+          decoratedOptionTree.value.length === 0 &&
+          Boolean(props.searchText) &&
+          (Boolean(props.noResultsText) || Boolean(instance.proxy.$scopedSlots['no-results']))
```
LianaHarris360, line 188: This piece of code is also used on lines 5-7, and lines 45-47, perhaps you could create a computed function and re-use that to clean it up/make it more readable?

lib/candidate/multiselect/KMultiSelect/index.vue
```diff
@@ -0,0 +1,725 @@
...
+    name: 'KMultiSelect',
+
+    components: {
+      KMultiSelectInput,
+      KMultiSelectDropdown,
+    },
+
+    setup(props, { emit }) {
+      const instance = getCurrentInstance();
+      const uid = instance.proxy._uid;
```
LianaHarris360: This isn't a problem as-is, but is there any reason not to reuse the same approach as [KListboxOption](https://github.com/learningequality/kolibri-design-system/blob/c0295416e9047dbe8512368c0456239bc42b9b32/lib/candidate/listbox/KListboxOption/index.vue#L71) of setting a counter directly within the module:
``` 
  let optionCount = 0;

      const uid = optionCount++;
      const optionId = `klistbox-option-${uid}`;
```
Mostly because in Vue 3, `_uid` is moved to `instance.uid` instead, and since this is used for the screenreader component Ids, it might cause silent failures after a future migration.

LianaHarris360 (dismissed):
I had a few comments to improve the code quality, but didn't see anything major, thanks @Prashant-thakur77 !

lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue
Prashant-thakur77: I've updated both to use showEmptyState directly and reused it.

lib/candidate/multiselect/KMultiSelect/index.vue
Prashant-thakur77: Updated as mentioned above

lib/candidate/multiselect/KMultiSelect/index.vue
Prashant-thakur77: done

radinamatic:
Still need to perform a few more QA rounds, but here are my first impressions:

1. **CountryField multi select** looks and is performing well, but it might be missing some additional screen reader output in cases when user makes the selection(s), but then exits the component. Upon return inside the component, the focus lands directly on the button of the first selected item (`Remove 'item label' button`) without any other context, which can be confusing depending on how much navigating the user did before re-entering the component. 

Win11 2606 (start)  Running  - Oracle VirtualBox_012

Could we add some screen reader output before that button is announced? Something like name/label of the multi select and how many items are already selected? So instead of the flat output from the screenshot above:

`clickable Remove Albania button `

User would get better context with something like the following upon re-entering the component:

```
Countries  combo box 4 items selected
clickable Remove Albania button
```

2. The above is also valid for the **CountryField (Single)**, as navigating back to the component that already has an item selected does not inform the user which one has been selected. [Combobox example on the APG patterns page](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/) also suggests that the selected item should be a part of the screen reader output.

Win11 2606 (start)  Running  - Oracle VirtualBox_013

radinamatic:
There is no corresponding pattern on the APG website for the combobox with multiple selections and autocomplete, but I think it may be better if we cleared the previous typed search once the user has made their selection. As of now it still hangs back in the field and can cause unnecessary _spelling errors_ for the user who tries to type in a new search not expecting that there are still some characters from the previous search.

Win11 2606 (start)  Running  - Oracle VirtualBox_006

Prashant-thakur77:
@radinamatic @LianaHarris360 Changes made:
1) Better screen reader context on re-entry: When a user tabs back into the component, screen reader now announces the component label and selection count (multi-select) or selected value (single-select).

2) Search input clears after chip selection: After selecting an item in multi-select, the typed search text is automatically cleared so the user can immediately start a fresh search.

3) Auto-select parent when all children are selected: In tree/hierarchical mode, when a user selects all children of a parent, the parent is automatically selected.


4) Implemented new autoSelectChild prop: When passed, selecting a parent node automatically selects all its descendants at every level.Currently used in CategoryOption example in docs

5) Leading/trailing spaces ignored in search: Spaces at the start or end of the search query are trimmed before filtering, preventing confusing empty results.

https://github.com/user-attachments/assets/7ee177d6-17ab-4207-9719-cc60bdd83ceb

radinamatic:
- Better screen reader context on re-entry: Working much better for multi-select, thank you! ✔️ 🎉 👏🏽 

   If I can suggest one last improvement for single-select: the selected value is now being announced, but the output would be better if the `selected` state was more explicit, so instead of:
   
   ``Country Mexico  combo box  collapsed  has auto complete  editable  blank``
   
   Win11 2606 (start)  Running  - Oracle VirtualBox_020

   Let's make it:
   
   ``Country combo box  collapsed  has auto complete  editable  selected Mexico``
   
   Similar as in the [example on APG patterns](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/#ex_label) page:
   
  Win11 2606 (start)  Running  - Oracle VirtualBox_019

   Another thing that might need some checking is why in the KDS single select example for countries above there is `Country Mexico grouping` output in NVDA? 🤔 There are no grouped options in this single select, right?
   
   Both of these comments (more explicit `selected` state and the unclear _grouping_) are valid for the examples LanguageDropdown (Single), and ActivityDuration (both)
   

- Search input clears after chip selection ✔️ 
- Auto-select parent when all children are selected ✔️
- Implemented new autoSelectChild prop ✔️
- Leading/trailing spaces ignored in search ✔️

Prashant-thakur77:
Thanks for the feedback @radinamatic 
Both issues should now be fixed.
For the selected state, the value now lives in the input itself instead of a separate element, so NVDA should now announce something like:
> **Country** combo box collapsed has auto complete editable selected **Mexico**

if mexico is already selected.
Could you verify this once and let me know if the announcement matches what you hear?

For the "Country Mexico grouping" announcement, The component was always rendering a `role="group"` wrapper, which caused NVDA to announce it as a group. That wrapper is now only used for multi-select, so single-select should no longer announce any grouping.So this is also fixed

The same fixes apply to all affected examples.

radinamatic:
Excellent @Prashant-thakur77, all single selection examples now render the selected value more explicitly! 🎉 

   ``Country combo box  collapsed  has auto complete  editable  selected Mexico``

Win11 2606 (start)  Running  - Oracle VirtualBox_025


As for the `grouping` output, it is not announced anymore in all single selection examples, and that's perfect 👍🏽 

But what I'm thinking is that the `role="group"` wrapper that is producing that output is actually not necessary (and may be confusing) in multi select examples **UNLESS** there are indeed some grouped options. As of now the `grouping` output is announced upon re-entering all multi select examples by default, when it is necessary and informative only for the **CategoryOptions (Dropdown Tree)** that actually has grouped options. If we had countries grouped by continents, or languages by origin (Romanic, Indo-Aryan, Sinitic, etc.), it would make sense, but in the absence of grouped option items, we should remove the `role="group"` wrapper for the multi select examples too.

Win11 2606 (start)  Running  - Oracle VirtualBox_024

Prashant-thakur77:
Thanks for checking this again @radinamatic @LianaHarris360 

I took a closer look, and there are actually two different `role="group"` elements here:

1) The wrapper around the chips and input, which is present in every multi-select and is what gets announced when you re-enter the field.
2) The groups inside the dropdown itself (like in CategoryOptions), which are what announce grouped options while navigating the list.

So removing the wrapper wouldn't affect the grouped-option announcements in CategoryOptions as those come from the dropdown groups, not the wrapper.

The only catch is that the wrapper is also what provides context before the first chip. Without it, re-entering lands straight on `"Remove Andorra button"`, and you don't hear `"Countries, 3 items selected"` until a few tab stops later where you reach the input box

One alternative would be to put that context on each chip instead (e.g. `"Remove Andorra button, Countries, 3 items selected"`). That removes the grouping announcement while keeping the context, though it would be repeated for every chip.

So I think the options are:

* Remove the wrapper: no grouping, but no context before the first chip.
* Keep it: context on entry, but with the grouping announcement.
* Move the context to each chip as a description: no grouping, context preserved, but it comes after the chip name and repeats on every chip.
* Move the context to the chip's label, only on entry, no grouping, context first, not repeated.
One issue that might come in option 4: it relies on focus actually moving, so it behaves as above when tabbing. In browse mode, where the virtual cursor moves without changing focus, the context may still be read on each chip,but might need to verify this

I am also looking for some other way to resolve this but this is what i have comeup with yet.:)

lib/candidate/multiselect/KMultiSelect/__tests__/index.spec.js
```diff
@@ -0,0 +1,275 @@
...
+  });
+
+  it('does not open when disabled=true', async () => {
+    renderKMultiSelect({ disabled: true });
+    await userEvent.click(getToggle());
+    expect(getListbox()).toBeNull();
+  });
+});
+
+describe('KMultiSelect — keyboard dispatch (onInputKeydown)', () => {
```
LianaHarris360: Can you update the test descriptions that reference internal functions names and remove them to be more general? We want to describe user-visible behavior, rather than the exact implementation. Also, if the function names ever change, the test descriptions might not be updated, but will continue to pass and reference old internal names. 

For example, this test description could be renamed to simply test "keyboard interactions". These function names should be removed from the descriptions as well:
`onSearchInput`
`onInputBlur`
`computedListboxMessages`
`suppressFilter`

lib/candidate/multiselect/KMultiSelect/index.vue
```diff
@@ -0,0 +1,757 @@
...
+
+      let blurTimer = null;
+      function onInputBlur() {
+        inputFocused.value = false;
+        blurTimer = setTimeout(() => {
+          if (containerEl.value && !containerEl.value.contains(document.activeElement)) {
+            closeDropdown();
+            emit('blur');
+          }
+        }, 150);
```
LianaHarris360: How did you decide upon 150? Is it to wait for the focus to move? An alternative could be to use `requestAnimationFrame(() => ... )` and wait until the next frame, if it isn't enough time and we really need the 150ms, to wait for the focus, it should be defined in a variable.

lib/candidate/multiselect/KMultiSelect/__tests__/index.spec.js
```diff
@@ -0,0 +1,275 @@
...
+    await userEvent.click(getToggle());
+    expect(screen.getByRole('listbox')).toBeInTheDocument();
+    await userEvent.tab();
+    await waitFor(() => expect(getListbox()).toBeNull(), { timeout: 500 });
+  });
+
+  it('emits blur after focus leaves the container', async () => {
+    const component = renderKMultiSelect();
+    getCombobox().focus();
+    getCombobox().blur();
```
LianaHarris360: Could we use `await userEvent.click(getCombobox())` and `await userEvent.tab()` here instead?

lib/candidate/multiselect/KMultiSelect/__tests__/KMultiSelectDropdown.spec.js
```diff
@@ -0,0 +1,172 @@
...
+    expect(screen.getByRole('option', { name: 'Apple' })).toBeInTheDocument();
+    expect(screen.getByRole('option', { name: 'Banana' })).toBeInTheDocument();
+    expect(screen.getByRole('option', { name: 'Vegetable' })).toBeInTheDocument();
+    expect(screen.getByRole('option', { name: 'Carrot' })).toBeInTheDocument();
+  });
+
+  it('renders children indented under their parent (depth > 0)', async () => {
+    renderDropdown({ options: HIERARCHICAL_OPTIONS });
+    const apple = await screen.findByRole('option', { name: 'Apple' });
+    expect(apple).toHaveStyle({ paddingInlineStart: '40px' });
```
LianaHarris360: This is a nitpick, but these tests will fail if these numbers ever change, but the tree would still be correct. Could we possibly assert that the `child padding > parent padding` instead? Or at least move the exact pixel values into a constant at the top of the file. Same for the two `KMultiSelectNode` tests.

LianaHarris360 (commented):
Thanks for adding tests across all of the composables and `KMultiSelect` components, they look solid. Could you look over the Kolibri [frontend unit testing docs](https://kolibri-dev.readthedocs.io/en/develop/frontend_architecture/unit_testing.html#unit-testing) and align the specs with a few of the conventions there? Specifically the string related items and using `userEvent` over direct calls.

There's some hardcoded strings that are repeated throughout the `KMultiSelect/__tests__/index.spec.js` file, that could be moved to a constant to be reused. There also some regex text matching in `KMultiSelectInput_spec.js` that should check the exact, expected string.

None of this truly blocking, but might be worth cleaning up before we merge this in.

lib/candidate/multiselect/KMultiSelect/__tests__/index.spec.js
Prashant-thakur77: done:)

lib/candidate/multiselect/KMultiSelect/index.vue
Prashant-thakur77: I have switched to requestAnimationFrame instead as you suggested above. I checked after implementing it and it was working as expected.

lib/candidate/multiselect/KMultiSelect/__tests__/KMultiSelectDropdown.spec.js
Prashant-thakur77: I've moved the pixel values into a constant at the top of the file:) I went with the constant helper rather than asserting child padding > parent padding because the relational check could still pass even if the values were visually broken for eg: if they accidentally shrank to 1px increments, the tree structure would still be technically correct but the indentation would be imperceptible to users.

lib/candidate/multiselect/KMultiSelect/__tests__/index.spec.js
Prashant-thakur77: I've switched the focus part to await userEvent.click(getCombobox()).
I did try await userEvent.tab() for moving focus away, but it doesn't quite match the browser behaviour here. Pressing Tab closes the dropdown and moves focus out of the component in one step, whereas in jsdom userEvent.tab() still lands on the listbox as the dropdown is closing, so it takes an extra Tab to actually leave. Since this test is only checking that blur is emitted when focus leaves the component, I used await userEvent.click(document.body) instead and added a short comment explaining why.

Prashant-thakur77:
@LianaHarris360 @radinamatic I have done the required changes in the docs and the names are as follows:
CountryFieldMultiple	to   Multiple selection
CountryFieldSingle	        to   Single selection
LanguageFilter	        to   Multi-field search
CategoryOptionsTree	to   Hierarchical tree
ActivityDurationShort      to    Primitive values
LanguageDropdownMultiple.vue this is deleted
LanguageDropdownSingle.vue this is deleted
ActivityDurationLong.vue this is deleted
what are your thoughts on this:)

lib/candidate/multiselect/KMultiSelect/useMultiSelectDropdown.js
rtibblesbot: Confirmed — the new `KMultiSelect — single mode pre-fill` block covers both halves: the option count stays at `OPTIONS.length` when a single-select dropdown opens with `value: "apple"` pre-filled, and filtering kicks back in once the user types (`ban` → one option, `Banana`). That was the issue’s called-out enhancement, so this closes the gap. Thanks. Resolving.

Prashant-thakur77:
And i have to also create the kmultiselectvisual test files,after the above is finalized:)

radinamatic:
@Prashant-thakur77 Thank you for renaming and cleaning up the list of examples, it's looking much better! 👍🏽 

On another hand, retesting the _Hierarchical tree_ I bumped into this inconsistency: while selecting only one last level item makes the parent checkbox **partially selected**, if all last level items are selected (parent checkbox is **fully selected**), and than the user deselects one of them, the parent checkbox does not change into **partially selected**.

Selecting | Deselecting
-- | --
Win11 2606 (start)  Running  - Oracle VirtualBox_039 | Win11 2606 (start)  Running  - Oracle VirtualBox_037
Win11 2606 (start)  Running  - Oracle VirtualBox_038 | Win11 2606 (start)  Running  - Oracle VirtualBox_036

Prashant-thakur77:
Hey @radinamatic, thanks for catching this🙏
I have fixed this and now when you deselect one child of a fully-selected parent, the parent goes back to the indeterminate state instead of staying fully checked.
While digging into it I also ran into a related case and fixed it: if you'd checked a parent yourself and then unchecked one of its children, the parent used to get wiped along with it. Now it stays checked, since you picked it on purpose ,only parents that got auto-checked (because all their children were selected) drop back to partial.
Let me know if it looks right on your end when you get a chance!
Also @LianaHarris360 ,these updates are done in the usemultiselectCascade files and ,some test are added to its repective file only:)

https://github.com/user-attachments/assets/132f1d8a-0826-4fe6-b156-412cc54b13b6

Prashant-thakur77:
@LianaHarris360 I have added the visual suite for kmultiselect:)

radinamatic:
@Prashant-thakur77 Not sure what went wrong, but there's still an issue with parent checkbox state: when a parent is checked directly, its state does not update when one or more children items are unchecked. Observed on both Firefox and Chrome on Windows 11 and 10.


https://github.com/user-attachments/assets/ac058446-02e6-4b1e-bd17-d149d0ce9a47

Prashant-thakur77:
Hi @radinamatic @LianaHarris360 , thanks for testing again. I missed the guard which caused this :)

The rule that keeps a directly-checked parent checked was leaking into the cascading tree example (Hierarchical tree first example), where it doesn't belong, so now I added a guard so it only runs when autoSelectChild is off. So unchecking a child of a directly-checked parent drops it to indeterminate.

While at it I also added two new examples: one Independent selection, that is what we are going to have in category options, and Automatic parent selection which symbolizes autoSelectParent prop usage, and there is one which you already checked, the first one in Hierarchical tree group with the autoSelectChild prop.
what are your thoughts on these 3 usecases?

I also added a dropdown flip when space is not enough,what are your thoughts on this:)

https://github.com/user-attachments/assets/0bc35f91-c478-4141-ae97-17672c3e6bdf

Also, adding on to my discussion with @LianaHarris360 in the last meet, about parent selection behavior in CategoryOptions:In the old component, selecting all children of a category did not tag the parent. The parent only looked checked in the dropdown, but was never written to the stored data. KMultiSelect currently does auto-select the parent in this case, which means it would also get saved.

I think the old behavior was intentional: each category tag is an explicit statement by the curator, and since category IDs are hierarchical paths, a resource tagged with the children is already discoverable under the parent. 

So the question is: should selecting all children of a category implicitly tag the parent? One of the solution i can think of is an opt-out prop in KMultiSelect so CategoryOptions can disable the auto-selection, keeping the stored data exactly what the curator clicked. Or we could keep the current behavior as is, since auto-selecting the parent does reflect what's actually happening. Would like your inputs on this, what do you think?

Would appreciate a re-check when you get a chance!

docs/pages/kmultiselect.vue
rtibblesbot: Confirmed on the latest commit — this one's closed from my side.

`useMultiSelectCascade.js` now derives parent state in both directions: a parent is promoted when `directChildren.every(c => resultSet.has(c))`, and demoted back to indeterminate when that stops holding, with the `!props.autoSelectChild && !derivedParents.has(ancestor)` guard so a directly-checked parent keeps its own state only in the non-cascading modes. The docs section no longer says "cascading" generically either — it splits into Cascading selection / Independent selection / Automatic parent selection with the parent-state rule spelled out per mode, which matches what the code does. My original point was that the copy overclaimed relative to the demo; that gap is gone. Resolving.

LianaHarris360:
To confirm my understanding: in the current Studio code, when all child categories are selected in `CategoryOptions`, the parent appears checked in the UI but is not included in the stored data. Is that correct?

And if a child resource can already be found under its parent without the parent being explicitly tagged, there's a chance that auto-saving the parent will duplicate information.

Whether to leave `KMultiSelect` as-is (auto-selecting and saving the parent) depends on the side effects this might cause across Studio, in every place `CategoryOptions` is used. If we want the flexibility to change this, while also keeping the option to revert to the old behavior in case of unforeseen side effects, then adding an opt-out prop would be a useful addition.

Prashant-thakur77:
> To confirm my understanding: in the current Studio code, when all child categories are selected in `CategoryOptions`, the parent appears checked in the UI but is not included in the stored data. Is that correct?

yes that is correct,just a demo below for reference;

https://github.com/user-attachments/assets/56734bb7-b0c2-407b-b256-a53a03addec3

Prashant-thakur77:
@LianaHarris360 I have added the opt-in prop and also implemented it in the independent selection example under Hierarchical tree.

LianaHarris360:
Thanks @Prashant-thakur77, it looks great! I tried running the Percy visual test, but realized that the KMultiSelect visual test still needs to be registered into `/jest.conf/visual.load-test-components.js`, step 4 under the [Create visual tests for a component](https://github.com/learningequality/kolibri-design-system/blob/develop/dev_docs/07_visual_testing_guide.md#create-visual-tests-for-a-component) section in the dev docs

Prashant-thakur77:
> Thanks @Prashant-thakur77, it looks great! I tried running the Percy visual test, but realized that the KMultiSelect visual test still needs to be registered into `/jest.conf/visual.load-test-components.js`, step 4 under the [Create visual tests for a component](https://github.com/learningequality/kolibri-design-system/blob/develop/dev_docs/07_visual_testing_guide.md#create-visual-tests-for-a-component) section in the dev docs

I have added it now;) thanks @LianaHarris360

examples/KMultiSelect/MultipleSelection.vue
```diff
@@ -0,0 +1,72 @@
...
+    { id: 'NG', name: 'Nigeria' },
+    { id: 'GB', name: 'United Kingdom' },
+    { id: 'US', name: 'United States' },
+  ];
+
+  export default {
+    components: { KMultiSelect },
+    data() {
+      return {
+        selected: [],
```
LianaHarris360: Can you add a few pre-selected items to this array, to allow the `KChip` element to show in the visual test as well? If you don't want it to effect the example shown in the KDS docs, you could also pass a prop of pre-selected items to assign to `selected`.

lib/candidate/multiselect/KMultiSelect/__tests__/components/KMultiSelectVisualTest.vue
```diff
@@ -0,0 +1,31 @@
...
+    />
+    <VisualTestExample
+      title="Multi-field search"
+      width="400px"
+      loadExample="KMultiSelect/MultifieldSearch.vue"
+    />
+    <VisualTestExample
+      title="Hierarchical tree"
+      width="400px"
+      loadExample="KMultiSelect/HierarchicalTree.vue"
```
LianaHarris360, line 29: Can you add an id attribute to this Hierarchical tree example, and use it to query and [click](https://github.com/learningequality/kolibri-design-system/blob/develop/dev_docs/07_visual_testing_guide.md#simulate-interactions
) the element, in the KMultiSelect `visual.spec.js`? 

This should allow the dropdown to be displayed in the Percy snapshots. You might have to explicitly include the [height](https://github.com/learningequality/kolibri-design-system/blob/d3321cb187f0b47dcbb071d405185a1ca00ef2d9/jest.conf/components/VisualTestExample.vue#L68) attribute as well, to ensure the drop-down doesn’t cover the element above it.

examples/KMultiSelect/MultipleSelection.vue
Prashant-thakur77: Done,and i dont think so there is any problem with the chips if they get rendered in the docs.:)

lib/candidate/multiselect/KMultiSelect/__tests__/components/KMultiSelectVisualTest.vue
Prashant-thakur77: done,added the click ,id and the height attribute:)

LianaHarris360:
I can confirm that the `KMultiSelect` visual tests are good and my previous code review comments have been resolved. Once @radinamatic approves the wording and additional examples added to the "Hierarchical tree" section, this will be ready to merge in!

KMultiSelectVisualTest

radinamatic:
I have 2 comments to the latest changes, one conceptual, and the other nitpicky 😅 

1. In both **Cascading** and **Independent** selections, parent is marked as indeterminate if some of its children are not checked. In the **Automatic parent** selection, there does not seem to be an **_indeterminate_** state, and that seems inconsistent to me. 

   I understand the need to have a workflow where _selecting a child also selects all of its ancestors_, but would using a _more accurate **selection marker**_ for parents/ancestors that have more than one child, namely the **_indeterminate_** one, for some reason be not appropriate? 

Cascading | Independent | Automatic parent
-- | -- | --
Selection_103 | Selection_104 | Selection_105

2. The order of categories in the tree (dropdown) goes from more generic (_oldest_ ancestor) to more specific, but the chips of selected ones in the edit field appear in the opposite order: from the most specific to more generic. Wouldn't it be more consistent if the chips also appeared in the generic-to-specific order?

   Selection_106

Prashant-thakur77:
https://github.com/user-attachments/assets/abf4d44c-60b0-4463-9d95-bde9083ad134

Hello @radinamatic @LianaHarris360  ,The indeterminate state is now there in Automatic parent selection too, a parent shows the dash while only some options under it are selected, and turns checked once all of them are.
And the chip order is fixed, they now follow the same generic-to-specific order and the docs and tests are updated too. Would appreciate a re-check when you get a chance:)

radinamatic:
Well done, @Prashant-thakur77, both indeterminate state in Automatic parent selection, and the  chip display the same generic-to-specific order are now present! 👏🏽 

consistent-order

LianaHarris360 (approved):
Everything is looking good on my end, the `KMultiSelect` turned out great! Thank you for all the hard work you put into this @Prashant-thakur77

PR #1303, opened 2026-07-16
Expose theme values as CSS variables 

### LINKED ISSUE #1300: Emit theme values as CSS variables on `:root`

## Overview

Make theme values available as CSS variables so styles can reference them directly with `var(--tokens-X)`, without going through Aphrodite, kept in sync with `globalThemeState` at runtime. 
 
**Complexity:** Medium
**Target branch:** `release-v5`

### The Change
- Add a runtime module that writes `--tokens-*`, `--palette-*`, and `--brand-*` variables to the root element, derived from `globalThemeState`:
  - `--palette-*`: raw palette colors
  - `--brand-*`: LE brand references
  - `--tokens-*`: semantic tokens 
- Naming: CSS variable names will include the layer prefix + object path, with the dots replaced by hyphens.
	- Theme token names are used as-is (`primary` -> `--tokens-primary`, `focusOutline` -> `--tokens-focusOutline`). 
	- Version keys are emitted as `vN` (the JS keys will remain `v_N` until renamed in later issue): `palette.grey.v_500` -> `--palette-grey-v500`, `brand.primary.v_600` -> `--brand-primary-v600`. 
- Set up a watcher, initialized once when theming initializes, that listens for changes to `globalThemeState` and updates the matching `:root` variables. 
- SSR: write the initial variable values into a `<style>` tag in the document head so the first paint is themed before JavaScript runs. On the client, the watcher updates that same tag.

### Out of Scope
- Removing `computedClass`, `generateGlobalStyles`, or the Aphrodite dependency 
- Migrating component styles to use the new variables.

#### General
- [ ] `--palette-*`, `--brand-*`, and `--tokens-*` variables are set on `:root` after theming initializes.
- [ ] Theme variable names match the existing token names; version keys are emitted as `vN` (e.g., `--palette-grey-v500`).
- [ ] Calling `setBrandColors()` or `setTokenMapping()` updates the matching root variables.
- [ ] With SSR, the initial variable values are in a head `<style>` tag before JavaScript runs, and the watcher updates that tag on the client.

### Testing
- [ ] Unit tests cover initial emission, updates on brand/token changes, and the SSR case.

## AI usage
I used Claude to help draft this issue from my notes and implementation proposal. I reviewed and edited it for accuracy.

LianaHarris360:
## Description
This PR adds the file `themeCSSVariables.js`, and emits theme values (`--tokens-*`, `--brand-*`, and `--palette-*`) as CSS variables in a `:root` rule so they can be referenced directly with `var(--tokens-X)`, without going through Aphrodite. With server-side rendering, the initial values are rendered into a head <style> tag so the content is themed before any JavaScript runs.

#### Issue addressed

Addresses #1300 

## Changelog

  - **Description:** Theme values are now emitted as CSS variables (`--tokens-*`, `--brand-*`, and `--palette-*`) in a `<style id="k-theme-css-variables">` tag in the document head, initialized by `KThemePlugin`, and kept in sync with `globalThemeState` at runtime. Version keys are emitted as `vN` (e.g. `palette.grey.v_400` -> `--palette-grey-v400`); token names are used as-is (e.g. `--theme-primary`). Also exports `themeCssVariablesText()` from `lib/styles/themeCssVariables` for SSR head injection.
  - **Addresses:** [Emit theme values as CSS variables on :root #1300](https://github.com/learningequality/kolibri-design-system/issues/1300)
  - **Components:** N/A
  - **Breaking:** no
  - **Impacts a11y:** no
  - **Guidance:** Styles can now reference theme values directly with `var(--theme-primary)`, `var(--brand-primary-v600)`, `var(--palette-grey-v400)`, etc., as an alternative to `$themeTokens`/`$themeBrand`/`$themePalette`. Existing APIs are unchanged.

## Steps to test

1. Run `yarn dev` and open the docs.
2. In devtools, confirm that the document head contains `<style id="k-theme-css-variables">` with a `:root` rule that defines `--palette-*`, `--brand-*`, and `--tokens-*` variables, and that the variables accurately correspond to the token mapping found in the files `colorsDefault.js` and `colorsMaterial.js`.
3. View the page source (SSR output) and confirm the same `<style>` tag with the variables is present in the head markup itself, i.e. before any JavaScript runs.
4. Navigate between doc pages client-side and confirm that the tag remains present and unchanged (to verify that vue-meta head updates don't remove or rewrite it).

- `lib/styles/themeCssVariables.js`: draws the variables from `globalThemeState` (`v_N` -> `vN` key formatting for palette and brand values; tokens resolved through `tokenMapping`), serializes them to a `:root` rule, and writes it to a `<style id="k-theme-css-variables">` tag in the document head.
- `initThemeCssVariables()` is called once from `KThemePlugin`. A `$watch` on the serialized CSS text re-runs whenever any brand color or token mapping it derives from changes, and rewrites the tag.
- `setTokenMapping()` in `theme.js` now uses `Vue.set` instead of `Object.assign`, so tokens added at runtime are reactive and picked up by the watcher.
- The Nuxt plugin registers the initial CSS as a head style entry on the server only. On the client, the watcher takes over the server-rendered tag and strips the vue-meta's marker attributes (`data-n-head`, `data-hid`) so vue-meta never removes it or rewrites it with stale CSS.
- Token resolution is intentionally duplicated from `theme.js`. The resolved tokens object there is non-observable, so reading it would not register the reactive dependencies relied upon by the watcher. The values are also checked against characters that would corrupt the emitted CSS.

- The token resolution logic exists in both `theme.js` and `themeCssVariables.js` because it is necessary for reactivity, documented in a code comment.
- The `v_N` -> `vN` reformatting will be removed after the JavaScript keys are renamed.

## Testing checklist

## Reviewer guidance

## Comments
I verified the SSR head output, hydration takeover of the SSR tag, and client-side navigation against the docs app in addition to the unit tests. I used Claude 🤖 to do a review of my initial changes and updated the code based on its findings.

### DIFF
--- a/docs/pages/installation.vue
+++ b/docs/pages/installation.vue
@@ -1,4 +1,5 @@
 <template>
+
   <DocsPageTemplate>
     <DocsPageSection
       title="1. Install the plugin"
@@ -94,4 +95,5 @@
       </p>
     </DocsPageSection>
   </DocsPageTemplate>
+
 </template>
--- a/docs/pages/installation.vue
+++ b/docs/pages/installation.vue
@@ -42,10 +42,10 @@
         </li>
         <li>Globally registers all KDS Vue components.</li>
         <li>
-          Emits theme values as CSS variables (<code>--theme-*</code>, <code>--brand-*</code>, and
+          Emits theme values as CSS variables (<code>--tokens-*</code>, <code>--brand-*</code>, and
           <code>--palette-*</code>) to a <code>&lt;style&gt;</code> tag
           <code>#k-theme-css-variables</code> in an application's document head so that styles can
-          reference them directly, for example <code>var(--theme-primary)</code>. The variables are
+          reference them directly, for example <code>var(--tokens-primary)</code>. The variables are
           updated whenever the theme changes via <code>setBrandColors()</code> or
           <code>setTokenMapping()</code>.
         </li>
--- a/docs/pages/installation.vue
+++ b/docs/pages/installation.vue
@@ -1,5 +1,4 @@
 <template>
-
   <DocsPageTemplate>
     <DocsPageSection
       title="1. Install the plugin"
@@ -41,6 +40,14 @@
           Vue instances.
         </li>
         <li>Globally registers all KDS Vue components.</li>
+        <li>
+          Emits theme values as CSS variables (<code>--theme-*</code>, <code>--brand-*</code>, and
+          <code>--palette-*</code>) to a <code>&lt;style&gt;</code> tag
+          <code>#k-theme-css-variables</code> in an application's document head so that styles can
+          reference them directly, for example <code>var(--theme-primary)</code>. The variables are
+          updated whenever the theme changes via <code>setBrandColors()</code> or
+          <code>setTokenMapping()</code>.
+        </li>
         <li>
           Inserts assertive and polite ARIA live regions <code>#k-live-region</code> to an
           application's document body (see
@@ -87,5 +94,4 @@
       </p>
     </DocsPageSection>
   </DocsPageTemplate>
-
 </template>
--- a/lib/KThemePlugin.js
+++ b/lib/KThemePlugin.js
@@ -65,8 +65,6 @@ require('./grids/globalStyles.js'); // global grid styles
  * Also, set up global state, listeners, and styles.
  */
 export default function KThemePlugin(Vue) {
-  initThemeCssVariables();
-
   if (!isNuxtServerSideRendering() && process.env.VISUAL_TESTING !== 'true') {
     const onDomReady = () => {
       _mountLiveRegion();
@@ -80,6 +78,8 @@ export default function KThemePlugin(Vue) {
       onDomReady();
     }
   }
+
+  initThemeCssVariables();
 
   Vue.mixin({
     computed: {
--- a/lib/KThemePlugin.js
+++ b/lib/KThemePlugin.js
@@ -50,6 +50,7 @@ import KSnackbar from './KSnackbar/KSnackbar.vue';
 
 import { themeTokens, themeBrand, themePalette, themeOutlineStyle } from './styles/theme';
 import globalThemeState from './styles/globalThemeState';
+import initThemeCssVariables from './styles/themeCssVariables';
 
 import useKLiveRegion from './composables/useKLiveRegion';
 import _useOverlay from './composables/_useOverlay';
@@ -64,6 +65,8 @@ require('./grids/globalStyles.js'); // global grid styles
  * Also, set up global state, listeners, and styles.
  */
 export default function KThemePlugin(Vue) {
+  initThemeCssVariables();
+
   if (!isNuxtServerSideRendering() && process.env.VISUAL_TESTING !== 'true') {
     const onDomReady = () => {
       _mountLiveRegion();
--- a/lib/styles/themeCssVariables.js
+++ b/lib/styles/themeCssVariables.js
@@ -1,9 +1,13 @@
 import Vue from 'vue';
 import { isNuxtServerSideRendering } from '../utils';
 import globalThemeState from './globalThemeState';
+// side-effect import: loading `theme.js` sets `globalThemeState.tokens`,
+// which the `--tokens-*` CSS variables are read from
+import './theme';
 
 export const THEME_CSS_VARIABLES_STYLE_ID = 'k-theme-css-variables';
 
+const VALID_CSS_VARIABLE_NAME_PATTERN = /^--[\w-]+$/;
 // characters that would end a declaration or rule and corrupt the emitted CSS
 const UNSAFE_CSS_VALUE_PATTERN = /[;{}<>]/;
 
@@ -13,7 +17,7 @@ function warnVariableIssue(name, value) {
 }
 
 function setVariable(variables, name, value) {
-  if (UNSAFE_CSS_VALUE_PATTERN.test(value)) {
+  if (!VALID_CSS_VARIABLE_NAME_PATTERN.test(name) || UNSAFE_CSS_VALUE_PATTERN.test(value)) {
     warnVariableIssue(name, value);
     return;
   }
@@ -24,50 +28,33 @@ function formatPathSegment(key) {
   return key.replace(/^v_(\d+)$/, 'v$1');
 }
 
-function collectColorVariables(prefix, value, variables) {
+// Returns `{ '--name': value }` object for a color tree, walking nested
+// objects and joining path segments with hyphens.
+function collectColorVariables(prefix, value) {
+  const variables = {};
   if (typeof value === 'string') {
     setVariable(variables, prefix, value);
   } else if (value && typeof value === 'object') {
     Object.keys(value).forEach(key => {
-      collectColorVariables(`${prefix}-${formatPathSegment(key)}`, value[key], variables);
+      Object.assign(
+        variables,
+        collectColorVariables(`${prefix}-${formatPathSegment(key)}`, value[key]),
+      );
     });
   }
-}
-
-// Intentionally duplicates resolution from `theme.js`: its resolved tokens
-// object is non-observable, so reading it here would not register the
-// reactive dependencies the watcher relies on.
-function resolveTokenValue(mapString) {
-  // a value without dots is a CSS color value, not a path
-  if (mapString.indexOf('.') === -1) {
-    return mapString;
-  }
-  const refs = mapString.split('.');
-  let obj = globalThemeState.colors;
-  while (refs.length && obj) {
-    obj = obj[refs.shift()];
-  }
-  return typeof obj === 'string' ? obj : null;
+  return variables;
 }
 
 /**
- * Returns an object mapping CSS variable names to color values, from
- * the current `globalThemeState`.
+ * Returns a `{ '--name': value }` object of all theme CSS variables,
+ * from the current `globalThemeState`.
  */
 export function generateThemeCssVariables() {
-  const variables = {};
-  collectColorVariables('--palette', globalThemeState.colors.palette, variables);
-  collectColorVariables('--brand', globalThemeState.colors.brand, variables);
-  Object.keys(globalThemeState.tokenMapping).forEach(tokenName => {
-    const name = `--theme-${tokenName}`;
-    const value = resolveTokenValue(globalThemeState.tokenMapping[tokenName]);
-    if (value === null) {
-      warnVariableIssue(name, globalThemeState.tokenMapping[tokenName]);
-    } else {
-      setVariable(variables, name, value);
-    }
-  });
-  return variables;
+  return {
+    ...collectColorVariables('--palette', globalThemeState.colors.palette),
+    ...collectColorVariables('--brand', globalThemeState.colors.brand),
+    ...collectColorVariables('--tokens', globalThemeState.tokens),
+  };
 }
 
 /**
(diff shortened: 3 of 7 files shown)


lib/styles/themeCssVariables.js
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Returns an object mapping CSS variable names to color values, from
+ * the current `globalThemeState`.
+ */
+export function generateThemeCssVariables() {
+  const variables = {};
+  collectColorVariables('--palette', globalThemeState.colors.palette, variables);
+  collectColorVariables('--brand', globalThemeState.colors.brand, variables);
+  Object.keys(globalThemeState.tokenMapping).forEach(tokenName => {
+    const name = `--theme-${tokenName}`;
```
rtibblesbot: ✅ **Resolved** — addressed in the current code.


**suggestion:** `setVariable` guards the *value* against `[;{}<>]` (line 16), but the variable *name* is interpolated unchecked here into `--theme-${tokenName}` and into the declaration string on line 78. `tokenName` comes from `setTokenMapping()` keys — the most runtime-dynamic input to this module — so a key like `` `x: red; } body { color: red` `` would break out of the `:root` rule despite the value guard. Token names are developer-controlled, so this is low-risk, but worth closing the asymmetry: run the same pattern (or a stricter custom-property-name pattern) against `name` in `setVariable` and skip+warn on failure.

lib/styles/themeCssVariables.js
```diff
@@ -0,0 +1,113 @@
...
+    Object.keys(value).forEach(key => {
+      collectColorVariables(`${prefix}-${formatPathSegment(key)}`, value[key], variables);
+    });
+  }
+}
+
+// Intentionally duplicates resolution from `theme.js`: its resolved tokens
+// object is non-observable, so reading it here would not register the
+// reactive dependencies the watcher relies on.
+function resolveTokenValue(mapString) {
```
rtibblesbot: ✅ **Resolved** — addressed in the current code.


**suggestion:** `resolveTokenValue` returns `null` (→ warn + skip) for a mapping that doesn't resolve to a string, whereas `generateTokenToColorMapping` in `theme.js` *throws* on the same invalid path. Since `setTokenMapping` calls `generateTokenToColorMapping()` synchronously, an invalid mapping both throws there and (next tick) warns from here — divergent handling of the same condition in two places is a maintenance hazard. Consider a brief comment noting the intentional leniency, or centralizing the "resolve a mapping string against `globalThemeState.colors`" step so both call sites share it.

rtibblesbot (commented):
PR #1303 cleanly meets the acceptance criteria: `--palette-*`/`--brand-*`/`--theme-*` are emitted to a `#k-theme-css-variables` `<style>` tag with `vN` version keys, updates propagate via a reactive Vue watcher, and SSR head injection + client takeover is handled and tested. Manual QA confirmed SSR-before-JS, single-tag hydration takeover, `var()` resolution, and no console errors; the only a11y violations found are pre-existing on the unmodified home page.

CI: frontend tests, lint, and description checks pass; Percy visual tests still pending. Verdict is COMMENT (pending CI + two non-blocking suggestions), not APPROVE.

Two suggestions inline; nothing blocking:
- Variable *names* aren't sanitized the way values are (`setVariable`).
- Divergent error handling between `resolveTokenValue` here and `generateTokenToColorMapping` in `theme.js`.

---

</details>

lib/styles/themeCssVariables.js
AlexVelezLl: (nitpick) Took me some time to understand the reference to the theme.js module and what this method does 😅. I think we can just write the JS docs referencing what this method does, without even mentioning that this is a duplicate of theme.js (it doesn't have this exact method), and a simple example of resolving things like "brand.primary.v_500" would make this more understandable.

Relatedly, could we make the `tokens` object in `theme.js` an observable and set its values with `set`, so that we can do `collectColorVariables('--tokens', tokens, variables)`?

lib/styles/themeCssVariables.js
AlexVelezLl: Is `--theme` the best prefix here? In `theme.js` we expose three parallel functions `themeTokens`, `themeBrand` and `themePalette`, so I think the corresponding prefix here would be `--tokens`, right?

lib/styles/themeCssVariables.js
```diff
@@ -0,0 +1,113 @@
...
+    const name = `--theme-${tokenName}`;
+    const value = resolveTokenValue(globalThemeState.tokenMapping[tokenName]);
+    if (value === null) {
+      warnVariableIssue(name, globalThemeState.tokenMapping[tokenName]);
+    } else {
+      setVariable(variables, name, value);
+    }
+  });
+  return variables;
+}
```
AlexVelezLl, line 58: (nitpick) Is passing the `variables` object everywhere really needed here? Could it be a bit more readable if we do something like:

```JavaScript
export function generateThemeCssVariables() {
  const variables = {
    ...getColorCssVariables('--palette', globalThemeState.colors.palette),
    ...getColorCssVariables('--brand', globalThemeState.colors.brand),
  };
  Object.keys(globalThemeState.tokenMapping).forEach(tokenName => {
    const _tokenName = validateName(tokenName);
    const name = `--theme-${_tokenName}`;
    const value = resolveTokenValue(globalThemeState.tokenMapping[tokenName]);
    if (value === null || _tokenName === null) {
      warnVariableIssue(name, globalThemeState.tokenMapping[tokenName]);
    } else {
      variables[name] = value;
    }
  });
  return variables;
}
```

The idea would be to keep the functions as [pure](https://en.wikipedia.org/wiki/Pure_function) as possible, as they are slightly easier to understand.

With this, we can also fix what rtibblesbot flagged about not validating the name on the `--theme` name.

lib/styles/themeCssVariables.js
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Emits theme CSS variables to a `<style>` tag in the document head, reusing
+ * the tag written during server-side rendering if there is one, and keeps them
+ * in sync with `globalThemeState`.
+ */
+export default function initThemeCssVariables() {
+  if (watcherVm || isNuxtServerSideRendering() || typeof document === 'undefined') {
+    return;
+  }
+  updateStyleTag(themeCssVariablesText());
```
AlexVelezLl: _I think_ we can also set immediate: true on the watcher to avoid this duplicate call. It's mostly just in case we need to add more statements for the same watch; we don't need to duplicate code for the first execution.

lib/KThemePlugin.js
```diff
@@ -64,6 +65,8 @@ require('./grids/globalStyles.js'); // global grid styles
  * Also, set up global state, listeners, and styles.
  */
 export default function KThemePlugin(Vue) {
+  initThemeCssVariables();
```
AlexVelezLl: (nitpick) Could we have this call below the `if (!isNuxtServerSideRendering() && process.env.VISUAL_TESTING !== 'true') {` so that all related code that runs unconditionally is grouped together?

docs/pages/installation.vue
AlexVelezLl, line 1: Just wondering if we should update the [colors](http://deploy-preview-1303--kolibri-design-system.netlify.app/colors) page too to show the CSS variables? I think this would be a bit difficult to demonstrate, but perhaps adding a line parallel to the "brand.primary.v_100" lines may help 
(although it may be a bit overwhelming to have this repeated everywhere 😅)

Image

If there are plans to better communicate these CSS variables on this page/another page in the future, then ignore this comment 😅.

AlexVelezLl (commented):
Looks great! Mostly nitpicks/questions :)

lib/styles/themeCssVariables.js
LianaHarris360: Yes, you're right, that would be the most equivalent. I'll update the PR with this change

docs/pages/installation.vue
LianaHarris360: Yep, I plan to update the colors doc page to demonstrate the updated API, along with migration guidance.

lib/styles/themeCssVariables.js
LianaHarris360: It doesn’t duplicate an exact function in `theme.js`. Instead, it copies the output of the mapped tokens. The comment was meant to prevent refactors to use the resolved objects from `theme.js`, since it would break the reactivity. The comment could definitely be simplified though, to be more clear.

Making the tokens obj in `theme.js` observable breaks reactivity for newly added tokens (it caused the `emits newly added tokens` test to fail), but I could instead update `globalThemeState`, storing the resolved tokens there, to access in this file.

AlexVelezLl: That'd be very nice!

docs/pages/installation.vue
```diff
@@ -41,6 +41,14 @@
           Vue instances.
         </li>
         <li>Globally registers all KDS Vue components.</li>
+        <li>
+          Emits theme values as CSS variables (<code>--theme-*</code>, <code>--brand-*</code>, and
```
rtibblesbot: **blocking:** Docs say the plugin emits `--theme-*` and give the example `var(--theme-primary)` (also line 48), but the code emits the semantic tokens under `--tokens` (`collectColorVariables('--tokens', globalThemeState.tokens)`, `themeCssVariables.js:56`). Manual QA confirmed `var(--theme-primary)` resolves to `""` while `var(--tokens-primary)` → `#4368F5`. A consumer copying this example gets a broken, unresolved custom property. Change both occurrences to `--tokens-*` / `var(--tokens-primary)` (and fix the same stale name in the PR-body changelog).

rtibblesbot (changes requested):
PR #1303 — one prior finding remains open; the rest are resolved.

The prior round's feedback is substantially addressed: `--theme`→`--tokens` prefix rename landed in code and tests, token resolution is centralized in `globalThemeState.tokens` (removing the duplicated resolve logic and its null-vs-throw divergence), CSS variable names are now validated, and the watcher uses `immediate: true`. SSR→hydration takeover verified working end to end in manual QA.

One blocking issue remains — see inline. Note: Percy visual CI is failing (not passing).

- **[blocking]** `docs/pages/installation.vue:45` — new doc copy documents `--theme-*` / `var(--theme-primary)`, but the code emits `--tokens-*`. QA confirmed `var(--theme-primary)` resolves to empty in the running app while `var(--tokens-primary)` → `#4368F5`. A consumer copying the example gets a broken variable. The PR-body changelog carries the same stale naming.

RESOLVED — themeCssVariables.js — variable name not validated (name-guard asymmetry)
RESOLVED — themeCssVariables.js — resolveTokenValue null vs theme.js throws (divergent handling)
RESOLVED — themeCssVariables.js — --theme prefix renamed to --tokens (code + tests)
UNADDRESSED — docs/pages/installation.vue:45 — docs still reference --theme-* (see blocking above)
RESOLVED — themeCssVariables.js — pass variables everywhere; prefer pure functions
RESOLVED — themeCssVariables.js — set immediate: true on the watcher
RESOLVED — themeCssVariables.js / theme.js — make resolved tokens observable for runtime additions
ACKNOWLEDGED — themeCssVariables.js — JSDoc clarity for the theme.js reference
ACKNOWLEDGED — KThemePlugin.js — group initThemeCssVariables() with SSR-guarded code
ACKNOWLEDGED — docs/pages/colors — also update the colors doc page (deferred to follow-up)

</details>


---

</details>

lib/styles/themeCssVariables.js
```diff
@@ -24,50 +28,33 @@ function formatPathSegment(key) {
...
-    } else {
-      setVariable(variables, name, value);
-    }
-  });
-  return variables;
+  return {
+    ...collectColorVariables('--palette', globalThemeState.colors.palette),
+    ...collectColorVariables('--brand', globalThemeState.colors.brand),
+    ...collectColorVariables('--tokens', globalThemeState.tokens),
+  };
```
AlexVelezLl, line 57: :chefs-kiss:

AlexVelezLl (approved):
LGTM! Code changes look good, and smoke tested this on Kolibri and seems to be working great!

PR #1310, opened 2026-07-31
[floating-production feaure branch] Add useKFloatingInteraction

MisRob:
## Description

Prepares `useKFloatingInteraction` for production.

Also see https://github.com/learningequality/kolibri-design-system/issues/1311 for broader context.

Compared to the draft version - while the high-level approach stays, this production-ready version deserves full review. It has lots of new logic and in some areas I deviated. Rather than comparing to draft, I will invite you to go through the documentation and code where I prepared lots of explanatory comments, and ask in case there's any area of doubt. I also updated related documentation pages here, so please re-read.

## Changelog

  - **Description:** Add useKFloatingInteraction
  - **Addresses:** -
  - **Components:** -
  - **Breaking:** no
  - **Impacts a11y:** -
  - **Guidance:** -

## Steps to test

- See https://github.com/learningequality/kolibri-design-system/issues/1311
- Preview final documentation pages
   - https://deploy-preview-1310--kolibri-design-system.netlify.app/browsersupport
   - https://deploy-preview-1310--kolibri-design-system.netlify.app/floatingelements
   - https://deploy-preview-1310--kolibri-design-system.netlify.app/usekfloatingposition
   - https://deploy-preview-1310--kolibri-design-system.netlify.app/usekfloatinginteraction
- Code review

## AI use

First I tried Claude to generate a better robust version of my draft, but it was very far away from what I imagined ~ mostly useful to start documentation and some initial file skeleton.

Then I gradually worked & tested myself, with Claude more as an assistance for repetitive tasks, tests, and documentation. Finally, I reviewed the whole final diff & made final testing.

(I experimented with Claude solving some problems, but rarely moved forward with - it doesn't seem to have a realistic sense in terms of having to do some trade-offs, and tends to end up with very complex solutions, probably in attempts to satisfy every possible path, even those unlikely ones)

I also used Claude to do few performance checks - see _Latest > Performance_ section in the roadmap issue in https://github.com/learningequality/kolibri-design-system/issues/1311


MisRob:
Test failure seems unrelated

lib/composables/useKFloatingInteraction/index.js
```diff
@@ -0,0 +1,1193 @@
...
+
+  removeUnusedDocumentListeners();
+}
+
+// ==================================================================
+// Main function
+
+export default function useKFloatingInteraction(floatingId, floatingRef, options = {}) {
+  if (isNuxtServerSideRendering()) {
+    return;
```
rtibblesbot, line 1042: ✅ **Resolved** — addressed in the current code.


**blocking:** Returning `undefined` breaks the documented call form on the server. Every documented use destructures the return value — `docs/pages/usekfloatinginteraction.vue:18` and all 12 examples do `const { isActive, activatorEl } = useKFloatingInteraction(...)` — so on the server that throws `Cannot destructure property 'isActive' of undefined`, and a KDS component built on this composable can't render in a Nuxt universal app. This repo's docs site is `mode: 'universal'` (`nuxt.config.js:8`); it survives today only because `DocsExample` loads examples via `import()` in `created()`, which never resolves during server render.

The pattern elsewhere is to guard internally and still return the API — `lib/composables/_useOverlay/index.js:46` returns early inside `getOverlayEl()` while the composable always returns `{ mountOverlay, getOverlayEl }`, and `useKResponsiveWindow` guards inside `initProps()`. An inert shape would match:

```js
if (isNuxtServerSideRendering()) {
  return {
    isActive: computed(() => false),
    activatorEl: computed(() => null),
    setActive: () => {},
    setEnabled: () => {},
  };
}
```

lib/composables/useKFloatingInteraction/index.js
```diff
@@ -0,0 +1,1193 @@
...
+// This makes the floating element activated by another of its
+// activator elements
+function switchActivator(floatingId, newActivatorEl) {
+  const entry = _registry[floatingId];
+  const oldActivatorEl = entry.activeActivatorEl.value;
+  if (!oldActivatorEl || oldActivatorEl === newActivatorEl) {
+    return;
+  }
+
+  entry.activeInteractions.value.filter(deactivatesOnLeavingActiveArea).forEach(interaction => {
```
rtibblesbot, line 427: ✅ **Resolved** — addressed in the current code.


**blocking:** This re-homes the deactivation listeners of *all* currently active area-based interactions onto the new activator, but `activeActivatorEl` is one element per floating element rather than one per interaction. When two activators share a floating element and hold it active through *different* interactions, a live listener gets moved off the element the user is actually on.

Trace with `interactions: { hover: true, focus: true }` and activators A and B both carrying `data-floating-id="f"`:

1. Pointer enters A → `mouseleave` added on A, `activeInteractions = ['hover']`, `activeActivatorEl = A` (index.js:842-844).
2. Keyboard focus lands on B → `activateInteraction('f', B, FOCUS)` → `switchActivator('f', B)`: `hover` is active and area-based, so `mouseleave` is **removed from A and added to B**.
3. Pointer leaves A. No handler on A any more, so `hover` stays in `activeInteractions` indefinitely.
4. Focus leaves B → only `FOCUS` is removed. The element stays active with the pointer nowhere near it, until the user happens to hover *and* leave B, presses Escape, or the activator unmounts.

That's the "tooltip anchored to the info icon but activated by focusing the input" pattern #1311 §7 names as a target use case, and it isn't covered — the multi-activator test (`__tests__/index.spec.js:624`) uses `hover` alone and never interleaves two interactions across two activators.

Suggested direction: track the activator per active interaction (e.g. `activeInteractions` as a Map of interaction → element), have `switchActivator` move only the interaction actually changing activator, and have `deactivateInteractions` remove each listener from the element it was attached to. A "hover A, focus B, leave A, blur B" test would lock it in. Derived from reading the code rather than a browser run — if you think the interleaving can't occur in practice, could you say why?

lib/composables/useKFloatingPosition/index.js
```diff
@@ -61,12 +64,20 @@ export default function useKFloatingPosition() {
...
     }
 
     _instances[floatingId] = {
       cleanup: autoUpdate(anchorEl, floatingEl, updatePosition),
       floatingEl,
     };
+
+    // Promise that lets callers wait until the floating element is
+    // actually in place, e.g. before moving focus into it
+    return firstUpdate;
```
rtibblesbot, line 83: ✅ **Resolved** — addressed in the current code.


**blocking:** `firstUpdate` is only assigned from inside `updatePosition`, so this return value depends on `autoUpdate` invoking its callback synchronously during setup. Real `@floating-ui/dom@1.7.6` does, but the contract is implicit and already broken in this repo: the existing spec mocks `autoUpdate: jest.fn(() => jest.fn())` (`__tests__/index.spec.js:5`), which never calls the callback — so under that double `initPosition` returns `null` and the documented `initPosition(...).then(...)` form (docs page, and `examples/useKFloatingInteraction/CustomEscape.vue:76`) throws `Cannot read properties of null (reading 'then')`. Any consumer mocking floating-ui the same way hits this.

Cheap fix: `return firstUpdate || Promise.resolve();`

Two related points — no test covers the new return value at all (the spec still only awaits the extracted update fn), and because the chain ends in `.catch()` the promise also resolves when positioning *failed*, so `@returns {Promise} Resolves once the position has been applied` overstates it. `CustomEscape.vue` would move focus into an unpositioned popup. Worth either rejecting on failure or softening the wording.

lib/composables/useKFloatingInteraction/__tests__/index.spec.js
```diff
@@ -0,0 +1,968 @@
...
+  const activator = document.createElement('button');
+  activator.setAttribute('data-floating-id', floatingId);
+  activator.textContent = activatorLabel;
+  document.body.appendChild(activator);
+  return activator;
+}
+
+// Set activation and deactivation delays to 0 for testing purpose
+// Otherwise some assertions would be too complex
+function withoutDelays(interactions) {
```
rtibblesbot, line 65: ✅ **Resolved** — addressed in the current code.


**suggestion:** `withoutDelays()` is applied to every `createPopup()` call and there are no fake timers anywhere in the suite, so `delays.js` is entirely uncovered — `scheduleChange`'s `delay` branch plus all of `flushPendingChange` / `cancelPendingChange` / `cancelAnyPendingChange` / `isActivationPending` (delays.js:22-113).

That includes the trickiest invariant in the PR: there is a single page-wide `pendingChange`, and scheduling for a different floating element or interaction flushes the previous one early (delays.js:55-62). That flush is what protects against an interaction holding its floating element active with nothing left to release it, and it's also what makes the WCAG Hoverable grace period conditional.

`jest.useFakeTimers()` plus the existing helpers should cover the important cases cheaply:
- pointer leaves the activator, returns before `deactivateDelay` elapses → stays active (the `shouldRun` re-check);
- pointer leaves activator A (deactivation pending) then enters activator B of another floating element → A's deactivation is flushed, B activates;
- activation pending, activator unmounts before the timer fires → the `isRegistered` guard prevents the run;
- `setActive` / `setEnabled(false)` while a change is pending → pending change is cancelled.

lib/composables/useKFloatingInteraction/validation.js
```diff
@@ -0,0 +1,128 @@
...
+      throw new Error(
+        `[useKFloatingInteraction] 'deactivateOn.${key}' needs to be a ${expectedType}.`,
+      );
+    }
+  });
+
+  return { ...DEACTIVATE_ON_DEFAULTS, ...deactivateOn };
+}
+
+export function normalizeInteractions(interactions) {
```
rtibblesbot, line 69: **suggestion:** `normalizeInteractions` and `normalizeDeactivateOn` throw eight distinct errors and are pure functions with no DOM or lifecycle dependency — the cheapest thing in the PR to test, and currently untested (the spec imports `normalizeInteractions` only as a helper). Since these messages are the primary developer-facing contract for the options API, a small table-driven describe block would be worth it. Same for `warnIfManyNonDelegated`'s `NODE_ENV` gate — the warning is tested, the "silent outside development" path isn't.

Separately, falsy values are handled three different ways:
- `interactions: undefined | null | false` → `{}` (no interactions), here at line 70;
- `deactivateOn: false | null` → **all defaults**, i.e. `esc: true` (line 41-43) — someone writing `deactivateOn: false` to mean "nothing dismisses this" silently gets Escape dismissal;
- `interactions: { hover: false }` → **throws** (line 103-107).

Given `hover: true` means "on", `hover: false` reads naturally as "off", and it's the ergonomic form when the value is computed (`{ hover: !isTouchDevice }`). Everything else here is strict-and-loud, which makes the silent `deactivateOn: false` fallback the odd one out. Suggest treating a falsy per-interaction value as "not configured" and rejecting non-object truthy `deactivateOn`, or documenting the accepted shapes precisely.

lib/composables/useKFloatingInteraction/index.js
```diff
@@ -0,0 +1,1193 @@
...
+// The shared first part of the deactivation handlers that watch
+// the active area. Their listener sits either on the activator
+// element or on the floating element itself.
+function resolveDeactivation(event, interaction) {
+  let activatorEl = getActivatorElFromEvent(event);
+  let floatingId;
+  if (activatorEl) {
+    floatingId = getFloatingId(activatorEl);
+  } else {
+    floatingId = event.currentTarget && event.currentTarget.id;
```
rtibblesbot, line 797: **suggestion:** A *missing* `id` throws a clear error here, but a *mismatched* one fails silently: `isInteractionActive(<wrong id>, interaction)` is just `false`, `resolveDeactivation` returns `null`, and the interaction is never released. Because leaving the activator toward the floating element is suppressed by the `isInsideActiveArea(event.relatedTarget, …)` check (index.js:856), the result is a tooltip that opens normally and then can never be closed by hover or blur, with nothing in the console.

The composable already holds the correct element via `getFloatingEl(floatingId)`, so a development-only check in `onMounted` (index.js:1068) would turn this into an immediate, obvious failure:

```js
if (process.env.NODE_ENV === 'development' && floatingRef.value && floatingRef.value.id !== floatingId) {
  console.warn(`[useKFloatingInteraction] Floating element's "id" ("${floatingRef.value.id}") must match floatingId ("${floatingId}")`);
}
```

Related, lower stakes: `getFloatingId` (utils.js:110) throws for `data-floating-id=""`, and it's reached from the document-wide `onClickDeactivate`/`onTouchDeactivate` handlers — an empty binding would throw on every click anywhere on the page while any click-activated floating element is open.

lib/composables/useKFloatingInteraction/index.js
```diff
@@ -0,0 +1,1193 @@
...
+  for (const floatingId of _active) {
+    if (deactivatesOnEsc(floatingId)) {
+      topmost = floatingId;
+    }
+  }
+  return topmost;
+}
+
+// Closes the innermost floating element that dismisses on Escape
+function onEscKeydown(event) {
```
rtibblesbot, line 199: **suggestion:** Two gaps in the Escape story, both worth resolving given `KTooltip`/`KDropdownMenu` are the intended consumers.

**Layering stops at this registry.** This is a capture-phase `document` keydown listener that neither stops propagation nor calls `preventDefault()`. The docs promise (`docs/pages/usekfloatinginteraction.vue:337-340`) that nested floating elements close "one layer at a time, innermost first" holds for two elements both managed here, but not once a KDS component is in the stack: `lib/KModal.vue:14` dismisses on `@keyup.esc.stop` — a *keyup* handler, so even adding `stopPropagation()` on keydown wouldn't suppress it — and `lib/keen/UiPopover.vue:321` registers keydown-27 on `document`. A tooltip open inside a modal closes both on one press. A shared dismissal stack, or at minimum `preventDefault()` plus a documented caveat, would keep the docs claim honest.

**No focus restoration.** `closeAll` flips `isActive` false; the consumer hides or unmounts the floating element, and if focus was inside it the browser drops focus to `<body>` — the user loses their place in the tab order (WCAG 2.4.3). Reachable from the shipped examples: in `Nested.vue` tab to the `KIconButton` inside the popup and press Escape twice. Notably the only two examples that restore focus (`CustomEscape.vue:60-64`, `OutsidePress.vue:51-55`) do it by hand *because* they opted out of the built-in behaviour — nothing tells a consumer they need to do this on the default path. Either restore focus to `activatorEl` when dismissal happens while focus is inside the active area, or add an a11y banner saying it's the consumer's job.

docs/pages/usekfloatinginteraction.vue
```diff
@@ -0,0 +1,508 @@
...
+
+      <h4>Lazy</h4>
+
+      <p>
+        Note that <code>setTimeout</code> is needed in the lazy implementation, so the fade-out
+        animation can finish.
+      </p>
+
+      <DocsBanner a11y>
+        When rendering lazily, use <code>aria-label</code>/ <code>aria-description</code>.
```
rtibblesbot, line 262: **suggestion:** `aria-description` is an ARIA 1.3 **working-draft** attribute — not in ARIA 1.2, not implemented in Firefox or Safari/VoiceOver, and flagged as unknown by validators. KDS's browserslist floor is `Safari >= 11.1`, `Firefox >= 52`, `iOS >= 9.3`, which is also the population this PR's own new `/browsersupport` page tells readers to design for. Verified in-browser on the lazy example: the activator has no accessible description at all — `aria-label`, `aria-labelledby` and `aria-describedby` are all `null` on `[data-floating-id="lazy-tooltip"]`.

The diagnosis in the banner is right (a dangling `aria-describedby` is worse), but the prescription needs something that works today: `aria-label` on the trigger where it has no visible text, or a visually-hidden copy of the description kept permanently in the DOM with `aria-describedby` pointing at *that* while the visible tooltip renders lazily. Same applies to `examples/useKFloatingInteraction/Lazy.vue:6`. Since this banner is the design system's official answer to "how do I make a lazy tooltip accessible", it's worth getting right.

examples/useKFloatingInteraction/Basic.vue
```diff
@@ -0,0 +1,105 @@
...
+<style lang="scss" scoped>
+
+  @import '../../lib/styles/definitions';
+
+  .tooltip {
+    @extend %dropshadow-2dp;
+
+    position: absolute;
+    top: 0;
+    left: 0;
```
rtibblesbot, line 90: ✅ **Resolved** — addressed in the current code.


**suggestion:** Two base-style issues, both repeated across all 12 examples (`Eager.vue:90`, `KeyboardOnlyFocus.vue:94`, `NoToggle.vue:86`, `ScrollDeactivate.vue:87`, `ManualActivation.vue:102`, `OutsidePress.vue:103`, `Lazy.vue:107`, `Enabling.vue:115`, `CustomEscape.vue:121`, `Nested.vue:134` and `:151`, `Tour.vue:187`).

**`left: 0` is RTL-flipped.** `left` is directional, so the RTL build rewrites it to `right: 0` (`docs/pages/writing.vue:94-101` documents the webpack-rtl-plugin doing exactly this), while `useKFloatingPosition` writes `left: <x>px` as an **inline** style that isn't flipped. In an RTL build the element then has `right: 0` from the stylesheet *and* `left: Xpx` inline — for an absolutely positioned `width: auto` box that stretches the used width between the two edges rather than shrink-to-fit, so it's mis-sized and mis-placed, only in Arabic/Hebrew. `lib/KTooltip/index.vue:160-166` avoids this by declaring only `position: absolute` and leaving placement to the inline styles. Either match `KTooltip`, or keep the `top: 0; left: 0` origin with `/*rtl:ignore*/` and a comment so consumers who copy it keep the guard.

**`width: max-content` was dropped.** `lib/KTooltip/index.vue:164` sets it and so did the deleted `examples/useKFloatingPosition/Basic.vue`; none of the 12 new examples do. Floating UI treats it as part of the required base styles — without it `computePosition` measures a shrink-to-fit width against the containing block that then changes once the element moves, visible as jitter on first open and premature wrapping near viewport edges. The `max-width: 240px` cap limits but doesn't remove it.

examples/useKFloatingInteraction/CustomEscape.vue
```diff
@@ -0,0 +1,130 @@
+<template>
+
+  <div>
+    <KButton :data-floating-id="FLOATING_ID">Click or focus me</KButton>
```
rtibblesbot: **suggestion:** The tooltip examples are annotated well (`role="tooltip"` + `aria-describedby`/`aria-labelledby`), but the click-activated popup examples have no ARIA at all — trigger and panel are a plain `KButton` and a plain `div`. Also in `ManualActivation.vue:4`, `NoToggle.vue:4`, `OutsidePress.vue:4`, `ScrollDeactivate.vue:4`, `Nested.vue:4`, `Tour.vue:5-15`.

A disclosure-style trigger needs `aria-expanded` bound to `isActive` (usually `aria-controls` too), and the panel needs a role and an accessible name. This example is the strongest case: it programmatically moves focus into the panel (`:81 inputRef.value.focus()`), which is dialog behaviour, but the panel has neither `role="dialog"` nor a label, so a screen-reader user is dropped into an unannounced textbox. These are the files consuming apps copy — the ARIA wiring is the part they most need shown.

docs/tableOfContents.js
```diff
@@ -259,8 +259,15 @@ export default [
         path: '/usekfloatingposition',
         title: 'useKFloatingPosition',
         isCode: true,
+        candidate: true,
         keywords: [...compositionRelatedKeywords, ...floatingRelatedKeywords],
       }),
+      new Page({
+        path: '/usekfloatinginteraction',
```
rtibblesbot, line 266: ✅ **Resolved** — addressed in the current code.


**nitpick:** `useKFloatingPosition` gained `candidate: true` two entries up plus a developer note ("Don't use in production yet. Browser testing in Kolibri isn't finished"), and `developerNotes` only renders inside the candidate banner (`docs/common/DocsPageTemplate/index.vue:36-47`). This page carries neither — verified in-browser, no badge in the nav or the header — yet all 12 of its examples call `useKFloatingPosition`, and three of its own warning banners are about `initPosition`/`destroyPosition`. A reader following the docs end-to-end adopts the candidate composable without ever seeing the candidate warning.

Is the asymmetry deliberate? If the interaction composable is meant to be stable on its own, a line on the page pointing at the position composable's candidate status would close the gap; otherwise mirroring `candidate: true` (or a `#developerNotes` entry pointing at #1311) would.

docs/pages/usekfloatinginteraction.vue
```diff
@@ -0,0 +1,508 @@
...
+          [
+            'Activation listeners',
+            '3 per activator element (total 60: 20 mouseenter, 20 touchstart, 20 focus)',
+            '3 page-wide (1 mouseenter, 1 touchstart, 1 focus)',
+          ],
+        ]"
+      />
+
+      <p>
+        The main trade-off of non-delegated mode is a MutationObserver present all the time, and a
```
rtibblesbot: ✅ **Resolved** — addressed in the current code.


**nitpick:** The trade-off names "a MutationObserver present all the time, and a much higher number of listeners", but the cost that actually decides the choice is what happens in the handler: any element added anywhere on the page invalidates *every* non-delegated cache entry and triggers one full `document.querySelectorAll('[data-floating-id]')` scan plus a `syncActivationListeners` pass per non-delegated floating element (index.js:658-697). That scales with (number of non-delegated floating elements × DOM churn), which is the thing to weigh against the delegated mode's "handler runs for every event of that type". One sentence would make the choice much easier to reason about. (The single-scan `sharedQuery` design and its lazy trigger are a nice optimisation — this is about documenting the residual cost, not changing it.)

lib/composables/_useElementsCache/index.js
```diff
@@ -0,0 +1,37 @@
...
+export default function _useElementsCache() {
+  const cache = new Map();
+
+  // Retrieves query result from cache if available.
+  // Otherwise runs query, caches and returns its result.
+  // Clears elements that are not in the DOM from
+  // the cache automatically, so they're not kept in memory.
+  function cachedQuery(key, query) {
+    const cached = cache.get(key);
+    if (cached) {
```
rtibblesbot, line 14: ✅ **Resolved** — addressed in the current code.


**nitpick:** `if (cached)` is truthy for `[]`, so once a key is cached as empty, `cachedQuery` keeps returning `[]` and never re-runs the query no matter what appears in the DOM. That's fine in this PR's usage — the MutationObserver calls `forgetActivatorEls` on any element addition (index.js:668) — but the header comment above advertises this as a generic per-feature cache that "clears elements that are not in the DOM from the cache automatically", which reads as if staleness is handled for you. Worth stating the invariant ("an empty result stays cached until you invalidate it — the caller must invalidate on element additions"), and a `returns a stale empty result until the key is cleared` test in the new cache spec would pin the behaviour deliberately rather than incidentally.

examples/useKFloatingInteraction/Enabling.vue
```diff
@@ -0,0 +1,130 @@
...
+        if (active) {
+          // Performance: Position needs to be initialized
+          // only when the tooltip becomes active (= visible),
+          // not when it's mounted
+          nextTick(() => {
+            initPosition(FLOATING_ID, tooltipRef.value, activatorEl.value, {
+              placement: 'bottom',
+              middleware: [offset(8), flip()],
+            });
+            Object.assign(tooltipRef.value.style, { opacity: '1', pointerEvents: 'auto' });
```
rtibblesbot, line 71: ✅ **Resolved** — addressed in the current code.


**nitpick:** Six examples mutate the DOM node's style object directly inside a `watch` (also `Basic.vue:54,57`, `Eager.vue:54,57`, `KeyboardOnlyFocus.vue:57,60`, `Lazy.vue:64,67`, `Nested.vue:94,97`). This duplicates state `isActive` already holds, bypasses Vue's rendering, and needs a null guard the moment the element is conditionally rendered. `:style="{ opacity: isActive ? 1 : 0, pointerEvents: isActive ? 'auto' : 'none' }"` (or a `:class` toggle) is equivalent, keeps the CSS transition, and is what the rest of KDS does. The examples otherwise model very good practice — the `initPosition`/`destroyPosition` lifecycle comments in particular — this is the one habit worth not propagating.

Separately at `:106`: flex `gap` needs Chrome 84+/Safari 14.1+/Firefox 63+, while `browserslist-config-kolibri@0.18.0` declares `Chrome >= 49`, `Safari >= 11.1`, `iOS >= 9.3`. Also `ManualActivation.vue:93`, `Nested.vue:137`, `Tour.vue:174`. Cosmetic and examples-only, but `margin` on the children degrades rather than collapses to zero.

lib/composables/useKFloatingInteraction/index.js
```diff
@@ -0,0 +1,1193 @@
...
+  if (isNeeded && !documentHasScrollListener) {
+    document.addEventListener('scroll', onDocumentScroll, CAPTURE_PASSIVE);
+    documentHasScrollListener = true;
+  } else if (!isNeeded && documentHasScrollListener) {
+    document.removeEventListener('scroll', onDocumentScroll, CAPTURE_PASSIVE);
+    documentHasScrollListener = false;
+  }
+}
+
+// ==================================================================
```
rtibblesbot, line 266: ✅ **Resolved** — addressed in the current code.


**praise:** The `callers` list, the `setCurrentCaller` handover (including re-attaching the deactivation listener to the incoming caller's floating element, :340-348), and the numbered walkthrough of the `:key`/`v-if` overlap address a race that would otherwise be very hard to diagnose. The `onBeforeUnmount` / `onUnmounted` split is the right seam and the comment says why — as does `sameArrayItems` existing specifically because the cache hands back a fresh array of identical elements. That reasoning usually gets lost.

rtibblesbot (dismissed):
PR #1310 — a large, genuinely careful composable. The caller/registry handover, the multi-activator race handling, and the comments explaining *why* are well above the usual bar, and the 968-line Testing Library suite is the model the other composable specs should follow. Manual QA in a real browser confirmed every documented behaviour on all 12 examples (hover/focus/click, keyboard-only focus, layered Escape, outside press, tour, lazy render, RTL, mobile) with zero console output and no new axe violations.

Three things look like real defects rather than polish:

- **blocking** — the SSR early return hands back `undefined`, so the documented `const { isActive } = useKFloatingInteraction(...)` call form throws on the server (`index.js:1042`).
- **blocking** — `switchActivator` migrates a still-live hover deactivation listener onto a different activator, leaving the floating element stuck active (`index.js:427`).
- **blocking** — `initPosition`'s newly documented promise is `null` under this repo's own `autoUpdate` test double, so the documented `.then()` form throws (`useKFloatingPosition/index.js:80`).

Beyond those: `delays.js` and `validation.js` — the two most branch-heavy new modules — have no test coverage at all, the Escape story doesn't survive contact with `KModal`/`UiPopover` or restore focus, and the examples teach a couple of patterns (RTL-unsafe `left: 0`, `aria-description`, no `aria-expanded` on popup triggers) that consumers will copy verbatim. Details inline.

CI is red on `lib/KDateRange/__tests__/KDateRange.spec.js` ("November 2022" vs "December 2022") — a timezone-sensitive assertion in a file this branch doesn't touch, 457 passed / 1 failed. Not yours, but it blocks the merge mechanically.

---

</details>

rtibblesbot (dismissed):
PR #1310 is unchanged since the last review — HEAD is still `4555260c`, with no new commits and no replies to the open threads.

**0 of 15 prior findings resolved; 14 still open.** No new findings: nothing new was pushed to review, and every open point already has a live inline thread carrying its detail, so I haven't re-posted them.

Still open, by severity (see the existing threads):

- **blocking** — SSR early return hands back `undefined`, breaking the documented destructuring call form (`useKFloatingInteraction/index.js:1042`); `switchActivator` re-homes a still-live area deactivation listener onto a different activator (`index.js:427`); `initPosition` returns `null` when `autoUpdate` doesn't invoke its callback synchronously (`useKFloatingPosition/index.js:80`).
- **suggestion** — `delays.js` has no coverage (every test zeroes both delays, so only the `if (!delay)` fast path runs); `validation.js` error paths untested; mismatched floating `id` fails silently; Escape dismissal doesn't restore focus; `aria-description` recommended despite ARIA 1.3 working-draft status; `left: 0` is RTL-flipped across all 12 examples; click-popup examples carry no disclosure semantics.
- **nitpick** — interaction page missing the `candidate` badge its dependency carries; non-delegated trade-off omits the per-mutation rescan cost; `_useElementsCache` empty-result caching invariant undocumented; six examples write inline styles from a `watch`.

Manual QA re-confirmed the Escape focus-loss defect in two examples (focus lands on `<body>` after dismissing a popup you've tabbed into) — that's the concrete repro behind the `index.js:199` thread.

CI is still red on `Frontend tests`, and the failure is unrelated to this branch: `KDateRange.spec.js` (1 failed, 457 passed), a timezone-sensitive assertion in a file this PR doesn't touch — matching @MisRob's earlier read. It still blocks the merge mechanically. Percy is pending.

UNADDRESSED — lib/composables/useKFloatingInteraction/index.js:1042 — SSR early return returns `undefined`, breaking documented destructuring
UNADDRESSED — lib/composables/useKFloatingInteraction/index.js:427 — `switchActivator` re-homes a live area deactivation listener
UNADDRESSED — lib/composables/useKFloatingPosition/index.js:80 — `initPosition` returns `null` on async `autoUpdate`
UNADDRESSED — lib/composables/useKFloatingInteraction/__tests__/index.spec.js:65 — `delays.js` has no coverage
UNADDRESSED — lib/composables/useKFloatingInteraction/validation.js:69 — validation module untested
UNADDRESSED — lib/composables/useKFloatingInteraction/index.js:797 — mismatched floating `id` fails silently
UNADDRESSED — lib/composables/useKFloatingInteraction/index.js:199 — no focus restoration on Escape dismissal
UNADDRESSED — docs/pages/usekfloatinginteraction.vue:262 — `aria-description` is an ARIA 1.3 working draft
UNADDRESSED — examples/useKFloatingInteraction/Basic.vue:90 — `left: 0` RTL-flipped; `width: max-content` dropped
UNADDRESSED — examples/useKFloatingInteraction/CustomEscape.vue:4 — click popup examples carry no ARIA
UNADDRESSED — docs/tableOfContents.js:266 — interaction page has no `candidate` badge
UNADDRESSED — docs/pages/usekfloatinginteraction.vue:119 — non-delegated trade-off omits rescan cost
UNADDRESSED — lib/composables/_useElementsCache/index.js:14 — empty result stays cached; invariant undocumented
UNADDRESSED — examples/useKFloatingInteraction/Enabling.vue:71 — examples mutate DOM style objects in a `watch`
ACKNOWLEDGED — lib/composables/useKFloatingInteraction/index.js:266 — praise for the callers/`setCurrentCaller` handover

</details>

---

</details>

lib/composables/useKFloatingInteraction/index.js
MisRob: I skipped. It's not breaking anything now, and more broadly, it'd be better achieve setup where there's no for these `isNuxtServerSideRendering` at all - from the point of consumers, it's not neccessary (and often a point of confusion in KDS reviews)

lib/composables/useKFloatingPosition/index.js
MisRob: > return firstUpdate || Promise.resolve();

Would give impression that it may happen that `firstUpdate` is not ready, and that's just confusing.

Rather than adding it for tests to here, I updated tests to have more precise mock.

lib/composables/useKFloatingInteraction/index.js
rtibblesbot: Understood — deferring to your call. Agreed that removing the need for `isNuxtServerSideRendering` guards entirely is the better direction than adding another one.

lib/composables/useKFloatingInteraction/index.js
MisRob: My intuition would be to not worry about this right now and rather observe in real deployments if there are any situations that need to be treated. Not to say that robustness doesn't matter ~ the point of this PR is to make the composable robust. Generally, there are actually more places like this in the composable where I chose to accept a trade-off rather than trying to cover for absolutely everything. The composable already does a ton to make it very easy on consumers~ if I should get it ready for all possible combinations then it'd grow very complex (which also increases chances of performance regressions now and in the future updates). When considering what situations to cover and what not, I usually put effort into a11y, support  for highly async loaded & reactive environment, and common scenarios.

Specifically here, users usually don't move pointer to point A, then focus to B, then leave pointer from A, and then move away focus from B :) Furthermore, more activators sharing one floating element is very rare (supporting it now is rather a nice benefit, but I don't think it's necessary ~ alternative would be to remove the support and document the constraint).

That said, another way to look at it that it points at a weak spot, and perhaps larger changes in implementation approach in some  areas would make the composable better. For this part I will only say that this version is the best I was able to achieve, and if anyone has any ideas for improvements, go ahead :)

rtibblesbot: Understood — deferring to your call as maintainer. Withdrawing the blocking label; the interleaving I traced (hover A → focus B → leave A → blur B, with two activators sharing one floating element) is a genuinely rare combination, and "observe in real deployments" is a reasonable place to draw the complexity line. Thanks for spelling out the criteria you use to decide what to cover — a11y, reactive environments, common scenarios — that is useful calibration for future reviews here.

docs/pages/usekfloatinginteraction.vue
MisRob: I improved how this is phrased. But I won't explain handlers here. Under some conditions, they run, under some, they don't, sometimes they're cached, sometimes they're not. Every developer knows that the point is that there's some cost based on how frequently something runs & what runs. This section is to give a very high-level overview on what to focus on when doing measurements, so it's vague on purpose. If someone needs to understand what's going on exactly under each condition, they can see code and in-code comments.

examples/useKFloatingInteraction/Enabling.vue
MisRob: > bypasses Vue's rendering

This is intentional here, so that everything is in once place (makes easier to understand timing when reading it). Also I forgot to take a note, but when I worked on PoC I think I observed that in this context, assigning style directly is much more performant than using Vue machinery.

rtibblesbot: Understood — that's a deliberate choice for the examples, and the timing-in-one-place rationale makes sense for teaching material. I can't disprove the perf observation either. Withdrawing this finding.

docs/pages/usekfloatinginteraction.vue
rtibblesbot: Fair — a deliberately high-level "what to measure" section rather than a cost model. Deferring to your framing; withdrawing this finding.

lib/composables/useKFloatingPosition/index.js
rtibblesbot: Makes sense — tightening the mock rather than loosening the return value is the right direction. Withdrawing this finding.

docs/tableOfContents.js
MisRob: `useKFloatingInteraction` is stable

rtibblesbot: Understood — the asymmetry is deliberate, `useKFloatingInteraction` ships stable while only `useKFloatingPosition` is candidate. Withdrawing this finding.

lib/composables/_useElementsCache/index.js
MisRob: It may be more intuitive to just remove the `key` as soon as the cached result is `[]`?

Not sure if that could cause any issues, I would guess it should be better than no.

lib/composables/useKFloatingInteraction/__tests__/index.spec.js
MisRob: In this particular test suite, I intentionally avoided having delays applied ~ tests are then complex and hard to read. Also, there's already lots of them.

But perhaps some critical pathways would be useful to test ~ somewhere in a separate file, focused particularly on delays.

rtibblesbot: Makes sense — keeping this suite delay-free is the right call for readability, and a separate delays-focused spec is a better home for the fake-timer cases. No objection to that split.

lib/composables/_useElementsCache/index.js
rtibblesbot: It would be correct — an empty result can only go stale via an element being *added*, and it'd never be held. But it costs the laziness the mutation handler is built around.

`onDOMMutations` only invalidates when a batch added an element (index.js:668), and `sharedQuery` is written to run the full `document.querySelectorAll(SELECTOR_ACTIVATOR)` scan only if some key actually missed ("Lazy — won't run when nothing was invalidated above", index.js:676-695). If `[]` is never cached, any non-delegated floating element whose activator isn't in the DOM right now (`v-if`'d out, not yet rendered) is a permanent miss, so every mutation batch on the page — including removal-only batches that today invalidate nothing — would trigger that whole-document scan.

And it buys no correctness here, since additions already call `forgetActivatorEls`. So I'd keep the caching and just state the invariant in the header comment; the empty entry is doing real work. Your call though — it's a nitpick either way.

examples/useKFloatingInteraction/Basic.vue
MisRob: As for `left: 0 is RTL-flipped`, I am not sure what would be best. I would expect that it flipping it to `right` would be what we want. But testing RTL reliably is not possible in KDS. I will add a note about to the roadmap issue so someone can try later in Kolibri or Studio.

As for `width: max-content was dropped` I removed it unintentionally. Will fix. Also I'll add a note to `useKFloatingPosition` documentation for developers to follow FloatingUI guidance for setting up styles.

rtibblesbot: Sounds good on both counts — thanks for picking up `width: max-content` and for adding the FloatingUI base-styles pointer to the docs.

On RTL: deferring to the roadmap issue is reasonable given there is no way to exercise an RTL build here. Agreed that a flip to `right` is the *intended* behaviour for a stylesheet rule; my concern was only about it combining with the un-flipped inline `left: <x>px`, which is exactly the kind of thing that needs a real RTL build to observe. Nothing further from me.

lib/composables/_useElementsCache/index.js
MisRob: Okay let's keep as is then, I checked on it and its true. It's more important to not lose cache at that place when not necessary.

examples/useKFloatingInteraction/Basic.vue
MisRob: Yes, the thing is that it's unclear if flipping to `right` for the inline style would work well with the `x` offset value that FloatingUI returns ~ it will need real-environment testing and more investigation. Just writing it down to the roadmap issue. Let's consider it solved here.

MisRob:
@rtibblesbot you can still re-review, to check that latest changes don't have any regressions

docs/pages/usekfloatinginteraction.vue
```diff
@@ -0,0 +1,520 @@
...
+        (value chosen based on manual testing and
+        <DocsExternalLink
+          text="Nielsen Norman Group guidance"
+          href="https://www.nngroup.com/articles/timing-exposing-content/"
+        />).
+      </p>
+
+      <p>
+        They also wait another <code>250ms</code> before deactivating, so the floating element
+        doesn't disappear while the pointer moves from the activator element to the floating element
```
rtibblesbot, line 290: **blocking:** 106 chars over `printWidth`; run `yarn lint-fix`.

docs/pages/usekfloatingposition.vue
```diff
@@ -51,30 +89,52 @@
...
-        to prevent severe performance problems.
-      </DocsBanner>
+      <!-- eslint-disable -->
+      <!-- prettier-ignore -->
+      <DocsShowCode language="css">
+        .tooltip {
+          position: absolute;
+          width: max-content; // or a fixed value
+          top: 0;
+          left: 0;
```
rtibblesbot, line 119: **suggestion:** RTLCSS flips `left: 0`; add `/* rtl:ignore */`.

examples/useKFloatingInteraction/Tour.vue
```diff
@@ -0,0 +1,197 @@
...
+      } = useKFloatingInteraction(SECOND_FLOATING_ID, secondRef, { interactions: ['click'] });
+
+      const { initPosition, destroyPosition, offset, flip, shift } = useKFloatingPosition();
+
+      function beginTour() {
+        tourStep.value = 1;
+        setFirstActive(true);
+      }
+
+      function continueTour() {
```
rtibblesbot, line 107: **suggestion:** Unmounting the focused button drops focus to `body`.

lib/composables/useKFloatingPosition/__tests__/index.spec.js
```diff
@@ -30,17 +36,14 @@ describe('useKFloatingPosition', () => {
 
   describe('initPosition', () => {
     it('stores the instance, sets up auto-updating, and updates the position', async () => {
-      initPosition('floating-1', floatingEl, anchorEl, { placement: 'bottom' });
+      await initPosition('floating-1', floatingEl, anchorEl, { placement: 'bottom' });
```
rtibblesbot, line 39: **suggestion:** Assert `initPosition` returns a promise.

examples/useKFloatingInteraction/ManualActivation.vue
```diff
@@ -0,0 +1,112 @@
...
+  };
+
+</script>
+
+
+<style lang="scss" scoped>
+
+  @import '../../lib/styles/definitions';
+
+  .buttons {
```
rtibblesbot, line 91: **suggestion:** Labels clip at 412px; add `flex-wrap: wrap`.

examples/useKFloatingInteraction/Basic.vue
```diff
@@ -0,0 +1,106 @@
...
+    min-width: 75px;
+    max-width: 240px;
+    padding: 8px;
+    font-size: 12px;
+    font-weight: normal;
+    line-height: 1.4;
+    text-align: center;
+    pointer-events: none;
+    border-radius: 8px;
+    opacity: 0;
```
rtibblesbot, line 102: **suggestion:** `opacity: 0` keeps this in the a11y tree; add `visibility: hidden`.

rtibblesbot (dismissed):
PR #1310 delta: 3 of 15 resolved, 7 acknowledged, 5 open.

RESOLVED — position:83
RESOLVED — Basic.vue:90
RESOLVED — index.js:266
ACKNOWLEDGED — index.js:1042
ACKNOWLEDGED — index.js:427
ACKNOWLEDGED — index.spec.js:65
ACKNOWLEDGED — tableOfContents.js:266
ACKNOWLEDGED — docs:119
ACKNOWLEDGED — cache:14
ACKNOWLEDGED — Enabling.vue:71
UNADDRESSED — validation.js:69
UNADDRESSED — index.js:797
UNADDRESSED — index.js:199
UNADDRESSED — docs:262
UNADDRESSED — CustomEscape.vue:4

</details>

---

</details>

docs/pages/usekfloatingposition.vue
MisRob: No need to repeat what was discussed https://github.com/learningequality/kolibri-design-system/pull/1310#discussion_r3695370544. This is already written down in the roadmap issue where it will be resolved. Consider closed for this PR.

rtibblesbot: You're right — this is the same `left: 0` / RTL point already closed at r3695370544 and tracked in the roadmap issue; I re-raised it on a different file path. Withdrawing this finding, and sorry for the noise.

docs/pages/usekfloatinginteraction.vue
```diff
@@ -0,0 +1,520 @@
...
+        the Dismissible, Hoverable, and Persistent requirements of the
+        <DocsExternalLink
+          text="WCAG Success Criterion"
+          href="https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html"
+        />.
+      </DocsBanner>
+    </DocsPageSection>
+
+    <DocsPageSection
+      title="Performance"
```
MisRob, line 51: I remember the request I received during the PoC work ~ to give some concrete numbers as guidance on which modes are better under what circumstances, and I thought about it a lot. However when I was re-working it for production, I got convinced that every answer based on a number is wrong, even if I communicated that it's generalization ~ there are many conditions in play, some of them mentioned below.

So I reframed the guidance as just pointers for developers to consider, and focused on communication what's roughly going on under the hood to help that. No comfy path here :)! Developers will need to consider for themselves. However, it's not complicated ~ performance measurement will be best anyway.

docs/pages/usekfloatinginteraction.vue
```diff
@@ -0,0 +1,520 @@
...
+        block
+      />
+
+      <h3>
+        Delay and animation
+        <DocsAnchorTarget anchor="#delay-and-animation" />
+      </h3>
+
+      <p>
+        By default, hover and focus wait for <code>300ms</code> before activating a floating
```
MisRob, line 279: The Nielsen Norman Group article linked below says to wait 0.3–0.5 seconds before displaying hidden content.

So 300ms is the lowest threshold. In manual testing I too had impression that anything below is insufficient.

I don't lean towards making the default larger, such as 400 or 500 ~ to me the tooltips start to feel laggy at such numbers.

If there's need to fine-tune the default value, best to consult with Jess :)

MisRob:
@rtibbles I resolved or pushed back on the bot feedback marked as important + few other comments + some additional tweaks. I reserved couple of hours for the bot review and final tweaks, and meanwhile the bot provided some new feedback ~ I haven't checked in detail, but at a quick glance seems some of is good. Nothing blocking or major though so I will stop here on my side. Thanks & take care

MisRob:
~~The bot likes to talk to me it seems, but I went ahead and removed things that weren't feedback to not overwhelm it here~~ Well, it's unstoppable :smile:

docs/pages/usekfloatinginteraction.vue
rtibblesbot: Understood — that reasoning holds. Any concrete threshold ("delegate above N activators") would be wrong the moment page size, mutation rate, or interaction mix changed, and stating it as a generalization does not stop readers from treating it as a rule. Pointers plus "measure it" is the honest framing. Withdrawing this finding.

docs/pages/usekfloatinginteraction.vue
rtibblesbot: You are right that 300ms sits inside the NN/g 0.3–0.5s range, so the default is not out of line with the cited guidance. Perceived lag at 400–500ms is a judgement call I cannot argue against, and fine-tuning it with Jess is the right route. Withdrawing this finding.

Issue #1318, opened 2026-08-21
[KMultiSelect] Add an always-expanded inline listbox mode (needed for Studio's Quick Edit categories modal)


Prashant-thakur77:
<img height="20px" src="https://i.imgur.com/0ZZG9qx.jpeg">

🙂 Looking for an issue? Welcome! This issue is open for contribution. If this is the first time you’re requesting an issue, please:

- **Read <a href="https://learningequality.org/contributing-to-our-open-code-base/" target="_blank">Contributing guidelines</a>** carefully. **Pay extra attention to [Using generative AI](https://learningequality.org/contributing-to-our-open-code-base/#using-generative-ai)**. **Pull requests and comments that don’t follow the guidelines won’t be answered.**
- **Confirm that you’ve read the guidelines** in your comment.

<img height="20px" src="https://i.imgur.com/0ZZG9qx.jpeg">

## Overview

KMultiSelect only renders its options as a floating dropdown that opens on focus and closes on blur, outside click, or Escape. Studio's CategoryOptions component also has an expanded mode, used by the Quick Edit "Edit categories" modal, where the full category list is permanently visible below the search field. Because KMultiSelect has no equivalent, the Studio migration (learningequality/studio#6092) could only convert the dropdown mode. The expanded mode still runs on VAutocomplete plus a hand rolled checkbox list, leaving the component a Vuetify/KDS hybrid.

KMultiSelect should support an opt-in expanded/inline mode so this last piece can be migrated, continuing the GSoC '26 Accessible MultiSelect work (#1259).

### Context

https://github.com/user-attachments/assets/09ec4423-2818-434c-8dd2-54150ac5e48f

What Studio needs to replace, in `shared/views/contentNodeFields/CategoryOptions.vue` when `expanded` is true:

- A `VAutocomplete` with its menu suppressed (`menu-props: { height: 0, maxHeight: 0 }`), used only for the search input, the selected chips, clear all, and the label.
- A hand rolled, permanently visible `KCheckbox` list below the field (250 px, scrollable) that renders the category tree filtered by the search text, with indentation while not searching and a "Category not found" empty state.

This duplicates filtering, hierarchical rendering, selection toggling, and the empty state outside of KMultiSelect, and there is no ARIA or keyboard relationship between the input and the list. Consumer: `channelEdit/components/QuickEditModal/EditCategoriesModal.vue` (`<CategoryOptions expanded ...>`).

Why KMultiSelect cannot do this today:

[...]

PR #1327, opened 2026-09-11
Add an always-expanded inline listbox mode to KMultiSelect

abhiraj75:
## Description

Adds an opt-in `expanded` prop to `KMultiSelect` that renders the option list inline and permanently visible below the field, plus `listMaxHeight` to cap the scroll height. This is what Studio needs to migrate the expanded branch of `CategoryOptions` off `VAutocomplete`.

**Base is `release-v5`, not `develop`.** The files only exist there (#1296).

#### Issue addressed

Addresses #1318

### Before/after screenshots

No real before/after: `expanded` is new, and with the flag off the dropdown is visually unchanged.

**New: `expanded` mode.** List in normal document flow, always visible, no toggle chevron.

Screenshot 2026-09-11 at 11 32 33 AM

## Changelog

  - **Description:** Add an opt-in `expanded` mode to `KMultiSelect` that renders the option list inline and permanently visible below the field, with no open/close lifecycle, plus a `listMaxHeight` prop to cap the list height in both modes. Also makes `Space` toggle the active option in both modes.
  - **Addresses:** https://github.com/learningequality/kolibri-design-system/issues/1318, https://github.com/learningequality/studio/pull/6092
  - **Components:** KMultiSelect
  - **Breaking:** no
  - **Impacts a11y:** yes
  - **Guidance:** Pass `expanded` to render the listbox in normal document flow, always visible. Blur, outside click, `Escape`, `Tab` and selection all leave it in place, and no toggle button is rendered. Use `listMaxHeight` (default `256px`) for the scroll cap. Per the WAI-ARIA APG combobox pattern the input keeps `role="combobox"` and reports `aria-expanded="true"` permanently. `messages.open` and `messages.close` are unused in this mode. Existing dropdown consumers are unaffected.

## Steps to test

`yarn dev`, then `http://localhost:4000/kmultiselect`, section **Expanded mode**.

1. List is visible on load, below the field, taking layout space.
2. Click away, `Tab`, `Escape`, select an option. It never hides.
3. `Down` then `Space` or `Enter` toggles, focus stays in the input.
4. Type `professional skills`. The space types normally, nothing toggles.
5. Type `zzz` for the empty state. Scroll the list, it caps at 250px.

Then confirm the existing dropdown examples still open, close, and float over the content below.

`isOpen` is seeded from `props.expanded` and `closeDropdown()` returns early in expanded mode. Every open path is already guarded by `if (!isOpen.value)`, and every close path funnels through `closeDropdown()`, so most of the behaviour falls out of those two changes.

Two structural bits:

- **New `.kmselect-field` wrapper.** The listbox rendered inside the bordered field box, so statically positioned it would inherit the grey background and border. It now sits in a wrapper alongside the box, with `containerEl` moved there to keep outside-click and blur containment working. Dropdown geometry verified unchanged in a browser.
- **`isPopupOpen`.** Splits "popup open" from "list visible", otherwise an expanded field renders permanently active.

### Deviations from the issue, please review

Five things go beyond the issue text. Happy to drop any:

1. **`disabled` in expanded mode.** Not in the issue. List stays rendered but inert. `pointer-events: none` turned out to miss synthetic clicks (how some AT activates elements), so `onListboxInput` refuses input when disabled. Easiest one to cut.
2. **`listMaxHeight` applies to both modes**, not just expanded. Default preserves current behaviour.
3. **Clear-button gutter.** Removing the chevron left a dead 40px gutter, so the clear button moves to `right: 4px`.
4. **`Escape` bubbles** in expanded mode instead of being swallowed, so the enclosing modal can close. Verified against a real `KModal`.
5. **`Space` toggles the active option, in both modes.** The AC requires it, but it worked in neither: `KListbox` handles `case ' '` only when the `<ul>` has DOM focus, which never happens. Follows the `Backspace` rule directly above it, empty search box means command, otherwise text, so multi-word search still works. **This changes dropdown-mode keyboard behaviour from #1296.**

The `v-show` gate and flip watcher were left in place rather than removed, since both already no-op when `isOpen` is permanently true.

- Studio migration of `CategoryOptions.vue` is a follow-up PR (step 8).

## Testing checklist

32 new tests, 159 passing in `KMultiSelect`, all 127 pre-existing unmodified. Browser-verified: slots, cascade, single mode, `disabled`, `listMaxHeight`, `Enter`, `Space`, and `Escape` inside a `KModal`.

## Reviewer guidance

- [ ] Are the five deviations acceptable, especially #1 (`disabled`) and #5 (`Space`)?
- [ ] Is the `.kmselect-field` wrapper the right way to lift the list out of the field box?
- [ ] Is permanent `aria-expanded="true"` right here? Agreed in the issue thread, worth a second look.
- [ ] LTR and RTL compliant? RTL untested, see tech debt.

## Comments

**AI usage:** I used Claude Code to survey the existing open/close paths and  draft the tests. I reviewed and verified every change myself, and the logical decisions  including the ones under "Deviations" are mine.


lib/candidate/multiselect/KMultiSelect/index.vue
```diff
@@ -390,6 +409,23 @@
...
+            if (
+              isOpen.value &&
+              !internalSearchText.value &&
+              dropdownComponent.value?.hasFocusedOption()
+            ) {
+              event.preventDefault();
+              dropdownComponent.value?.toggleFocusedOption();
+              break;
+            }
+            if (!isOpen.value) {
```
rtibblesbot, line 427: ✅ **Resolved** — addressed in the current code.


**blocking:** no `preventDefault()`, so the Space that opens the list is also typed. `internalSearchText` becomes `' '` — truthy — and the toggle guard above can never pass again.

On `#kmselect-input-0`: Space (opens, value `" "`) → ArrowDown → Space toggles nothing and types a second space. The placeholder vanishes, so it's visible. Same dead end by clicking open then pressing Space with no active option; `Enter` there falls back to `moveFocusByOne(1)` + toggle.

Add `event.preventDefault()` before `openDropdown()`, and mirror `Enter`'s fallback. Every Space test opens via the toggle or a click, so this path is untested — `expect(getCombobox()).toHaveValue('')` at `index.spec.js:455` catches it.

lib/candidate/multiselect/KMultiSelect/index.vue
```diff
@@ -390,6 +409,23 @@
             closeDropdown();
             break;
 
+          case ' ':
```
rtibblesbot, line 412: **suggestion:** no modifier guard, unlike `default:` below. Ctrl/Cmd/Alt/Shift+Space all toggle the active option, and Ctrl+Space opens a closed dropdown. Ctrl+Space is the IME switch on Windows/Linux — for Kolibri's language mix, switching input method inside the field silently changes the selection. Add `&& !event.ctrlKey && !event.metaKey && !event.altKey`.

lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue
```diff
@@ -3,7 +3,9 @@
   <div
     v-show="isOpen && (decoratedOptionTree.length > 0 || showEmptyState)"
     class="kmselect-dropdown"
-    :style="[{ backgroundColor: $themeTokens.surface }, flipStyle]"
+    :class="{ 'kmselect-dropdown-inline': expanded, 'is-disabled': expanded && disabled }"
```
rtibblesbot, line 6: ✅ **Resolved** — addressed in the current code.


**blocking:** `pointer-events: none` doesn't remove the listbox from the tab order, and `KListbox` renders `<ul role="listbox" tabindex="0">` unconditionally. In `disabled` + `expanded` the input is disabled, the clear button absent and the toggle `v-if="!expanded"` — so the `<ul>` is the control's only tab stop.

Tab-walked the disabled example: focus lands on `ul#kmselect-listbox-6`, then Arrow/Home/End move `aria-activedescendant` through all 16 options with the active ring painted and each announced, while Enter/Space do nothing. `aria-disabled` (line 8) sits on a role-less `<div>`; the focusable `<ul>` reports `ariaDisabled: null` — WCAG 4.1.2.

Fix: `tabindex="-1"` + `aria-disabled` on the `<ul>` when `expanded && disabled` (or `inert` on the wrapper). A `disabled` prop on `KListbox` that early-returns from `onListKeydown` would also fix `index.vue:498`, where the refusal runs after `toggleOption` has announced "Option deselected" and moved the active option.

Also: `opacity: 0.5` (line 400) fades the `surface` background, so page content shows through; the field beside it dims via `textDisabled`.

lib/candidate/multiselect/KMultiSelect/useMultiSelectDropdown.js
```diff
@@ -12,7 +12,7 @@ export default function useMultiSelectDropdown(
     getOptionText = o => o.label,
   },
 ) {
-  const isOpen = ref(false);
+  const isOpen = ref(props.expanded);
```
rtibblesbot: ✅ **Resolved** — addressed in the current code.


**blocking:** snapshotted at setup, never watched, so the public prop ignores updates. `false → true` on a mounted instance leaves `isOpen` false — list hidden, toggle already gone (`v-if="!expanded"`), a bare text field with no way to reach its options. `true → false` mounts a floating panel the user never opened. `:expanded="!windowIsSmall"` is an obvious consumer use, and nothing documents this as mount-only.

```js
const internalOpen = ref(false);
const isOpen = computed(() => props.expanded || internalOpen.value);
```

lib/candidate/multiselect/KMultiSelect/useMultiSelectDropdown.js
```diff
@@ -32,6 +32,16 @@ export default function useMultiSelectDropdown(
   }
 
   function closeDropdown() {
+    if (props.expanded) {
```
rtibblesbot, line 37: ✅ **Resolved** — addressed in the current code.


**blocking:** the early return never clears the active descendant, so it survives blur forever while the outline is correctly dropped and never restored. Dropdown mode clears it on close.

On `#kmselect-input-6`: focus + ArrowDown ×2 → `aria-activedescendant` = Mathematics, outlined. Blur → outline gone, attribute unchanged. Refocus → still Mathematics, still no outline; the next ArrowDown lands on Algebra, skipping two options.

A screen reader hears an active option while nothing is highlighted (WCAG 2.4.7). No close event exists to hang the reset on — clear `activeDescendantId` and the listbox's active index in `onInputBlur` when `props.expanded`.

lib/candidate/multiselect/KMultiSelect/index.vue
```diff
@@ -301,8 +313,12 @@
       watch(
         selectedOptionsData,
         options => {
-          if (props.multiple || isOpen.value) return;
-          setSearchText(options.length ? getOptionText(options[0]) : '');
+          if (props.multiple || isPopupOpen.value) return;
+          const label = options.length ? getOptionText(options[0]) : '';
+          // In expanded mode the list stays visible behind the field, so the label
+          // written back here must not filter it down to the single selection.
+          if (props.expanded) suppressFilter.value = Boolean(label);
```
rtibblesbot, line 320: ✅ **Resolved** — addressed in the current code.


**blocking:** this invariant is one keypress deep. `suppressFilter` is re-armed only here (on selection *change*) and in `closeDropdown()`'s expanded branch (on blur); `onSearchInput` clears it every keystroke (`:438`), and `openDropdown()` never runs again since `isOpen` starts `true`.

With the expanded example at `:multiple="false"`: click **Biology** (16 options, correct) → type a letter (0 options, 186px reflow) → Backspace back to exactly `Biology` → the list is now one option long: the item already selected.

Dead end from the keyboard — Escape breaks out at `:361`, clicking the last option is a single-mode no-op, nothing is announced. Recovery needs select-all-delete, the clear button, or a blur. Not pre-existing: the dropdown single example re-arms via `openDropdown()`.

Fix: re-arm whenever the query equals the selected label — treat `!multiple && expanded && query === selectedOptionText()` as unfiltered in `displayedOptions`.

lib/candidate/multiselect/KMultiSelect/index.vue
```diff
@@ -340,6 +356,9 @@
 
         switch (event.key) {
           case 'Escape':
+            // Expanded mode has no popup to dismiss, so let Escape reach the host
+            // (a modal, for instance) instead of swallowing it here.
+            if (props.expanded) break;
```
rtibblesbot, line 361: **suggestion:** the `break` drops the query reset too, not just the dismiss. With `geo` typed, expanded keeps `"geo"` and 1 option where dropdown restores the full list. The only keyboard way back is Backspace per character, and one too many silently drops a selection with no undo. APG assigns Escape to clearing the textbox when the listbox isn't dismissible — keep the event bubbling, but clear `internalSearchText` when non-empty.

lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue
```diff
@@ -3,7 +3,9 @@
   <div
     v-show="isOpen && (decoratedOptionTree.length > 0 || showEmptyState)"
```
rtibblesbot, line 4: **suggestion:** `showEmptyState` (`:185`) needs `noResultsText` or a `#no-results` slot, both optional — so an `expanded` consumer passing neither loses the whole list to `display: none` once the query stops matching, while `aria-expanded` stays `"true"`.

Even with `noResultsText`, height tracks the result count (`maxHeight`, no `min-height`): on `#kmselect-input-6`, `""` → field 311px, `"z"` → 125px. Everything below jumps 186px mid-typing and back on Backspace, and the same reflow fires under the pointer when selecting from a filtered list.

Reserve the space in expanded mode (`min-height`, or render the container unconditionally).

lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue
```diff
@@ -348,19 +373,33 @@
...
-    max-height: 256px;
-    padding: 8px 0;
     margin-top: 2px;
-    overflow-y: auto;
-    border-radius: 2px;
     @extend %dropshadow-2dp;
   }
 
+  .kmselect-dropdown-inline {
+    position: static;
```
rtibblesbot, line 399: **nitpick:** dead — the base rule no longer sets `position`, and `absolute` now lives only on `:not(.kmselect-dropdown-inline)`.

lib/candidate/multiselect/KMultiSelect/components/KMultiSelectInput.vue
```diff
@@ -336,4 +351,18 @@
     transform: translateY(-50%);
   }
 
+  // Without the toggle there is no gutter to reserve on the right, beyond
+  // whatever the clear button needs.
+  .kmselect-input-no-toggle {
+    padding-right: 10px;
+
+    &.has-clear {
+      padding-right: 36px;
```
rtibblesbot, line 360: **nitpick:** the clear button occupies the 4…36px band, so a 36px gutter ends exactly where the button begins and a full-width chip butts flush against it. Dropdown mode reserves 40px for a cluster ending at 36px. `padding-right: 40px` costs nothing — the gutter is empty without `has-clear`.

lib/candidate/multiselect/KMultiSelect/index.vue
```diff
@@ -285,6 +293,10 @@
         },
       );
 
+      // True only while the floating dropdown is showing. In expanded mode the list is always
+      // visible, so the active border, label colour and hover key off this instead of isOpen.
+      const isPopupOpen = computed(() => isOpen.value && !props.expanded);
```
rtibblesbot, line 298: ✅ **Resolved** — addressed in the current code.


**praise:** splitting "popup showing" from "list visible" in one computed keeps an expanded field from rendering permanently active without scattering `&& !expanded` through the template.

rtibblesbot (dismissed):
PR #1327 — reusing `closeDropdown()` as the single close seam carries the mode with little new state. Five blocking findings inline: Space, `disabled`, single-select, blur, prop reactivity. CI pending (Percy awaiting approval).

Two suggestions that don't map to changed lines:

**index.vue:161** — **suggestion:** `.kmselect-feedback` is a sibling of `.kmselect-field`, so `invalidText` renders 258px below its field in expanded mode (4px in dropdown mode) and jumps 186px as the list resizes. Place it above the inline list when `expanded`.

**index.vue:461** — **suggestion:** the `watch(isOpen)` that `select()`s the pre-filled label never fires when `isOpen` starts `true`, so typing after a single-mode selection appends (`"Biologya"`, 0 results). Select the label on focus instead.

**nitpicks:** `ExpandedMode.vue:56` passes `messages.open`/`close`, documented as unused here; `KMultiSelectDropdown.spec.js:180,193` and `KMultiSelectInput.spec.js:172` assert bindings they mirror (Percy covers positioning; jsdom can't); `index.spec.js:503` duplicates `KMultiSelectNode.spec.js:75-85`.

Manual QA against a live dev server: dismissal, flow, `listMaxHeight`, chevron removal, focus colouring, chip teardown and the unchanged floating panel all pass; axe AA found no new violations.

---

</details>

abhiraj75:
Hi @LianaHarris360,

Pushed fixes for the five blocking findings:

1. Make expanded reactive so it can be toggled after mount.
2. Clear the active descendant on blur in expanded mode
3. Stop Space from typing into the field when it opens the list
4. Restore the full list when the query is edited back to the selected label
5. Add a disabled prop to KListbox so a disabled inline list is not a tab stop

Three for your call:

1. Space doesn't mirror Enter's fallback - that contradicts the existing test at index.spec.js:455. preventDefault() alone fixes the defect. Can switch if you want them identical.

2. disabled is new public API on KListbox. Additive, guards toggleOption and onListFocus too. Say if you'd rather it stayed inside KMultiSelect.

3. The single-select fix assumes query-equals-label means "showing the selection", not "filtering by it".

Suggestions and nitpicks left alone pending your read.

rtibblesbot (commented):
PR #1327 — all 5 prior blocking findings resolved, each with a behavioural regression test; 11 prior findings at suggestion severity or below are still open, and 3 new suggestions on this round's additions (inline).

Not approving: CI has not verified this commit (`Javascript Tests` and `Lint` are `action_required`), manual QA did not run, and suggestions remain open. Locally `candidate/multiselect` + `candidate/listbox` are 208 passing.

Still open from prior rounds — existing threads, no new evidence:

- **index.vue:412** — no modifier guard on Space; Ctrl/Cmd/Alt+Space still toggles.
- **index.vue:361** — expanded Escape drops the query reset along with the dismiss.
- **KMultiSelectDropdown.vue:4** — empty-state gate hides the whole list; height reflows with the result count.
- **index.vue:161** — `.kmselect-feedback` renders far below its field in expanded mode.
- **index.vue:461** — pre-filled label never selected in expanded mode, so typing appends to it.
- **ExpandedMode.vue:56** — passes `messages.open`/`close`, documented as unused in expanded mode.
- **KMultiSelectDropdown.spec.js:180,193**, **KMultiSelectInput.spec.js:172**, **index.spec.js:503** — assert bindings/behaviour they mirror or that `KMultiSelectNode.spec.js:75-85` already covers.
- **KMultiSelectDropdown.vue:399** — dead `position: static`.
- **KMultiSelectInput.vue:360** — clear-button gutter is 36px against a 4…36px button.

RESOLVED — lib/candidate/multiselect/KMultiSelect/index.vue:427 — Space typed into the search box when it opens the list
RESOLVED — lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue:6 — disabled inline listbox is a focusable, operable tab stop
RESOLVED — lib/candidate/multiselect/KMultiSelect/useMultiSelectDropdown.js — `expanded` snapshotted at setup, ignores post-mount updates
RESOLVED — lib/candidate/multiselect/KMultiSelect/useMultiSelectDropdown.js:37 — active descendant survives blur forever in expanded mode
RESOLVED — lib/candidate/multiselect/KMultiSelect/index.vue:320 — `suppressFilter` re-arm dead end in expanded single mode
ACKNOWLEDGED — lib/candidate/multiselect/KMultiSelect/index.vue:298 — `isPopupOpen` splits "popup showing" from "list visible"
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/index.vue:412 — no modifier guard on Space
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/index.vue:361 — expanded Escape drops the query reset
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue:4 — empty-state gate hides the whole list
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/index.vue:161 — `.kmselect-feedback` renders far below its field in expanded mode
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/index.vue:461 — pre-filled label never selected in expanded mode
UNADDRESSED — examples/KMultiSelect/ExpandedMode.vue:56 — passes `messages.open`/`close`, unused here
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/__tests__/KMultiSelectDropdown.spec.js:180 — asserts a binding it mirrors
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/__tests__/KMultiSelectDropdown.spec.js:193 — asserts a binding it mirrors
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/__tests__/KMultiSelectInput.spec.js:172 — asserts a binding it mirrors
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/__tests__/index.spec.js:503 — duplicates KMultiSelectNode.spec.js:75-85
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue:399 — dead `position: static`
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/components/KMultiSelectInput.vue:360 — clear-button gutter is 36px against a 4…36px button
</details>

---

Ran an automatic code-only delta review triggered by new commits on a previously reviewed PR:
- Core review pass only — specialized frontend/backend lenses and manual QA run when a review is explicitly requested

</details>

lib/candidate/listbox/KListbox/index.vue
```diff
@@ -116,6 +117,7 @@
       }
 
       function toggleOption(value) {
+        if (props.disabled) return;
```
rtibblesbot, line 120: ✅ **Resolved** — addressed in the current code.


**suggestion:** `changeSelectAll` (`:176`) has no matching `disabled` guard, and the `@click.self` handler on `.k-listbox-select-all` (`:20`) reaches it directly. Against the built component:

```
props: { multiple: true, disabled: true }, scopedSlots: { selectAll }
click .k-listbox-select-all  ->  emitted input: [["a","b"]]
```

`selectAllStyles` also still resolves to `cursor: pointer` plus a hover background when disabled, so the row advertises itself as operable. No current consumer combines the two — KMultiSelect never passes `#selectAll` — but `disabled` is public API and four `examples/KListbox/*` use the slot. `if (props.disabled) return;` alongside the existing `hasOptions`/`multiple` guards, plus `disabled` folded into `selectAllStyles`, closes it.

lib/candidate/listbox/KListbox/index.vue
```diff
@@ -35,9 +35,10 @@
       :aria-label="ariaLabel"
       :aria-labelledby="ariaLabelledBy"
       class="k-listbox-list"
-      tabindex="0"
+      :tabindex="disabled ? -1 : 0"
```
rtibblesbot, line 38: **suggestion:** the ternary drops the `<ul>` out of the tab order only when `disabled`. In enabled expanded mode the inline list sits inside `containerEl` and stays `tabindex="0"`, so Tab from the combobox lands on it:

```
props: { expanded: true }
combobox.focus() -> tab()
  document.activeElement === ul.k-listbox-list  ->  true
  ul aria-activedescendant                      ->  "klistbox-option-4"
```

Dropdown mode has one tab stop because `v-show` hides the `<ul>` when closed; expanded mode has two, and the second drives its own `aria-activedescendant` in parallel with the combobox's. APG's combobox pattern keeps DOM focus on the combobox and the listbox out of the tab sequence. Deliberate? If not, `:tabindex="disabled || expanded ? -1 : 0"` (plumbed the way `disabled` already is) matches the pattern — the combobox's ArrowDown/Space path already covers everything the `<ul>`'s handlers do.

lib/candidate/multiselect/KMultiSelect/__tests__/index.spec.js
```diff
@@ -95,6 +95,38 @@ describe('KMultiSelect — keyboard interactions', () => {
...
+    await userEvent.keyboard(' ');
+    const emitted = component.emitted()['input'];
+    expect(emitted).toBeTruthy();
+    expect(emitted[emitted.length - 1][0]).toEqual([OPTIONS[0].value]);
+    expect(getCombobox()).toHaveValue('');
+  });
+
+  it('opens on Space without typing the space into the search input', async () => {
+    renderKMultiSelect();
+    await userEvent.click(getCombobox());
```
rtibblesbot: ✅ **Resolved** — addressed in the current code.


**suggestion:** this click bubbles to `onContainerClick`, which calls `openDropdown()` — the listbox is already present before line 112 runs, so the test named "opens on Space" never presses Space on a closed field.

The assertion still catches the reported defect (drop the new `preventDefault()` and line 114 fails with `" "`), so the test earns its place — but `index.vue:427`'s `if (!isOpen.value) openDropdown()` has no coverage in the file, and a refactor could delete it with this test still green. `getCombobox().focus()` in place of the click gives the name and the branch at once; verified it reaches the intended path (`listbox before: false`, `after: true`, input value `""`).

rtibblesbot (commented):
No new commits since my last review (`4933913`).

---

</details>

lib/candidate/listbox/KListbox/index.vue
```diff
@@ -172,6 +174,7 @@
       );
 
       function changeSelectAll(checked) {
+        if (props.disabled) return;
```
rtibblesbot, line 177: **suggestion:** the guard stops the effect but not the affordance. The `selectAll` slot scope is `{ allSelected, someSelected, setAllSelected }` — no `disabled` — so the checkbox every documented consumer renders there (`examples/KListbox/SelectAll.vue`, `Grouped.vue`, `Scrollable.vue`, `Customized.vue`) stays enabled, tabbable and announced as operable while `changeSelectAll` silently swallows the call. With the `<ul>` at `tabindex="-1"`, it is then the only tabbable thing in a disabled listbox; the row `div` also carries no `aria-disabled`.

Passing `:disabled="disabled"` into the slot scope (and `aria-disabled` on `.k-listbox-select-all`) lets consumers keep the two halves consistent. Not reachable from this PR — `KMultiSelect` never uses the `selectAll` slot — so this is about the new public API surface, not a live regression.

lib/candidate/listbox/KListbox/__tests__/index.spec.js
```diff
@@ -322,6 +322,61 @@ describe('KListbox', () => {
...
+      expect(component.emitted()['input']).toBeFalsy();
+    });
+
+    const renderWithSelectAll = (props = {}) =>
+      renderComponent({
+        props: { multiple: true, ...props },
+        scopedSlots: { selectAll: `<span slot-scope="s">All {{ s.allSelected }}</span>` },
+      });
+
+    const clickSelectAllRow = () => document.querySelector('.k-listbox-select-all').click();
```
rtibblesbot, line 365: **suggestion:** `document.querySelector('.k-listbox-select-all')` pins a behavior test to a scoped-style class name — a CSS rename breaks it. The guard under test lives in `changeSelectAll`, which the slot's `setAllSelected` also calls, so both assertions are reachable through the public path: render a control in the slot and click it by role.

```js
scopedSlots: { selectAll: `<button slot-scope="s" @click="s.setAllSelected(!s.allSelected)">Select all</button>` },
```

Keep one class-selector test only if the `@click.self` row-padding path is worth pinning separately. Minor: both tests are `async` with nothing awaited, and `multiple: true` is already the prop default.

rtibblesbot (commented):
PR #1327 — the delta adds a `disabled` guard to `changeSelectAll` and tightens two specs. Both new findings are suggestions about the guard's public surface; nothing blocking. CI is still pending (Percy), and manual QA did not run, so nothing here claims visual verification.

5 of 5 prior blocking findings resolved; 6 earlier suggestion/nitpick threads remain open and untouched by this delta.

**Still open from earlier rounds** (unchanged code — not re-raised inline):
- `KMultiSelect/index.vue:412` — `case ' '` still has no modifier guard, unlike `default:`.
- `KMultiSelect/index.vue:361` — the expanded-mode `break` on Escape still drops the query reset along with the dismiss.
- `KMultiSelectDropdown.vue:4` — expanded mode still collapses to nothing when neither `noResultsText` nor a `#no-results` slot is supplied.
- `KListbox/index.vue:38` — the `<ul>` stays tabbable in enabled expanded mode, giving the field two tab stops.
- nitpicks: `KMultiSelectDropdown.vue:399` dead `position`, `KMultiSelectInput.vue:360` 36px clear-button gutter.

Two new suggestions inline on the select-all guard.

RESOLVED — lib/candidate/multiselect/KMultiSelect/index.vue:427 — Space that opens the list is also typed (no preventDefault)
RESOLVED — lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue:6 — pointer-events leaves listbox in tab order
RESOLVED — lib/candidate/multiselect/KMultiSelect/useMultiSelectDropdown.js — prop snapshotted at setup, never watched
RESOLVED — lib/candidate/multiselect/KMultiSelect/useMultiSelectDropdown.js:37 — active descendant survives blur
RESOLVED — lib/candidate/multiselect/KMultiSelect/index.vue:320 — suppressFilter invariant only one keypress deep
RESOLVED — lib/candidate/listbox/KListbox/index.vue:120 — changeSelectAll had no disabled guard
RESOLVED — lib/candidate/multiselect/KMultiSelect/__tests__/index.spec.js — Space test opened the dropdown before pressing Space
ACKNOWLEDGED — lib/candidate/multiselect/KMultiSelect/index.vue:298 — popup-showing vs list-visible split (praise)
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/index.vue:412 — no modifier guard on Space
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/index.vue:361 — expanded Escape break drops the query reset
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue:4 — expanded empty state needs noResultsText or slot
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue:399 — dead position rule
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/components/KMultiSelectInput.vue:360 — 36px clear-button gutter
UNADDRESSED — lib/candidate/listbox/KListbox/index.vue:38 — ul tabbable in enabled expanded mode

</details>

---

</details>

docs/pages/kmultiselect.vue
```diff
@@ -11,6 +11,16 @@
...
+        <li>
+          In expanded mode the listbox is permanently visible, so the combobox keeps
+          <code>role="combobox"</code> and reports <code>aria-expanded="true"</code> at all times,
+          following the
+          <DocsExternalLink
+            href="https://www.w3.org/WAI/ARIA/apg/patterns/combobox/"
+            text="WAI-ARIA APG combobox pattern"
+          />. <code>messages.open</code> and <code>messages.close</code> are unused in that mode,
+          since no toggle button is rendered.
+        </li>
```
LianaHarris360, line 23: This explanation is already documented in the PR description, I think it's fine to remove from this section.

lib/candidate/multiselect/KMultiSelect/__tests__/KMultiSelectDropdown.spec.js
```diff
@@ -175,3 +175,35 @@ describe('KMultiSelectDropdown — empty state', () => {
...
+  it('applies the maxHeight prop to the list', () => {
+    const { container } = renderDropdown({ expanded: true, maxHeight: '250px' });
+    expect(container.querySelector('.kmselect-dropdown')).toHaveStyle({ maxHeight: '250px' });
+  });
+
+  it('marks the list inert when disabled in expanded mode', () => {
+    const { container } = renderDropdown({ expanded: true, disabled: true });
+    const panel = container.querySelector('.kmselect-dropdown');
+    expect(panel).toHaveAttribute('aria-disabled', 'true');
+    expect(panel).toHaveClass('is-disabled');
```
LianaHarris360, line 202: The new tests in this file use `container.querySelector('.kmselect-dropdown')` and then test a class name. But if these styles change, these test will break for a component that still works. 

For the disabled mode, because `KMultiSelectDropdown` passes `:disabled="expanded && disabled"` to `KListbox`, which puts *aria-disabled* and *tabindex* on the listbox element, can you try using `const listbox = await screen.findByRole('listbox')`, then checking that the listbox has `aria-disabled` and a negative tab index instead?

For the other tests, you can add `data-testid="kmselect-dropdown-panel"` to the panel, and query it with `screen.getByTestId()`. This is based on kolibri's unit testing guidance [here](https://kolibri-dev.readthedocs.io/en/develop/frontend_architecture/unit_testing.html).

This also applies to the new tests in `KMultiSelectInput.spec.js`. The second half of the tests use `kmselect-input-no-toggle` and `has-clear`. I would recommend removing the "still renders the clear button, moved into the freed gutter" test, because we can let the visual test cover layout assertions.

lib/candidate/multiselect/KMultiSelect/__tests__/index.spec.js
```diff
@@ -304,3 +337,251 @@ describe('KMultiSelect — single mode pre-fill', () => {
...
+    expect(screen.getByRole('listbox')).toBeInTheDocument();
+  });
+
+  it('filters the inline list as the user types', async () => {
+    renderExpanded();
+    await userEvent.type(getCombobox(), 'ban');
+    await waitFor(() => {
+      const visible = screen.getAllByRole('option');
+      expect(visible).toHaveLength(1);
+      expect(visible[0]).toHaveAccessibleName('Banana');
```
LianaHarris360, line 433: Rather than using a hardcoded string, reference the value from `HIERARCHICAL_OPTIONS` instead. Use this same pattern for the hardcoded strings on lines  517, 518, 528 and 532.

lib/candidate/multiselect/KMultiSelect/__tests__/index.spec.js
```diff
@@ -304,3 +337,251 @@ describe('KMultiSelect — single mode pre-fill', () => {
...
+  it('cascades a parent selection through the inline list with autoSelectChild', async () => {
+    const component = renderExpanded({
+      options: HIERARCHICAL_OPTIONS,
+      autoSelectChild: true,
+    });
+    await userEvent.click(await screen.findByRole('option', { name: 'Fruit' }));
+    const emitted = component.emitted()['input'];
+    expect(emitted).toBeTruthy();
+    expect(emitted[emitted.length - 1][0]).toEqual(
+      expect.arrayContaining(['fruit', 'apple', 'banana']),
```
LianaHarris360, line 532: `arrayContaining` passes if the array holds extra values, but this is the bug this test exists to catch, so using it here hides potential failures. Since the tree holds three options, the exact result is known, so the `expect.arrayContaining` can be removed.

lib/candidate/multiselect/KMultiSelect/__tests__/index.spec.js
```diff
@@ -304,3 +337,251 @@ describe('KMultiSelect — single mode pre-fill', () => {
...
+describe('KMultiSelect — expanded mode', () => {
+  const HIERARCHICAL_OPTIONS = [
+    { label: 'Fruit', value: 'fruit', level: 0 },
+    { label: 'Apple', value: 'apple', level: 1 },
+    { label: 'Banana', value: 'banana', level: 1 },
+  ];
+
+  const indentFor = depth => `${16 + depth * 24}px`;
+
+  const renderExpanded = (propsOverrides = {}, options = {}) =>
```
LianaHarris360, line 352: Rather than rebuilding what `renderKMultiSelect` already does, can you add a second argument to it and call it from `renderExpanded` instead? E.g.
```
function renderKMultiSelect(propsOverrides = {}, options = {}) {
  return render(KMultiSelect, {
    props: { ...BASE_PROPS, ...propsOverrides },
    ...options,
  });
}
```

LianaHarris360 (changes requested):
The code itself looks fine and the tests are in good shape. I just had a few clean up notes. Great work @abhiraj75! 

Also tagging @radinamatic for some QA on the new expanded mode usage.

lib/candidate/multiselect/KMultiSelect/__tests__/KMultiSelectDropdown.spec.js
abhiraj75: Done ! Two consequences worth flagging:

The panel's own is-disabled class and aria-disabled are no longer asserted anywhere. Both looked like styling and duplicate ARIA next to the listbox attributes, it can be discussed if we'd rather keep one assertion on the wrapper.

I deliberately did not assert tabindex="0" on the enabled test, since whether the <ul> should be tabbable in expanded mode is still open on another thread.

For KMultiSelectInput.spec.js, removed the clear-button test. The .closest('.kmselect-input') query at :133 predates this PR so I left it alone - happy to convert that one too if you'd like.

lib/candidate/multiselect/KMultiSelect/__tests__/index.spec.js
abhiraj75: One deviation on 433 itself: that test calls renderExpanded(), which renders the flat OPTIONS fixture from BASE_PROPS rather than the hierarchical one, so I used OPTIONS[1].label. HIERARCHICAL_OPTIONS[2].label is also 'Banana' and would have passed, but it would point at a fixture the test never renders. Happy to switch if you'd rather.

lib/candidate/multiselect/KMultiSelect/__tests__/KMultiSelectDropdown.spec.js
```diff
@@ -175,3 +176,33 @@ describe('KMultiSelectDropdown — empty state', () => {
...
+    renderDropdown();
+    expect(screen.getByTestId(DROPDOWN_PANEL_TESTID)).not.toHaveClass('kmselect-dropdown-inline');
+  });
+
+  it('applies the maxHeight prop to the list', () => {
+    renderDropdown({ expanded: true, maxHeight: '250px' });
+    expect(screen.getByTestId(DROPDOWN_PANEL_TESTID)).toHaveStyle({ maxHeight: '250px' });
+  });
+
+  it('marks the list inert when disabled in expanded mode', async () => {
```
rtibblesbot, line 196: **suggestion:** The disabled panel's `opacity: 0.5` now has zero coverage of any kind. The `.is-disabled` assertion was dropped as visual-test territory, but no Percy case renders a disabled dropdown. `KMultiSelectVisualTest.vue` mounts only `ExpandedMode.vue`, which sets no `disabled`. Add a disabled visual-test example, or keep the one `toHaveClass('is-disabled')` line here.

On your open question about the panel wrapper: don't re-add `aria-disabled` there. On a role-less `<div>` it is inert to AT. These listbox assertions are the real semantics.

lib/candidate/multiselect/KMultiSelect/__tests__/KMultiSelectInput.spec.js
```diff
@@ -156,3 +156,16 @@ describe('KMultiSelectInput', () => {
...
+    renderComponent({ expanded: true });
+    expect(screen.queryByRole('button', { name: LABELS.openLabel })).not.toBeInTheDocument();
+    expect(screen.queryByRole('button', { name: LABELS.closeLabel })).not.toBeInTheDocument();
+  });
+
+  it('still renders the toggle when expanded is off', () => {
+    renderComponent();
+    expect(screen.getByRole('button', { name: LABELS.openLabel })).toBeInTheDocument();
+  });
+});
```
rtibblesbot, line 171: **suggestion:** Nothing covers the clear button in expanded mode now. Its `getByRole(clearAllLabel)` assertion went out alongside the mirrored class assertion I asked you to drop. The Percy case doesn't cover it either: `ExpandedMode.vue:52` starts at `selected: []`, so the button never renders and the moved gutter is unsnapshotted. Keep the `getByRole` line, or seed the example with `selected: ['WORK']` — that puts the gutter in the snapshot for free.

lib/candidate/multiselect/KMultiSelect/__tests__/KMultiSelectDropdown.spec.js
```diff
@@ -175,3 +176,33 @@ describe('KMultiSelectDropdown — empty state', () => {
...
+    renderDropdown({ expanded: true, disabled: true });
+    const listbox = await screen.findByRole('listbox');
+    expect(listbox).toHaveAttribute('aria-disabled', 'true');
```
rtibblesbot, line 199: **praise:** Asserting `tabindex="-1"` and `aria-disabled` on the `<ul>` pins exactly the pair whose absence was the tab-order bug. The old class assertion would have passed with the listbox still focusable.

rtibblesbot (commented):
PR #1327: 8 of 23 prior findings are resolved. One of those was praise, acknowledged. The other 14 are still open on unchanged code, so those threads stand.

The delta applies @LianaHarris360's cleanup notes faithfully. Dropping `arrayContaining` does tighten that test: `HIERARCHICAL_OPTIONS.map(o => o.value)` is exactly the order the cascade emits. Ran `candidate/multiselect` + `candidate/listbox` locally: 209 passing, 2 visual suites skipped.

One new suggestion, inline in two files. Two states lost their assertions as visual-test territory, but neither the disabled panel nor the clear-button gutter is rendered in any Percy snapshot. Both now have zero coverage. Restore the assertion or extend the example in each case — specifics are in the inline comments.

CI pending. Manual QA did not run, so nothing here is visually verified.

RESOLVED — lib/candidate/multiselect/KMultiSelect/index.vue:427 — Space that opens the list is also typed (no preventDefault)
RESOLVED — lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue:6 — pointer-events leaves listbox in tab order
RESOLVED — lib/candidate/multiselect/KMultiSelect/useMultiSelectDropdown.js — `expanded` snapshotted at setup, never watched
RESOLVED — lib/candidate/multiselect/KMultiSelect/useMultiSelectDropdown.js:37 — active descendant survives blur
RESOLVED — lib/candidate/multiselect/KMultiSelect/index.vue:320 — `suppressFilter` invariant only one keypress deep
RESOLVED — lib/candidate/listbox/KListbox/index.vue:120 — `changeSelectAll` had no disabled guard
RESOLVED — lib/candidate/multiselect/KMultiSelect/__tests__/index.spec.js — Space test opened the dropdown before pressing Space
RESOLVED — lib/candidate/multiselect/KMultiSelect/__tests__/KMultiSelectInput.spec.js:172 — asserts a binding it mirrors
ACKNOWLEDGED — lib/candidate/multiselect/KMultiSelect/index.vue:298 — `isPopupOpen` splits "popup showing" from "list visible" (praise)
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/index.vue:412 — no modifier guard on Space
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/index.vue:361 — expanded Escape drops the query reset
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/index.vue:461 — pre-filled label never selected in expanded mode
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/index.vue:161 — `.kmselect-feedback` renders far below its field in expanded mode
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue:4 — empty-state gate hides the whole list
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue:399 — dead `position: static`
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/components/KMultiSelectInput.vue:360 — clear-button gutter is 36px against a 4…36px button
UNADDRESSED — lib/candidate/listbox/KListbox/index.vue:38 — `<ul>` stays tabbable in enabled expanded mode, two tab stops
UNADDRESSED — lib/candidate/listbox/KListbox/index.vue:177 — `selectAll` slot scope has no `disabled`, affordance stays operable
UNADDRESSED — lib/candidate/listbox/KListbox/__tests__/index.spec.js:365 — select-all guard tested through a scoped-style class name
UNADDRESSED — examples/KMultiSelect/ExpandedMode.vue:56 — passes `messages.open`/`close`, unused here
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/__tests__/KMultiSelectDropdown.spec.js:183 — asserts a class binding it mirrors
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/__tests__/KMultiSelectDropdown.spec.js:188 — asserts a class binding it mirrors
UNADDRESSED — lib/candidate/multiselect/KMultiSelect/__tests__/index.spec.js:513 — duplicates KMultiSelectNode.spec.js:75-85
</details>

---

Ran an automatic code-only delta review triggered by new commits on a previously reviewed PR:
- Core review pass only — specialized frontend/backend lenses and manual QA run when a review is explicitly requested

</details>

lib/candidate/multiselect/KMultiSelect/__tests__/index.spec.js
LianaHarris360: You're right, your change here is correct as-is, the test should point at the fixture it renders!

lib/candidate/multiselect/KMultiSelect/__tests__/KMultiSelectDropdown.spec.js
LianaHarris360: In expanded mode the list is the popup of the combobox, not a second widget. The input already carries `aria-controls` and `aria-activedescendant`, so it is the single tab stop and it drives the highlight. That means the list is `tabindex="-1"` in expanded mode whether or not it is disabled. We should assert `tabindex="-1"` on both expanded tests, the disabled one and the enabled one.

On the wrapper `aria-disabled`, is redundant. The listbox carries the same state and nothing focusable sits between them.

`is-disabled` is different. It carries `pointer-events: none` and `opacity: 0.5` and after the test change nothing covers it. `examples/KMultiSelect/ExpandedMode.vue` has no disabled variant either, so the snapshot doesn't show that state. Can you add one there?

The deleted clear-button test dropped two things, the class check is the part that needed to be removed. The other half asserted that the button renders in expanded mode, which is behavior. Can you restore that half as `getByRole('button', { name: LABELS.clearAllLabel })`, with no class check?

LianaHarris360:
@abhiraj75 Thanks for the quick turnaround, your fixes look right. As for your open questions:

1. Space doesn't mirror Enter's fallback - that contradicts the existing test at index.spec.js:455. preventDefault() alone fixes the defect. Can switch if you want them identical.

This is correct, in the current `CategoryOptions` implementation in Studio, the behavior of space and enter is different as well. Enter accepts: it commits the highlighted option, or the top match after a query. Space toggles, and it only takes the key when the field is empty, because a space is printable text. Making them identical would mean a single space press  select an option the user never saw. So `preventDefault()` is the right fix, and the `does not select anything on Space when no option is active` test asserts the behavior we want to keep, so it stays as is. If anything, could you add a comment at the Space branch that states why the two keys are different?

2. disabled is new public API on KListbox. Additive, guards toggleOption and onListFocus too. Say if you'd rather it stayed inside KMultiSelect.

`disabled` has to stay public. KListbox's root is a `div`, so nothing from outside reaches the `ul` that carries `role="listbox"`. Only KListbox can set `aria-disabled` and `tabindex` there. 

Can you split the tab stop into its own prop, `tabbable`, default true, and have `KMultiSelectDropdown` pass false? An enabled expanded list is not disabled. It is only not the focus owner, and `aria-disabled="true"` on it would be wrong.

It would be preferable to keep `toggleOption` and `changeSelectAll` in the KListbox, for cases where it is used on it's own. For example, in Kolibri, the `SelectableList` pairs KListbox with a `FilterTextbox`, a plain input with `aria-controls`, so the user tabs into the list to reach options. Both defaults would leave it unchanged. It is also why `tabbable` must default to true and stay separate from `disabled`

3. The single-select fix assumes query-equals-label means "showing the selection", not "filtering by it".

This assumption is fine. In single-select expanded mode, filtering the list down to the option you already picked doesn't tell the user anything.

abhiraj75:
Thanks @LianaHarris360! Addressed:

1. Added a comment at the Space branch explaining why it doesn't mirror Enter's fallback.
2. Split the tab stop into a new tabbable prop on KListbox (default true); disabled keeps aria-disabled and the guards. KMultiSelectDropdown passes :tabbable="!expanded", so the combobox is the single tab stop in expanded mode.
3. Removed the redundant aria-disabled from the panel wrapper; is-disabled stays.
4. Both expanded listbox tests now assert tabindex="-1"; restored the clear-button check as getByRole only.
5. Added a disabled variant to ExpandedMode.vue so Percy snapshots the is-disabled state.

rtibblesbot (commented):
PR #1327: all 8 prior findings resolved; no new findings.

Not approving yet: CI is pending and manual QA didn't run.

RESOLVED — lib/candidate/multiselect/KMultiSelect/index.vue:430 — Space that opens the list is also typed
RESOLVED — lib/candidate/multiselect/KMultiSelect/components/KMultiSelectDropdown.vue:6 — `pointer-events: none` leaves listbox in tab order
RESOLVED — lib/candidate/multiselect/KMultiSelect/useMultiSelectDropdown.js — `expanded` snapshotted at setup, never watched
RESOLVED — lib/candidate/multiselect/KMultiSelect/useMultiSelectDropdown.js:37 — early return never clears active descendant
RESOLVED — lib/candidate/multiselect/KMultiSelect/index.vue:320 — `suppressFilter` invariant one keypress deep
RESOLVED — lib/candidate/multiselect/KMultiSelect/index.vue:298 — praise: popup-showing vs list-visible split
RESOLVED — lib/candidate/listbox/KListbox/index.vue:120 — `changeSelectAll` lacks `disabled` guard
RESOLVED — lib/candidate/multiselect/KMultiSelect/__tests__/index.spec.js — click bubbles to `onContainerClick`

</details>

---

Ran an automatic code-only delta review triggered by new commits on a previously reviewed PR:
- Core review pass only — specialized frontend/backend lenses and manual QA run when a review is explicitly requested

</details>

